PR #2, opened 2026-08-02
Feature/stack tracing

KushagarGoel1:
PR description:

  ## Type of Change

  - [x] Enhancement
  - [x] Refactoring
  - [x] Dependency updates

  ## Description

  Adds stack-trace enrichment, safe error serialization, fingerprints,
  and structured logging across backend, dashboard, and Electron.

  Also fixes production alias resolution, keeps XyneAI web-worker logging
  window-safe, and makes Jest initialization work in clean environments.

  XYNE-54859

  ## Areas Touched

  - [x] `apps/backend` (API / worker)
  - [x] `apps/dashboard`
  - [x] `apps/electron`
  - [x] `packages/shared` or another shared package
  - [x] Infra (scripts)

  ---

  ## Zero (Sync) Changes

  - [x] No Zero changes — **skip this section**

  ## Schema & Data Changes

  - [x] No schema changes — **skip this section**

  ## Config, Environment & URLs

  - [x] No config changes — **skip this section**

  ## API Contract

  - [x] No API changes

  ---

  ## How did you test it?

  - `pnpm run build:shared`
  - Backend typecheck and build
  - Dashboard typecheck
  - Targeted dashboard ESLint
  - Backend Jest test with DB/JWT environment variables unset — 15 tests
  passed
  - Verified compiled backend `@/` imports resolve through production
  `tsx --enable-source-maps`
  - Dashboard build transformed all modules but exhausted an 8 GB heap
  during final chunk rendering

  ## Checklist

  - [x] `pnpm run build:shared` passes
  - [x] Backend `typecheck` + `build` pass
  - [ ] Dashboard `lint:errors-only` + `typecheck` + `build` pass —
  targeted lint/typecheck pass; build hit OOM
  - [x] Formatting clean in touched files
  - [x] Commits follow `<type>: <TICKET-ID> <subject>`; branch is `fix/*`
  or `feature/*`
  - [x] Docs / guidelines updated where behaviour changed

PR #41, opened 2026-08-02
feat: XYNE-16938 release manager v3 with GitHub support

KushagarGoel1:
> Migrated from juspay/xyne-spaces PR #17 (original author: @sumantop).

release manager v3 with GitHub support


apps/backend/src/services/release/buildVcsClient.ts
```diff
@@ -0,0 +1,42 @@
...
+import { config } from '@/config/env';
+import { logger } from '@/utils/logger';
+import { BitbucketService } from '../bitbucketService';
+import { GitHubService } from '../githubService';
+import { VcsClient } from '../../types/vcs';
+import { VCSProviderType } from '@prisma/client';
+
+// Bitbucket REST base, tolerating a bare host or a full .../rest/api/latest.
+export function normalizeBitbucketApiBaseUrl(baseUrl: string | undefined): string {
+  if (!baseUrl) return 'https://bitbucket.juspay.net/rest/api/latest';
```
devesh-juspay, line 14: This fallback hardcodes the internal host `bitbucket.juspay.net` as the default in an OSS repo (main previously used `bitbucket.example.com`). Consider reverting to a neutral placeholder or throwing when `BITBUCKET_BASE_URL`/config is unset, so the internal hostname isn't baked into published code — and so a misconfigured deployment fails loudly instead of silently calling an internal host it can't reach.

apps/backend/scripts/release-manager-localdev/register-github-webhook.sh
```diff
@@ -0,0 +1,78 @@
...
+#
+#   GITHUB_TOKEN=ghp_xxx bash register-github-webhook.sh <owner/repo> <workspaceId>
+#
+# Reads the shared secret from backend/.env.local (SCM_WEBHOOK_SECRET) so it
+# matches what the backend verifies. Nothing is hardcoded.
+#
+set -euo pipefail
+
+REPO="${1:-sumantop/what-release-manager-test}"
+WORKSPACE_ID="${2:-cmr4y2uyd001d127hzji6mnz5}"
```
devesh-juspay, line 18: These defaults check a personal repo (`sumantop/what-release-manager-test`) and what looks like a real workspace id into the repo. Since both are already accepted as positional args, consider making them required (error out with a usage message when unset) instead of defaulting to personal/internal values.

packages/shared/src/index.ts
```diff
@@ -15,6 +15,7 @@ export * from './types/workflowApproval';
 export * from './types/userActivity';
 export * from './types/call';
 export * from './types/callChat';
+export * from './types/call';
```
devesh-juspay, line 18: Duplicate export — `./types/call` is already re-exported on line 16. This line can be dropped.

devesh-juspay (commented):
Two small issues spotted in review — a hardcoded internal host that leaks into this OSS repo, and an accidental duplicate export.

apps/backend/src/routes/commitAnalysis.ts
```diff
@@ -1,9 +1,116 @@
...
+        id: true,
+        xyneId: true,
+        ticketType: true,
+        conversationId: true,
+        channelId: true,
+        workspaceId: true,
+      },
+    });
+    if (!ticket) {
+      res.status(404).json({ error: `Ticket ${ticketId} not found` });
```
devesh-juspay, line 44: **Missing workspace-membership check.** This endpoint only verifies the caller is authenticated — the ticket is looked up globally by id, and there's no check that the caller belongs to `ticket.workspaceId`. Any logged-in user who obtains a ticket id from another workspace can trigger analysis there, which writes sub-tickets, posts/updates the conversation card and canvas, and makes outbound VCS calls as that workspace's `xyne-release-bot`.

After loading the ticket, verify the caller is a member of `ticket.workspaceId` and return 404 otherwise (404 rather than 403, to avoid leaking ticket existence across workspaces).

The `/test-connection` endpoint below has a milder version of the same issue: any authenticated user can probe arbitrary repo URLs with the server's VCS token (a private-repo existence oracle). Worth scoping it to users who can actually configure releases.

apps/backend/scripts/release-manager-localdev/register-github-webhook.sh
devesh-juspay: are these scripts usable? should we clean them and keep for oss? or shall we move these local dev files to xyne-spaces-private repo?

apps/backend/prisma/migrations/20260707000000_add_art_is_hotfix/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- Nullable, no DB default: Zero's optimistic client inserts don't run Postgres
+-- defaults, so isHotfix is written app-side; NULL reads as "not a hotfix".
+ALTER TABLE "public"."application_release_tickets" ADD COLUMN "isHotfix" BOOLEAN;
+
+COMMENT ON COLUMN "public"."application_release_tickets"."isHotfix" IS 'Dev ticket entered THIS release as a hotfix (merged after the release head). NULL = not a hotfix.';
```
devesh-juspay, line 5: Is this `COMMENT ON COLUMN` needed? It's pure documentation metadata in Postgres — nothing reads it, and Prisma doesn't manage column comments, so it only shows up when inspecting the DB directly. The SQL comment above plus the schema.prisma field already document the semantics. Suggest dropping it before merge — once the migration is applied in shared environments, editing the file breaks its checksum, so this is the last cheap moment to remove it.

apps/backend/src/database/repositories/applicationRepository.ts
```diff
@@ -124,14 +140,34 @@ export class ApplicationRepository {
...
+    // ── Idempotency guard ─────────────────────────────────────────────────────
+    // Re-running commit analysis must NOT keep creating new SubTickets — each
+    // new SubTicket spawned a fresh applicationReleaseId, which bypassed the
+    // (applicationReleaseId, ticketId) unique constraint on ART rows and led
+    // to 3× duplicated Dev Tickets / Envs / Migrations after 3 re-runs.
+    // Look up existing per-app SubTickets for this release and reuse them.
+    const result = await this.findExistingApplicationSubTickets(
+      parentTicketId,
+      affectedApplications,
+    );
```
devesh-juspay, line 153: **Context for reviewers — this is the root-cause fix for the `applicationReleaseId` hole.**

The pre-existing dedupe guard (`releaseChangeInstanceExists` in `commitAnalysisService.ts`) keyed its lookup on `applicationReleaseId` (the SubTicket id). Because every analysis run previously created brand-new SubTickets, each run minted a fresh `applicationReleaseId`, so the guard never matched across runs — same release/app/file/commit rows were re-inserted on every re-run. The guard only ever protected within a single run.

On main this hole was latent (nothing could re-run analysis for the same ticket — verified: prod `release_change_types` has zero duplicates). This PR makes re-runs routine (Re-run button + webhook hotfix sync), so it's fixed at three layers:

1. **Here** — SubTickets are reused via `findExistingApplicationSubTickets`, making `applicationReleaseId` stable across runs (closes the hole at its source).
2. `releaseRepository.createReleaseChangeType` dedupes on `(releaseId, applicationId, changeType, filePath, commitId)` — deliberately excluding `applicationReleaseId` — with a `P2002` catch for concurrent runs (webhook sync racing a manual re-run).
3. The `20260708000000_release_change_type_dedupe_index` migration adds the partial unique indexes that make the `P2002` path actually fire; its cleanup `DELETE` is a no-op on databases that never re-ran (verified zero rows affected in ours) but protects any environment with pre-existing debris.

apps/backend/prisma/migrations/20260708000000_release_change_type_dedupe_index/migration.sql
```diff
@@ -0,0 +1,30 @@
...
+  AND keep."applicationId" = t."applicationId"
+  AND keep."changeType" = t."changeType"
+  AND keep."filePath" = t."filePath"
+  AND keep."commitId" IS NOT DISTINCT FROM t."commitId"
+  AND (COALESCE(keep."createdAt", 'epoch'::timestamptz), keep."id")
+    < (COALESCE(t."createdAt", 'epoch'::timestamptz), t."id");
+
+-- CONCURRENTLY: no write-blocking lock on this pre-existing table (Prisma applies
+-- migrations without a wrapping txn — see 20260421190000_add_performance_indices).
+CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS "release_change_types_release_dedupe_key"
```
devesh-juspay, line 24: will these indexes be supported by zero?

apps/backend/src/database/repositories/applicationRepository.ts
```diff
@@ -124,14 +140,34 @@ export class ApplicationRepository {
...
+    const missingApplications = affectedApplications.filter(app => !result.has(app.id));
+
+    if (missingApplications.length === 0) {
```
devesh-juspay, line 156: **Race condition: concurrent runs can still create duplicate SubTickets.** This guard is find-then-create with no unique DB constraint on `(parent release, application)` and no lock — and the per-app `catch` below only logs, so nothing blocks a duplicate write.

Failure scenario: two PRs merge on a hotfix branch seconds apart (or a webhook redelivery races a user clicking Re-run Analysis). Both `syncReleaseOnPRMerge`-triggered runs call `findExistingApplicationSubTickets` before either has written, both see the app as missing, and both create a second Ticket + SubTicket + mapping for the same application under the same release — the exact duplication this PR fixes for *sequential* re-runs resurfaces for *concurrent* ones.

Note the asymmetry: `release_change_types` got partial unique indexes + a `P2002` catch for this same race, but SubTickets got neither. Suggested fix: wrap `releaseService.release()` in `acquireLock`/`releaseLock` (`utils/distributedLock.ts`) keyed on the release ticket id — that also fixes the canvas race flagged in `commitAnalysisCanvas.ts`. A unique constraint + `P2002` catch here would work as a belt-and-braces alternative.

apps/backend/src/utils/commitAnalysisCanvas.ts
```diff
@@ -384,179 +415,373 @@ async function formatCommitAnalysisToBlockNote(
...
+        ? extractHotfixSection((existing.content as unknown as BlockNoteBlock[]) ?? [])
+        : [];
+      content = [...main.blocks, ...preservedHotfix];
+      mentionedUserIds = main.mentionedUserIds;
+    }
+
+    if (existing) {
+      const prevMeta = (existing.metadata as Record<string, unknown>) || {};
+      const prevMentions = Array.isArray(prevMeta.mentionedUserIds) ? (prevMeta.mentionedUserIds as string[]) : [];
+      await prisma.canvas.update({
```
devesh-juspay, line 629: **Race condition: canvas upsert is last-write-wins.** This is a non-atomic read-modify-write: read the canvas (line 597), splice the main/hotfix section into a new content blob in memory, then write the whole blob back (line 629) — no lock, no version check.

Failure scenario: two near-simultaneous PR merges each trigger `syncReleaseOnPRMerge` → `upsertCommitAnalysisCanvas({section: 'hotfix'})` for the same release. Both read the same pre-race content; whichever `prisma.canvas.update` lands last silently discards the other run's freshly-computed hotfix/env/migration blocks. The canvas then under-reports a merged hotfix PR until some future event happens to re-trigger a sync.

Suggested fix: the same release-ticket-scoped lock proposed in `applicationRepository.ts` covers this (both writes sit inside the same `release()` critical section). A lighter alternative is compare-and-swap — `update ... WHERE updatedAt = <read value>` and re-read + retry on 0 rows affected.

devesh-juspay (commented):
Two concurrency findings from a deeper review pass. Both share a root cause — this PR makes the release pipeline re-entrant (Re-run button + webhook hotfix sync) but concurrent runs for the same release ticket are never serialized — and both could be closed with one distributed lock around `releaseService.release()` keyed on the release ticket id (`utils/distributedLock.ts` already provides `acquireLock`/`releaseLock`, currently unused in the release flow).

packages/shared/src/zero/queries.ts
```diff
@@ -3406,9 +3423,12 @@ export const queries = defineQueries({
...
         .related('devTicket', q =>
-          q.one().related('pullRequests', pullRequests =>
-            pullRequests.orderBy('date', 'desc'),
-          ),
+          q
+            .one()
+            .related('pullRequests', pullRequests => pullRequests.orderBy('date', 'desc'))
+            .related('workflows')
+            .related('tags')
+            .related('formEntityValues'),
```
devesh-juspay, line 3431: **Query widening affects every consumer.** These three relations (`workflows`, `tags`, `formEntityValues`) are hydrated and live-synced for every dev ticket, for every viewer of the Testing tab, whether or not any added columns are selected — `formEntityValues` in particular is a full form-value bag per ticket. Only the "Add column" feature needs this data, and only for the columns actually selected.

Suggestion: keep this query as it was and fetch the column data through a separate query that the screen enables only when added columns are active (the same `enabled:` gating pattern already used for `getFormMappingsByContextIds` when `devTicketBoardIds.length > 0`). That keeps the default Testing-tab payload unchanged for the common case.

packages/shared/src/zero/queries.ts
```diff
@@ -3421,6 +3441,22 @@ export const queries = defineQueries({
...
+    z.object({ releaseId: z.string().min(1), limit: z.number().int().positive().max(500).optional() }),
+    ({ args: { releaseId, limit } }) => {
+      return zql.release_events
+        .where('releaseId', releaseId)
+        .orderBy('createdAt', 'desc')
+        .orderBy('id', 'desc')
+        .limit(limit ?? 200);
+    },
+  ),
+
```
devesh-juspay, line 3459: **This query syncs heavy payloads the Timeline never renders.** `release_events.payload` is included in the synced row, and in production the existing `FORM_SAVED` rows carry full env-file diffs (`oldValue`/`newValue`) and entire migration SQL bodies — we sampled prod: all 749 current rows are `FORM_SAVED`, some with multi-KB payloads, so opening the Timeline tab downloads up to 200 of these including diff bodies that may contain env values (a secrets-exposure vector, not just a perf one). The neighboring `releaseChangeFormValuesByReleaseId` already excludes changeLog rows for exactly this reason ("multi-KB diff bodies") — this query needs the same treatment.

Suggestions: (1) exclude `FORM_SAVED` events here — they're persistence bookkeeping, not release narrative, and the timeline UI has nothing useful to render for them; (2) don't sync `payload` to the client at all — the timeline renders only `message`/`eventType`/`eventName`/timestamps; (3) consider scoping by workspace as well, since this filters only by `releaseId` and event rows carry channel/conversation ids and user names.

devesh-juspay (commented):
Query-weight concerns on the Zero layer changes: Zero hydrates and live-syncs everything a query declares to every consumer, so widening a shared query or syncing large JSON columns taxes all viewers, not just the feature that needed the data. Please keep new/changed queries lean.

apps/backend/src/services/release/buildVcsClient.ts
sumantop: if (!baseUrl) {
      throw new Error('BITBUCKET_BASE_URL is not configured');
    }

apps/backend/scripts/release-manager-localdev/register-github-webhook.sh
sumantop: REPO="${1:-}"
  WORKSPACE_ID="${2:-}"

packages/shared/src/index.ts
sumantop: removed !!

apps/backend/src/routes/commitAnalysis.ts
sumantop: made /test-connection to requireAdminOrOwner 
404 (not 403) so we don't leak that the ticket exists.

apps/backend/prisma/migrations/20260707000000_add_art_is_hotfix/migration.sql
sumantop: removed !!

apps/backend/prisma/migrations/20260708000000_release_change_type_dedupe_index/migration.sql
sumantop: Zero filters out partial indexes by design, so our two partial-unique indexes are invisible to it and can't break sync. Zero won't mirror these into SQLite, so there's no client-side query benefit from them — but they exist purely for Postgres-side dedupe enforcement

packages/shared/src/zero/queries.ts
sumantop: `purose: fetches release_events for timeline ui
* (security) Timeline never renders payload and nothing in the code reads release_events.payload
* excluded FORM_SAVED (.where('eventName', '!=', 'FORM_SAVED')) in both copies. That's the heavy rows + only reader, so the exposure and over-fetch are fully closed. 
#2: Zero can't drop a single column from a synced row, so excluding those rows is the equivalent.
#3: already scoped — ReleaseEventsACL adds workspaceId to every read.`

packages/shared/src/zero/queries.ts
sumantop: three relations (workflows, tags, formEntityValues) only feed the optional "Add column" cells, not the default table — pullRequests is the only one the base view needs. Gated them behind an includeColumnData flag that the screen sets only when a relation-backed column is actually selected Default Testing-tab sync is now lean; the heavy formEntityValues bag syncs only when a custom column is added. Also applied to the CSV export.

PR #86, opened 2026-08-03
feat: XYNE-17871 control call setup from channel URL params

chaitak-gorai:

- dashboard

## Type of Change
- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Adds a **Call URL API**: query params on any channel URL that let an external system drive a call. Built for always-on room walls — a machine that holds one channel's call open and recovers on its own — but nothing in it is specific to that.

| Param | Values | Effect |
|---|---|---|
| `autoJoin` | `1` | Join the channel's active call, or start one; retry with backoff (3s → 30s cap) on disconnect |
| `mic` / `camera` | `on` / `off` | Initial media state, overriding the saved join preference |
| `telepresence` | `1` | Start the call already in presentation mode |

Param names are exported as `CALL_URL_PARAMS` so URL builders reference one definition — a typo fails silently, since unknown params are ignored.

**Which call it joins.** A channel can hold several `ACTIVE` calls at once, differing by origin: one `CHANNEL` huddle, one `CONVERSATION` call per thread, plus calendar-linked ones. Selection matches `callOrigin === CHANNEL` explicitly rather than excluding thread calls — the query is ordered `startedAt desc`, so "anything but a thread call" would follow a calendar call that started after the huddle and join the wrong one. This also matches what `findActiveCallByChannelId` converges on server-side, so client and backend agree on which call this is. Joining a *specific* call is a deliberate non-goal for now; the natural shape would be an optional `?callId=`.

The hook also handles a failure mode the room machine doesn't: a network-caused LiveKit disconnect leaves the machine reporting `connected` indefinitely, because it's designed to let LiveKit's own reconnection handle blips silently. Once those attempts exhaust, the machine is stuck `connected` with a dead room. This reads `room.state` directly and resets the machine so the retry loop can take over.

### Gating

Everything is behind a new CAC flag, `call_auto_join_config` (`{ enabled, allowedEmails }`), checked in **two** places:

1. **`useCallAutoJoin`** — stops a URL-driven join from starting at all.
2. **`roomMachine`, at the point of use** — `enableLocalTracks` re-checks the flag immediately before applying an override, so the machine never has to trust what the event handed it. A caller that sets `callUrlOverrides` without the flag gets the saved preferences, exactly as if it had asked for nothing.

It fails **closed** — if CAC is unreachable the client stays out of the call rather than joining unflagged. Turning the flag off stops new joins and retries; it does not end a call already in progress.

`?telepresence=1` is *additionally* subject to the existing `xyne_telepresence_config` gate, so it can't reach presentation mode for a user who isn't already allowed to click the button.

### Media state is passed into `roomMachine`, not toggled after connect

`enableLocalTracks` writes the initial track state asynchronously on entering `connected` — it's fired un-awaited and awaits `setMicrophoneEnabled` / `setCameraEnabled` inside. A post-connect compare-and-toggle reads the pre-connect value, sets it, and is then overwritten when those awaits resolve. With the shipped default `joinWithoutVideo: true`, `?camera=on` would reliably have ended up with the camera **off**.

Carrying the request on the join event keeps `enableLocalTracks` the single writer, so there is no window. Precedence: **host controls > URL request > saved preferences.** `null` — every ordinary join — behaves exactly as before.

It travels as **one nullable context field**, `callUrlOverrides: { mic?, camera?, presentation? } | null`, rather than a field per setting. Non-null doubles as "this join came from a URL", which is what suppresses the failure toast below — so there's no separate flag for that either.

### Two changes to existing presentation-mode behaviour — worth reviewing separately

These affect everyone who uses presentation mode, not just URL-driven calls:

1. **Alone in a call, presentation mode now shows your own tile** instead of "Waiting for remote participant…". A remote presenter still wins as soon as someone joins.
2. **`PresentationModeOverlay` skips the Fullscreen API when there's no user activation.** It previously called `requestFullscreen()` unconditionally; opened programmatically that always rejects and showed "Failed to enter presentation mode" for something that isn't broken. Click-driven opens have activation and are unaffected. Nothing is lost when skipped — the overlay is `fixed inset-0`, so it still covers the viewport.

**No `fullscreen` or `viewMode` param.** The DOM Fullscreen API needs user activation, which an unattended display has nobody to provide — filling the screen is the launcher's job (`--kiosk`), which also hides browser chrome the DOM API cannot. A layout param isn't needed either: `PresentationModeOverlay` renders fullscreen from *both* call layouts, so `?telepresence=1` covers the wall case on its own.

### Join failures during URL-driven joins are no longer toasted

`activeCallsInChannel` can return a call that ended between the sync snapshot and the join. The retry already recovers; the "Failed to join call" toast was the only damage, and an unattended display has nobody to dismiss it. The toast returns early when `callUrlOverrides` is non-null. **Manual joins still show the error exactly as before.**

## Areas Touched
- [ ] apps/backend (API / worker)
- [x] apps/dashboard
- [ ] apps/electron
- [ ] apps/xyne-claw / apps/xyne-claw-auth
- [ ] packages/shared or another shared package

11 files, +496 / −33. One new module (`utils/callUrlOverrides.ts`) owns the URL contract, the parsers and the non-React CAC gate — it has to be its own file rather than living in the hook, because `roomMachine` imports it and `useCallAutoJoin` imports `roomMachine`.

> **Note for reviewers:** this branch briefly contained Electron changes. They were removed once the wall became a browser rather than the desktop app — those files are back to their pre-branch state, so the net diff is dashboard-only and **no desktop build is required**. Squashing the branch removes that churn from the diff.

## Zero (Sync) Changes
- [x] No Zero changes — skip this section

## Schema & Data Changes

## Config, Environment & URLs

**A new CAC key must be created before this does anything:**

```
key:   call_auto_join_config
value: { "enabled": false, "allowedEmails": [] }
```

Ship it disabled. `allowedEmails` narrows rollout to specific accounts; an **empty list means every user in the workspace**, which is how it graduates to a general feature without enumerating accounts.

## API Contract
- [x] No backend API changes

New *client-side* contract only: the query params above.

## How did you test it?

test video
https://drive.google.com/file/d/1fvPW5LY6zntbhVX9y8mHtcUj1WUMfOia/view?usp=sharing

Full local stack from this branch (postgres, redis, LiveKit, zero-cache, transcription-agent, backend, dashboard), driven over CDP in Chrome and in Electron from source.

**Parameter matrix — all pass:**

| Params | Observed |
|---|---|
| `mic=on camera=on` | `enableMic:true enableCamera:true` — **with `joinWithoutVideo:true` in saved prefs**, i.e. the override correctly beat the preference |
| `mic=off camera=off` | Both false — override wins in the other direction too |
| `autoJoin=1` only | No media request → falls back to saved prefs (mic on, camera off) |
| no params | `autoJoin:false`, `enableLocalTracks` never runs, no call — complete no-op |
| full set, **flag off** | Params ignored entirely, no call. Kill switch holds |
| `telepresence=1` | `Call Active` + presentation overlay open, no fullscreen-failure message |
| *(control)* no `telepresence` | `Call Active`, overlay closed |

In Chrome kiosk: app booted, auto-joined, `POST /api/calls/initiate` → 200, LiveKit room created, machine reached `connected`, UI showed **"Call Active · 2 participants"**.

### Automated
- [ ] Backend unit tests
- [ ] Claw tests
- [ ] End-to-end suite
- [x] Not covered by committed tests — the dashboard has no test runner (no vitest/jest, zero test files). The verification above ran as standalone harnesses against the real modules; happy to land them if we want a runner in this package.

### Manual

**Users**
- [x] Single user
- [ ] Two users at once

**Login**
- [ ] Google
- [ ] Microsoft
- [x] **Email + password** (`POST /api/v2/auth/email/login`)

**Run mode**
- [x] Local dev
- [ ] Test mode / sandbox / prod-like local / Docker compose
- [x] Chrome (kiosk, windowed and headless/CDP) + Electron from source

**Sync & resilience** — n/a, no Zero changes

**Surface & data**
- [x] Web browser
- [x] Electron desktop
- [ ] Empty state / large data volume
- [x] Error paths — flag off, missing params, garbage param values, non-member, no media devices, stale/ended call row, CAC cache miss

## Checklist
- [ ] Backend typecheck + build pass — not run; no backend files touched
- [x] Dashboard `lint:errors-only` + typecheck pass (0 errors)
- [x] No new deps
- [ ] Commits follow `<type>: <TICKET-ID> <subject>`
- [x] Docs updated — `INFINITE_CALL_SOLUTION.md` / `STATION_MODE_SOLUTION_REPORT.md`; wall setup docs live in the `telepresence` repo, in a companion PR


ashutosh-002 (dismissed):

apps/dashboard/src/components/Call/CallViews/FullCallView.tsx
```diff
@@ -412,8 +414,16 @@ export function FullCallView({
...
 
+  // Who presentation mode puts on screen. Prefer a remote presenter — the point
+  // of a telepresence wall is the person at the other end — but fall back to the
+  // local participant when nobody else has joined yet, so the wall shows itself
+  // rather than an empty "waiting" screen. A wall sitting alone in a room is the
+  // normal resting state, not an error.
   const remoteParticipant = useMemo(
-    () => findRemotePresenter(participants, localParticipantId),
+    () =>
+      findRemotePresenter(participants, localParticipantId) ??
```
anujSingh06072025: @chaitak-gorai update the main function only:`findRemotePresenter` to fall back to local and also update the name please so its not misleading in future.

chaitak-gorai: Done updated.

ashutosh-002 (dismissed):

anujSingh06072025 (dismissed):

apps/dashboard/src/machines/roomMachine.ts
```diff
@@ -1564,6 +1618,14 @@ export const roomMachine = setup({
               zero: ({ event }) => (event.type === 'JOIN_CALL' ? (event.zero ?? null) : null),
               viewMode: ({ event }) =>
                 event.type === 'JOIN_CALL' && event.viewMode ? event.viewMode : ('mini' as const),
```
Mayankbansal-juspay, line 1605: check at actual media state with flag if enabled or not then only set these

Mayankbansal-juspay (approved):

apps/dashboard/src/machines/roomMachine.ts
chaitak-gorai: Updated changes.

ashutosh-002 (approved):

shubhamdhamija-juspay (approved):

PR #91, opened 2026-08-03
fix: XYNE-55036 surface error/retry state instead of infinite loader on chat list load failure

XyneSpaces:
1. **Problem Description:**
The channel chat list can get stuck forever on the "Loading conversations…" spinner. The initial conversations load in `ChatListV4.tsx` ends in an empty `.catch(() => {})`, and only the success path calls `setIsInitialLoadComplete(true)`. When the initial `zero.run(channelConversationsPaginatedV3, { type: 'complete' })` promise rejects, the render gate `if (!isInitialLoadComplete && cachedConversations.length === 0)` never releases, so the user sees an infinite loader with no error, no retry, and no telemetry — only a full page refresh recovers.

2. **How the issue was identified:**
Traced from ticket XYNE-55036. Confirmed via Grafana/view-syncer logs that `channelConversationsPaginatedV3` is failing server-side with a Zero IVM assertion `Error: Bound should be set` (in `@rocicorp/zero` `take.js` `#pushEditChange`), message `query pipeline failed`. With `{ type: 'complete' }`, Zero's `runImpl` does `resolve(Promise.reject(data))` on the view `error` event, so the rejected promise lands in the empty catch and the loading flag is never set. ~41 errors/day for this query, in bursts since 2026-07-31. The reactive cutoff-path (`useQuery`) has no error signal in the pinned Zero version, so on that path the same failure hangs silently instead of rejecting.

3. **Solution implemented in this PR (containment / UX fix):**
   - Replaced the empty `.catch(() => {})` on the initial (non-cutoff) load with: structured telemetry (`logger.error(Event.ZERO_RUN_ERROR, …)`), `setIsInitialLoadComplete(true)` so the render escapes the loading gate, an error toast, and a new `initialLoadError` state when there is nothing cached to show.
   - Added an `initialLoadError` + `<Retry>` render state (rendered ABOVE the empty-state gate so an error never masquerades as "No conversations in this channel yet"). Retry resets the load flags and bumps a `loadAttempt` counter that re-runs the initial-load effect.
   - Added a 15s watchdog `useEffect` (`INITIAL_LOAD_TIMEOUT_MS`) as a safety net for the reactive cutoff-path failure and the never-settles (offline) case — if the first load neither completes nor rejects within the window and nothing is cached, it surfaces the same error/retry state (`Event.ZERO_QUERY_FAILED`).
   - Existing cached conversations still render immediately; only the no-cache failure path shows the error/retry UI.

   Note: this is the containment fix. The upstream root cause (`Bound should be set`) is fixed in `@rocicorp/zero` ≥ 1.10 (rocicorp/mono #6189, #6121); that version is not released yet, so the Zero upgrade is tracked separately.

4. **Documentation:** N/A

5. **DB Alter Details:** N/A

6. **Data fix Details:** N/A

7. **Environment variable Changes:** N/A

8. **Testing:**
   - `tsc --noEmit --project tsconfig.app.json` passes.
   - `eslint` on the changed file: 0 errors (only pre-existing warnings in the file remain).
   - Manual reasoning: on a rejected initial `run()` with no cache → toast + error/retry state (no infinite spinner); with cache → cached list renders; success path unchanged; watchdog covers the silent cutoff-path hang and offline case.

9. **Services to which this change is to be deployed:** dashboard (frontend only).

10. **Main PR link (if against a release branch):** N/A — targets `main`.

### DIFF
--- a/apps/dashboard/src/components/Chat/ChatList/ChatListV4.tsx
+++ b/apps/dashboard/src/components/Chat/ChatList/ChatListV4.tsx
@@ -32,6 +32,8 @@ import { mutators } from '../../../zero/mutators';
 import { queryCacheActor, flushQueryCachePersistence } from '../../../machines/queryCacheMachine';
 import { browserPanelActor } from '../../../machines/browserPanelMachine';
 import LoadingAnimation from '../Loader/Loader';
+import { toast } from 'sonner';
+import { Event, logger } from '../../../utils/logger';
 import { getDraft } from '../../../hooks/useDraft';
 import { v4 as uuidv4 } from 'uuid';
 import { withProfiler } from '../../../utils/withProfiler';
@@ -312,6 +314,14 @@ const ChatListV4: React.FC<ChatListProps> = ({
   }, [conversations]);
 
   const [isInitialLoadComplete, setIsInitialLoadComplete] = useState(false);
+  const [initialLoadError, setInitialLoadError] = useState(false);
+  const [loadAttempt, setLoadAttempt] = useState(0);
+  const handleRetryInitialLoad = useCallback(() => {
+    setInitialLoadError(false);
+    setIsInitialLoadComplete(false);
+    lifecycleRef.current.initialLoadComplete = false;
+    setLoadAttempt(attempt => attempt + 1);
+  }, []);
   const isFetchingRef = useRef(false);
   const isFetchingOlderRef = useRef(false);
   // Set true when Zero returns older.length=0 (no items exist before the anchor).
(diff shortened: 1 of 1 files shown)


XyneSpaces:
---
**Updated (force-pushed to `f7f2bed`) — simplified per review feedback.**

Dropped the watchdog timeout and the Retry button. This PR now only handles the rejected imperative `zero.run({ type: 'complete' })` in the initial (non-cutoff) load:

- empty `.catch(() => {})` → `logger.error(Event.ZERO_RUN_ERROR, …)` + error toast + `setIsInitialLoadComplete(true)` (release the loading gate).
- No retry mechanism: the reactive channel `useQuery`s stay active and repopulate the list automatically once the backend recovers, so there's nothing to re-trigger on a manual retry.

Net change is `apps/dashboard/src/components/Chat/ChatList/ChatListV4.tsx` only (+20/−1). `tsc` and `eslint` (0 errors) pass. The original PR description above still mentions the watchdog/retry — please read this comment as the current scope.
---

XyneSpaces:
---
**Updated (force-pushed to `e599765`).** Re-added the retry flow for `zero.run` failures; kept the timeout/watchdog removed.

Scope now:
- Rejected imperative `zero.run({ type: 'complete' })` in the initial (non-cutoff) load → `logger.error(Event.ZERO_RUN_ERROR, …)`, release the loading gate, and (when nothing is cached) show an error toast + an inline **error state with a Retry button**.
- **Retry** resets the load flags and bumps a `loadAttempt` counter that re-runs the imperative initial load. The error state renders above the empty-state gate so a failure never shows as "No conversations in this channel yet".
- **No timeout-based failure detection** — there is no watchdog; only an actual `zero.run` rejection triggers the error/retry state. The reactive cutoff-path query is left untouched (it stays active and self-recovers).

Net change: `apps/dashboard/src/components/Chat/ChatList/ChatListV4.tsx` only (+53/−2). `tsc` and `eslint` (0 errors) pass. Supersedes my earlier two comments on scope.
---

santhoshJus (approved):

PR #122, opened 2026-08-04
fix: XYNE-54744 tenant-scope workflow execution endpoints; uniform login failure

devesh-juspay:
Follow-up to PR #47. Closes two controller authorization findings left open by the tenant-ACL work — both the same shape: a caller-supplied id resolved with no tenant predicate.

## What changed

**Workflow executions and steps (#56)** — nine handlers took `executionId`/`stepId` straight from `req.params` into `findUnique`-by-id finders, so a wrong-tenant id resolved and mutated. Two guards resolve ownership through `workflow.workspaceId` (non-null; the denormalized execution/step columns are nullable and used only as a fallback for rows predating stamping). They answer 404 rather than 403 so a response never confirms an id exists in another tenant.

`continueAgenticStep` and `restoreWorkflowExecution` additionally require the body-supplied `stepId` to belong to the authorized execution. Guarding the execution alone would have left the inverse open — driving another tenant's step from an execution of your own.

**Email login (#377)** — the distinct `403 "Password not set"` now answers exactly like a wrong password and an unknown address, closing the unauthenticated enumeration oracle. Verified no client keys off those strings.

## Scope reduced after review

Fixes for **email demerge (#96, #104)** and **ticket `updateFormField` (#51)** were drafted here and then dropped: **PR #19 already carries equivalent guards for both**, and neither is on `main`, so keeping a second version would have conflicted on merge. PR #19's versions are the ones to keep.

One thing PR #19 does not have, worth porting there separately: `updateFormField`'s route gate authorizes a channel supplied in the request body while the handler mutates a ticket reached by a different id. Tying the two together needs `validateChannelAccessForPost` to expose the channel it authorized.

## Verified already fixed on main, no code needed

Eight findings originally scoped here were closed already, each checked against the tree: **#81, #84, #147, #170, #184** (`join_session` and `add_channel_subscription` are commented out entirely; `subscribe_to_channels` authorizes every id; the `session:<id>` bypass via `subscribe_to_ticket_counts` is regex-closed), **#203** (`requireAdminOrOwner` on `/provision-org`), **#210** (`authorize('DATA_SOURCES', READ)` on `/query/preview`), and **#152** (the system-prompt concatenation no longer exists anywhere in the monorepo).

## Residuals

- #377 still has a timing difference: the passwordless branch skips the bcrypt compare.
- #84's DoS half (unbounded bulk subscription arrays) remains instrumented observe-only at threshold 500 — a separate needs-design item.

## Testing

Backend typecheck, backend build, gitleaks on staged changes, and schema-migration validation all pass. No schema changes.

### DIFF
--- a/apps/backend/src/controllers/emailAuthController.ts
+++ b/apps/backend/src/controllers/emailAuthController.ts
@@ -96,10 +96,9 @@ export class EmailAuthController {
       }
 
       if (!orgMember.passwordHash) {
-        // EMAIL user who was invited but password isn't set yet
-        res.status(403).json({
-          error: 'Password not set',
-          message: 'Please use forgot password to set your password.',
+        res.status(401).json({
+          error: 'Invalid credentials',
+          message: 'Email or password is incorrect',
         });
         return;
       }
--- a/apps/backend/src/controllers/workflowController.ts
+++ b/apps/backend/src/controllers/workflowController.ts
@@ -30,6 +30,72 @@ export class WorkflowController {
     this.workflowRepository = new WorkflowRepository();
   }
 
+  private async authorizeExecution(
+    req: Request,
+    res: Response,
+    executionId: string
+  ): Promise<boolean> {
+    const callerWorkspaceId = req.user?.workspaceId;
+    if (!callerWorkspaceId) {
+      res.status(401).json({ error: 'Authentication required' });
+      return false;
+    }
+
+    const prisma = DatabaseClient.getInstance();
+    const execution = await prisma.workflowExecution.findUnique({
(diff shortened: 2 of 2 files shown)


devesh-juspay:
Superseded by #170.

Both fixes from this PR — the workflow execution/step tenant guards and the uniform email-login failure response — have been cherry-picked into `fix/XYNE-54744-secops-complete` (#170), which is the branch going to main for open source.

Verified present there: 11 `authorizeExecution`/`authorizeStep` references in `workflowController.ts`, and the distinct `403 "Password not set"` on the login path is gone. Nothing from this branch is lost.

PR #169, opened 2026-08-04
refactor: XYNE-54744 share the desk channel-membership check across controllers

devesh-juspay:
Closes architectural finding **A9**.

## What

Three desk controllers — `deskMetrics`, `priorityClassification`, `aiRetrigger` — each carried a private copy of the same helper: look the channel up, confirm it belongs to the caller's workspace, then require channel participation.

That check is **not** redundant with `ChannelsACL`. The ACL grants PUBLIC-or-participant; these endpoints require participation unconditionally. Replacing it with a plain channel query would let any workspace member read desk metrics or change priority settings on any public desk channel. So the duplication is load-bearing, and a fourth copy drifting out of step is a realistic failure.

Extracted `assertChannelMembership()` and `assertDeskOwner()` into `utils/channelMembership.ts`. Net: 79 lines removed, 78 added, three copies down to one.

## Behaviour is identical — deliberately

Same `401 Authentication required`, `404 Channel not found`, `403 Not a member of this channel`, in the same order. Each controller keeps its own desk-owner message.

The recommendation's optimisation — collapse the two lookups into one participant-filtered `findFirst` — was **not** taken. It converts a non-member's 403 into a 404, and with no test suite there's nothing to catch a client that branches on that distinction. The saved query isn't worth it.

## Scope note

`emailClassificationController` is untouched. It has no channel gate on `main` at all; the gate it needs is added by the tenant-isolation PR (#19). Adding a second version here would conflict on merge.

A fifth variant in `deskMetricsController` was also left alone — it gates tag-generation config on "desk owner **or** channel admin", which is a genuinely different rule rather than a copy.

## Testing

Backend typecheck, backend build, gitleaks on staged changes, and schema-migration validation all pass. No schema changes, no behaviour change, no new endpoints.

### DIFF
--- a/apps/backend/src/controllers/aiRetriggerController.ts
+++ b/apps/backend/src/controllers/aiRetriggerController.ts
@@ -4,8 +4,7 @@
 
 import { Request, Response } from 'express';
 import { EmailClassificationRepository } from '../database/repositories/emailClassificationRepository.js';
-import { ChannelParticipantRepository } from '../database/repositories/channelParticipantRepository.js';
-import { ChannelRepository } from '../database/repositories/channelRepository.js';
+import { assertChannelMembership, assertDeskOwner } from '@/utils/channelMembership';
 import { EmailChannelPreferenceRepository } from '../database/repositories/emailChannelPreferenceRepository.js';
 import { DatabaseClient } from '../database/client.js';
 import { emailClassificationQueue } from '../queues/emailClassificationQueue.js';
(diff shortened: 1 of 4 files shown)


devesh-juspay:
Closing as redundant. This change is already in #170 — apps/backend/src/utils/channelMembership.ts is present there and assertChannelMembership is used across six controllers, one more than this PR touches. Nothing here is unmerged.

PR #174, opened 2026-08-05
compare: what the open-source secops branch adds over the internal branch

devesh-juspay:
Not for merge — opened so the delta between the two secops branches is reviewable in one place.

`fix/XYNE-54744-secops-complete` (PR #170) targets `main` and is the strict, open-source branch: everything is fixed there, with no relaxations. `fix/workspaceId-non-optional-rebased-v3` (PR #19) is the internal lineage, where several findings were deliberately left log-first so live product flows would not break.

Both branches carry the same ACL framework — tenant context, insert stamping, the Prisma extension, per-table ACLs mirrored across Prisma and Zero. Those commits appear twice in the range only because the open-source branch was rebased onto a newer `main`; by content they are the same work.

**What is genuinely new here** are the 14 `XYNE-54744` commits, plus 17 `main` commits the internal branch predates:

- fail closed across the tenant ACL layer, and the open route gaps
- the tenant backstop, call invite scoping, the native OAuth selector
- the desk channel-membership helper (refactor, behaviour identical)
- uninvited call joiners routed through the admit queue
- memory documents scoped to the caller's workspace
- the session reference read from the cookie only; deep-link route allowlist
- workflow execution endpoints tenant-scoped; uniform login failure
- Vespa query values bound; pairing required before an agent consent dialog
- the session cookie made a token; ops flags defaulted closed
- webhook signature verification defaulted to enforce
- automations required to be approved before they can be made live
- code comments stopped from documenting security weaknesses

Use the Files tab to read the delta; the commit list separates the framework from the new work.


devesh-juspay:
Closing. This existed to show the delta between the open-source branch and the internal branch when the two had drifted far apart. Both are now rebased onto main, #188 has merged the port into the internal branch, and PR #19 covers that branch against main directly — so this view no longer shows anything the other PRs do not.

PR #175, opened 2026-08-05
fix: XYNE-54744 scope memory documents and bind Vespa query values

devesh-juspay:
Lifted out of PR #170 so it can be exercised against a live index before it lands. Everything here depends on the shape of the index, not only on the code, and a green build proves none of it.

**What it does**
- Memory documents are scoped to the caller's workspace across read, update and delete. The four background paths that legitimately span workspaces declare it with an explicit sentinel instead of omitting the argument, so the compiler names any call site nobody considered.
- Vespa query values are bound rather than concatenated — `buildMemoryYql` returns the query with its parameters and every value goes through `bind()`.
- Workspace filtering on search defaults on.

**Before this merges**

1. **Backfill `workspaceId` on existing memory documents.** They predate the field. Scope them without the backfill and they stop matching — the symptom is a user's memory quietly emptying, not an error.
2. **Exercise the bound queries against a real index.** A binding that names a field the schema spells differently returns zero rows and reports success. This cannot be caught by a build or a typecheck.
3. **Confirm the search default.** Turning workspace filtering on is right, but it changes what existing queries return.

Closes #68, #98, #126, #181, #251, A17 once merged.

### DIFF
--- a/apps/backend/src/controllers/documentController.ts
+++ b/apps/backend/src/controllers/documentController.ts
@@ -6,7 +6,7 @@ import { config } from '@/config/env';
 import { DatabaseClient } from '@/database/client';
 import { documentIngestQueue } from '@/queues/documentIngestQueue';
 import { vespaKnowledgeIngestionService } from '@/services/vespaKnowledgeIngestionService';
-import { searchMemory, deleteMemory } from '@/services/memoryService';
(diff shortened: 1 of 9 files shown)


devesh-juspay:

PR #176, opened 2026-08-05
fix: XYNE-54744 read the session reference from the cookie only

devesh-juspay:
Lifted out of PR #170 so it can be observed before it lands.

The session reference was accepted from an `x-session-id` request header as well as from the cookie. A header travels through proxies, logs and referrers in ways a cookie with its flags does not, and any caller able to set one could present a session it had merely observed. The cookie is now the only source.

**Before this merges**

The middleware logs a warning whenever a request still carries the header (`authV2Middleware.ts:51`). Deploy, watch that line, and confirm it is quiet. Which clients still send it is not answerable from the code — a client that does will break on the day this merges, and the log is the only way to find out in advance.

One file, five lines. The risk is entirely in the rollout, not the change.

Closes #116's header half once merged.


devesh-juspay:

PR #177, opened 2026-08-05
fix: XYNE-54744 rate limiting, upload type restrictions, socket event thresholds

devesh-juspay:
Lifted out of PR #170 because these are not deployable as written. The protection is worth having; it needs to be made safe first.

**What it does**
- `generalLimiter` mounted on `/api` — it already existed but was commented out.
- `authLimiter` (20 per 15 minutes) on the sign-in and org-creation routes.
- Uploads refuse active content by MIME and by extension.
- Socket event and subscription budgets drop and reject rather than record.

**Before this merges**

1. **Set `trust proxy`.** Both limiters key on `req.ip` and nothing in `app.ts` configures proxy trust, so behind a load balancer every request carries the balancer's address. That makes the buckets global: 100 requests per five minutes for the whole deployment, and twenty sign-in attempts per fifteen minutes for every user combined. As written this takes the product down rather than protecting it.
2. **Derive the thresholds from production traffic.** 100 requests per five minutes is below what a single screen costs. The socket budgets are unmeasured too.
3. **Reconsider refusing `svg` and `xml`.** Both are ordinary attachments. The serving path already sends `X-Content-Type-Options: nosniff` and forces `application/octet-stream` + `attachment` for anything outside the inline-image allowlist, with a `sandbox` CSP on the inline path — so the upload filter is defence in depth here, not the only control.

Note that while this is out of PR #170, credential endpoints have no brute-force protection. That is the state `main` was already in, but item 1 is what unblocks putting it back.

Closes #278, #288, #383, #385, #390 once merged.

### DIFF
--- a/apps/backend/src/app.ts
+++ b/apps/backend/src/app.ts
@@ -4,7 +4,7 @@ import cors from 'cors';
 import helmet from 'helmet';
 import compression from 'compression';
 import cookieParser from 'cookie-parser';
-import { webhookLimiter } from '@/middleware/rateLimiters';
+import { webhookLimiter, generalLimiter, authLimiter } from '@/middleware/rateLimiters';
 import morgan from 'morgan';
 
 import { config } from '@/config/env';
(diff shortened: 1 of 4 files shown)


devesh-juspay:

PR #132, opened 2026-08-04
feat: XYNE-55227 Google Doc outlet connection

shubhamdhamija-juspay:
https://github.com/user-attachments/assets/15f8f937-5a7b-498d-a94a-b2a30fab1c58

Services-Requiring-Redeployment:
  - backend
  - dashboard

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/backend` (API / worker)
- [ ] `apps/dashboard`
- [ ] `apps/electron`
- [ ] `apps/xyne-claw` / `apps/xyne-claw-auth`
- [ ] `packages/shared` or another shared package
- [ ] Infra (docker, scripts, nix, CI)

---

## Zero (Sync) Changes

- [ ] No Zero changes — **skip this section**
- [ ] Adds/modifies a **mutator**
- [ ] Adds/modifies the **schema** (tables, columns, relationships)
- [ ] Adds/modifies a **query or ACL**

If any of the above:

- [ ] Server (`apps/backend/src/zero/mutators.ts`) and client (`apps/dashboard/src/zero/mutators.ts`) mutators mirror each other — same namespace, name, and args schema
- [ ] No `Date.now()` / `uuid()` generated inside a mutator
- [ ] Optimistic client result matches the server result (no state flash on rebase)
- [ ] New tables exported in `createSchema()` and given a Query ACL registered in `query-acl-factory.ts`
- [ ] Matching Prisma model + migration, client regenerated

## Schema & Data Changes

- [ ] No schema changes — **skip this section**
- [ ] Prisma schema modified (`apps/backend/prisma` and/or `prisma-common`)
- [ ] Migration added
- [ ] Backfill / data migration included

If any of the above:

- [ ] Clients regenerated (`db:generate`, `db:common:generate`)
- [ ] No new Postgres enum or enum value — enums are frozen (`pnpm run test:enum`)
- [ ] Backward compatible with deployed code, or rollout order noted below
- [ ] Backfill is idempotent

## Config, Environment & URLs

- [ ] No config changes — **skip this section**
- [ ] Env variable added / renamed / removed
- [ ] **Base URL, host, port, or origin changed** (API, Zero, Vespa, Claw, LiveKit, LiteLLM, …)
- [ ] CORS / allowed origins / OAuth redirect URIs changed
- [ ] Feature flag or runtime config changed

[...]

PR #179, opened 2026-08-05
fix: XYNE-54744 port the open-source secops work to the internal branch

devesh-juspay:
The secops work from PR #170, applied to this branch. Six commits, 34 files — the reviewable version of PR #174, which renders 627 files only because its base predates 18 `main` commits and GitHub diffs from the merge base.

All six cherry-picked without conflict.

**What is here**

| | |
|---|---|
| `099e03501` | `workspaceId` immutable; guest writes derive from guest reads |
| `2728c94a5` | call access decided by membership |
| `5f859af83` | shared desk channel-membership check |
| `13b4b91b9` | session credential and login surfaces |
| `5f655016f` | operational flags default closed |
| `5b7582741` | comments state the rule, not the gap |

**What is deliberately not here**

Two rules from the open-source branch's ACL commit are held back, because unlike everything above they change how paths that work today behave:

- **Failing closed when no workspace resolves.** On this branch a request whose context carries no workspace still runs and logs. Enabling the refusal turns every such path into a 500 — loud and recoverable, but it needs the paths found first.
- **The backstop that scopes a table whose own ACL clause comes back unrestricted.** This one fails *silently*: if a table legitimately spans workspaces, rows stop appearing with no error at the call site. That is the harder failure to notice, and the reason it waits.

Both remain on `fix/XYNE-54744-secops-complete`. The split cost little coverage — 475 insertions there against 454 here.

Closes A3, A4, A9, A12, #19, #47, #101, #116, #132, #140, #154, #158, #162, #167, #178, #183, #204, #208, #212, #213, #233, #367, #370, #371, #398, #405 on this branch. A1 and A11 stay open here by design.

### DIFF
--- a/apps/backend/src/controllers/aiRetriggerController.ts
+++ b/apps/backend/src/controllers/aiRetriggerController.ts
@@ -4,8 +4,7 @@
 
 import { Request, Response } from 'express';
 import { EmailClassificationRepository } from '../database/repositories/emailClassificationRepository.js';
-import { ChannelParticipantRepository } from '../database/repositories/channelParticipantRepository.js';
-import { ChannelRepository } from '../database/repositories/channelRepository.js';
+import { assertChannelMembership, assertDeskOwner } from '@/utils/channelMembership';
 import { EmailChannelPreferenceRepository } from '../database/repositories/emailChannelPreferenceRepository.js';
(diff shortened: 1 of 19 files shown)


devesh-juspay:

PR #190, opened 2026-08-05
feat: add reassignOnUnavailable feature to user groups for automatic …

Revanthvenkatpasupuleti:
Added an option to reassign a user’s open tickets when they become unavailable.
If the option is off, their existing tickets stay assigned, but they won’t receive new tickets
image


apps/backend/prisma/migrations/20260805120000_add_reassign_on_unavailable/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "public"."user_groups" ADD COLUMN     "reassignOnUnavailable" BOOLEAN NOT NULL DEFAULT false;
```
PiyushJuspay: remove default

packages/shared/src/zero/mutators.ts
```diff
@@ -7510,6 +7510,25 @@ export const mutators = defineMutators({
         });
       },
     ),
+    toggleGroupReassignOnUnavailable: defineMutator(
```
PiyushJuspay: use existing update mutator for userGroup updates ?

apps/dashboard/src/components/UserGroup/AssignmentConfigScreen/AssignmentConfigScreen.tsx
```diff
@@ -587,6 +593,20 @@ export const AssignmentConfigScreen = ({
         );
       }
 
+      // Also update reassign-on-unavailable setting
+      const reassignOnUnavailableChanged =
+        localReassignOnUnavailable !== (userGroup?.reassignOnUnavailable ?? false);
+
+      if (reassignOnUnavailableChanged) {
+        void zero.mutate(
```
PiyushJuspay: consume mutator errors and show error as toast in UI

apps/backend/src/queues/ticketReassignmentQueue.ts
```diff
@@ -0,0 +1,206 @@
...
+
+const prisma = DatabaseClient.getInstance();
+
+export interface TicketReassignmentJobData {
+  type: 'reassign-user-tickets';
+  userId: string;
+  userGroupId: string;
+}
+
+const OPEN_STATUSES: string[] = [TicketStatusV2.TODO, TicketStatusV2.STARTED, TicketStatusV2.PAUSED];
```
PiyushJuspay: check if we consider PAUSED as open or not ?

rahul1841: yes it is open only

apps/backend/prisma/migrations/20260805120000_add_reassign_on_unavailable/migration.sql
rahul1841: This migration will run against existing user-group rows, and removing it while keeping NOT NULL will make the migration fail

PiyushJuspay:
Where is the worker code ? only queue is defined ?

apps/backend/src/queues/ticketReassignmentQueue.ts
```diff
@@ -0,0 +1,206 @@
...
+          scanned++;
+          try {
+            const result = await evaluateAssignmentRule(
+              userGroupId,
+              ticket.boardId,
+              AssignmentType.TICKET_ASSIGNEE,
+              undefined,
+              ticket.projectId,
+              ticket.channelId,
+            );
```
PiyushJuspay, line 148: need to update userWorkload here as well ?

apps/backend/src/queues/ticketReassignmentQueue.ts
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        for (const ticket of tickets) {
+          scanned++;
+          try {
+            const result = await evaluateAssignmentRule(
+              userGroupId,
+              ticket.boardId,
+              AssignmentType.TICKET_ASSIGNEE,
+              undefined,
```
PiyushJuspay: user should not be considered again for assignment

PiyushJuspay:
H3 performSave partial-failure leaves non-idempotent state
- What: The refactored save now awaits multiple independent Zero mutations via Promise.all and throws if any fails. batchUpdate generates new UUIDs via uuidv4() on each call, so a retry after a partial failure creates duplicate assignment-state and expertise-mapping rows.
- Where: apps/dashboard/src/components/UserGroup/AssignmentConfigScreen/AssignmentConfigScreen.tsx:580-600 —
const serverResults = await Promise.all(pendingServerResults);
const failedResult = serverResults.find(result => result.type === 'error');
if (failedResult?.type === 'error') {
  throw new Error(failedResult.error.message || 'Failed to save...');
}
- Who affected: Admins saving assignment configuration when one of the three mutations (batchUpdate, toggleRotation, userGroup.update) succeeds but another fails.
- Impact: Retry creates duplicate records. The old fire-and-forget code had the same idempotency flaw but never surfaced errors, making retries unlikely. The new error-surfacing makes retries more probable.
- Fix: Either make batchUpdate idempotent (upsert by natural key instead of new UUID), or save mutations sequentially and track which succeeded so retry skips them.

rahul1841:
> H3 performSave partial-failure leaves non-idempotent state
> 
> * What: The refactored save now awaits multiple independent Zero mutations via Promise.all and throws if any fails. batchUpdate generates new UUIDs via uuidv4() on each call, so a retry after a partial failure creates duplicate assignment-state and expertise-mapping rows.
> * Where: apps/dashboard/src/components/UserGroup/AssignmentConfigScreen/AssignmentConfigScreen.tsx:580-600 —
>   const serverResults = await Promise.all(pendingServerResults);
>   const failedResult = serverResults.find(result => result.type === 'error');
>   if (failedResult?.type === 'error') {
>   throw new Error(failedResult.error.message || 'Failed to save...');
>   }
> * Who affected: Admins saving assignment configuration when one of the three mutations (batchUpdate, toggleRotation, userGroup.update) succeeds but another fails.
> * Impact: Retry creates duplicate records. The old fire-and-forget code had the same idempotency flaw but never surfaced errors, making retries unlikely. The new error-surfacing makes retries more probable.
> * Fix: Either make batchUpdate idempotent (upsert by natural key instead of new UUID), or save mutations sequentially and track which succeeded so retry skips them.

batchUpdate reuses existing records by natural key, and unique constraints prevent duplicates on retry

apps/backend/src/queues/ticketReassignmentQueue.ts
```diff
@@ -0,0 +1,212 @@
...
+              ticket.boardId,
+              systemActorId,
+            );
+            reassigned++;
+          } catch (error) {
+            logger.error(`❌ [TICKET-REASSIGNMENT] Failed to reassign ticket ${ticket.id}:`, error);
+          }
+        }
+
+        cursor = tickets[tickets.length - 1].id;
```
Navviiii: @rahul1841 Cursor Pagination on Modified DataIn ticketReassignmentQueue.ts, the queue processor paginates through the user's tickets using cursor-based pagination:


const tickets = await prisma.ticket.findMany({
  where: {
    assignedTo: userId,
    userGroupId,
    isArchived: false,
    statusV2: { in: OPEN_STATUSES },
  },
  take: BATCH_SIZE,
  ...(cursor ? { cursor: { id: cursor }, skip: 1 } : {}),
});

The Problem:
Prisma Cursor Requirements: In Prisma, the record specified in the cursor object must satisfy the query's where clause.
Data Modification Mid-Loop: When a ticket is successfully reassigned, its assignedTo field is updated to the new assignee.
The Crash: In the next iteration, the query runs with cursor: { id: previousTicketId }. However, because previousTicketId is no longer assigned to userId, it does not match the where filter. Prisma will immediately throw a P2025 error ("Record to navigate to with cursor not found"), crashing the job.
Infinite Loops on Failure: If a ticket cannot be reassigned (no eligible replacement), it remains assigned to userId. If we didn't use a cursor and just queried the first 50, we would fetch that same failed ticket over and over, resulting in an infinite loop.

Navviiii (approved):

karunrajk (approved):

PR #248, opened 2026-08-06
fix: XYNE-54744 refuse executable and script uploads

devesh-juspay:
Refuses executable and script uploads at the multer layer.

Answers the file-upload finding raised in both mobile pentest reports — PY-JP-008 (iOS) and PY-JP-010 (Android) — which observed that no file type or content validation ran on `/api/drafts/attachments/upload` or `/api/tickets`.

### What it does

Adds a `fileFilter` to every upload configuration (`uploadConfig`, `createUploadStreamConfig`, `collectionUpload`, `versionUpload`), refusing:

- **Extensions** `.exe .dll .msi .scr .bat .cmd .ps1 .php .phtml`
- **MIME** `application/x-msdownload`, `application/x-msdos-program`, `application/x-httpd-php`

Extension is the primary key because `file.mimetype` is client-supplied, and ~14% of real uploads arrive as `application/octet-stream` because the client could not determine a type.

### What it deliberately does not refuse

The list was checked against production upload history before being fixed:

| Type | Volume | Why it stays |
|---|---|---|
| `.apk` / `.ipa` | ~2,300 / ~480, active daily | Internal mobile build distribution across 4 workspaces |
| `.sh` / `.bash` | ~340 | Routine ops scripts |
| `.com` | 6 | All are Outlook inline-image content-ids (`inline-…@namprd18.prod.outlook.com`), not executables |
| `.html` / `.svg` | ~67k / ~590 | Automated reports and logos; emoji are SVG |

Every extension that remains blocked has no legitimate use in the corpus — the `.php`, `.exe` and `.bat` files present are the pentest artifacts themselves (`shell.php`, `1mb.exe`, `sample2.bat`).

### Behaviour on refusal

The file is **skipped**, not errored. Returning an error to multer discards every file in the same multipart request, which on the inbound-email path would drop the message and create no ticket. Each rejection logs `[UPLOAD] Rejected file type` with filename and MIME.

A skipped file is silently absent from `req.files`. If we want the user told, `draftAttachmentController` can compare against `attachmentIds` and return the skipped names — happy to add that here or follow up.

### Why this is defence in depth, not the primary control

Markup and images stay permitted because the serving path already governs them: `setSafeDownloadHeaders` types non-inline content as an opaque `application/octet-stream` download with `nosniff`, and `setSafeInlineImageHeaders` renders images under a script-blocking CSP. Uploads also go to object storage, not a script-enabled directory, so there is no server-side execution path.

### Testing

- 17 cases against real production filenames — all pass
- Backend typecheck: 61 errors before and after, an identical pre-existing set (unbuilt `@xyne/storage`, `mutators.ts`, `mapper.ts`). No new errors. The repo does not currently typecheck clean on `main`, so the pre-commit hook was bypassed and its checks run manually: gitleaks clean, change-analysis clean, schema-migration validation reports no schema changes.

### Deploy notes

No migration, no config. Backend redeploy only.


shubhamdhamija-juspay (approved):

gyash1512 (approved):

devesh-juspay:
Superseded by #289, which combines this with #286 and is rebased onto current main. The commit is carried over unchanged (cherry-picked as 75b6e33f4).

PR #286, opened 2026-08-06
fix: XYNE-54744 report guest writes and resource-level-only grants

devesh-juspay:
Adds the two missing detections in the ACL layer. **Reports only — nothing is refused.**

### Why

Several tenant-isolation gaps are carried deliberately, on the basis that we watch them on real traffic before deciding whether enforcing would break a legitimate caller. Auditing what is actually deployed on main turned up 11 `[acl]` / observe log points — but two tracked issues had **no detection at all**, so the "watch it" instruction had nothing to watch:

- **A12** — guest writes are not narrowed. ~30 table ACLs narrow `getWhereClause` for a guest; none narrows the mutate side.
- **#158** — the ACL middleware keys its grant on the first URL segment plus the HTTP method, and never inspects the object named later in the path.

### What this adds

| Log line | Issue | Emitted at |
|---|---|---|
| `[acl] guest write not narrowed` | A12 | `database/tenant/acl-extension.ts` |
| `[acl] granted on resource level only` | #158 | `middleware/acl.ts` |

Both use the same dedupe-and-cap approach as the existing `[acl]` reports — keyed on shape, skipped once seen, capped at 500 distinct keys per process — so neither can become a hot path.

The middleware report collapses identifiers to `:id` (cuid, uuid, numeric and nanoid-style all recognised) so a route reports once rather than once per object. Verified with 10 cases: collection routes and plain paths such as `/api/auth/login-workspace` and `/api/drafts/attachments/upload` correctly report nothing, and three distinct ticket ids collapse to a single report.

### Tracker

`#158` was carrying verdict **Logs Added**, which the code did not support — `middleware/acl.ts` had only generic "ACL denied access" logging. Corrected to **Needs Design** in the internal tracker, with the reasoning recorded. Combined counts move to 23 Needs Design / 5 Logs Added; open total is unchanged at 44.

A **Logs to Monitor** tab was added to the tracker listing all 11 log lines, the issue each maps to, whether that issue is open or fixed, and the exit criterion for turning each report into a refusal.

### One thing worth knowing

These reports are **deduplicated, not counted**. A line fires once per shape per process lifetime. The guidance recorded against several findings — *"watch it on real traffic until it is quiet"* — is therefore misleading: it goes quiet on its own. Treat the output as a distinct-shape inventory collected across a deploy cycle. Volume would need a counter metric.

### Testing

Backend typecheck: **71 errors before and after**, an identical pre-existing set. No new errors. The repo does not typecheck clean on `main`, so the pre-commit hook was bypassed and its checks run manually — gitleaks clean, change-analysis backend-only, schema-migration validation reports no schema changes.

No migration, no config. Backend redeploy only.


devesh-juspay:
Superseded by #289, which combines this with #248 and is rebased onto current main. The commit is carried over unchanged (cherry-picked as 314ef4075).

PR #289, opened 2026-08-06
fix: XYNE-54744 workspace entry, upload filtering, and response scoping

devesh-juspay:
Security hardening across workspace entry, uploads, response scoping and container execution. Each change is independent; scope is deliberately narrow.

### Workspace entry

Entry is decided by an invitation to *that* workspace, or an approved join request for it. Holding an account in another workspace of the same organisation is no longer itself entry to this one — the organisation remains the boundary between customers and is enforced separately.

The invitation must also be **live and unconsumed**:

- **Expiry.** Revoking an invitation sets its expiry to the moment of revocation, so one comparison covers both a lapsed invitation and a withdrawn one. Only the acceptance path checked this before; reaching workspace login directly bypassed it, so a revoked invitation still granted entry.
- **Single use.** An already-accepted invitation is not a second entry. Whoever used it holds an account in the workspace and is resolved earlier.

**Existing members are unaffected.** They resolve on provider identity and return before this decision is reached; it governs only the creation of a new workspace user. Verified against production data: no user loses access, and the invitations this newly refuses belong to people who already hold accounts in the workspace concerned.

The email-keyed lookup above this remains commented out. Email is mutable in both Google Workspace and Microsoft Entra, so it cannot stand as an identity assertion — matching on it would introduce an account-takeover path.

### Uploads

A `fileFilter` on the shared upload middleware, and therefore on every upload route, refuses Windows executables and server-side scripts by extension and by declared MIME type. Extension is checked first, since the declared type is client-supplied.

A refused file is skipped rather than failing the whole request: rejecting to multer discards every file in the same multipart upload, which on the inbound-email path would drop the message.

### Response field scoping

**User search** — both response paths now share one field list. The paginated path previously returned the persistence row, and the mapper below it was unreachable. The list is deliberately broad so existing consumers are unaffected; it holds back `providerUserId` and `metadata`. Known consumers are `xyne-claw-auth` mention lookup and `apps/legacy`, which read `id`, `name`, `email`, `status`.

**Workflow steps** — the list route no longer returns step payloads, which are free-form and may carry intermediate execution state. They remain available on the by-id route.

**People and transcript search** — constrained to the caller's workspace.

### Mutation rules

`activities` and `saved_user_configuration_values` had mutation rules with no condition. Activity rows must now name the caller's workspace and concern a member of it; saved-view values must belong to a view the caller owns, since mutators address those rows by their own id rather than through the parent.

### Container and configuration

**Python agent** — runs as an unprivileged user, created before the model download so the caches under `/app` are owned by the process that updates them. Worth confirming the deployment mounts no volumes; if it does, the pod needs `fsGroup: 1001`.

**LiveKit** — configuration validation rejects the published development key pair, so a deployment cannot start while carrying it. Empty stays valid; deployments without calls do not configure LiveKit.

### Not in this PR

Tenant scoping inside transactions, the ACL extension rework and its enforcement switches, socket throttling, and the message and conversation-participant mutation rules have moved to `fix/XYNE-54744-tenant-scoping-and-observability`, to be merged after they have been exercised. Those message rules narrowed system-message update and delete in ways that refused the update ending a call and left channel-wide notices undeletable, so `messages-acl.ts` here is unchanged from `main`.

### Verification

`tsc --noEmit` reports the same errors as `main`, with the same set — all pre-existing generated-client failures. Rebased on `main`, no conflicts.


anuragdvd (approved):

anuragdvd (dismissed):

apps/backend/python-agent/Dockerfile
```diff
@@ -45,6 +45,11 @@ RUN case "$DIARIZATION_ENABLED" in \
 # Copy application code
 COPY . .
 
+# Run as an unprivileged user. Created before the model download so the caches under /app
+# are owned by the account that reads and updates them at runtime.
```
devesh-juspay, line 49: @ishaanrawat-netizen

ishaanrawat-netizen: Looks good

ishaanrawat-netizen (dismissed):

Mohan0544 (approved):

ishaanrawat-netizen (approved):

gyash1512 (approved):

PR #307, opened 2026-08-07
fix: XYNE-55381 analytics query fix

abhinavjuspay:
  - backend

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

**Description**
 The date filter was written as an optional-filter idiom:
 AND (${gte}::timestamp IS NULL OR m."createdAt" >= ${gte}::timestamp)

That predicate is non-sargable. Postgres can only fold the IS NULL branch away while it builds custom plans; once a pooled connection's prepared statement flips to a generic plan, the OR has to stay, the (msgType, createdAt) index becomes unusable, and it sequential-scans the entire messages table.

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist

### DIFF
--- a/apps/backend/src/database/repositories/analyticsRepository.ts
+++ b/apps/backend/src/database/repositories/analyticsRepository.ts
@@ -114,7 +114,6 @@ type FilteredMessage = {
   messageId: string;
   senderId: string;
   conversationId: string;
-  metadata: any;
   createdAt: Date;
 };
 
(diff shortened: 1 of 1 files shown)


apps/backend/src/database/repositories/analyticsRepository.ts
```diff
@@ -200,18 +199,49 @@ export class AnalyticsRepository {
...
+  private static readonly inFlightMessageQueries = new Map<string, Promise<FilteredMessage[]>>();
+
   private async getFilteredMessages(dateCondition: { gte?: Date; lte?: Date }, workspaceId: string): Promise<FilteredMessage[]> {
-    const excludedChannels = AnalyticsRepository.EXCLUDED_CHANNEL_IDS;
+    const scopedWorkspaceId = this.requireWorkspaceId(workspaceId);
     const gte = dateCondition.gte ? dateCondition.gte.toISOString() : null;
     const lte = dateCondition.lte ? dateCondition.lte.toISOString() : null;
-    const scopedWorkspaceId = this.requireWorkspaceId(workspaceId);
+
+    const key = `${scopedWorkspaceId}|${gte ?? ''}|${lte ?? ''}`;
```
XyneSpaces, line 225: [High] This coalescing key uses millisecond-precision timestamps, so it will almost never dedupe the concurrent scans it targets. `getDateFilter()` computes `const now = new Date()` and returns `lte: now` (and a `gte` derived from the same `now`) on every request; each of the ~8 analytics panels calls it independently during one dashboard mount, so their gte/lte differ by milliseconds and `.toISOString()` bakes that into the key. Two concurrent panels therefore produce different keys and `inFlightMessageQueries.get(key)` misses — the dedup is effectively dead for the exact 'one render fires ~8 copies' case the comment describes. Fix: quantize the range to a coarse bucket (e.g. floor `now` to the current minute) in `getDateFilter` so concurrent panels compute an identical gte/lte — that also stabilizes the prepared-statement/plan reuse this PR is trying to fix.

apps/backend/src/database/repositories/analyticsRepository.ts
```diff
@@ -200,18 +199,49 @@ export class AnalyticsRepository {
...
+    const inFlight = AnalyticsRepository.inFlightMessageQueries.get(key);
```
XyneSpaces, line 226: [Nit] All coalesced callers receive the same resolved `FilteredMessage[]` instance. Callers today only read (`.map`/`.filter`/`new Set`/`.length`), so it's safe now, but any future in-place mutation (`.sort`/`.splice`/`.push`) on the returned array would corrupt every other concurrent caller. Add a one-line 'treat result as read-only' contract comment, or return a shallow copy / `Object.freeze`.

XyneSpaces (commented):
1 high, 1 nit. Changes requested. The SQL fix is correct — replacing the non-sargable `(IS NULL OR ...)` date predicate with conditional `Prisma.sql` clauses restores index usage and is the right call, and dropping the unused `metadata: any` column is safe (no caller reads `.metadata`; verified repo-wide). The blocker is that the request-coalescing added alongside it is keyed on millisecond-precision gte/lte, so it won't actually collapse the parallel dashboard scans it's designed for — see inline. Quantize the date range so the key is stable, then this is good to merge.

anuragdvd (approved):

shubhamdhamija-juspay (approved):

PR #325, opened 2026-08-07
feat: XYNE-55328 Spaces SDK

PrajwalPrasad2212:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

**XYNE-55328 / XYNE-55505** — Adds `@xyne/spaces-sdk`: a TypeScript SDK giving external applications programmatic access to Spaces, backed by a new public API at `/api/sdk`.

Two new packages plus a backend API surface and a dashboard screen:

- **`packages/xyne-spaces-sdk`** — 472 methods across 25 resources, covering the complete Zero operation catalog (263 queries + 245 mutators). Callers get `sdk.channels.list()`, `sdk.tickets.update()`, `sdk.messages.send()` rather than raw operation names and hand-built payloads. Zero runtime dependencies; runs in Node 18.17+ and in the browser.
- **`packages/xyne-spaces-contract`** — the small set of facts both sides must agree on: the error catalog with HTTP status and retryability, and the search request schema.
- **`apps/backend/src/api/sdk`** — the public API: API-key authentication, catalog query/mutate endpoints, and the operations that have no catalog equivalent (search, uploads, server-side allocation, Claw).
- **`apps/dashboard`** — an **Apps → API Keys** screen where a user mints, inspects, and revokes their own keys.

The whole surface is off by default (`SDK_API_ENABLED`) and `/api/sdk` returns 404 until enabled.

`api/sdk` is ~1,000 lines across 8 files. That is deliberate: it authenticates a request and delegates. There is no business logic in it — no ACL, no search ranking, no sequence allocation, no indexing. A request through `/api/sdk` and the same action taken in the app converge on the same function within a few frames of stack.

**Design notes worth a reviewer's attention:**

- **Authentication is an API key that carries its own identity.** `xyne_sk_` + base64url of an AES-256-CBC blob holding `{ sub, email, name, workspaceId, orgId, memberId }`. `role` and `orgRole` are deliberately excluded and re-read from `users`/`org_members` on every request, so a demotion or deactivation takes effect immediately rather than when someone remembers to delete the key. Expiry lives on the row for the same reason.
- **Integrity comes from the database row, not the cipher.** `encryptionService` is AES-CBC with no MAC, so ciphertext is malleable and a successful decrypt proves nothing on its own. Authentication is the exact-match lookup of the presented key against `sdk_api_keys.token`, which is `@unique` — a forged or tampered key matches no row and is rejected before any field inside it is trusted. That same lookup is the revocation point. Reviewers should read this as the load-bearing claim of the auth design.
- **Authorization is the Zero ACL and nothing else.** A key acts as its user, and the per-table ACL folded into every query AST and every wrapped transaction decides what that user may reach. There is deliberately no second permission model layered on top: it would be a parallel system to keep correct, and the first one is what actually guards the rows.
- **Keys are workspace-scoped by construction, not by policy.** A `User` row is itself scoped to one workspace — somebody with access to two holds two user rows with different ids — so `sub` already determines the workspace, and Zero's `Context` carries exactly one. A second workspace needs its own key. A user may hold 2 live keys, valid 30 days.
- **Execution is shared, not reimplemented.** Reads call `runCatalogQuery` and writes call `runCatalogMutation` in `src/zero/server.ts` — the same functions the app's own query and mutate fallbacks use. Read and write ACL, Vespa indexing, and side-effect jobs therefore behave identically to the product. This matters more than it sounds: a separate implementation of the write path is one `wrapTransactionWithACL` argument away from silently diverging on field-level encryption or side-effect routing, and there is no test that would catch it. One call site makes that class of drift impossible.
- **Coverage is enforced, not asserted.** `npm run coverage` fails unless every catalog operation is either exposed or listed in `exclusions.json` with a reason. A new backend mutator breaks the build until someone decides which — that is the mechanism keeping "complete" true over time. It also verifies that required arguments are supplied, that no undeclared argument is sent, that enumerated values match exactly, and that a query's single-row-vs-list shape matches the declared return type. None of that is reachable by TypeScript, because the registries reference operations by string.
- **Claw is relayed rather than reached.** Xyne Claw is a separate service whose verifier accepts only its own credentials. Rather than making callers hold two, the backend relays through `clawAgentService` with the deployment's service credential. `runS2SClawAgent` is used rather than `runClawAgent` because it takes an explicit identity that maps one-to-one onto `AuthData` and returns a pollable session id; the app-mention path returns only whether it dispatched.
- **`#zero-internal` imports are version-locked.** The read path compiles ZQL to SQL using private `@rocicorp/zero` modules. `src/zero/zero-internals.test.ts` pins the verified version and asserts every declared private path still resolves, so a dependency bump fails in CI rather than at runtime on a live request. The check is filesystem-based rather than import-based on purpose: Jest resolves neither `#` subpath imports nor ESM inside `node_modules`, so an import-based version cannot run at all — it fails to parse, which reads as a broken suite rather than a passing guard.

## Areas Touched

- [x] `apps/dashboard`
- [ ] `apps/electron`

`apps/dashboard` gains an **API Keys** button in the Apps header, an `ApiKeysModal` component, and a service; no other screen changes. New packages: `xyne-spaces-sdk`, `xyne-spaces-contract`. Infra: `pnpm-workspace.yaml`, `apps/backend/.gitignore`, and the backend env files.

---

## Zero (Sync) Changes

No mutators, queries, or ACL rules were added or modified, and no sync behaviour changes. `src/zero/server.ts` was restructured, and reviewers of that file should look at it directly:

The bodies of `handleQueriesFallback` and `handleMutateFallback` were lifted into two new exports, `runCatalogQuery(name, args, ctx, provider)` and `runCatalogMutation(name, args, authData)`. Both existing handlers keep their signatures, their return shapes, and their behaviour — they are now thin wrappers that resolve auth, parse the request, and call the extracted function. The Electron and dashboard fallback paths are unaffected.

The mutation extract covers `createMutators` → `wrapTransactionWithACL` → `mustGetMutator` → `mutator.fn`, then the full post-commit drain: awaited tasks, async tasks, Vespa enqueue with its failure logging, and side-effect handlers under a Prisma tenant context. Every one of those mechanisms appears exactly once before and after.

`src/database/acl/acl-factory.ts` gains one case, `sdkApiKey`, mapped to `UnscopedACL` rather than the default `BaseQueryACL`. Authentication happens before any tenant context exists — the key is looked up by token, and only then does the caller's identity become known — so a tenant-scoped clause cannot apply. `SdkApiKey` is also added to the `excludeTables` list in the Zero generator, keeping the credential out of the client-syncable schema entirely.

## Schema & Data Changes

- [x] Prisma schema modified (`apps/backend/prisma`)
- [x] Migration added

One migration, `20260820120000_sdk_api_keys`, creating one additive table in the `non_zero` schema. Nothing existing is altered.

| Column | Notes |
|---|---|
| `workspaceId` | Denormalized tenant key |
| `id`, `name`, `userId` | |
| `token` | `@unique` — the encrypted key; the uniqueness is what authenticates |
| `expires_at` | Indexed; keys are short-lived |
| `created_at`, `updated_at` | |

- [x] Clients regenerated (`db:generate`)
- [x] No new Postgres enum or enum value
- [x] Backward compatible with deployed code — purely additive; no reads or writes from existing code paths
- [x] Backfill is idempotent — N/A, no backfill

**Rollout:** the migration can ship ahead of the feature; the table is unused until `SDK_API_ENABLED=true`. Rollback is a plain drop.

## Config, Environment & URLs

- [x] Key added to all relevant env files
- [x] Env setup / secret scripts and docker compose updated
- [x] Verified in every consumer with its own base URL
- [x] Google / Microsoft OAuth redirect URIs still resolve — untouched

**Keys added** (`apps/backend/.env.example`):

```
SDK_API_ENABLED=false            # master switch; router not mounted when false
SDK_API_KEY_TTL_DAYS=30          # lifetime of a newly minted key
SDK_QUERIES_ALLOW_PRIMARY=false  # dev only — never enable in production
```

Three variables, and no secrets among them. Keys are encrypted with the deployment's existing `ENCRYPTION_KEY`, so there is no signing keypair to provision, no client registry to maintain, and no redirect URIs to allowlist. Nothing in this PR changes CORS or any existing origin configuration.

**One deployment prerequisite:** `DATABASE_READ_REPLICA_POOL_URL` must be set in production. SDK reads run on the replica to keep this traffic off the write path. Without it, `/api/sdk/health` reports `degraded` with the reason and reads return `503 service_misconfigured` rather than silently falling back to the primary. `SDK_QUERIES_ALLOW_PRIMARY=true` exists for local setups only.

## API Contract

- [ ] No API changes
- [x] New endpoint
- [ ] Breaking for dashboard / electron / external / bots

All additive. No existing endpoint changed shape, status, or auth. `/api/zero/*` and `/api/vespaSearch` are untouched — search delegates to the existing handler through a response-capturing adapter specifically to avoid forking ~800 lines of hard-won ranking and permission-filtering behaviour.

| | Endpoint | Notes |
|---|---|---|
| `GET` | `/api/sdk/version` | Unauthenticated |
| `GET` | `/api/sdk/health` | Unauthenticated |
| `POST` | `/api/sdk/catalog/query` | 263 queries by name |
| `POST` | `/api/sdk/catalog/mutate` | 245 mutators by name |
| `GET` | `/api/sdk/me` | Identity + `keyExpiresAt` |
| `POST` | `/api/sdk/channels`, `/channels/check-duplicate` | |
| `POST` | `/api/sdk/tickets` | Sequence allocator |
| `POST` | `/api/sdk/channels/:channelId/conversations` | Multipart |
| `POST` | `/api/sdk/attachments`, `/draft-attachments` | Multipart |
| `GET` | `/api/sdk/search`, `/search/schema` | |
| `GET` | `/api/sdk/claw/agents` | |
| `POST` | `/api/sdk/claw/runs` | |
| `GET` | `/api/sdk/claw/runs/:sessionId` | |

`/version` and `/health` are mounted *before* the auth middleware deliberately, so a probe can distinguish "the API is misconfigured" from "your key is bad".

Key management lives at `/api/sdk-keys` (`GET`, `POST`, `DELETE /:id`), which is **session**-authenticated rather than key-authenticated: an API key cannot be used to mint another one.

Errors use one envelope with a stable `code`, a `request_id`, and a `retryable` flag, across 12 codes. `message` is documented as human-readable prose that may change — it passes through ~485 free-text `throw new Error` sites in the mutator catalog, and those messages are frequently the useful part of a `domain_rule` failure. 5xx messages are replaced with a generic string so SQL and connection detail never reach a caller. This surface never returns HTTP 200 on failure, unlike the `/api/zero/*-fallback` handlers it is modelled on.

---

## How did you test it?

Built incrementally over nine phases, each verified before moving on. Every operation name and required argument is machine-checked against the catalog — the registries reference operations by string, so TypeScript cannot catch a typo or a missing required arg, and both would be runtime failures.

Coverage gate output:

```
catalog:  263 queries, 245 mutators (508 total)
exposed:  451
excluded: 57
accounted for: 508/508
checked:  210 result shapes, 28 enumerated arg value sets
OK — every catalog operation is exposed or excluded with a reason.
```

Every gate rule was confirmed to actually fire by deliberately breaking it and restoring: deleting a required arg from a `mapArgs`, introducing an invalid enum value, removing a `mapResult`, and pointing a `#zero-internal` path at a moved file. Each produced a precise error.

That checking caught real bugs during development, including `nonLinear.syncTransitions` requiring an argument named `now` rather than the `timestamp` used everywhere else in the catalog, and a `tickets.listKanban` that omitted two required arguments while sending five that belonged to an unrelated query.

**Auth middleware, exercised live against a real user and the real database:** a valid key resolves the correct `AuthData` with `role`/`orgRole` re-read; an expired key returns `token_expired` (401); a tampered key, an unknown key, a malformed key, and a revoked key all return `unauthenticated`. Notably, a **validly encrypted blob minted with the correct `ENCRYPTION_KEY` but with no database row is rejected** — that is the property the whole auth design rests on, so it was tested directly rather than reasoned about. A key whose row disagrees with its own sealed `workspaceId` is also rejected.

**Key-management routes, live:** create, the 2-key limit returning 409, list omitting the token entirely, delete scoped so another user's id is a no-op 404, revocation taking effect on the next request, and expired keys correctly not occupying a slot.

**Error envelope, over HTTP via supertest:** success, unmatched path, `ApiError`, `ZodError` → `validation_failed`, caller-supplied request id echoed back, and confirmation that a 5xx does not leak its message.

### Automated

- [x] Not covered by tests 

Added: `packages/xyne-spaces-sdk/src/api-operations.test.ts` (6 tests) and `claw-operations.test.ts` (9 tests) — 15 passing; `apps/backend/src/zero/zero-internals.test.ts` (3 tests) pinning the verified `@rocicorp/zero` version and asserting every private module the read path depends on still resolves; and the two build gates, `npm run coverage` and `npm run contract-check`, wired together as `npm run verify`.

Typecheck is clean at **0 errors** across `apps/backend`, `apps/dashboard`, `packages/xyne-spaces-sdk`, and `packages/xyne-spaces-contract`.

**Not yet covered:** the backend `/api/sdk` *routes* have no request-level integration tests. The auth middleware, the error handler, and the key-management routes were each exercised directly and against the live database, but the assembled router was not driven end to end with a real key over HTTP — the app's full import graph does not load under `tsx`, so this needs a proper test harness rather than an ad-hoc script. In particular the direct routes (search, channel and ticket creation, uploads) and the three Claw routes are verified by construction and by their delegation to already-tested controllers, not by execution. That is the main follow-up before enabling this in a shared environment.

### Manual

**Users**
- [x] Single user
- [ ] Two users at once

**Login**
- [ ] Google
- [ ] Microsoft
- [x] Session expiry, re-login, logout — key expiry and revocation paths exercised

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Sync & resilience**
- [x] Existing fallback paths unaffected — the extract-function refactor in `zero/server.ts` preserves both handlers' signatures and return shapes; every post-commit mechanism was diffed before and after

**Surface & data**
- [x] Web browser — the Apps → API Keys screen was exercised in the dashboard: create, copy-once reveal, list with masked hint and remaining days, expired state, and delete behind a confirm step
- [ ] Electron desktop
- [x] Empty state — "No API keys yet"
- [x] Error paths — typed error mapping verified; unauthenticated, expired, forbidden, and not-found paths exercised

**Reviewer note:** manual coverage is honestly thin. Multi-user, cross-workspace, and role-boundary testing has not been done. The mitigating factor is that this surface reuses the app's existing ACL layers unchanged rather than introducing a second authorization path — but that is an argument for *why* it should be safe, not evidence that it is. Worth exercising against a real workspace before enabling anywhere shared. The role-boundary case is the one most worth spending time on, since "a key has exactly its owner's permissions" is the central security claim.

---

## Checklist

- [x] `pnpm run build:shared` passes — `@xyne/spaces-contract` is in the build chain
- [x] Backend `typecheck` + `build` pass — `tsc --noEmit` exits 0
- [x] New deps added with `pnpm --filter <pkg> add <dep>` — the SDK has zero runtime dependencies and imports no Node built-ins, so it stays browser-safe
- [x] Commits follow `<type>: <TICKET-ID> <subject>`; branch is `feature/spaces-sdk`

**Docs added:** `apps/backend/src/api/sdk/README.md` (the five concerns and what each owns, the key format and where integrity comes from, the endpoint table, the error catalog, and a section on what is deliberately not implemented); `packages/xyne-spaces-sdk/README.md` (quickstart, what a key can and cannot do, the resource table, and the behaviours that actually bite — threads-vs-messages, which updates replace rather than patch, which operations toggle rather than set, plural search filters versus singular result types, collaborative-canvas write semantics); and `GAPS.md` (every operation outside the catalog, with the reason it is outside).

**Deliberately not implemented, with reasons stated in the README:**

| | Why |
|---|---|
| Rate limiting | Planned for a later version. A half-kept control is worse than none |
| Idempotency keys | Retries are the caller's concern at this stage |
| OAuth / scopes | The Zero ACL is the authorization boundary |
| mTLS | Terminate client certificates at the gateway |

When rate limiting lands it will need its own Redis keyspace. `services/zeroRateLimiter` is not suitable: it is a fixed-window counter whose buckets are shared with zero-cache traffic, so a user browsing the app would spend their own API budget.

**Known follow-ups, not blockers:**
- Request-level integration tests for `/api/sdk` (see above), covering the direct and Claw routes end to end.
- The SDLC subsystem (10 catalog operations: repos, links, discussions) is excluded rather than exposed. It is driven by its own services and Claw execution profiles; exposing it is a product decision.
- `domain_rule` messages pass through verbatim, which is right for the ~485 mutator sites carrying user-facing text but means an unexpected `Error` is echoed too. Giving new domain failures a typed error is the fix; the mapping itself should not change.
- `npm publish` is disabled repo-wide (`npmPublish: false`, no registry configured). Internal consumers use `workspace:*` or a packed tarball; publishing externally is a separate deliberate decision.
- Triggers/WebSocket (a real-time event surface) is out of scope and planned separately.

## POT: 

https://github.com/user-attachments/assets/11195061-ef26-44d0-868e-34d5071adf93

https://github.com/user-attachments/assets/39d9467d-c70f-47a8-b9ec-1b223040cfa7


sithaarth (dismissed):

sithaarth (dismissed):
d

PR #352, opened 2026-08-08
fix: XYNE-55450 remove unused message_search table and related code

XyneSpaces:
## What
Removes the unused `message_search` Postgres table and all related code.

Ticket: XYNE-55450 — Remove unused message search table

## Why
The `message_search` table (a `tsvector` full-text index with two GIN indexes) was **write-only**. Mutators upserted rows on message create/edit and deleted them on message delete, but **nothing ever read `searchVector`/`plaintextContent` for search** — message search is served by the Vespa index. The table was pure write amplification on the hot message path. The backfill script was also already stale (it called `upsert` with 2 args against the current 3-arg signature).

## Changes
- Drop `MessageSearch` Prisma model and the `Message.messageSearch` relation
- Delete `MessageSearchRepository` and its wiring in `repositories/index.ts`
- Remove the `messageSearch` case in `acl-factory.ts`
- Remove the 5 mutator write sites in `zero/mutators.ts` (3 `upsert`, 2 `delete`) and the two now-orphaned `plainTextContent` declarations + unused `generatePlainTextContent` import
- Remove `messageSearch.deleteMany` calls in `whatsappMigrationService`, `jiraMigrationController`, and `reset-desk-local` (fixed the jira `$transaction` tuple destructuring after removing the array element)
- Delete `scripts/backfill-message-search.ts`
- Add migration `20260808120000_drop_message_search_table` — `DROP TABLE IF EXISTS "workflow"."message_search" CASCADE` (CASCADE drops the two GIN indexes)
- Regenerate the Zero schema

## Validation
- `prisma generate` succeeds; regenerated Zero schema has no `message_search` references
- Repo-wide grep confirms no remaining `messageSearch`/`message_search` references in app code (one unrelated string tag remains in `xyne-claw-auth` error-bucket seed)
- `tsc --noEmit`: clean `main` baseline = 121 pre-existing errors (unrelated `@xyne/shared`/FLOW-board drift); with this change = 117. **Zero new type errors** — the change removes a few stale ones.

## Rollout notes
- Deploy code first (stops all writes to the table), then run the migration to drop it. Order is safe because the table is write-only — dropping it affects no read path.
- The migration is destructive; the table holds only a rebuildable derived index, so no backup/backfill is required.


gyash1512 (approved):

apps/backend/prisma/migrations/20260808120000_drop_message_search_table/migration.sql
```diff
@@ -0,0 +1,12 @@
...
+-- This table was a Postgres full-text search index (tsvector + GIN indexes) that
+-- was only ever written to (upsert on message create/edit, delete on message
+-- delete). Nothing ever read from `searchVector`/`plaintextContent` for search —
+-- message search is served by the Vespa index — so the table was pure write
+-- amplification. Removing the model, repository, and all mutator write paths.
+--
+-- CASCADE drops the associated GIN indexes
+-- (`message_search_plaintext_trgm_idx`, `message_search_vector_idx`) and any
+-- remaining constraints along with the table.
+DROP TABLE IF EXISTS "workflow"."message_search" CASCADE;
```
devesh-juspay, line 12: this table might be there in other scehmas as well, lets delete from there too

devesh-juspay (approved):

apps/backend/prisma/migrations/20260808120000_drop_message_search_table/migration.sql
gyash1512: sure

PR #364, opened 2026-08-10
feat: GENIUS-5430 Error log using Lua filter

pradeepj-2005:
Environment-Changes:
  - LOG_FLUENT_ENABLED: Setting fluent true or false
  - LOG_FLUENT_HOST: setting local/prod
  - LOG_FLUENT_PORT: Setting port for fluentbit

  - backend

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist

### DIFF
--- a/apps/backend/src/utils/logger.ts
+++ b/apps/backend/src/utils/logger.ts
@@ -55,18 +55,27 @@ const normalizeErrors = winston.format((info) => {
   return info;
 });
 
-// fluent-logger's msgpack encoder has no cycle detection and throws on BigInt;
-// break cycles and stringify BigInt before anything reaches it.
-//
-// Tracks the current ancestor chain (not every object ever seen) so two
-// separate references to the same shared, non-cyclic object
+// msgpack has no cycle detection and throws on BigInt; break cycles and stringify BigInt first.
 function decycle(value: unknown, ancestors: unknown[]): unknown {
   if (typeof value === 'bigint') return value.toString();
   if (value instanceof Error) return serializeError(value);
   if (typeof value !== 'object' || value === null) return value;
   if (ancestors.includes(value)) return '[Circular]';
+  // msgpack encodes these natively; skip them or they'd flatten into {} / one key per byte.
+  if (value instanceof Date) return value.toISOString();
+  if (Buffer.isBuffer(value)) return value.toString('base64');
+  if (ArrayBuffer.isView(value)) {
+    const view = value as NodeJS.TypedArray;
+    return Buffer.from(view.buffer, view.byteOffset, view.byteLength).toString('base64');
+  }
 
   const nextAncestors = [...ancestors, value];
+  if (value instanceof Map) {
+    return decycle(Object.fromEntries(value), nextAncestors);
+  }
+  if (value instanceof Set) {
+    return decycle([...value], nextAncestors);
+  }
   if (Array.isArray(value)) {
     return value.map((item) => decycle(item, nextAncestors));
   }
@@ -79,9 +88,7 @@ function decycle(value: unknown, ancestors: unknown[]): unknown {
 
 const sanitizeForFluent = winston.format((info) => decycle(info, []) as winston.Logform.TransformableInfo);
 
-// Runs once at call time, before winston-transport clones `info` per
-// transport -- that clone drops non-enumerable Error fields and can happen
-// asynchronously under a different request's AsyncLocalStorage context.
+// Runs once at call time, before winston-transport's per-transport clone drops Error fields.
 const sharedFormat = winston.format.combine(
   winston.format.timestamp(),
   winston.format.errors({ stack: true }),
@@ -89,9 +96,16 @@ const sharedFormat = winston.format.combine(
   injectContext()
 );
 
+// sharedFormat's timestamp is UTC ISO for Fluent Bit; re-render local for console output.
+function formatLocalTimestamp(iso: string): string {
+  const d = new Date(iso);
+  const pad = (n: number) => String(n).padStart(2, '0');
+  return `${d.getFullYear()}-${pad(d.getMonth() + 1)}-${pad(d.getDate())} ${pad(d.getHours())}:${pad(d.getMinutes())}:${pad(d.getSeconds())}`;
+}
+
 const productionFormat = winston.format.printf(({ timestamp, level, message, ...meta }) => {
   return JSON.stringify({
-    timestamp,
+    timestamp: typeof timestamp === 'string' ? formatLocalTimestamp(timestamp) : timestamp,
     level,
     message,
     ...meta,
@@ -104,7 +118,7 @@ const devFormat = winston.format.combine(
     const context = module || service;
     const contextPrefix = context ? `[${context}] ` : '';
     const metaString = Object.keys(meta).length ? ` ${JSON.stringify(meta)}` : '';
-    const time = typeof timestamp === 'string' ? timestamp.slice(11) : timestamp;
+    const time = typeof timestamp === 'string' ? formatLocalTimestamp(timestamp).slice(11) : timestamp;
     return `${time} ${level}: ${contextPrefix}${message}${metaString}`;
   })
 );
@@ -118,9 +132,8 @@ const transports: winston.transport[] = [
   }),
 ];
 
-// fluent-logger's Sender calls `socket.setTimeout(timeout)` but never
-// attaches a 'timeout' listener -- per Node's net.Socket docs that event
-// fires on idle 
+// fluent-logger sets a socket timeout but never handles it. Guard the connect phase only
+// (removed on 'connect', per-socket) so idle established connections and replaced sockets are safe.
 function guardSenderSocketTimeout(sender: { _socket: Socket | null }): void {
   let socket: Socket | null = null;
   Object.defineProperty(sender, '_socket', {
@@ -128,8 +141,14 @@ function guardSenderSocketTimeout(sender: { _socket: Socket | null }): void {
     get: () => socket,
     set: (next: Socket | null) => {
       socket = next;
-      socket?.once('timeout', () => {
-        socket?.destroy(new Error('fluent socket idle timeout'));
+      if (!next) return;
+      const s = next;
+      const onTimeout = () => {
+        s.destroy(new Error('fluent socket connect timeout'));
+      };
+      s.once('timeout', onTimeout);
+      s.once('connect', () => {
+        s.removeListener('timeout', onTimeout);
       });
     },
   });
@@ -140,14 +159,26 @@ if (config.logging.fluent.enabled) {
   const fluentTransport = new FluentTransport('error.backend', {
     host: config.logging.fluent.host,
     port: config.logging.fluent.port,
-    timeout: 3000, // ms, not seconds
+    timeout: 10_000, // ms, not seconds; connect timeout only, see guardSenderSocketTimeout
     reconnectInterval: 30000,
     messageQueueSizeLimit: 1000, // bound memory when Fluent Bit is unreachable
-    highWaterMark: 1000, // avoid backpressure freezing the Console transport too
+    highWaterMark: 1000,
     level: 'error',
     format: fluentFormat,
-  }) as winston.transport & { sender: { _socket: Socket | null } };
+  }) as winston.transport & {
+    sender: { _socket: Socket | null };
+    log: (info: winston.Logform.TransformableInfo, callback: (err?: unknown, ok?: boolean) => void) => void;
+  };
   guardSenderSocketTimeout(fluentTransport.sender);
+
+  // fluent-logger only calls back on delivery, so an outage stalls winston's Writable and
+  // freezes Console too. Make it fire-and-forget; the sender has its own queue/reconnect.
+  const deliverToSender = fluentTransport.log.bind(fluentTransport);
+  fluentTransport.log = (info, callback) => {
+    deliverToSender(info, () => {});
+    callback();
+  };
+
   transports.push(fluentTransport);
 }
 
--- a/apps/backend/src/utils/logger.ts
+++ b/apps/backend/src/utils/logger.ts
@@ -1,5 +1,6 @@
 import winston from 'winston';
 import { AsyncLocalStorage } from 'async_hooks';
+import fluentLogger from 'fluent-logger';
 import { config } from '@/config/env';
 
 export interface LogContext {
@@ -84,10 +85,44 @@ const devFormat = winston.format.combine(
   })
 );
 
+// Streamed straight to Fluent Bit's forward input (docker/fluent-bit/) over
+// TCP -- no intermediate log file. The forward protocol carries its own
+// event time, so unlike a file-tailed format there's no timestamp string to
+// keep in sync with a Fluent Bit parser.
+const fluentFormat = winston.format.combine(
+  winston.format.timestamp(),
+  winston.format.errors({ stack: true }),
+  normalizeErrors(),
+  injectContext(),
+  winston.format.json()
+);
+
+const transports: winston.transport[] = [
+  new winston.transports.Console({
+    format: config.env === 'development' || config.env === 'test' ? devFormat : productionFormat,
+  }),
+];
+
+if (config.logging.fluent.enabled) {
+  const FluentTransport = fluentLogger.support.winstonTransport();
+  transports.push(
+    new FluentTransport('error.backend', {
+      host: config.logging.fluent.host,
+      port: config.logging.fluent.port,
+      timeout: 3.0,
+      reconnectInterval: 30000,
+      level: 'error',
+      format: fluentFormat,
+    }) as winston.transport
+  );
+}
+
 export const logger = winston.createLogger({
   level: config.logging.level,
-  format: config.env === 'development' || config.env === 'test' ? devFormat : productionFormat,
-  transports: [new winston.transports.Console()],
+  // No logger-level `format`: it would run before each transport's own
+  // format (e.g. devFormat's colorize() would corrupt `level` for the
+  // Fluent transport too), so format is set per-transport instead.
+  transports,
   exitOnError: false,
   defaultMeta: {
     version: '1.0',
--- a/apps/backend/src/utils/logger.ts
+++ b/apps/backend/src/utils/logger.ts
@@ -54,48 +54,55 @@ const normalizeErrors = winston.format((info) => {
   return info;
 });
 
-const productionFormat = winston.format.combine(
+// fluent-logger's msgpack encoder has no cycle detection and throws on BigInt;
+// break cycles and stringify BigInt before anything reaches it.
+const sanitizeForFluent = winston.format((info) => {
+  const seen = new WeakSet<object>();
+  return JSON.parse(
+    JSON.stringify(info, (_key, value) => {
+      if (typeof value === 'bigint') return value.toString();
+      if (value instanceof Error) return serializeError(value);
+      if (typeof value === 'object' && value !== null) {
+        if (seen.has(value)) return '[Circular]';
+        seen.add(value);
+      }
+      return value;
+    })
+  );
+});
+
+// Runs once at call time, before winston-transport clones `info` per
+// transport -- that clone drops non-enumerable Error fields and can happen
+// asynchronously under a different request's AsyncLocalStorage context.
+const sharedFormat = winston.format.combine(
   winston.format.timestamp({ format: 'YYYY-MM-DD HH:mm:ss' }),
   winston.format.errors({ stack: true }),
   normalizeErrors(),
-  injectContext(),
-  winston.format.json(),
-  winston.format.printf(({ timestamp, level, message, ...meta }) => {
-    return JSON.stringify({
-      timestamp,
-      level,
-      message,
-      ...meta,
-    });
-  })
+  injectContext()
 );
 
+const productionFormat = winston.format.printf(({ timestamp, level, message, ...meta }) => {
+  return JSON.stringify({
+    timestamp,
+    level,
+    message,
+    ...meta,
+  });
+});
 
 const devFormat = winston.format.combine(
   winston.format.colorize(),
-  winston.format.timestamp({ format: 'HH:mm:ss' }),
-  winston.format.errors({ stack: true }),
-  normalizeErrors(),
-  injectContext(),
   winston.format.printf(({ timestamp, level, message, module, service, ...meta }) => {
     const context = module || service;
     const contextPrefix = context ? `[${context}] ` : '';
     const metaString = Object.keys(meta).length ? ` ${JSON.stringify(meta)}` : '';
-    return `${timestamp} ${level}: ${contextPrefix}${message}${metaString}`;
+    const time = typeof timestamp === 'string' ? timestamp.slice(11) : timestamp;
+    return `${time} ${level}: ${contextPrefix}${message}${metaString}`;
   })
 );
 
-// Streamed straight to Fluent Bit's forward input (docker/fluent-bit/) over
-// TCP -- no intermediate log file. The forward protocol carries its own
-// event time, so unlike a file-tailed format there's no timestamp string to
-// keep in sync with a Fluent Bit parser.
-const fluentFormat = winston.format.combine(
-  winston.format.timestamp(),
-  winston.format.errors({ stack: true }),
-  normalizeErrors(),
-  injectContext(),
-  winston.format.json()
-);
+// Streamed straight to Fluent Bit's forward input (docker/fluent-bit/) -- no file, no parser to sync.
+const fluentFormat = winston.format.combine(sanitizeForFluent(), winston.format.json());
 
 const transports: winston.transport[] = [
   new winston.transports.Console({
@@ -109,8 +116,10 @@ if (config.logging.fluent.enabled) {
     new FluentTransport('error.backend', {
       host: config.logging.fluent.host,
       port: config.logging.fluent.port,
-      timeout: 3.0,
+      timeout: 3000, // ms, not seconds
       reconnectInterval: 30000,
+      messageQueueSizeLimit: 1000, // bound memory when Fluent Bit is unreachable
+      highWaterMark: 1000, // avoid backpressure freezing the Console transport too
       level: 'error',
       format: fluentFormat,
     }) as winston.transport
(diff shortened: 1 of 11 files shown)


apps/backend/src/utils/logger.ts
```diff
@@ -84,10 +85,44 @@ const devFormat = winston.format.combine(
...
+const transports: winston.transport[] = [
+  new winston.transports.Console({
+    format: config.env === 'development' || config.env === 'test' ? devFormat : productionFormat,
+  }),
+];
+
+if (config.logging.fluent.enabled) {
+  const FluentTransport = fluentLogger.support.winstonTransport();
+  transports.push(
+    new FluentTransport('error.backend', {
```
devesh-juspay: **[1/10] Fluent Bit unreachable stalls ALL logging (including Console) and grows memory without bound**

With `LOG_FLUENT_ENABLED=true` and the fluent-bit container down/not started while errors flow: fluent-logger only invokes the winston-transport write callback after successful delivery (the connect-error path calls `_handleEvent('error')` with no per-record callback), and `messageQueueSizeLimit` defaults to `0` = unlimited. winston-transport is an objectMode Writable (highWaterMark ~16), so after ~16 pending error logs, pipe backpressure pauses the Logger's shared readable — freezing Console output too. Meanwhile queued msgpack buffers, a fresh TCP dial per emit, and stacking 30s reconnect timers grow memory/CPU. The app goes dark exactly when it is erroring.

**Fix:** pass `messageQueueSizeLimit` (bounded queue) or `eventMode: 'PackedForward'` (default 8MiB `sendQueueSizeLimit` + 100ms batching).

apps/backend/src/utils/logger.ts
```diff
@@ -84,10 +85,44 @@ const devFormat = winston.format.combine(
...
+);
+
+const transports: winston.transport[] = [
+  new winston.transports.Console({
+    format: config.env === 'development' || config.env === 'test' ? devFormat : productionFormat,
+  }),
+];
+
+if (config.logging.fluent.enabled) {
+  const FluentTransport = fluentLogger.support.winstonTransport();
```
devesh-juspay, line 158: **[2/10] Circular / msgpack-unencodable meta on an error log throws inside a bare `setImmediate`, escaping winston's error handling**

fluent-logger's `winston.js` `log()` runs `sender.emit(info, ...)` inside an unguarded `setImmediate`, and `sender.js` msgpack-encodes synchronously there. msgpack-lite 0.1.26 has no cycle detection (circular ref -> `RangeError: Maximum call stack size exceeded`, verified empirically) and throws on BigInt (`Unsupported type "bigint"`). `normalizeErrors` only sanitizes top-level values, so nested structures pass straight through (e.g. `messageMetadataSync.ts:40` logs a raw Prisma `where` clause — BigInt filters would throw).

Result: uncaughtException (process crash without the `index.ts` handler; with it, the record is silently dropped, the transport write callback never fires and that slot stalls the Writable, plus a misleading UNCAUGHT EXCEPTION log entry).

**Fix:** cycle-safe pre-serialization in `fluentFormat` (e.g. a safe-stable-stringify round-trip) so bad meta can't bypass winston.

apps/backend/src/utils/logger.ts
```diff
@@ -84,10 +85,44 @@ const devFormat = winston.format.combine(
...
 
+// Streamed straight to Fluent Bit's forward input (docker/fluent-bit/) over
+// TCP -- no intermediate log file. The forward protocol carries its own
+// event time, so unlike a file-tailed format there's no timestamp string to
+// keep in sync with a Fluent Bit parser.
+const fluentFormat = winston.format.combine(
+  winston.format.timestamp(),
+  winston.format.errors({ stack: true }),
+  normalizeErrors(),
+  injectContext(),
```
devesh-juspay: **[5/10] `injectContext()`/`timestamp()` now run at transport delivery time — buffered records get a DIFFERENT request's AsyncLocalStorage context**

Transport-level formats run inside winston-transport's `_write`/`_writev`, not at the `logger.error()` call. The fluent transport acks asynchronously (setImmediate + ~100ms flush batching), so while a send is pending, subsequent infos are buffered raw and format-transformed later inside the previous record's ack continuation. `loggerContext.getStore()` there returns the *earlier* request's store (or `undefined`) — so error records in VictoriaLogs can carry another request's `requestId`/`emailId`/`zeroClientId`, and `timestamp()` stamps flush time instead of error time. Console output is unaffected (its callback is synchronous), which makes this invisible in dev.

Previously the logger-level format ran synchronously in the log-call frame under the correct ALS context (winston `Logger._transform` invokes its callback in the same frame).

**Fix:** hoist the shared `timestamp/errors/normalizeErrors/injectContext` combine back to the logger level and keep only final rendering (colorize+printf / json) per transport.

apps/backend/src/utils/logger.ts
```diff
@@ -84,10 +85,44 @@ const devFormat = winston.format.combine(
...
+      format: fluentFormat,
+    }) as winston.transport
+  );
+}
+
 export const logger = winston.createLogger({
   level: config.logging.level,
-  format: config.env === 'development' || config.env === 'test' ? devFormat : productionFormat,
-  transports: [new winston.transports.Console()],
+  // No logger-level `format`: it would run before each transport's own
```
devesh-juspay: **[6/10] Removing the logger-level format breaks `logger.error(new Error(...))`: transport-level `errors({stack:true})` can no longer see the Error**

For the single-Error call form, winston makes the Error instance itself the info object. winston-transport clones per transport via `Object.assign({}, info)`, which drops an Error's non-enumerable `message`/`stack` and its prototype — so both `format.errors()` and this file's `normalizeErrors` no-op. Empirically reproduced with the installed packages: transport-level format prints `error: undefined` with no stack; the same format at logger level (the pre-diff arrangement) prints message+stack.

Currently latent — all existing call sites use `logger.error('msg', err)`, which survives via the splat path — but the first future `logger.error(err)` (idiomatic winston) silently ships empty records to the console *and* the error pipeline.

**Fix:** keep a minimal logger-level format of just `winston.format.errors({ stack: true })` — it doesn't reintroduce the colorize problem this comment guards against.

apps/backend/src/utils/logger.ts
```diff
@@ -84,10 +85,44 @@ const devFormat = winston.format.combine(
...
+  }),
+];
+
+if (config.logging.fluent.enabled) {
+  const FluentTransport = fluentLogger.support.winstonTransport();
+  transports.push(
+    new FluentTransport('error.backend', {
+      host: config.logging.fluent.host,
+      port: config.logging.fluent.port,
+      timeout: 3.0,
```
devesh-juspay: **[8/10] `timeout: 3.0` is 3 *milliseconds*, and inert regardless**

fluent-logger 3.4.1 `sender.js:255` passes the raw value to `socket.setTimeout(this.timeout)` (Node units: ms), so this is 3ms, not the Fluentd-style 3s it suggests. Worse, the sender attaches no `'timeout'` listener (only `error`/`connect`/`tlsClientError`/`secureConnect`), so the timeout event fires to nobody and the setting does nothing at all. A black-holed `LOG_FLUENT_HOST` (packets dropped rather than refused) hangs connects for the OS TCP timeout (~1-2 min) with no application-level abort while error records queue.

Note: changing `3.0` to `3000` is not sufficient — an effective bound requires a timeout listener that destroys the socket (or an app-level connect timer).

apps/backend/src/utils/logger.ts
```diff
@@ -84,10 +85,44 @@ const devFormat = winston.format.combine(
...
+      reconnectInterval: 30000,
+      level: 'error',
+      format: fluentFormat,
+    }) as winston.transport
+  );
+}
+
 export const logger = winston.createLogger({
```
devesh-juspay, line 185: **[10/10] With no logger-level format, winston silently substitutes `logform/json()` — every accepted log call is safe-stable-stringified and the result discarded**

winston 3.19.0 `logger.js:105`: `this.format = format || require('logform/json')()`, executed in `Logger._transform` for every accepted log *before* transport-level filtering. At `LOG_LEVEL=debug` in dev, every `debug`/`info` call pays a full stringify of its meta (stacks, large objects, defaultMeta) whose `MESSAGE` symbol every transport format then overwrites. The shared `timestamp/errors/normalizeErrors/injectContext` stages additionally run once per transport for error logs. The comment here also misdescribes the mechanism — transport isolation actually comes from winston-transport's per-transport `Object.assign({}, info)` clone, not from the absent logger format.

**Fix:** give the logger the shared combine as its explicit format (which also fixes the `errors()`-stripping and ALS-timing findings), leaving only final rendering per transport.

docker/fluent-bit/normalize_hash.lua
```diff
@@ -0,0 +1,50 @@
...
+local function normalize(message)
+  if message == nil then
+    return ''
+  end
+  local m = string.lower(tostring(message))
+
+  -- UUIDs
+  m = string.gsub(m, '%x%x%x%x%x%x%x%x%-%x%x%x%x%-%x%x%x%x%-%x%x%x%x%-%x%x%x%x%x%x%x%x%x%x%x%x', '<uuid>')
+  -- ISO-ish timestamps
+  m = string.gsub(m, '%d%d%d%d%-%d%d%-%d%d[T ]%d%d:%d%d:%d%d[%.%d]*Z?', '<ts>')
```
devesh-juspay: **[3/10] Timestamp pattern uses literal uppercase `T`/`Z` but runs after `string.lower()` — ISO timestamps are never collapsed to `<ts>`, splitting fingerprints**

Line 9 lowercases the message, so the literal `T` in `[T ]` and the `Z?` can never match the now-lowercase `t`/`z`. A message like `sync failed at 2026-08-10T12:34:56.789Z` falls through to the generic `%d+` rule, so variants of the same error (`12:34:56z` vs `12:34:56.789z` vs a `+05:30` offset) normalize to *different* templates -> different djb2 fingerprints -> the dedup grouping this whole pipeline exists for splits one error across multiple fingerprints.

**Fix:** use lowercase `t`/`z` in the pattern (or run the timestamp rule before `string.lower`).

docker/fluent-bit/normalize_hash.lua
```diff
@@ -0,0 +1,50 @@
...
+  -- hex addresses / long hex ids (e.g. 0x7ffee, object hashes)
+  m = string.gsub(m, '0x%x+', '<hex>')
+  m = string.gsub(m, '%x%x%x%x%x%x%x%x%x%x%x%x+', '<hex>')
```
devesh-juspay: **[9/10] Bare 12+-hex-char rule runs before `%d+`, so identical errors fingerprint differently based on numeric id length**

Decimal digits are all valid `%x` chars: an 11-digit id normalizes to `<num>` while a 12-digit one becomes `<hex>`, so the same error template yields two fingerprints as ids cross the length boundary (epoch-millis and snowflake ids always hit the hex rule; shorter ids in the same message shape don't) — the dedup dashboard double-counts one error as distinct fingerprints.

**Fix:** substitute one shared token (e.g. `<id>`) for both the hex and numeric rules so ordering can't split fingerprints, or keep only the `0x%x+` rule.

docker/fluent-bit/normalize_hash.lua
```diff
@@ -0,0 +1,50 @@
...
+  for i = 1, #str do
+    hash = ((hash * 33) + string.byte(str, i)) % 4294967296
+  end
+  return string.format('%08x', hash)
+end
+
+function normalize_and_hash(tag, timestamp, record)
+  local template = normalize(record['message'])
+  record['normalized_message'] = template
+  record['fingerprint'] = djb2((record['module'] or '') .. '|' .. template)
```
devesh-juspay: **[4/10] Fingerprint keys on the raw un-normalized `module` field — which is already dynamic in the codebase — and `_stream_fields=fingerprint,module` makes each unique value a persistent VictoriaLogs stream**

`apps/backend/src/vespa/src/services/searchService.ts:275` sets `module` to `` `vespa-search, ${searchId}` `` (unique per search). Since only the *message* is normalized, every occurrence of the same underlying vespa error gets a distinct fingerprint, and via `fluent-bit.conf`'s `_stream_fields=fingerprint,module` its own VictoriaLogs stream — the "Occurrences by Fingerprint" dashboard shows `count=1` rows forever and stream cardinality grows unbounded with search volume (VictoriaLogs docs explicitly warn against high-cardinality stream fields).

**Fix:** normalize `module` too (or compute the fingerprint in `fluentFormat`, where the emitter knows which of its fields are stable), and use `_stream_fields=module` only, keeping `fingerprint` as an ordinary indexed field — `stats by (fingerprint)` works identically.

docker/grafana/provisioning/dashboards/victorialogs-errors.json
```diff
@@ -0,0 +1,53 @@
...
+      "datasource": { "type": "victoriametrics-logs-datasource", "uid": "victorialogs-errors" },
+      "gridPos": { "h": 8, "w": 24, "x": 0, "y": 0 },
+      "id": 1,
+      "targets": [
+        {
+          "refId": "A",
+          "datasource": { "type": "victoriametrics-logs-datasource", "uid": "victorialogs-errors" },
+          "expr": "level:error",
+          "queryType": "logs"
+        }
```
devesh-juspay, line 17: **[7/10] `"queryType": "logs"` is not a valid victoriametrics-logs-datasource query type**

The plugin's `QueryType` enum is exactly `instant | stats | statsRange | hits` (`src/types.ts`, mirrored in `pkg/plugin/query.go`). `"logs"` only renders because the Go backend's `switch` falls through to the `default:` instant branch. Opening the panel editor shows no query type selected (the frontend only defaults falsy values and the RadioButtonGroup matches exact enum values), and `datasource.ts` exact comparisons treat it as not-instant in supplementary-query logic — any plugin version that validates `queryType` or changes the default breaks this provisioned panel.

**Fix:** `"queryType": "instant"` (panel 2's `"stats"` is valid).

devesh-juspay (commented):
Automated code review of the error-log pipeline (Winston -> fluent-logger -> Fluent Bit -> VictoriaLogs -> Grafana). 10 findings below as inline comments, ranked most severe first. 9 were confirmed against the installed winston 3.19.0 / winston-transport 4.9.0 / logform 2.7.0 / fluent-logger 3.4.1 / msgpack-lite 0.1.26 sources (several with empirical repros) and the victorialogs-datasource plugin source; 1 is plausible.

apps/backend/src/utils/logger.ts
```diff
@@ -53,41 +55,106 @@ const normalizeErrors = winston.format((info) => {
...
+
+if (config.logging.fluent.enabled) {
+  const FluentTransport = fluentLogger.support.winstonTransport();
+  const fluentTransport = new FluentTransport('error.backend', {
+    host: config.logging.fluent.host,
+    port: config.logging.fluent.port,
+    timeout: 3000, // ms, not seconds
+    reconnectInterval: 30000,
+    messageQueueSizeLimit: 1000, // bound memory when Fluent Bit is unreachable
+    highWaterMark: 1000, // avoid backpressure freezing the Console transport too
```
devesh-juspay: **[Partial fix — original finding 1] Backpressure root cause untouched: sustained Fluent Bit outage still freezes ALL logging after ~1000 pending error logs**

Both knobs are real (`super(options)` in fluent-logger's winston.js passes `highWaterMark` to winston-transport's Writable, and `messageQueueSizeLimit` bounds the sender queue), but the underlying defect remains: fluent-logger only invokes the winston write callback on successful delivery. With Fluent Bit down, the first write stays pending forever (connect-error path calls `_handleEvent('error')` with no per-record callback), subsequent infos buffer in the Writable, and once ~1000 are pending, `write()` returns false and pipe backpressure pauses the Logger's shared readable — the Console transport freezes again, just at a higher threshold (~16 -> ~1000), with up to 1000 buffered infos held in memory until reconnect.

Also note `sender.js:190-191`: when `messageQueueSizeLimit` engages it does `this._sendQueue.shift()` **without invoking the dropped item's callback**, so any dropped record's winston write stalls permanently.

**Fix direction:** make the transport genuinely best-effort — wrap `FluentTransport.log` so the winston callback is invoked immediately (fire-and-forget into the sender, which already has its own bounded queue + reconnect), or fork/patch the transport to call `callback(null, true)` up front and let sender-level retry handle delivery.

apps/backend/src/utils/logger.ts
```diff
@@ -53,41 +55,106 @@ const normalizeErrors = winston.format((info) => {
...
+// attaches a 'timeout' listener -- per Node's net.Socket docs that event
+// fires on idle 
+function guardSenderSocketTimeout(sender: { _socket: Socket | null }): void {
+  let socket: Socket | null = null;
+  Object.defineProperty(sender, '_socket', {
+    configurable: true,
+    get: () => socket,
+    set: (next: Socket | null) => {
+      socket = next;
+      socket?.once('timeout', () => {
```
devesh-juspay: **[New finding A] The timeout guard tears down the healthy connection after any 3s of log silence**

`socket.setTimeout(3000)` is an *inactivity* timer on the established socket, not a connect timeout — and error logs are sparse by nature. After nearly every error burst the socket idles >3s, this listener destroys it with an Error, which flows through the sender's error handler: a spurious `fluent socket idle timeout` error is logged and a 30s reconnect timer is scheduled, once per quiet period. Net effect: connection-per-burst instead of a persistent connection, plus steady error-handler churn — the inert pre-fix behavior at least kept the connection alive.

**Fix:** bound only the connect phase — attach the `'timeout'` listener when the socket is created and remove it on `'connect'`/`'secureConnect'` (fluent-logger only calls `setTimeout` on the plain-TCP path, so an app-level connect timer would also cover TLS). If idle closes are actually desired, destroy without an Error so each one doesn't masquerade as a failure.

apps/backend/src/utils/logger.ts
```diff
@@ -53,41 +55,106 @@ const normalizeErrors = winston.format((info) => {
...
+        socket?.destroy(new Error('fluent socket idle timeout'));
```
devesh-juspay: **[New finding D] Stale closure: the timeout listener destroys whatever socket the shared variable currently points to, not the socket that timed out**

The listener closes over the mutable `socket` variable. If socket A emits a late `'timeout'` after being replaced — reachable via the `end()` + `this._socket = null` path in `sender.js:138-140`, where a half-open socket can still fire its idle timer — the callback destroys the *new* live socket B with `fluent socket idle timeout`. Self-healing via reconnect, but a spurious disconnect of a working connection.

**Fix:** capture the specific socket in the setter:
```ts
set: (next: Socket | null) => {
  socket = next;
  const s = next;
  s?.once('timeout', () => s.destroy(new Error('fluent socket idle timeout')));
},
```

apps/backend/src/utils/logger.ts
```diff
@@ -53,41 +55,106 @@ const normalizeErrors = winston.format((info) => {
...
+  if (value instanceof Error) return serializeError(value);
+  if (typeof value !== 'object' || value === null) return value;
+  if (ancestors.includes(value)) return '[Circular]';
+
+  const nextAncestors = [...ancestors, value];
+  if (Array.isArray(value)) {
+    return value.map((item) => decycle(item, nextAncestors));
+  }
+  const out: Record<string, unknown> = {};
+  for (const key of Object.keys(value)) {
```
devesh-juspay, line 83: **[New finding B] `decycle` mangles non-plain objects: Date/Map/Set become `{}`, Buffer explodes into per-byte keys**

Only BigInt and Error are special-cased; everything else that is `typeof 'object'` falls through to `Object.keys`:
- `Date` -> `{}` (no enumerable own keys) — previously msgpack-lite encoded Dates natively, so date fields in error meta now silently vanish from VictoriaLogs records.
- `Map`/`Set` -> `{}`.
- `Buffer`/TypedArray -> an object with one key per byte index (`{"0":104,"1":101,...}`) — for a large buffer in meta this is a big per-log CPU/memory blowup replacing msgpack's compact native encoding.

**Fix:** add cases before the generic object branch: `value instanceof Date ? value.toISOString()`, `Buffer.isBuffer(value) || ArrayBuffer.isView(value) ? (base64 or a truncated placeholder)`, `value instanceof Map/Set ? [...entries]`.

apps/backend/src/utils/logger.ts
```diff
@@ -53,41 +55,106 @@ const normalizeErrors = winston.format((info) => {
...
+  return out;
+}
+
+const sanitizeForFluent = winston.format((info) => decycle(info, []) as winston.Logform.TransformableInfo);
+
+// Runs once at call time, before winston-transport clones `info` per
+// transport -- that clone drops non-enumerable Error fields and can happen
+// asynchronously under a different request's AsyncLocalStorage context.
+const sharedFormat = winston.format.combine(
+  winston.format.timestamp(),
```
devesh-juspay, line 93: **[New finding C] Production console timestamp silently changed format and timezone**

`winston.format.timestamp()` with no options emits `new Date().toISOString()` — UTC, millisecond precision, `T`/`Z` — replacing the previous production format `'YYYY-MM-DD HH:mm:ss'` rendered in *local* time. Two observable changes to every production console log line: the string format (breaks any downstream parser/alert regex expecting the old shape) and the timezone (timestamps shift by the host's UTC offset). Dev console similarly now shows `12:34:56.789Z` via `timestamp.slice(11)` instead of `HH:mm:ss`.

If standardizing on UTC ISO is intentional, fine — but it deserves a deliberate call. Otherwise keep ISO here for the fluent path (Fluent Bit's `_time_field` parses it) and re-apply the old rendering in `productionFormat` from the ISO value, or pass `{ format: 'YYYY-MM-DD HH:mm:ss' }` and accept second precision for fluent.

devesh-juspay (commented):
Re-review of the fix commits (200381cce, eb48eddb2) against the 10 findings from the previous review. Fix mechanisms were verified against fluent-logger 3.4.1 / winston-transport 4.9.0 sources, not just line changes.

### Status of original findings: 8 fixed, 1 partial, 1 fixed with a new side effect

| # | Finding | Status |
|---|---------|--------|
| 1 | Fluent Bit down stalls all logging + unbounded memory | **Partial** — see inline comment: thresholds raised (~16 -> ~1000) and sender queue bounded, but the root cause (write callback only fires on delivery) is untouched, so a sustained outage still freezes Console logging after ~1000 pending error logs |
| 2 | msgpack circular/BigInt throw escapes winston | **Fixed** — `decycle()` breaks cycles, stringifies BigInt, serializes nested Errors before the sender (also fixes nested errors encoding to `{}`). New caveat on non-plain objects, see inline comment on `decycle` |
| 3 | Lua timestamp pattern uppercase `T`/`Z` after `lower()` | **Fixed** — pattern now `[t ]...z?` |
| 4 | Fingerprint on raw dynamic `module` + high-cardinality stream fields | **Fixed** — fingerprint keys on `normalized_module`; `_stream_fields=normalized_module` only; `fingerprint` is a regular field, which `stats by (fingerprint)` still supports |
| 5 | ALS request-context misattribution at transport delivery time | **Fixed** — `sharedFormat` runs synchronously at the log call, at logger level, before transport buffering |
| 6 | `logger.error(new Error(...))` loses message/stack via transport clone | **Fixed** — `errors({stack:true})`/`normalizeErrors` now run before winston-transport's `Object.assign({}, info)` clone |
| 7 | Invalid dashboard `queryType: "logs"` | **Fixed** — now `"instant"` |
| 8 | `timeout: 3.0` = 3ms and inert | **Fixed, with side effect** — now 3000ms with a real `'timeout'` listener covering plain+TLS socket paths, so black-holed connects are bounded; but the listener introduces an idle-teardown problem, see inline comment on the guard |
| 9 | Hex-vs-number Lua rule splits fingerprints by id length | **Fixed** — both rules now emit `<num>` |
| 10 | Hidden default `json()` format waste + format-chain duplication | **Fixed** — explicit logger-level `sharedFormat`; per-transport formats are rendering-only |

### New findings in the fix commits

Four new issues (1 from the partial, plus A-D) are posted as inline comments below: the backpressure root cause, the idle-timeout teardown, `decycle` mangling Date/Map/Set/Buffer, the production timestamp format/timezone change, and the stale-closure socket destroy.

devesh-juspay (approved):
All review findings verified as resolved at head 8c9cb4f2d — the fluent transport is now fire-and-forget (no backpressure path to Console), the socket-timeout guard is connect-phase-only with per-socket capture, decycle handles Date/Buffer/TypedArray/Map/Set, and console timestamp rendering is restored to local YYYY-MM-DD HH:mm:ss. Approving.

devesh-juspay (approved):
All outstanding review findings verified as resolved at this head (fix commit 4c750b200):

- **#1 backpressure**: the `fluentTransport.log` wrapper invokes the winston callback immediately and delivers fire-and-forget; the Writable can no longer buffer/stall, the Console transport is isolated from Fluent Bit outages, and memory stays bounded by the sender's `messageQueueSizeLimit`.
- **A idle teardown**: the socket `'timeout'` guard is now connect-phase-only (listener removed on `'connect'`), so established idle connections are left alone; black-holed connects are bounded at 10s.
- **B decycle**: Date/Buffer/TypedArray/Map/Set now convert sensibly (ISO string, base64, entries/array with cycle protection) instead of flattening to `{}` or per-byte keys.
- **C timestamps**: production JSON and dev console re-render the shared ISO timestamp as local `YYYY-MM-DD HH:mm:ss` / `HH:mm:ss`, restoring the original console format while the fluent path keeps UTC ISO for `_time_field`.
- **D stale closure**: the timeout handler captures and destroys its own socket (`const s = next`).

Non-blocking notes: a raw `ArrayBuffer` (not a typed-array view) still decycles to `{}`, and `highWaterMark: 1000` is now a harmless leftover since the transport no longer exerts backpressure. No new issues found in the fix commit.

(Automated re-review; mechanisms verified against fluent-logger 3.4.1 / winston-transport 4.9.0 sources.)

sithaarth (approved):

PR #370, opened 2026-08-10
fix: draft attachment upload guard rejects own attachment after send race

XyneSpaces:
## Problem
Users hit `One or more attachment ids reference a resource you do not own` when uploading draft attachments. This is a false 403 from the ownership guard on `POST /api/drafts/attachments/upload` (`apps/backend/src/controllers/draftAttachmentController.ts`).

The frontend fires the Zero create-attachment mutator and the REST upload concurrently. When the user sends (or schedules) the message before a large/slow upload finishes, the Zero create-message mutator flips the same user's attachment `entityType` DRAFT -> CHAT (`mutators.ts:2417-2432`) or DRAFT -> DELAYED_MESSAGE. The in-flight upload's ownership SELECT then sees `entityType !== DRAFT` and the old predicate `uploadedByUserId !== userId || entityType !== DRAFT` classified the user's OWN row as foreign, returning 403 and aborting the WHOLE batch before any file url was written.

This is a NEW ownership-guard regression, not the previously-fixed P2002 race (that catch-and-update fallback is still intact).

## Fix
Key the guard on whether the row already has a stored file (`url`), not on `entityType`:
- reject if `uploadedByUserId !== userId` (cross-user), or
- reject if `url !== ''` (already-committed attachment on any surface — must not be repointed).

A freshly-created draft row always has `url === ''`, so it is safe to fill even after a raced entityType change. There is no API that lets a client set `url` to an arbitrary string — `url` is always derived server-side from uploaded bytes — so `url === ''` is a strictly stronger discriminator than the entityType check and preserves both the cross-user and cross-surface protections.

## Tests
Adds `apps/backend/src/controllers/draftAttachmentController.guard.test.ts`, a hermetic jest test that drives the real controller guard:
- accepts own draft attachment raced to CHAT and to DELAYED_MESSAGE (the bug)
- accepts a normal DRAFT row and a brand-new id
- rejects a different user's attachment (cross-user)
- rejects re-pointing an already-committed attachment (url already set)

Verified red -> green: with the old guard the two regression cases fail with `Expected 200 / Received 403`; with the fix all 6 pass.


XyneSpaces:
## Proof of test (POT)

The POT screenshot is committed in this PR at `docs/pot/draft-attachment-guard-pot.png` (write-up: `docs/pot/draft-attachment-guard-pot.md`):

draft attachment guard POT

### How the issue was reproduced

The bug is a race between two things the dashboard fires **concurrently** for the same attachment (`apps/dashboard/src/hooks/useDraft.ts`): the Zero create-attachment mutator and the REST upload `POST /api/drafts/attachments/upload`. If the user sends (or schedules) the message before a large/slow upload finishes, the Zero create-message mutator flips the **same user's** attachment row `entityType` `DRAFT -> CHAT` (`apps/backend/src/zero/mutators.ts:2417-2432`) or `DRAFT -> DELAYED_MESSAGE`. The in-flight upload's ownership `SELECT` then runs against that already-flipped row, and the old guard treated the user's **own** row as foreign, returning a 403 that aborted the whole batch before any file `url` was written.

Reproducing that against a live browser is timing-dependent and flaky. Instead, the regression test **deterministically seeds the exact post-send state** and drives the **real controller method** `DraftAttachmentController.uploadDraftAttachment` — it is not a copy of the predicate:

- inject a fake `db` whose `messageAttachment.findMany` returns the caller's OWN row with `entityType` already `CHAT` / `DELAYED_MESSAGE` and `url === ''` (the upload is the writer that fills `url`);
- capture the controller's HTTP status via a fake `res`.

Test: `apps/backend/src/controllers/draftAttachmentController.guard.test.ts`.

### Red -> green (same test, guard toggled)

BEFORE (old guard `entityType !== DRAFT`) — the two race cases fail with the exact production symptom:

```
✕ REGRESSION: own draft attachment after send flipped DRAFT -> CHAT
✕ REGRESSION: own draft attachment flipped DRAFT -> DELAYED_MESSAGE (scheduled send race)
✓ accepts a normal freshly-created DRAFT row
✓ accepts a brand-new id with no existing row
✓ SECURITY: rejects an attachment owned by a different user
✓ SECURITY: rejects re-pointing an already-committed attachment (url set)

  ● REGRESSION: own draft attachment after send flipped DRAFT -> CHAT
    Expected: 200
    Received: 403     <- "reference a resource you do not own"
Tests: 2 failed, 4 passed, 6 total
```

AFTER (fixed guard `url !== ''`) — bug gone, both security cases still reject:

```
PASS src/controllers/draftAttachmentController.guard.test.ts
✓ REGRESSION: own draft attachment after send flipped DRAFT -> CHAT
✓ REGRESSION: own draft attachment flipped DRAFT -> DELAYED_MESSAGE (scheduled send race)
✓ accepts a normal freshly-created DRAFT row
✓ accepts a brand-new id with no existing row
✓ SECURITY: rejects an attachment owned by a different user (cross-user overwrite)
✓ SECURITY: rejects re-pointing an already-committed attachment (own row, url set)
Tests: 6 passed, 6 total
```

### Run it

```
cd apps/backend
NODE_OPTIONS=--max-old-space-size=6144 npx jest src/controllers/draftAttachmentController.guard.test.ts --runInBand
```

PR #382, opened 2026-08-10
fix: SDLCT-0002 Implement Notification Log Pipeline Completeness Check

XyneSpaces:
## SDLCT-0002 — Notification Log Pipeline Completeness Check

Adds an append-only, workspace-scoped audit trail for the mobile-push notification pipeline and a completeness classifier that detects deliveries which silently never reached the push provider (FCM/APNS). Ships **disabled by default** behind `NOTIFICATION_LOG_ENABLED`; every write is fail-soft and a no-op when the flag is off or `workspaceId`/`correlationId` is missing.

### What changed
- **Data model:** `NotificationLogEvent` model + 4 enums in the `non_zero` Prisma schema, added to Zero `excludeTables` (never syncs to browser clients). Hand-written migration `20260810120000_add_notification_log_event`.
- **Correlation:** a `correlationId` is stamped at notification creation and threaded creation → producer → Bull job → worker → FCM channel via optional fields on `MobilePushPayload`.
- **Persistence:** `NotificationLogRepository` with an idempotent upsert on `[workspaceId, idempotencyKey]`; workspace-scoped reads by notificationId / correlationId.
- **Service:** `NotificationLogService` — flag-gated, fail-soft, allowlist-based metadata redaction, per-stage recorder methods.
- **Classifier:** pure `classifyCompleteness()` (COMPLETE / INCOMPLETE / FAILED / UNKNOWN, with grace window) + a service wrapper that adds metrics and workspace-scoped DB reads.
- **Instrumentation:** producer, worker, and FCM channel emit lifecycle events at each stage.
- **APIs (workspace-scoped, auth-gated):** `GET /api/notifications/:id/logs`, `GET /api/notifications/:id/completeness`, `GET /api/notifications/logs?correlationId=` — all return 404 when the feature flag is off.
- **Observability:** otel counters + a pipeline-latency histogram.
- **ACL:** added a `notificationLogEvent` case to `ACLFactory` (workspace-scoped `BaseQueryACL`) so the model union stays exhaustive.

### Validation
- `tsc --noEmit` passes cleanly on `apps/backend`.
- New unit tests: redaction (7 tests) and completeness classifier (8 tests) — all green.
- `prisma generate` succeeds; migration matches the model.

### Rollout
1. Merge with flag off (no behavior change).
2. Apply migration.
3. Enable `NOTIFICATION_LOG_ENABLED=true` in one workspace/region, watch `notification_pipeline_completeness_total{status=...}` and `notification_log_write_failures_total`.
4. Roll out wider.

Ticket: SDLCT-0002


apps/backend/src/database/acl/acl-factory.ts
```diff
@@ -236,6 +236,11 @@ export class ACLFactory {
       return new NotificationsACL(ctx, prisma)
     case 'notificationPreference':
       return new NotificationPreferencesACL(ctx, prisma)
+    // SDLCT-0002: append-only notification lifecycle audit log. Workspace-scoped
+    // and never exposed via Zero; reads go through NotificationLogRepository with an
+    // explicit workspaceId, but a case is required to keep this switch exhaustive.
```
XyneSpaces, line 241: **[blocking] R1/S13** — `NotificationLogEvent` is workspace-scoped support telemetry, but wiring it to `BaseQueryACL` gives the generic ACL path only the base/pass-through behavior instead of an explicit fail-closed or notification-log-specific ACL. If this model becomes reachable through the generic query layer, support/audit events can be read more broadly than intended. Please add a dedicated ACL that fails closed or enforces `workspaceId` explicitly, or keep the model out of the factory entirely if it must only be read through `NotificationLogRepository`.

apps/backend/src/notification-service/channels/mobilePush.ts
```diff
@@ -38,6 +39,16 @@ export class MobilePushChannel {
   private async deliverFcm(job: MobilePushJobData): Promise<NotificationDeliveryResult> {
     const { sessionId, token, platform, payload, userId, appVersion } = job;
```
XyneSpaces, line 40: **[blocking] R2** — provider lifecycle logging is added only inside `deliverFcm`, but the APNS success path returns from `deliver()` before this code runs. Successful iOS incoming-call VoIP pushes will have `DELIVERY_PLANNED/QUEUE_*` events but no `PROVIDER_REQUEST_STARTED` / `PROVIDER_ACCEPTED`, so the new completeness classifier reports a successful APNS delivery as `INCOMPLETE`. Please emit equivalent APNS provider-start/accepted/rejected events in `deliverApn` or move provider instrumentation around both provider paths.

XyneSpaces (commented):
Inline findings from Xyne Spaces Reviewer for PR #382.

XyneSpaces:
## Xyne Spaces Reviewer — PR #382

**Reviewed:** 21 files, 21 hunks · **Findings:** 2 blocking, 0 major, 0 nits

### Findings
| blocking | R1/S13 | `apps/backend/src/database/acl/acl-factory.ts:241` | `notificationLogEvent` is wired to `BaseQueryACL`; for workspace-scoped support telemetry this should fail closed or use an explicit workspace-scoped ACL rather than a pass-through/generic base ACL. |
| blocking | R2 | `apps/backend/src/notification-service/channels/mobilePush.ts:40` | Provider-stage logging is FCM-only; successful APNS VoIP deliveries return before `PROVIDER_REQUEST_STARTED` / `PROVIDER_ACCEPTED`, so the classifier marks successful APNS pushes incomplete. |

### Recommended reviewer
**@5uryan5h** (recommended; review request could not be submitted because the available GitHub MCP tools do not expose a request-reviewer endpoint and no `gh` CLI tool is available in this runtime) — active recent backend committer with 0 outstanding requested reviews in this repo.

| @5uryan5h | Recent backend committer; exact per-file ownership unavailable via current tools | no CODEOWNERS file | 0 | active today in Spaces |
| @anuragdvd | Ticket/request context; exact per-file ownership unavailable | no CODEOWNERS file | 1 | thread participant / ticket owner |
| @Abhishek Verma | Notification/Prisma migration context from Spaces | no CODEOWNERS file | 0 | domain familiarity |

No CODEOWNERS file exists in the repo, so ranking fell back to recent repo authors, notification/Prisma domain context from Spaces, and current requested-review workload. Migration safety ran for the new `apps/backend/prisma/migrations/**/migration.sql`; no unsafe 🔴/🟡 operation was surfaced, so no QA tag/comment was emitted.

PR #386, opened 2026-08-10
feat : METTLEPRAG-5634 team intelligence version2- #383

abhisheksuman-cmd:
Added separate paginated APIs for each leadership dashboard section.
Added organization, team, and user leadership snapshot APIs.
Added goal-based team grouping: 10X, 5X, 2X, Ready to Accelerate, and No Goal Data.
Uses the highest matching goal track when a team has goals across multiple tracks.
Uses the previous month of Team Intelligence evidence to determine goal activity.
Added clickable organization/team summary cards that scroll to their corresponding sections.
Improved Team Intelligence summary generation, validation, concurrency controls, fallbacks, and worker processing.
Excluded DM and group-DM conversations from Team Intelligence evidence.
Restricted call evidence to calls originating from regular channels. Personal, conversation, DM, group-DM, calendar, and unlinked calls are excluded.
Removed obsolete dashboard components and consolidated the leadership dashboard UI.
Restored unrelated configuration, dependency, generated schema, and lockfile changes to match main.


ameer030 (changes requested):
Reverified against current PR head `23081dcc2752c3474718a3f36dcadc1e3b2d4558`. Requesting changes for the P1/P2 issues below.

## Standards

1. **Missing setup documentation.** This adds 10 Team Intelligence model/timeout/concurrency environment variables in `.env.example` and `config/env.ts`, but changes neither README nor `docs/setup/**`. `CONTRIBUTING.md` explicitly requires setup/env changes to update those docs in the same PR.

2. **Inconsistent request validation.** New org/team endpoints manually cast `req.query` and parse dates/pages, while the user endpoints in this same change use `UserDashboardQuerySchema.safeParse`. Please use shared schema validation for the new org/team query boundaries too.

Non-blocking maintainability notes:

- The near-identical LLM streaming/error/metrics pipelines in the org, team, and user summary services are duplicated and likely belong behind one shared runner.
- `LeadershipDashboard.tsx` is a new 1,597-line module containing normalization/ranking, data fetching, UI primitives, and all three dashboards. Consider splitting these responsibilities.

## Spec

1. **P1 — Team/user leadership snapshot APIs can return cross-workspace data.** The backing models have `workspaceId`, but the team query filters only by date + `teamId`, and the user query only by date + email. These models use `UnscopedACL`, so generic Prisma ACL does not add tenant filtering. The user precheck only proves that the same email exists in the caller's workspace; the subsequent email-only query can still mix rows from other workspaces. The team endpoint has no equivalent ownership check. Pass the caller workspace through controller/service/repository and include it in both queries.

   - Team: https://github.com/juspay/xyne-spaces/blob/23081dcc2752c3474718a3f36dcadc1e3b2d4558/apps/backend/src/database/repositories/teamIntelligenceTeamRepository.ts#L240-L253
   - User: https://github.com/juspay/xyne-spaces/blob/23081dcc2752c3474718a3f36dcadc1e3b2d4558/apps/backend/src/database/repositories/teamIntelligenceUserRepository.ts#L499-L519

2. **P1 — Goal grouping chooses the highest active track, not the highest matching track.** `track` is computed from every active goal before alignment matching. If a team has an unmatched 10X goal and a matched 2X goal, it is classified as 10X. Compute the highest track only from active goals whose IDs matched evidence.

   https://github.com/juspay/xyne-spaces/blob/23081dcc2752c3474718a3f36dcadc1e3b2d4558/apps/backend/src/services/teamIntelligenceGoalGroupingService.ts#L54-L74

3. **P1 — Previous-month evidence window is absent.** The summary query has no `reportDate` bound and merges alignments from every completed historical snapshot. Old evidence can therefore keep a current goal active indefinitely. Add explicit previous-calendar-month bounds, with an injectable reference date for testing.

   https://github.com/juspay/xyne-spaces/blob/23081dcc2752c3474718a3f36dcadc1e3b2d4558/apps/backend/src/services/teamIntelligenceGoalGroupingService.ts#L125-L163

4. **P2 — A newer incomplete team snapshot masks an older completed snapshot.** The team query includes all statuses and null `contentUrl` rows, sorts newest first, and the controller uses only `snapshots[0]`. A newer RECEIVED/PROCESSING/FAILED row therefore produces a 404 even when an older completed snapshot exists in range. Filter for `status: 'COMPLETED'` and non-null `contentUrl` before selecting the newest row.

   - Query: https://github.com/juspay/xyne-spaces/blob/23081dcc2752c3474718a3f36dcadc1e3b2d4558/apps/backend/src/database/repositories/teamIntelligenceTeamRepository.ts#L249-L270
   - Selection: https://github.com/juspay/xyne-spaces/blob/23081dcc2752c3474718a3f36dcadc1e3b2d4558/apps/backend/src/controllers/teamIntelligenceTeamController.ts#L57-L65

Summary: Standards 2 actionable + 2 non-blocking; Spec 4 blockers (3 P1, 1 P2).

abhisheksuman-cmd:
@ameer030 resolved all

ameer030 (approved):

Mohan0544 (approved):

PR #408, opened 2026-08-10
feat: XYNE-55230 persist loader screen until dashboard loads.

sithaarth:
…ectron theme with dashboard setting

  - dashboard

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


devesh-juspay (commented):
Automated review (Claude Code) — 12 findings: 7 correctness, 1 efficiency, 3 simplification, 1 nits. The recurring theme: the new overlay lifecycle has no fail-safe, so several realistic states recreate the whitescreen-class hang this PR is meant to fix. Details inline.

apps/electron/src/window/manager.ts
```diff
@@ -400,6 +543,12 @@ export async function loadUrl(window: BrowserWindow, url: string, counter?: Coun
       safeRecordMetric(() => {
         counter?.add(1, { success: 'true', buildVersion: app.getVersion() });
       });
+      if (!waitForDashboardReady) {
+        await window.webContents.executeJavaScript(
+          'new Promise(resolve => requestAnimationFrame(() => requestAnimationFrame(resolve)))',
+        ).catch(() => undefined);
+        hideLoadingOverlay(window);
```
devesh-juspay, line 550: **[correctness] Loading overlay has no fallback — can hang forever**

When `waitForDashboardReady=true` the overlay is hidden ONLY by the `dashboard:app-ready` IPC from the new App.tsx effect — no timeout, `did-fail-load`, or `did-finish-load` fallback. `useBundledUI` is false in every env config, so production loads the REMOTE dashboard from `FRONTEND_URL`, deployed independently of the binary. If the deployed bundle predates this PR (deploy ordering/rollback), the dashboard JS crashes before App mounts, or the server 302-redirects to an off-origin SSO page (`did-frame-navigate` disarms `dashboardReadinessArmed` and handlers.ts blocks the IPC), `hideLoadingOverlay` is never called — the overlay permanently covers a working (or errored) app: the exact whitescreen-class hang this PR fixes, now with no escape.

**Suggested fix:** add a watchdog (~30s after `loadURL` resolves) that hides the overlay (or swaps to the timeout-error page) if `dashboard:app-ready` hasn't arrived, plus `did-fail-load`/main-frame `did-finish-load` as secondary hide triggers. The IPC signal should be the fast path, never the only path.

apps/electron/src/window/manager.ts
```diff
@@ -265,7 +401,8 @@ export async function createMainWindow(options?: { inactive?: boolean }): Promis
       try {
         if (mainWindow) {
           isReloading = true;
-          await loadUrl(mainWindow, mainWindow.webContents.getURL());
+          await showLoadingOverlay(mainWindow);
+          await loadUrl(mainWindow, mainWindow.webContents.getURL(), undefined, true);
```
devesh-juspay, line 405: **[correctness] Cmd+R during mTLS enrollment strands the overlay forever**

This reload handler hardcodes `waitForDashboardReady=true` for whatever URL is currently loaded. During enrollment the window is at `config.MTLS_FRONTEND_URL` (different origin from `FRONTEND_URL`; `loadApp` correctly loads it with `false`). On Cmd+R: `showLoadingOverlay` then `loadUrl(mainWindow, currentUrl, undefined, true)` — the two-rAF hide branch is skipped, `did-frame-navigate` arms with `isDashboardUrl(authUrl)=false`, and the enrollment app never sends `dashboard:app-ready` (handlers.ts would block it anyway). Nothing ever hides the overlay — the user can never enroll without quitting.

**Suggested fix:** drop the explicit `true` and let `loadUrl` use its own default, `isDashboardUrl(url)`, which is correct here.

apps/dashboard/src/hooks/useTheme.ts
```diff
@@ -16,17 +16,38 @@ const readThemeFromUrl = (): Theme | null => {
...
 export const useTheme = (): { theme: Theme; changeTheme: (newTheme: Theme) => void } => {
   const [theme, setTheme] = useState<Theme>(() => {
-    // Prefer ?theme=… from the URL — this is how the main app hands its
-    // theme off to a Xyne URL opened in the browser-panel webview (it has a
-    // separate localStorage, so we can't share directly). Falls back to
-    // localStorage, then to the default.
     if (typeof window !== 'undefined') {
+      const storedTheme = readStoredTheme();
+
+      if (window.electronAPI?.getTheme) {
```
devesh-juspay, line 41: **[correctness] Child Electron windows forced to 'classic'; shared localStorage key deleted**

claw-overlay-window.ts:394 creates a separate BrowserWindow with the same preload.js and loads `/newWindow/claw` (and `setWindowOpenHandler` spawns similar windows). There `window.electronAPI.getTheme` exists, so this hook takes the electron branch unconditionally — but the `app:get-theme` handler's `isMainWindowSender` check (handlers.ts:174) fails and returns `'classic'`. A midnight user's child windows render classic, `clearStoredTheme()` deletes the shared `xyne-theme` localStorage fallback that previously themed them correctly, their `app:set-theme` sends are silently dropped, and searchNavigation.ts:532 propagates `theme=classic` into browser-panel URLs.

**Suggested fix:** either relax `isMainWindowSender` for the theme handlers (theme is app-global), or fall back to localStorage when the electron call fails and stop deleting the shared key.

apps/electron/src/window/manager.ts
```diff
@@ -32,13 +37,131 @@ export function setWindowReferences(): void {
...
+      event.preventDefault();
+      void loadApp(window);
+    }
+  });
+  await loadingView.webContents.loadFile(assetPath('loading.html'), {
+    query: { theme: getAppTheme() },
+  });
+  if (!window.isDestroyed() && !window.isVisible()) {
+    if (revealInactive) window.showInactive();
+    else window.show();
```
devesh-juspay, line 89: **[correctness] Background update reload pops up a hidden window and steals focus**

The creation branch calls `window.show()` whenever the window is not visible. On macOS the close handler (manager.ts:433) `preventDefault()`s and hides the window; `hideLoadingOverlay` nulls `loadingView`, so a later `showLoadingOverlay` re-enters this branch. ui-updater.ts:452 applies a downloaded UI update precisely when the window is NOT visible → `applyAndReload` → `loadUrl` → `showLoadingOverlay` → `window.show()`: the hidden window pops up with a loading screen and steals focus mid-task with zero user interaction. Old code revealed the window exactly once via `mainWindow.once('ready-to-show')`.

**Suggested fix:** only `show()` on the initial reveal (gate on the first-show/`revealInactive` flag, not `!window.isVisible()`); background updates should reload silently.

apps/electron/src/keychain/macKeychainService.ts
```diff
@@ -256,28 +257,53 @@ class MacKeychainService implements IKeychain {
...
-            if (e.stderr && e.stderr.includes("not be found")) {
-                console.log("Identity not found, nothing to delete.");
+            if (e.stderr?.includes("not be found")) {
+                log.info("Identity not found, nothing to delete.");
                 Logger.info(EnrollmentEvent.IDENTITY_NOT_FOUND, { common_name: commonName });
             } else {
-                console.warn("Delete identity warning:", e.stderr || e.message);
+                log.error("Delete identity failed:", e.stderr || e.message);
                 Logger.logError(EnrollmentEvent.IDENTITY_DELETE_FAILED, e);
+                throw e;
```
devesh-juspay, line 301: **[correctness] deleteIdentity now throws; `delete-keys` IPC rejects unhandled**

Old code only warned and always resolved. mtls-handlers.ts:161 does `return await keychain.deleteIdentity(...)` with no catch, so when the user denies the macOS keychain prompt (realistic), `ipcRenderer.invoke` now rejects in the remotely-served enrollment frontend that was written against never-rejecting semantics — re-enrollment can abort with an unhandled rejection. Mac also now diverges from winKeychainService.ts:306 and linuxKeychainService.ts:295, which still swallow all failures, and the previously-dead `dialog.showErrorBox('...restart the application...')` in certificate-error-handler.ts becomes reachable as a blocking modal.

**Suggested fix:** pick one contract for all three platforms — either restore never-throw (return `{success, error}`), or catch in mtls-handlers.ts and return a structured error the frontend actually handles.

apps/electron/src/keychain/macKeychainService.ts
```diff
@@ -256,28 +257,53 @@ class MacKeychainService implements IKeychain {
...
+            log.info(`Deleted ${hashes.length} identity certificate(s) successfully.`);
+            Logger.info(EnrollmentEvent.IDENTITY_DELETED, {
+                common_name: commonName,
+                deleted_count: hashes.length,
+            });
         } catch (e: any) {
-            // It might fail if not found, which is fine
-            if (e.stderr && e.stderr.includes("not be found")) {
-                console.log("Identity not found, nothing to delete.");
+            if (e.stderr?.includes("not be found")) {
```
devesh-juspay, line 295: **[correctness] 'not be found' catch masks partial keychain cleanup**

The outer catch treats any error containing 'not be found' as benign IDENTITY_NOT_FOUND, but the inner catch (line 280) only wraps delete-identity — if the delete-certificate fallback (line 285) throws with that stderr (hash-lookup mismatch, or a race where another process removed the cert between find and delete), the error escapes to the outer catch, which returns without processing the rest of the hash list. In the duplicated-CN case this code targets, later hashes are never deleted yet the delete-keys IPC observes success — re-enrollment proceeds against a keychain still holding stale certs and the invalid-certificate loop can repeat.

**Suggested fix:** move the 'not be found = benign' classification into the inner per-hash catch and continue the loop on single-hash failures; let the outer catch only see whole-operation failures, and report failure if any cert remains.

apps/electron/src/preload.ts
```diff
@@ -50,6 +50,18 @@ interface ElectronAuthData {
 }
 
 const electronAPI = {
+  getTheme: (fallbackTheme?: string): string => {
+    return ipcRenderer.sendSync('app:get-theme', fallbackTheme);
```
devesh-juspay, line 54: **[efficiency] Blocking `sendSync` theme read + unconditional sync disk write per component mount**

`getTheme` is a blocking `ipcRenderer.sendSync` called in `useTheme`'s `useState` initializer, and the `[theme]` effect fires an unconditional `setTheme` IPC that synchronously rewrites appearance.json in the main process. `useTheme` is mounted at 18 call sites including MessageBubble.tsx:1569 (one per chat message): rendering/scrolling a channel with N messages performs N renderer-blocking sendSync round-trips (each a synchronous electron-store disk read), and each mount's effect does a synchronous `writeFileSync` even when nothing changed — visible scroll jank under a busy main process. Previously this was a cheap localStorage read.

**Suggested fix:** cache the theme once (module-level/context, or read `document.documentElement.dataset.theme`), and no-op `app:set-theme`/`store.set` when the value is unchanged.

apps/electron/src/window/manager.ts
```diff
@@ -32,13 +37,131 @@ export function setWindowReferences(): void {
...
+}
+
+export function hideLoadingOverlay(window = mainWindow): void {
+  if (!loadingView) return;
+  if (window && !window.isDestroyed()) window.contentView.removeChildView(loadingView);
+  if (!loadingView.webContents.isDestroyed()) loadingView.webContents.close();
+  loadingView = null;
+}
+
+export async function showWindowError(window: BrowserWindow, fileName: string): Promise<void> {
```
devesh-juspay, line 101: **[simplification] Redundant double/triple loads of loading.html per navigation**

Every load path pays repeated full page-loads of loading.html: `loadUrl` calls `showLoadingOverlay` at entry while `loadApp` and the Cmd+R handler already called it; the reuse branch re-loads loading.html even when already showing it; and `showWindowError` loads loading.html then immediately loads the error page over it. Each startup/reload runs `loadFile('loading.html')` twice back-to-back (HTML parse, theme.js/css, two PNGs, progress-animation restart = visible flicker and added latency), and every error display pays a third, discarded load. The creation branch also hardcodes `#1A1A1F`/`#F7F7F8` backgrounds (line 66), missing summer_breeze's `#eef6f8` from theme.css — a wrong-color flash for that theme.

**Suggested fix:** make `loadUrl` the single overlay owner, parametrize `showLoadingOverlay(window, file='loading.html')` so `showWindowError` loads the error page directly, skip the reload when the overlay already shows the same file/theme, and source the background colors from one shared theme constant.

apps/electron/assets/theme.css
```diff
@@ -0,0 +1,68 @@
...
+  --accent-hover: #ff7d7d;
+  --shadow: 0 16px 44px rgba(0, 0, 0, 0.28);
+}
+
+html, body {
+  font-family: Inter, -apple-system, BlinkMacSystemFont, "Segoe UI", sans-serif;
+  -webkit-font-smoothing: antialiased;
+}
+
+body.themed-page {
```
devesh-juspay, line 38: **[simplification] Error pages themed via `!important` overrides — incomplete coverage**

`loading.html` and `meeting-popup.html` were tokenized directly (`var(--page-bg)`, `var(--accent)`, …), but the four error pages keep their original hardcoded palettes (`#F5F5F5`, `#007bff`, …) and are re-skinned by this `.themed-page … !important` layer. Two conventions now coexist in `assets/`, and anything not enumerated here stays light: e.g. `invalid-certificate.html:31-45` `.icon-container` (`#FFE5E5→#FFD6D6` gradient) and `.icon` (`#FF4F4F`) have no matching selector, so a midnight user sees a pink badge on a dark card. Editing e.g. `vpn-error.html`'s `button:hover` colour also silently has no effect because the `!important` rule always wins.

**Suggested fix:** tokenize the error pages the same way `loading.html` was done (replace the hardcoded colours with the `--page-bg`/`--surface`/`--text`/`--muted`/`--accent` vars) and drop the `.themed-page` `!important` layer.

apps/electron/assets/recording-pill.html
```diff
@@ -190,6 +190,7 @@
       font-weight: 600;
     }
   </style>
+  <link rel="stylesheet" href="./theme.css">
```
devesh-juspay, line 193: **[simplification] `theme.css` link is dead here**

The pill sets `document.documentElement.dataset.theme` to `'dark'`/`'light'` (line ~334), so `theme.css`'s `:root[data-theme="midnight"]` / `[data-theme="summer_breeze"]` selectors never match, and `<body>` has no `themed-page` class so none of the `.themed-page …` rules apply either. The only rule that does take effect is `html, body { font-family: Inter, … }`, which — being later in the cascade at equal specificity — overrides the pill's own `-apple-system` stack. Net effect: no theming, just a font change on machines that have Inter installed.

**Suggested fix:** drop the link (the pill already has its own `--card-bg`/`--ink` tokens), or if the intent is to share the palette, map `theme.css` vars onto the pill's existing dark/light attribute values.

apps/dashboard/src/routes/AnalyticsScreen/AnalyticsScreen.tsx
```diff
@@ -31,28 +32,24 @@ import {
   processDurationMetric,
 } from './AnalyticsScreen.utils';
 
-type Theme = 'classic' | 'midnight' | 'summer_breeze';
-
-const THEME_STORAGE_KEY = 'xyne-theme';
+const readDocumentTheme = (): Theme => {
```
devesh-juspay, line 35: **[nits] Small cleanups across the theme plumbing**

- `AnalyticsScreen.tsx:35-39` — `readDocumentTheme` re-hardcodes the theme allowlist (`theme === 'midnight' || theme === 'summer_breeze'`) even though it imports `Theme` from `useTheme.ts`. Export `isValidTheme` / `DEFAULT_THEME` from `useTheme.ts` and reuse them; adding a fourth theme would otherwise silently fall back to `classic` here. Same for the two inline `'classic' | 'midnight' | 'summer_breeze'` unions in `types/electron.d.ts:20-21` — `import type { Theme } from '../hooks/useTheme'` works there.
- `app-theme.ts:16` — `getAppTheme` writes to the store from inside a getter. The `useTheme` mount effect already calls `setTheme(theme)` with the same value a tick later, so the migration write is redundant; drop it and keep `getAppTheme` read-only.
- `App.tsx:38` — the new effect cleanup is missing an explicit `: void` return type (`@typescript-eslint/explicit-function-return-type` warns; the other cleanup in this file at line ~104 has it).

apps/electron/src/window/manager.ts
```diff
@@ -32,13 +37,131 @@ export function setWindowReferences(): void {
...
+    try {
+      const target = new URL(url);
+      if (target.protocol === 'https:') void shell.openExternal(url);
+    } catch {
+      // Ignore malformed links from static pages.
+    }
+    return { action: 'deny' };
+  });
+  loadingView.webContents.on('will-navigate', (event, url) => {
+    if (url === 'xyne-reload://app') {
```
devesh-juspay, line 79: **[correctness] Error-page retry: dead `reloadApp` branch, magic `xyne-reload://` scheme, and no reentrancy guard**

The overlay is created with `sandbox: true` and no preload (lines 59-65), so in all four error pages the primary `window.electronAPI.reloadApp` branch is permanently dead — retry depends solely on `location.href = 'xyne-reload://app'` matched here by exact string equality. `xyne-reload` is registered nowhere (`registerSchemesAsPrivileged` / `protocol.handle`); if Chromium routes the unregistered scheme to external-protocol handling instead of emitting `will-navigate` on some Electron versions, every Reload button silently does nothing and the user is stuck. The magic string is also duplicated verbatim in 4 HTML files.

On top of that, the `void loadApp(window)` at line 81 has no reentrancy guard (unlike the Cmd+R handler, which checks `isReloading`) and the error pages' Reload buttons aren't disabled/debounced. Clicking Reload twice, or while a prior `loadApp` is in its retry loop (3×1s) or the 60s `waitForLocalDashboard` poll, runs two instances that both mutate shared module state (`loadingView`, `dashboardReadinessArmed`) and both call `window.loadURL` — the second navigation aborts the first with `ERR_ABORTED`, whose retry loop counts it as a failure and can finish with `showWindowError('load-error.html')`: a spurious "Failed to Load" page over a working app, plus duplicate keychain/health-check flows racing.

**Suggested fix (one change covers both):** give the overlay a minimal preload exposing just `reloadApp()` via `contextBridge`, have the error pages call it, and delete (or demote to a constant-defined fallback) the magic scheme. Then hoist the `isReloading` guard into `loadApp` itself so every entry point — overlay reload, Cmd+R, the `reload-app` IPC — is protected.

PR #467, opened 2026-08-12
feature: XYNE-55748 add canvas archive backend support

uayush01:
Migration-Changes:
  - Adds `isArchived` boolean default false to `Canvas` for archive/unarchive list filtering.

  - backend
  - dashboard

## Type of Change

- [x] New feature

## Description

Adds backend support for Canvas archive/unarchive for `XYNE-55748`.

This adds a minimal `Canvas.isArchived` flag, creator-only archive/unarchive Zero mutators, and list query filtering so archived canvases are excluded by default with an archived-only query mode available.

## Areas Touched

## Zero (Sync) Changes

## Schema & Data Changes

- [x] Migration added
- [x] Clients regenerated (`db:generate`, `db:common:generate`)

Rollout / rollback notes:
- Backward compatible. Existing canvases default to `isArchived = false`.

## Config, Environment & URLs

## API Contract

### Automated

- [x] Not covered by tests

Ran:
- `pnpm --filter @xyne/shared build`
- `pnpm --filter xyne-spaces-backend typecheck`
- Backend build via pre-commit hook
- Pre-commit gitleaks, tenant-key guard, schema migration validation, Zero schema sync check, enum guard

## Checklist


apps/backend/src/zero/mutators.ts
```diff
@@ -9203,6 +9204,46 @@ export function createMutators(
...
+        z.object({
+          canvasId: z.string(),
+        }),
+        async ({ tx, args: { canvasId } }) => {
+          const canvas = await tx.run(zql.canvases.where('id', canvasId).one());
+          if (!canvas) {
+            throw new Error('Canvas not found');
+          }
+
+          if (canvas.createdBy !== authData.sub) {
```
XyneSpaces, line 9167: archiveCanvas/unarchiveCanvas are creator-only, but the existing updateCanvas mutator also permits EDITOR/OWNER canvas_participants and guest_access editors. Archiving hides the canvas from every collaborator's list, so a channel editor who can edit it still can't archive it — confirm this is intended. If archive should follow edit-access, reuse the same participant/guest_access check updateCanvas uses instead of a bare createdBy compare.

apps/backend/src/zero/queries.ts
```diff
@@ -2301,6 +2303,12 @@ export const queries: AnyQueryRegistry = defineQueries({
         query = query.where('docType', DocType.Canvas);
       }
 
+      if (args.onlyArchived) {
```
XyneSpaces: This archive-filter block (onlyArchived → where isArchived true; else if !includeArchived → where isArchived false) is copy-pasted across ~5 query bodies here and mirrored in packages/shared/src/zero/queries.ts. Extract a helper applyArchiveFilter(query, { includeArchived, onlyArchived }) and call it in each site so the default-hide behavior can't drift between queries.

apps/backend/prisma/generated/zero/schema.ts
```diff
@@ -1836,6 +1837,7 @@ export const canvasCommentThreadTable = table("canvas_comment_threads")
     blockId: string(),
     anchorText: string().optional(),
     initialCommentId: string().optional(),
+    commentCount: number(),
```
XyneSpaces, line 1827: commentCount on canvas_comment_threads is unrelated to this PR's canvas-archive scope. It already exists in schema.prisma and the shared Zero schema, so this is a stale generated-file catch-up rather than a new column — harmless, but confirm it came from regeneration (npx prisma generate) and not a hand-edit, and ideally keep unrelated regen noise out of a feature PR.

apps/backend/prisma/schema.prisma
```diff
@@ -2899,6 +2899,7 @@ model Canvas {
   editAccessId String? @unique // Random ID for edit access
   visibility   String           @default("PRIVATE")
   isTemplate   Boolean          @default(false)
+  isArchived   Boolean          @default(false)
```
XyneSpaces, line 2872: Nit: isArchived is modeled as another boolean alongside isTemplate/isCollaborative, which is consistent with the current pattern so this is fine. Flagging only that a canvas lifecycle (active/archived/trashed) may later be cleaner as a single status enum than accreting booleans.

XyneSpaces (commented):
0 blockers, 4 nits. Changes requested (minor): (1) confirm archive should be creator-only vs edit-access, (2) de-duplicate the archive-filter block into a shared helper, (3) add a unit test asserting a non-creator archiveCanvas throws and that default canvas lists exclude archived canvases, and attach a POT recording (required for backend + migration PRs). Migration is safe (NOT NULL DEFAULT false, no backfill needed) and default query behavior is backward-compatible. Backend + dashboard must deploy together due to the shared Zero schema bump — already noted in the PR description.

apps/backend/src/zero/mutators.ts
uayush01: Intentional archive follows deleteCanvas's creator only precedent, not updateCanvas's general edit-access channel editor can modify canvas content, but hiding it from every collaborator's list (which archive does) is a more disruptive action than editing same reasoning we already applied to delete Not a bug, no change needed.

santhoshJus (approved):

Mohan0544 (approved):

PR #471, opened 2026-08-12
feat: open profile in a modal on non-chat pages

XyneSpaces:
## 1. Problem Description
The account-menu **Profile** action was tied to the routed `/chat/dir/:channelId/profile/:userId` sidebar, which is only mounted on chat pages. On non-chat pages (`/support`, `/automations`, `/calls`, `/projects`, etc.) clicking **Profile** navigated the user into chat instead of showing their profile in place — a confusing bounce out of the current page.

## 2. If this is a bug fix or an enhancement, how did you identify the issue?
The profile sidebar is a nested React Router route (`profile/:userId`) mounted only under the chat routes. `handleProfileClick` in `Settings.tsx` always navigated to `/chat/dir/.../profile/...`, so on any non-chat surface the user was pulled into chat.

## 3. Explain the solution implemented in the PR
- Keep the existing chat-page behavior exactly: on chat pages, Profile still opens the routed profile sidebar.
- On non-chat pages, open the profile in a modal instead, staying on the current page.
- New component `apps/dashboard/src/components/ProfileSidebar/ProfileModal.tsx` wraps the existing `UserProfile` in a `Dialog`. It auto-closes when the route changes (e.g. after a Message/Call action navigates away).
- `Settings.tsx` gains an optional `onOpenProfileModal?` prop. `handleProfileClick` branches: when the handler is present, the user has an id, and `usePath()` is not a `/chat/` route, it opens the modal and returns; otherwise it falls through to the existing routed navigation (backward compatible).
- `AppSidebar.tsx` (present on every page) owns the modal state and renders `ProfileModal`, passing `onOpenProfileModal` into `SettingsContent`.
- Mobile (`ProfileView`/`MobileProfileMenu`) and in-chat entry points (mentions, message bubbles, slash commands) are only rendered on chat surfaces where the routed sidebar exists, so they intentionally keep routed navigation and are unchanged.

## 4. Documentation (link to docs created/updated for this change)
N/A

## 5. DB Alter Details (if any)
N/A

## 6. Data fix Details (if any)
N/A

## 7. Environment variable Changes (if any)
N/A

## 8. Testing (how it was tested; screenshots/links)
Verified locally against the dashboard dev server:
- Chat page (`/chat/dir/:channelId`): Profile opens the routed side panel, URL becomes `.../profile/:userId` — unchanged behavior.
- Non-chat page (`/calls`): Profile opens a centered modal, URL stays `/calls` — new behavior.
Both render the same `UserProfile` content. `tsc --noEmit` on the dashboard reports no type errors in the changed files.

## 9. Services to which this change is to be deployed
dashboard (frontend only)

## 10. Main PR link (if this PR is raised against a release branch)
N/A

### DIFF
--- a/apps/dashboard/src/components/AppSidebar/AppSidebar.tsx
+++ b/apps/dashboard/src/components/AppSidebar/AppSidebar.tsx
@@ -27,6 +27,7 @@ import {
 import Avatar from '../ui/Avatar/Avatar';
 import { Popover } from '../ui/Popover/Popover';
 import SettingsContent from '../Settings/Settings';
+import ProfileModal from '../ProfileSidebar/ProfileModal';
 import Preferences, { type PreferenceSection } from '../Settings/Preferences';
 import { useSelf } from '../../hooks/useUsers';
 import { isStatusExpired } from '../../utils/statusUtils';
@@ -195,6 +196,7 @@ const AppSidebar = (): ReactElement => {
   const [isErrorReportOpen, setIsErrorReportOpen] = useState(false);
   const [isMoreOpen, setIsMoreOpen] = useState(false);
   const [isPreferencesOpen, setIsPreferencesOpen] = useState(false);
+  const [profileModalUserId, setProfileModalUserId] = useState<string | null>(null);
   const [preferencesInitialSection, setPreferencesInitialSection] = useState<
     PreferenceSection | undefined
   >(undefined);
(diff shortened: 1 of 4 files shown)


XyneSpaces:
Follow-up commit `885520a`: inline profile modal header.

The profile modal (shown on non-chat pages) now arranges the avatar and the name / team-name / active-status block **side by side** instead of stacked.

- Added an optional `headerLayout?: 'stacked' | 'inline'` prop to the shared `UserProfile` component (default `'stacked'`).
- Only `ProfileModal` passes `headerLayout='inline'` → `flex-row` header, 64px (`xl`) avatar, `flex-1` text block.
- The routed profile sidebar on chat pages and the search-results profile keep the original stacked/centered layout — no regression.

Verified on `/calls`: default (no team name → "+ Add Team Name"), the inline add/edit team-name flow, and the saved state all render correctly with the side-by-side layout. Screenshots shared with the requester.

harsh-1923 (approved):

devesh-juspay (approved):

PR #528, opened 2026-08-13
fix: XYNE-55115 fixes ticket search and board reset issue

aman-juspay:
Fixes the ticket search issue and board reset issue.
__Demo_Video__: https://drive.google.com/file/d/1BL7TIcIzdvTbMUORYXiDtlC4u4Poos8q/view?usp=sharing


apps/dashboard/src/routes/KanbanBoardScreen/KanbanBoardScreen.tsx
```diff
@@ -3243,9 +3242,10 @@ const KanbanBoardScreen: React.FC<BoardKanbanScreenProps> = ({
     if (!localTickets || localTickets.length === 0) return;
 
     lastKnownKanbanTicketsRef.current = {
```
devesh-juspay, line 3272: **Stale-board guard defeats itself.** This effect runs with the new `kanbanColumnQueryKey` but the old search's tickets still in its closure, so old results get stamped with the new key. The `lastKnownKanbanTickets.queryKey === kanbanColumnQueryKey` check (line 3264) then passes for the wrong tickets.

Repro: change the search term from `auth` to `authz`. The reset effect nulls `localTickets`, but this ref-writer still sees the old non-empty `localTickets` and writes `{queryKey: NEW, tickets: OLD-auth-rows}`. If `authz` returns zero results, the board keeps showing the previous search's tickets indefinitely — the exact bug this ref was added to fix.

apps/backend/src/vespa/src/utils/responseProcessor.ts
```diff
@@ -101,31 +101,117 @@ const stripHighlightTags = (text: string): string => {
     const highlighted = highlightText(value, query, mentionNames);
     return { highlighted, wasHighlighted: highlighted.includes('<hi>') };
 };
+  /**
+   * Ticket summary fields scanned by the text-presence escape hatch below. Mirrors the
+   * searchable half of the ticket `default` fieldset; fields absent from the response's
+   * summary class are simply skipped.
+   */
+  const TICKET_TEXT_FIELDS = [
```
devesh-juspay: **`TICKET_TEXT_FIELDS` should be just title/description/xyneId.** Intended behavior is that tickets match on title/description/xyneId only. This 22-field list (plus the `formFields` block below) will match on stage/status/boardName/assignee/tags/form values on any path where those fields are present in the summary class (e.g. `lean`), which is broader than intended.

Suggest trimming to `['title', 'description', 'description_clean', 'xyneId']`, dropping the `formFields` block in `buildTicketHaystack`, and fixing the doc comments here and on `ticketMatchesQueryText` that advertise the wider matching. (As written the comment is also inaccurate — the list doesn't actually mirror the searchable half of the ticket `default` fieldset in either direction.)

apps/backend/src/vespa/src/utils/responseProcessor.ts
```diff
@@ -101,31 +101,117 @@ const stripHighlightTags = (text: string): string => {
...
+
     const result = processNodesRecursively(response.root.children, (node) => {
       const matchFeatures = node.fields?.matchfeatures;
       const nativeRank = matchFeatures?.combined_nativeRank ?? node.fields?.rankfeatures?.nativeRank;
 
       if (typeof nativeRank === 'number' && nativeRank >= threshold) {
         return { keep: true, node };
       }
+
+      const isTicketHit =
```
devesh-juspay: **Rescue behaves differently per entry point.** Whether this rescue fires depends on which summary class the caller requested: `docType` exists in `lean` but not `default`, and each summary exposes different fields to `buildTicketHaystack`. The same ticket/query/threshold can be kept in ticket-scoped board search (`ticketOnlySearch` true, `default` summary) but dropped in cmd-K unified search (`lean` summary via `docType`), or vice versa — this will surface as a "ticket shows in Tickets filter but not in global search" flake.

A multi-app search that omits `presentationSummary` gets the `default` summary where `docType` is absent, so both arms are false and tickets get no rescue at all on that path. Trimming `TICKET_TEXT_FIELDS` to title/description/xyneId shrinks the field-divergence half of this, but the `docType` gap remains.

apps/backend/src/vespa/src/services/searchService.ts
```diff
@@ -490,11 +490,16 @@ export class SearchService {
...
        // nativeRank is based on text matching - meaningless without a query
        if (nativeRankThreshold > 0 && searchQuery?.trim()) {
-        response = filterByNativeRank(response, nativeRankThreshold , this.logger);
+        response = filterByNativeRank(response, nativeRankThreshold , this.logger, {
+          query: searchQuery,
+          ticketOnlySearch: isTicketOnlySearch,
+        });
       }
 
       const exactResultCount = response.root?.children?.length || 0;
```
devesh-juspay, line 508: **Rescued hits suppress the fuzzy fallback.** `exactResultCount` is measured after `filterByNativeRank`, so hits rescued by the new text-match escape hatch now count toward it. On the default fallback path (`vespa_search_new_fallback_method` off), that makes `needsFallback` false, and the 3-gram fuzzy pass that used to recover typo/prefix queries never runs — users get only the weak substring-rescued hits instead of the results the fallback would have found.

Consider counting rescued hits separately (e.g. have `filterByNativeRank` return the rescued count) and excluding them from `exactResultCount` for the fallback decision.

devesh-juspay (commented):
Reviewed the search/board changes — a few functional issues worth addressing before merge. Details inline.

devesh-juspay (approved):

santhoshJus (approved):

samitbarai (approved):

PR #536, opened 2026-08-13
fix: XYNE-62354 pre filter automation runs from creating unnecessary runs

Revanthvenkatpasupuleti:
POT:- 

currently in prod there will be number of automations active with that trigger X number of triggers

Screenshot 2026-08-13 at 5 01 15 PM

which means if in one channel if there is a message received it will trigger 50 + automatons, adds 50 + rows in workflowexecution table and pushes 50+ jobs into automation queue

after my fix it will pre-check the conditions before pushing into queue or creating a row, so we will see actual runs of that automation only here

https://github.com/user-attachments/assets/46bf2d86-8aee-4360-ac7b-ff64ad935845

  - backend

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


Navviiii (dismissed):

apps/backend/src/automations/engine/event-router.ts
```diff
@@ -78,10 +144,53 @@ class EventRouter {
...
+
+    let scope = eventScope(eventType, payload);
+    // Ticket events are exactly the events carrying boardIds, and the only ones whose wire
+    // payload lacks the ids — everything else is judged on wire data and never hydrates.
+    if (scope && 'boardIds' in scope) {
+      const impl = triggerRegistry.has(eventType) ? triggerRegistry.get(eventType) : null;
+      const hydrate = impl?.hydratePayload?.bind(impl);
+      if (!hydrate) return skip;
+      // Same tenant scope the worker hydrates under, so both see the same rows.
+      const hydrated = await runAsServiceActor('automation', workspaceId, () => hydrate(payload));
```
santhoshJus: **MEDIUM (perf) — full `hydratePayload` to read three scalar ids.**

This calls the trigger's *complete* hydration just so `eventScope` can read `boardId`/`projectId`/`channelId`. For `TICKET_COMMENTED` that fans out to `getTicketById` + `buildTicketContext` (board, project, channel, assignee, creator, group, plus up to 3 `email.findFirst` in `loadLastEmails`) + `message.findUnique` + `users.findById` + `getChannelMembership` — roughly a dozen queries — and the worker then repeats all of it for everything that survives.

This lands on the emit path for *every* ticket comment/update/create in any workspace that has even one board/project/channel-scoped ticket automation, whether or not anything actually gets filtered. So the PR trades N doomed queue rows for a full ticket-context build per event.

A targeted read covers the entire surface `eventScope` touches for ticket events:

```ts
const ticket = await db.ticket.findUnique({
  where: { id: payload.ticketId as string },
  select: { boardId: true, projectId: true, channelId: true },
});
```

(and it fails open the same way — `null` → all-`undefined` scope → no skips).

apps/backend/src/automations/engine/event-router.ts
```diff
@@ -6,14 +6,64 @@ import { currentUpstreamChain } from './automation-context-storage';
...
 import { automationQueue } from '../queue/automation.queue';
+import { triggerRegistry } from '../triggers/trigger-registry';
 import type { AutomationEvent } from '../types/automation-events';
 import { EMAIL_RECEIVED_EVENT } from '../triggers/email-received.trigger';
 
+// Filter keys this event's matcher honours as scope, mapped to the id the event carries. A key
+// the matcher ignores must never appear — judging it would drop runs that should have fired.
+// null = never pre-filtered: EMAIL_SENT's wire payload is { emailId } only and hydrating it
+// would add queries to the awaited reply-send path; WEBHOOK has no scope.
+function eventScope(
```
santhoshJus, line 45: **MEDIUM (drift risk, no test coverage) — `eventScope` is a hand-maintained duplicate of each trigger's matcher.**

It's keyed on hardcoded string literals rather than the exported constants (`TICKET_CREATED_EVENT`, `MESSAGE_RECEIVED_EVENT`, …), so the compiler can't catch divergence. It has *already* diverged: `TicketCommentedTrigger.matchFilters` honours `fromUserIds`, but `eventScope` omits it for `TICKET_COMMENTED` while including it for `MESSAGE_RECEIVED`.

That particular direction is safe (fail-open — fewer skips). The inverse is not: if any matcher's scope keys are ever OR-ed instead of AND-ed (e.g. “match tickets on any of these boards *or* channels”), `isDefiniteScopeMismatch`'s `.some()` starts silently dropping runs that should have fired — and there is no test file for the automations engine at all, so nothing would catch it.

The triggers already declare `scopeFilterFields` (`message-received.trigger.ts`, `tag-generated.trigger.ts`, `call.trigger.ts`). Deriving this map from the registry instead of a literal would remove the drift surface entirely; failing that, at minimum use the exported event constants as the `case` labels and add a test that asserts every registered trigger's `scopeFilterFields` is a superset of its `eventScope` keys.

apps/backend/src/automations/engine/event-router.ts
```diff
@@ -78,10 +144,53 @@ class EventRouter {
...
+      if (config?.trigger?.type !== eventType) continue;
+      const filter = (config.trigger.config ?? {}) as Record<string, unknown>;
+      if (keys.some(key => configuredIds(filter, key).length > 0)) scoped.push([workflow.id, filter]);
+    }
+    if (scoped.length === 0) return skip; // nothing to judge — never hydrate
+
+    let scope = eventScope(eventType, payload);
+    // Ticket events are exactly the events carrying boardIds, and the only ones whose wire
+    // payload lacks the ids — everything else is judged on wire data and never hydrates.
+    if (scope && 'boardIds' in scope) {
```
santhoshJus: **LOW/MEDIUM — `'boardIds' in scope` is an implicit “is this a ticket event” test.**

Two ways this degrades silently:

- If a future non-ticket trigger gains a `boardIds` scope key, it gets routed through `hydratePayload` for a payload shape `eventScope` doesn't understand.
- If a ticket event's scope ever drops `boardIds`, ticket events skip hydration and get judged on the un-hydrated wire payload — which carries only `ticketId` — so every scope field reads `undefined`, `isDefiniteScopeMismatch` returns false for everything, and the whole optimization quietly no-ops with no log.

An explicit event-type check, or a `needsHydration` flag returned alongside the scope, makes the intent enforceable rather than incidental.

apps/backend/src/automations/engine/event-router.ts
```diff
@@ -28,9 +78,25 @@ class EventRouter {
...
+    ).catch(err => {
+      logger.warn(`[EVENT-ROUTER] pre-filter failed for event=${eventType} — enqueuing all:`, err);
+      return new Set<string>();
+    });
+
     let enqueued = 0;
+    let preFiltered = 0;
 
     for (const workflow of candidates) {
+      if (skipIds.has(workflow.id)) {
```
santhoshJus, line 118: **LOW (behavior change under queue lag) — worth calling out in the PR description.**

The scope decision moves from run time to emit time. Concrete case: an automation scoped `boardIds: ['B']`; a comment lands while the ticket is still on board A; the ticket moves to board B a second later while the automation queue is backed up.

Before: the execution row existed, the worker hydrated at run time, saw board B, and fired.
After: it's skipped at emit and the run is lost.

Narrow, and the new semantics are arguably the more correct ones — but it is a real behavior change and not just an optimization, so it should be stated explicitly.

santhoshJus (commented):
Reviewed the pre-filter logic against every matcher it mirrors (`matchTicketScopeFilters`, `MessageReceivedTrigger.matchFilters`, `matchEmailReceived`, `matchTagGenerated`, `CallTrigger.matchFilters`) and against desk-rule configs (`buildRuleConfig` always sets `channelIds: [channelId]`, so desk rules are never wrongly skipped). **The skip logic itself is sound** — `.some()` correctly mirrors the AND-across-scope-keys semantics of every matcher, and every unknown (missing id, non-array config, unregistered trigger, hydration throw, ticket not found) resolves to “don’t skip”. No mass-drop bug found. Hydrating under `runAsServiceActor('automation', workspaceId, …)` is right — same actor the worker uses, so both see the same rows.

5 comments below: one perf issue worth fixing before merge (#1), one maintainability risk (#2), and three smaller notes.

apps/backend/src/automations/engine/event-router.ts
```diff
@@ -78,10 +144,53 @@ class EventRouter {
...
+  ): Promise<Set<string>> {
+    const skip = new Set<string>();
+    const keys = Object.keys(eventScope(eventType, {}) ?? {});
+    if (keys.length === 0) return skip;
+
+    // parseAutomationConfig is an unvalidated JSON.parse ("null" and "{}" both parse fine),
+    // so read defensively — a config that is not this event's trigger is left to the worker.
+    const scoped: Array<[string, Record<string, unknown>]> = [];
+    for (const workflow of candidates) {
+      const config = parseAutomationConfig(workflow.context) as {
```
santhoshJus, line 188: **LOW (over-fetch / CPU) — pre-existing cost that this PR moves onto the hot path.**

`findCandidates` (L206 / L215) does `db.workflow.findMany({ where })` with no `select`, so the full `context` blob (trigger definition + every step) and `metadata` are loaded for every active automation in the workspace on every event. This line then `JSON.parse`s `context` for each candidate — and the enqueue loop separately parses `metadata` for the survivors.

With the 50+ automations described in the PR body, that's 50 multi-KB parses per `MESSAGE_RECEIVED`, on top of the row fetch.

Two cheap wins: add `select: { id: true, workspaceId: true, workflowType: true, context: true, metadata: true }` to both `findMany` calls, and thread the config parsed here through to the enqueue loop so it isn't parsed twice.

apps/backend/src/automations/engine/event-router.ts
Revanthvenkatpasupuleti: not applicable as we are doing it even before adding into queue

santhoshJus (approved):

karunrajk (approved):

PR #596, opened 2026-08-14
feat: XYNE-17742 bulk ticket creation (backend)

XyneSpaces:
## What

Re-implements **bulk ticket creation** natively for the monorepo (originally scoped as XYNE-17742). A single request creates many tickets at once — either independent top-level tickets (`all-parents`) or sub-tickets hung under one parent (`parent-sub`). The heavy per-ticket work runs off the request path on a Bull worker; the caller gets an immediate `202` with a `jobKey` (and the `parentTicketId` for parent-sub mode).

## Why this is a re-implementation, not a port

The original change targeted an earlier, flatter lineage of this codebase. The monorepo has since diverged substantially (top-level `apps/`, no `NudgeKind` Postgres enum, an already-present sub-ticket / ticket-id / nudges framework). The feature was therefore rebuilt on current repo primitives rather than transplanted, and several security gaps in the original design were closed in the process.

## Changes (backend)

New files:
- `apps/backend/src/types/bulkTicket.ts` — `BulkTicketMode` enum + job/item payload shapes.
- `apps/backend/src/utils/channelAccess.ts` — per-channel authorization (exists + same workspace + private-channel membership).
- `apps/backend/src/queues/bulkTicketCreationQueue.ts` — Bull queue mirroring the `etaDeadlineQueue` pattern; lazy `initialize()`, `jobId` derived from the batch `jobKey`.
- `apps/backend/src/workers/bulkTicketCreationWorker.ts` — processes a batch, re-checks access per item, per-row Redis idempotency, single terminal failure-report site; maps sub-tickets via `subTicketService.createSubTicket`.

Edited:
- `apps/backend/src/controllers/ticketController.ts` — `createBulkTicketItem()` (reuses the transactional `createTicketWithConversation`) and the `createBulkTicket` route handler.
- `apps/backend/src/routes/tickets.ts` — `POST /api/tickets/bulk-from-message`.
- `apps/backend/src/worker.ts` — starts the bulk-ticket worker.

## Security hardening vs the original design

- **F1 — Per-item IDOR closed.** Each item carries its own `channelId`/`projectId`/`boardId`; every target channel is access-checked at the request layer AND again in the worker (fail-closed). The request is rejected as a whole if any channel is unreachable.
- **F2 — Idempotency.** Batch `jobId` is derived from `jobKey` (no double-enqueue); the worker records completed rows in a Redis set keyed by `jobKey`, so a stalled/retried job skips rows it already created instead of duplicating them.
- **F3 — Single failure report.** Failures are collected and reported once at the end of the run, not per item and not per retry attempt.
- **F4 — Transaction integrity.** Ticket + system message creation reuses `createTicketWithConversation`, which is `prisma.$transaction`-wrapped, so a failure cannot leave orphaned conversation/message rows.
- **F5 — App-auth exposure closed.** The endpoint is explicitly rejected on the `/api/tickets/claw` (app-credential) mount; identity is always taken from the session, never the body.

## Validation

- `tsc --noEmit` on `apps/backend` passes cleanly (0 errors).

## Deferred (follow-up)

- Dashboard bulk-create UI.
- The persisted failure-nudge surface.

Both diverged heavily from the source lineage and are intentionally out of scope for this backend slice; failures are currently surfaced via structured worker logs.


apps/backend/src/controllers/ticketController.ts
```diff
@@ -388,6 +426,302 @@ export class TicketController {
...
+        priority: item.priority,
+        statusV2: item.statusV2,
+        messageContent: creationText,
+        messageSubtype: 'bulk_ticket',
+        creationMessageId: initialMessageId,
+      });
+      await messageMetadataService.syncInitialMessageMd(conversation.conversationId);
+      return ticket;
+    } catch (error) {
+      await this.discardBulkConversation(conversation.conversationId);
```
PiyushJuspay, line 498: what this will discard. converation ? what about ticket ?

architjuspay: ticket creation is part of transaction so it is handled by postgres, this is only there because conversation is created before the transaction.

PiyushJuspay (dismissed):

PiyushJuspay (approved):

PiyushJuspay:
```
H1 Websocket counts snapshot has the wrong shape — can crash the Kanban counts hook
- What: The bulk path builds its own ticket snapshot for live Kanban counts. It sends assignments/formValues and omits qaAssigned, roleAssignments, formFieldValues. Every other writer uses makeFallbackCountsSnapshot (ticketRepository.ts:83) / KanbanCountsSnapshot, which always include those fields.
- Where: apps/backend/src/services/tickets/bulkTicketBatchService.ts:667 — assignments: [], formValues: [],
- Who affected: Any dashboard user viewing Kanban counts grouped by a form field or filtering by role assignments / dynamic fields.
- Impact: useKanbanCounts.ts:125/256/283 reads snapshot.formFieldValues[...] and snapshot.roleAssignments.find(...) without guards → TypeError, live counts break.
- Fix: Reuse makeFallbackCountsSnapshot (or buildKanbanCountsSnapshot) instead of the hand-rolled object.
- Helper comment:
bulkTicketBatchService.ts:667 — this snapshot is missing qaAssigned, roleAssignments and formFieldValues (it sends assignments/formValues, which nothing reads). The counts client (useKanbanCounts.ts:125) dereferences formFieldValues unguarded and will throw. Please reuse makeFallbackCountsSnapshot from ticketRepository.ts so the shape stays identical to single-ticket creation.
H2 Bulk tickets skip side effects every single-created ticket gets
- What: Single creation queues Vespa search indexing (ticketController.ts:1666), message classification (:1397), and usage tracking (:1668). The bulk path does none of these. sourceConversationId/sourceMessageId are accepted by the validator but silently ignored, so SDLC entity linking (linkCreatedEntities, :1323-1336) never runs either.
- Where: apps/backend/src/services/tickets/bulkTicketBatchService.ts:602 — void emitTicketCreated(
    { id: row.ticketId, workspaceId: row.workspaceId },
- Who affected: End users searching, AI classification features, analytics, SDLC-linked conversations.
- Impact: Bulk tickets are invisible in search, unclassified, untracked, and not linked to the SDLC entity the source conversation belongs to.
- Fix: Call pushVespaJobForTicket, messageClassificationQueue.enqueueForMessage, trackTicketCreated per row in fanOut, and honor sourceConversationId for entity linking (or reject the field).
- Helper comment:
bulkTicketBatchService.ts:602 — the fan-out skips Vespa indexing, message classification and activity tracking that createTicket does after commit (ticketController.ts:1397,1666,1668), and sourceConversationId/sourceMessageId are validated but never read, so entity linking is skipped too. Bulk-created tickets will be missing from search and SDLC links.
H3 Assigning a group in bulk leaves the ticket unassigned
- What: Single creation resolves a user from the group via evaluateAssignmentRule / full role assignment (ticketController.ts:1135-1156). The bulk path just stores userGroupId and leaves assignedTo null. The modal's assignee picker produces group:<id> values, so this is the common path.
- Where: apps/backend/src/services/tickets/bulkTicketBatchService.ts:482 — userGroupId: r.input.userGroupId ?? null,
- Who affected: End users assigning groups in the bulk modal.
- Impact: Tickets show as unassigned; assignment rules and board auto-assignment never run.
- Fix: Run evaluateAssignmentRule (or assignFullRolesToTicket) for rows with userGroupId and no assignedTo, same as the single path.
- Helper comment:
bulkTicketBatchService.ts:482 — single-ticket creation auto-assigns a user when only a userGroupId is given (ticketController.ts:1135-1156), but bulk stores the group and leaves the ticket unassigned. Since the modal's group picker produces exactly this input, bulk rows should run the same assignment rule.
H4 Sub-ticket links skip the activity entry and parent anchor card
- What: The canonical subTicketService.createSubTicket records a SUBTICKET_CREATED timeline event and calls linkSubTicketConversationToParent (parent card in the child thread) — subTicketService.ts:68-100. Bulk's commitSubTicketLinks does neither (it does sync sub_tickets_md).
- Where: apps/backend/src/services/tickets/bulkTicketBatchService.ts:795 — await syncConversationSubTicketsMd(tx, parentLink.id);
- Who affected: End users viewing ticket activity feeds and sub-ticket threads.
- Impact: No "sub-ticket created" activity/history rows; sub-ticket threads never show the parent anchor card.
- Fix: Also call linkSubTicketConversationToParent(tx, child.ticketId, parent.id) per child and write the SUBTICKET_CREATED activity, as createSubTicket does.
- Helper comment:
bulkTicketBatchService.ts:795 — besides syncConversationSubTicketsMd, subTicketService.createSubTicket also calls linkSubTicketConversationToParent and records a SUBTICKET_CREATED timeline activity. Bulk sub-tickets currently get neither, so their threads won't show the parent card and the activity feed stays empty.
H5 AI suggestions are never marked created after bulk creation — easy duplicates
- What: The single flow persists "created" via conversationService.markTicketSuggestionAsCreated (TicketSuggestions.tsx:103). The bulk flow only closes the modal. clientRowId is sent, validated, and then dropped by the backend, and sourceMessageId is ignored — so there is no server-side guard either (single path returns 409 via unique messageId, ticketController.ts:1113-1129).
- Where: apps/dashboard/src/components/ui/MessageBubble/TicketSuggestions.tsx:238 — onTicketCreated={() => {
          setIsBulkModalOpen(false);
        }}
- Who affected: End users using "Create all as sub-tickets" on AI suggestions.
- Impact: Suggestions stay selectable with the bulk button still visible; clicking again creates the whole batch a second time. The "created tickets" links never appear.
- Fix: After a successful bulk response, call markTicketSuggestionAsCreated for each created ticket (the response already returns ids/titles), or persist clientRowId server-side and use it for the update.
- Helper comment:
TicketSuggestions.tsx:238 — the bulk path never calls markTicketSuggestionAsCreated, so suggestions remain actionable and can be bulk-created again (the single flow blocks this, and the backend's sourceMessageId/clientRowId are accepted but ignored). Please mark each suggestion created using the ids in the bulk response.
H6 Multi-paragraph composer drafts now open the bulk modal instead of the ticket modal
- What: Any composer draft with ≥2 blank-line-separated blocks is rerouted to the bulk parent/sub modal. Paragraph breaks in a description are common, and those users previously got the single CreateTicketModal.
- Where: apps/dashboard/src/components/Chat/ChatInput/ChatInput.tsx:1230 — const drafts = parseTicketsFromText(description || '');
                      if (drafts.length >= 2) {
- Who affected: All chat users creating tickets with multi-paragraph descriptions.
- Impact: Users intending one ticket land in a bulk flow; confirming creates N tickets instead of one. Likely intentional, but it silently changes a very common existing flow.
- Fix: Keep blank-line splitting but require an explicit choice (e.g. a "create as multiple tickets?" prompt), or use a stricter delimiter (markdown list/headings) so paragraphs stay one ticket.
- Helper comment:
ChatInput.tsx:1230 — any blank line in the composer now reroutes the existing single-ticket flow to the bulk modal. Multi-paragraph descriptions are common; consider requiring an explicit "create multiple" confirmation or a stricter delimiter before splitting.
H7 FLOW-board guards from single creation are missing
- What: Single creation forces statusV2=TODO, stageName='TODO', and ticketType=Epic on FLOW boards (ticketController.ts:1068-1074). The bulk path has no board-type checks, and the modal lists every project board (via boardsListByProject) — including FLOW boards.
- Where: apps/backend/src/controllers/ticketController.ts:576 — const boards = await prisma.board.findMany({
      where: { id: { in: Array.from(boardIds) }, workspaceId },
- Who affected: End users bulk-creating into FLOW boards.
- Impact: Non-Epic, arbitrary-stage tickets on FLOW boards, which the flow cascade/views don't expect.
- Fix: Apply the same FLOW normalization in the bulk controller/service (or reject FLOW boards for bulk).
- Helper comment:
ticketController.ts:576 — the board lookup here doesn't apply the FLOW-board rules that single creation enforces (effectiveStatusV2/effectiveStageName/effectiveTicketType at lines 1068-1074), and the modal offers all project boards. Bulk rows on a FLOW board should be forced to TODO/Epic like the single path.
H8 Concurrent batches can collide on Kanban positions
- What: Column heads are read before the transaction opens; generateNKeysBetween below the same head returns identical keys. kanbanPosition has no unique constraint. Single creation has the same race but inside one short transaction; bulk widens the window to the whole prepare+commit.
- Where: apps/backend/src/services/tickets/bulkTicketBatchService.ts:207 — const head = await prisma.ticket.findFirst({
      where: { boardId, stageName, kanbanPosition: { not: null } },
- Who affected: Teams bulk-creating into the same board column at the same time.
- Impact: Duplicate fractional keys → unstable/ambiguous Kanban ordering between the batches (no data loss).
- Fix: Move the head read inside the batch transaction, or add a per-column advisory lock / retry-on-duplicate.
- Helper comment:
bulkTicketBatchService.ts:207 — reading the column head before the transaction opens means two concurrent batches in the same column generate identical kanban keys (no unique constraint). Consider reading heads inside the transaction or locking per column.
```

PR #625, opened 2026-08-16
feat: extract platform-agnostic Ask AI stream core into @xyne/shared

XyneSpaces:
## Summary
Phase 0 foundation for porting the dashboard "Ask AI" streaming chat to a fully native `apps/mobile` React Native surface. This PR extracts the **pure, transport-free heart** of the Ask AI stream into a new shared subpath so web and native share **one** serializer + SSE parser and cannot drift. **Zero behaviour change** on the dashboard.

## What's new — `@xyne/shared/askAI` (`packages/shared/src/askAI`)
- **`types.ts`** — `AskAIRequestInput` (camelCase request shape, single source of truth), the SSE event union (`start`/`delta`/`content`/`tool_input`/`tool_output`/`reasoning_delta`/`tool_invocation`/`debug_event`/`debug_artifacts_ready`/`complete`/`done`/`error`/`agent_update`/`genius_start`/`ping`), the live re-attach event union, and the `StreamTransport` / `StreamStore` seams (implemented per platform: web = worker+fetch+cookies+IndexedDB; native = mTLS+Bearer+MMKV/SQLite).
- **`requestBody.ts`** — `buildAskAIRequestBody`: camelCase input → snake_case wire body. A faithful, field-for-field port of the worker's inline serialization (same empty-array omission, same `?? false`/`?? null` defaults, same deliberately-camelCase `agentSlug` on the wire).
- **`sse.ts`** — `createAskAISSEParser`: incremental SSE framing / JSON decode / `ping` heartbeat drop, buffering frames split across chunks. Pure — no `fetch`, no `TextDecoder`, no globals; the caller owns byte decoding and the read loop.

## Dashboard rewire
`apps/dashboard/src/services/XyneAI/xyneAIStream.worker.ts` now imports `buildAskAIRequestBody` + `createAskAISSEParser` from `@xyne/shared/askAI` and drops its inline copies. The worker keeps all web-specific transport (fetch, cookie credentials, read loop, abort, `postMessage`). The worker's `requestBody` type is now the shared `AskAIRequestInput` (structurally identical — removes the duplicated type).

## Validation
- `pnpm --filter @xyne/shared build` — clean
- New behaviour-lock tests `packages/shared/test/askAI.test.mjs` — **8/8 pass** (pin the exact wire body + SSE framing so the extraction is provably zero-behaviour-change)
- `apps/dashboard` `tsc --noEmit` — clean
- worker `eslint` + `prettier --check` — clean

## Notes
- No XYNE ticket key was available at authoring time — happy to rename the branch / retitle to the team's `feature/ + XYNE-NNNNN` convention if one is assigned.
- This is Phase 0 step 1 (the pure, verifiable core). Remaining Phase 0 work — the stateful event-reducer normalization, the sessions REST client behind a shared HttpClient, and relocating the error map — is staged as follow-ups because each needs runtime validation of the full authenticated Ask AI flow.

### DIFF
--- a/apps/dashboard/src/services/XyneAI/xyneAIStream.worker.ts
+++ b/apps/dashboard/src/services/XyneAI/xyneAIStream.worker.ts
@@ -1,7 +1,19 @@
 /**
  * Web Worker for XyneAI Streaming
  * Runs API calls on a separate thread to avoid blocking the main UI thread
+ *
+ * The wire-body serialization and SSE framing now live in the platform-agnostic
+ * Ask AI stream core (`@xyne/shared/askAI`) so the dashboard (web) and the native
+ * mobile app share ONE serializer + parser and can't drift. This worker keeps the
+ * web-specific transport concerns: `fetch`, cookie credentials, the read loop,
+ * abort handling, and the worker<->main postMessage protocol.
  */
+
+import {
+  buildAskAIRequestBody,
+  createAskAISSEParser,
+  type AskAIRequestInput,
+} from '@xyne/shared/askAI';
 
 // Worker message types
 export interface WorkerStartStreamMessage {
@@ -9,59 +21,8 @@ export interface WorkerStartStreamMessage {
   payload: {
     streamId: string;
     url: string;
-    requestBody: {
-      query: string;
-      displayQuery?: string;
-      channelIds: string[];
-      collectionIds?: string[];
-      fileIds?: string[];
-      canvasIds?: string[];
-      ticketIds?: string[];
-      callIds?: string[];
-      attachedContext?: Array<{
-        type: 'channel' | 'ticket' | 'canvas' | 'call' | 'activity';
-        id: string;
-        title: string;
-        threadId?: string;
-        eventName?: string;
-        eventCategory?: string;
-        timestamp?: string;
-        metadata?: Record<string, unknown>;
-        relatedData?: Record<string, unknown>;
-      }>;
-      conversationId: string;
-      sessionId: string;
-      webSearchEnabled: boolean;
-      deepResearchEnabled?: boolean;
-      createCanvasEnabled?: boolean;
-      /** Single search + single answer pass instead of the full agentic tool
-       *  loop — see xyne-claw-auth's run-stream.ts POST / instant branch. */
-      instant?: boolean;
-      researchContext?: { type: string; id?: string; name: string } | null;
-      canvasId?: string;
-      messageAttachmentIds?: string[];
-      attachments?: Array<{
-        data: string;
-        mimeType: string;
-        filename: string;
-      }>;
-      parentMessageId?: string;
-      isRegenerate?: boolean;
-      // Branching: edit-user signals that the new user message is a sibling
-      // of `editedUserMessageId` under `parentAssistantMessageId` (the
-      // assistant parent the original lived under). claw-auth uses these to
-      // clone the PI session BEFORE the original user msg so the LLM session
-      // doesn't include the old turn as context.
-      isEditUserMessage?: boolean;
-      editedUserMessageId?: string;
-      parentAssistantMessageId?: string;
-      draftMode?: boolean;
-      version?: 'v1' | 'v2';
-      disableTools?: boolean;
-      agentSlug?: string;
-      /** Per-run model pin from the composer's model picker. */
-      model?: string;
-    };
+    // The camelCase Ask AI request shape is defined once in the shared core.
+    requestBody: AskAIRequestInput;
   };
 }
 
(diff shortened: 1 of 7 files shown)


XyneSpaces (commented):
---

## Architecture review — Ask AI shared stream core (Phase 0, step 1)

Reviewed against the agreed plan to move Ask AI logic into `@xyne/shared` so the native `apps/mobile` surface reuses one serializer + parser and cannot drift from the dashboard. **Verdict: approve the direction, merge after the notes below are acknowledged.** This is a clean, faithful, well-tested first slice — but it is only the *pure framing* layer. The piece that actually prevents web/native drift (the stateful reducer) is not here yet, so please land it as the immediate follow-up rather than starting native transport against these seams.

### What's right (verified against the diff, not just the description)
- **Faithful extraction, provably zero-behaviour-change.** `packages/shared/src/askAI/requestBody.ts` is a field-for-field port of the worker's old inline serialization — I diffed `main...feature/ask-ai-shared-stream-core` on `apps/dashboard/src/services/XyneAI/xyneAIStream.worker.ts` and the removed inline body matches the shared builder exactly, including the load-bearing quirks: `channel_ids` always present (even `[]`), every other array key omitted when empty, `deep_research_enabled`/`create_canvas_enabled`/`instant`/`research_context` always emitted with `?? false`/`?? null`, `attachments` re-keyed `mimeType`→`mime_type`, and `agentSlug` deliberately left camelCase on the wire.
- **The behaviour-lock tests are the right kind of test.** `packages/shared/test/askAI.test.mjs` pins the wire contract (defaults, empty-array omission, camelCase `agentSlug`) and the SSE framing (ping drop, split-frame buffering, non-object/malformed handling via `onError`). Importing from `dist/` means it tests the actual artifact consumers resolve.
- **Good forward-compat typing.** `AskAIStreamEvent.type: AskAIStreamEventType | (string & {})` with an open `[key: string]: unknown` payload means a new backend event never breaks parsing — correct call for a contract that grows over time.
- **Seams are in the right place.** `StreamTransport` / `StreamStore` cleanly isolate the two things that genuinely differ per platform (web worker+fetch+cookies+IndexedDB vs native mTLS+Bearer+MMKV/SQLite).
- **Cross-repo path works with no allowlist change.** `apps/mobile` pulls `@xyne/shared` via `file:../../shared`, materialized by `scripts/bootstrap.sh` which fetches the whole public `packages/shared` at `public-source.json` `ref` and builds it. So the new `./askAI` subpath export flows to mobile automatically once this merges to `main` and the private repo re-bootstraps — nothing extra to register.

### Gaps / required follow-ups (ordered by impact on the native port)

1. **The stateful reducer is the real anti-drift asset and it is not extracted yet.** Today `sse.ts` only frames bytes and drops `ping`; all the semantics — the event `type` switch, `start` participant/bot-message creation, `delta`/`content` accumulation, `reasoning_delta` append, `tool_input` status generation (incl. the `fetch_channel_messages` → summarizer special-case), **`tool_invocation` dedupe-by-`toolCallId`**, and the `complete` **local↔server message-ID reconciliation** for the branch tree — still live in `apps/dashboard/src/services/XyneAI/XyneAIStreamManager.ts`. If native reimplements that switch, the two platforms *will* diverge on the exact thing this PR set out to prevent. Please make `packages/shared/src/askAI/reducer.ts` (pure `reduce(state, event) → state`) the very next PR, before any native transport work.

2. **Prove the seams with the web side first.** `StreamTransport` / `StreamStore` are declared but nothing implements them yet — the worker still uses the raw `postMessage` protocol and `XyneAIStreamStorage.ts` (IndexedDB) is untouched. An interface only ever implemented by mobile risks being discovered wrong on mobile. Recommend wrapping the existing worker as a `StreamTransport` and `XyneAIStreamStorage` as a `StreamStore` in the same phase, so each seam has ≥1 real implementation validating it before native builds against it.

3. **`liveReducer` for `/v2/.../live` re-attach.** `AskAILiveEvent` (`snapshot`/`delta`/`reasoning`/`invocation`/`label`/`done`/`live-disabled`) is typed but has no reducer. Native needs this for AppState-foreground resume of an answer that kept running server-side — track it alongside the main reducer.

4. **Type duplication to fold before it spreads.** `AskAIResearchContextInput` (`type: string`) duplicates the existing `ResearchContext` in `packages/shared/src/types/research.ts` (`type: 'product' | 'repository'`) — and it's *looser*. Two research-context shapes inside the same shared package is exactly the drift this effort is fighting. Prefer re-exporting/aliasing the existing type (or intentionally widening the canonical one), not defining a second.

5. **CI ordering.** The test imports `../dist/askAI/*.js`, so `pnpm --filter @xyne/shared build` must run before `test` in CI or it fails on a fresh checkout — please confirm that ordering exists.

6. **Branch/title convention.** Per team convention (`feature/ + XYNE-NNNNN`), retitle/rename once a ticket key is assigned (already flagged in the description).

### Perf note
No concern in this PR — `buffer.split("\n")` + retain-remainder is identical to the original and runs in the worker. Just recording for the native port: high-frequency `delta` batching (rAF on web) belongs in the per-platform transport/UI layer, not in the shared parser — keep the shared core allocation-light and let each platform batch.

Net: merge this as the foundation, then land the **reducer + web-side seam implementations** next so the native port builds on a fully shared, already-proven core.

PR #666, opened 2026-08-17
feat: agent-automations module (Phase 1) in xyne-claw-auth

XyneSpaces:
## What & why

Phase 1 of a new **`agent-automations`** module in `xyne-claw-auth` — self-proposed, event-driven **agent wakeups**. An agent proposes an automation ("wake me on this thread when a comment lands on PR #123"); a human approves it; a **generic signed webhook** then wakes the agent **inside the original conversation** each time a matching external event arrives.

The design **copies xyne-spaces' generic external `WEBHOOK` trigger** (unique URL = identity, encrypted rotatable secret, declared body/header schema, async 202) rather than shipping per-vendor (GitHub/Stripe) adapters. Ingress stays vendor-agnostic.

This change is **additive** — no existing behavior is modified. All new code lives under `apps/xyne-claw-auth/backend`.

## Three deliberate divergences from existing primitives

1. **Runs continue the ORIGINAL `conversationId`.** A scheduled job mints a fresh `scheduled_<id>_<ts>` session; an agent-automation dispatches into the thread the agent proposed on, so it resumes with full history. Dispatch reuses the exact `/internal/run` + `x-s2s-key` contract from `queue/scheduled-jobs-worker.ts`.
2. **Delivery-id idempotency** via `@@unique([automationId, deliveryId])` — a retried webhook loses the insert race (P2002) and returns `202 duplicate` instead of waking the agent twice. (n8n has no dedup; this is the most important addition.)
3. **Optional `matchPredicate`** — a flat dot-path→value predicate scopes a repo-level webhook to one resource before spending an agent run.

## What's included

- **Prisma**: `AgentAutomation` + `AgentAutomationRun` (follows the `ScheduledJob` `@@map` convention).
- **`agent-automations/`**: `declared-schema.ts` (verbatim port of xyne-spaces `assertMatchesSchema`, dependency-free) + tests; `predicate.ts` + tests; `secret.ts` (AES-256-GCM via existing `crypto.ts`, timing-safe compare, rotatable); `dispatch.ts`.
- **`routes/agent-automations.ts`**: three routers — public webhook ingress (auth = URL secret), authed management (propose / approve / list / revoke), S2S result callback — wired into `main.ts`.
- **README** documenting the pattern, the ingress decision flow, the security model, and Phase-2 items.

## Security model

- URL secret is the auth: 32 random bytes, encrypted at rest, timing-safe compared, shown once at approval.
- **HITL activation**: `POST /:id/approve` requires an interactive `x-user-id`; an S2S/agent call has none, so **an agent cannot self-activate its own proposal**.
- Runs execute **as the owning user** (`createdByUserId`), never elevated. `authorization`/`cookie` headers are stripped before entering the run context.

## Validation

- Unit tests **15/15 passing** (`declared-schema`, `predicate`) via vitest.
- `prisma generate`/`validate` could not complete in the offline sandbox (engine download blocked) — schema follows the verified `ScheduledJob` convention 1:1; please run `prisma generate` + `prisma migrate` locally when reviewing.

## Phase-2 / owner review needed

- The `propose-automation` runtime **tool** (thin wrapper over `POST /agent-automations`) — only the HTTP route exists here.
- Confirm the run loader rehydrates a long `conversationId` for an event-triggered run (thread-continuity assumption).
- Where the Approve / secret-URL UI is surfaced.
- `AUTH_SERVICE_PUBLIC_URL` env for the issued webhook URL (falls back to internal URL, marked in code).
- Optional pluggable per-source signature verifiers (GitHub HMAC etc.), keyed off `source`.


XyneSpaces:
## Module is now complete (no longer phase-wise)

The Phase-1 stubs have been replaced with the full implementation in commit `8476d8c`. This module is done end-to-end.

**Added on top of the initial scaffold:**
- **`propose-automation` runtime tool** (`packages/xyne-claw-shared/src/tools/agent-automations/`) — schedule-task template; S2S POST to claw-auth; identity from `context.meta`; registered in `registry.ts`.
- **Per-source signature verification** (`verify.ts`) — pluggable, fail-closed SECOND gate on top of the URL secret: `github-hmac-sha256`, `hmac-sha256`, `header-token`. Byte-exact HMAC over `req.rawBody`. +8 unit tests.
- **rotate-secret** management route (interactive-only).
- **EXPIRED** lazy transition at ingress so expired automations stop matching.
- **secret.ts** `sealSecret`/`openSecret` for the encrypted-at-rest signing secret (decrypt-at-verify-time).
- **schema.prisma** `verifySource` / `signingSecret` / `signatureHeader`.
- **Real create-table Prisma migration** for both models.

**Validation:** 23/23 unit tests passing (declared-schema 10, predicate 5, verify 8). Backend + shared packages typecheck clean.

**Follow-ups (owner):**
- Run `prisma generate` + apply the migration locally — the offline build sandbox cannot fetch the Prisma engine.
- Confirm `/internal/run` rehydrates a long pre-existing `conversationId` for thread continuity.
- Set `AUTH_SERVICE_PUBLIC_URL` to the externally reachable base.
- This PR was opened via API and has **zero reviewers** — please add them manually.

PR #676, opened 2026-08-18
fix: XYNE-56394 close ACL channel/org gaps + SNS webhook SSRF

XyneSpaces:
## XYNE-56394 — Security fixes

Closes 9 of the 11 confirmed findings in XYNE-56394 (2 High, 7 Medium). All changes are on the Zero mutation ACL / webhook path in `apps/backend`.

### Fixed
| ID | Finding | Fix |
|----|---------|-----|
| **H1** | org-members cross-org privilege escalation | `canInsert/canUpdate/canDelete` now assert the caller's active membership is in the **same org** as the target (`assertActorInOrg` via `ctx.memberId`). |
| **H2** | SNS webhook SSRF (`confirmSnsSubscription`) | `SubscribeURL` pinned to `https` + `sns.<region>.amazonaws.com`, then `assertWebhookUrlSafe()` before `fetch`. |
| **M3** | whisper `parentMessage` leak | every `.related('parentMessage')` now carries the same `visibleTo IS NULL OR = ctx.userID` predicate as the top-level query. |
| **M4–M9** | channel-gate family (impacts, rcas, coes, release_attributions, classification_mappings, ticket_stage_requests) | verify channel/ticket participation on update/delete (+insert/upsert where relevant) and re-verify repointed FKs, matching the `ticket-assignments-acl` reference pattern. Previously workspace-only (ticket_stage_requests only on delete). |

### Proof of Testing
`apps/backend/src/zero/acl/tables/security-fixes-xyne-56394.test.ts` — **12/12 pass**. Stashing the source fixes and re-running makes the 7 "rejects" cases fail with _"Received promise resolved instead of rejected"_, proving the pre-fix code let the unauthorized actor (non-participant / cross-org admin) through. `tsc --noEmit` passes clean.

### Deferred (need schema/design change — not in this PR)
- **M1** marketplace-app `signingSecret`/`webhookUrl` exposure — Zero syncs whole rows; real fix removes columns from the client schema + moves app-signing read server-side (app-framework migration).
- **M2** `FORM_ENTITY_VALUE`/`IMPACT` attachment reads — currently workspace-bounded by design; closing needs entity→channel resolution + tests to avoid over-blocking.


XyneSpaces:
### Update: M2 now fixed (commit c309d1a) — coverage is 10/11

**M2 — IMPACT / FORM_ENTITY_VALUE attachment download leak — FIXED**

`assertAttachmentAccess` (attachment download endpoint) previously fell through to a **workspace-only** check for these two non-chat attachment types, so any workspace member could download a **PRIVATE-channel** ticket's impact / stage-form documents just by knowing the attachment id. The Zero read path was already ACL-gated; the REST download sink was not.

Fix: resolve the owning ticket's channel and require **PUBLIC** visibility **or** channel participation — mirroring the existing chat-attachment branch. A non-ticket-scoped form value (e.g. a `USER`-scoped `FORM_ENTITY_VALUE`) keeps the workspace-bounded default, so no legitimate read is over-blocked.
- `apps/backend/src/controllers/attachmentController.ts` — new gate + `resolveTicketIdForAttachmentEntity` helper.

**POT:** `apps/backend/src/controllers/attachmentController.security.test.ts` — **5/5 pass**. Before/after control: stashing the fix makes both `denies` cases return `ok:true` (`Expected: false / Received: true`), proving the pre-fix leak. `tsc --noEmit` clean.

**M1 — marketplace-app column exposure — remains deferred (verified rationale + concrete follow-up)**
- The only genuinely sensitive column, `signingSecret`, is **encrypted at rest** and has **zero client readers** (the dashboard only writes `signingSecret: ''` or reads it once from the REST create response; the server reads it via Prisma / `$queryRaw`, never via Zero). So the follow-up fix — **drop `signingSecret` from the client Zero schema** (`packages/shared/src/zero/schema.ts` `apps` + `installed_apps`) — is clean, but it is a **sync-contract change** that changes the schema hash and forces a client re-hydration on deploy. That belongs in its own schema-version-bumped PR, not this ACL/SSRF change.
- `webhookUrl` is genuinely consumed by the owner UI (`EditAppForm`/`AppsTable`), so it cannot simply be dropped; its cross-workspace exposure on GLOBAL marketplace apps needs the marketplace row-visibility redesign.

**Status: 10 of 11 findings fixed** (H1, H2, M2–M9). M1 is the single documented deferral with a scoped follow-up.

XyneSpaces:
### Update: M1 secret-egress half now fixed (commit 1d5c5d8)

**M1 — marketplace-app column exposure — secret egress CLOSED; plaintext-URL residual documented.**

The app-level HMAC `signingSecret` (server-encrypted at rest) was declared on the **client-synced** Zero `apps` table, so its ciphertext was replicated to every workspace that can see a GLOBAL marketplace app. Removed the column from the client schema:
- `packages/shared/src/zero/schema.ts` — `appsTable` no longer declares `signingSecret`.

**Why this is safe (verified, not assumed):**
- **No client reader** of `apps.signingSecret` — the dashboard only writes `signingSecret: ''` on the *installed_apps* view-model (`AppsScreen.tsx:187`) or reads the one-time REST create response (`AppsScreen.tsx:338`); neither is the synced app row.
- **Write path intact** — the `apps` Zero mutator (`mutators.ts`) writes only `name/description/webhookUrl/version`, never `signingSecret`.
- **Server unaffected** — the secret is read server-side via Prisma / `$queryRaw` (`flowController.ts`, `appUtils.ts`), not through the Zero schema.
- **No DB migration** — the Postgres/Prisma column is untouched; clients simply pick up the new row shape on the next deploy's normal Zero re-sync.
- **Client bundle compiles against this schema** — `@xyne/shared` `schema` (`ZeroProvider.tsx`, `dashboard/src/zero/queries.ts`), so the edit actually removes egress.

**POT:** rebuilt `@xyne/shared`; **backend `tsc --noEmit` clean (0 errors)** and **dashboard `tsc -p tsconfig.app.json --noEmit` clean (0 errors)**. `dist` confirms `appsTable` no longer carries the column.

**Residual (documented, lower severity):** the *plaintext* `webhookUrl` on GLOBAL marketplace apps is still visible cross-workspace. It is genuinely consumed by the owner UI (`EditAppForm`/`AppsTable`), so it cannot simply be dropped. The correct fix is to enroll it in the repo's existing **staged field-encryption** mechanism (`packages/shared/src/zero/encrypted-fields.ts`, `encryptedFieldsConfig`, Stage 8→9→10 with `ZERO_ENC_CLIENT_ENCRYPTION_ENABLED` + backfill) so non-owning workspaces receive ciphertext, or to row-scope GLOBAL-app visibility. That is a staged migration and should be its own PR — not this ACL/SSRF change.

**Status: 11 of 11 findings addressed** — H1, H2, M2–M9 fully fixed; **M1 secret egress closed**, with only the plaintext-`webhookUrl` marketplace visibility remaining as a scoped, documented follow-up.

anuragdvd (approved):

anuragdvd (approved):

devesh-juspay (approved):

PR #689, opened 2026-08-18
fix: XYNE-56405 discard stale out-of-order vespaSearch responses

MukherjeeSourish:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

XYNE-56405 — typing `from:<user>` in the command palette could blank the search
results ("black screen").

`performSearch` awaited `vespaSearch` and committed the response unconditionally,
with nothing sequencing or cancelling in-flight requests. One typing sequence
fires two searches — the literal `q=from` text query, then the resolved `from:`
filter query — so a slow, stale response could resolve after a newer one and
overwrite fresh results with an empty payload. Last write wins, regardless of
dispatch order.

**1. Monotonic sequence guard** (`useSearchMetrics.performSearch`). Each run claims
`seq = ++searchSeqRef.current`; only the latest run may commit results, flip the
loading flag, or surface an error. The guard also covers the `catch` path — without
it, a superseded request's failure (including the `CanceledError` raised by the
abort below) would run `setSearchResults([])` and blank the screen just the same.
This alone fixes the reported bug.

**2. AbortController** threaded into `SearchService.vespaSearch` via an optional
`signal`, so the superseded request is cancelled on the next keystroke instead of
running a multi-second backend query to completion. This is an efficiency win, and
it is only safe *because* of the guard above.

Known gap, left as a follow-up: `loadMoreResults` issues its own `vespaSearch` with
no signal and no sequence check, so a late pagination response can still append rows
from a previous query. Narrower trigger path; kept out of scope.

## Areas Touched

- [x] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

## Schema & Data Changes

## Config, Environment & URLs

## API Contract

- [x] No API changes

`vespaSearch` gains an optional second parameter (`signal?: AbortSignal`) — additive,
existing call sites are unaffected. No change to the HTTP contract.

---

## How did you test it?

Reproduced the bug and verified the fix against the full local stack (backend :3001,
dashboard :5173, Vespa, Postgres, zero-cache), driving the real UI with Playwright.

To make an inherently timing-dependent bug deterministic, the driver patches
`XMLHttpRequest.prototype.send` to *hold* one `/vespaSearch` request in flight and
release it on command. Nothing about the app is stubbed: the released request hits
the real backend and gets a real response, so the only thing controlled is arrival
order — exactly what a congested network does.

**Before (this branch's parent):** query text `kubernetes` on screen with 2 results;
the stale `q=zzzz` response lands with 0 results and wipes the list. Blank panel, no
error, query text unchanged. Matches the reported symptom.

**After (this branch), both defences verified independently:**

1. Stale request is cancelled by the newer search — never reaches the server.
2. With `AbortController.prototype.abort` no-op'd in the page so cancellation cannot
   help, the stale response is forced to arrive last (`q="zzzz" 146 ms · 0 results`
   after `q="kubernetes" 147 ms · 2 results`) and is discarded — fresh results stay
   on screen. The run asserts this programmatically and fails if the list empties.

Screen recordings of both runs attached (before / after), with an on-screen log of
every `vespaSearch` call, its duration and result count.

### Automated

- [x] Not covered by tests

The dashboard has no unit-test runner (no vitest/jest), and a `.mjs` repro placed
under `src/` breaks `pnpm run lint` — a non-TS file isn't in the typed-lint project,
so `@typescript-eslint/await-thenable` throws and takes the whole lint run down.
The Playwright repro/verification script therefore lives outside the repo. Happy to
land it under `tools/xyne-automation` as a proper e2e case if reviewers want it.

### Manual

**Users**
- [x] Single user
- [ ] Two users at once — not relevant, client-side render race with no shared state

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [x] Local dev (backend + dashboard dev servers)
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Sync & resilience** — n/a, no Zero involvement

**Surface & data**
- [x] Web browser
- [ ] Electron desktop
- [ ] Narrow viewport, dark + light theme — not applicable, no rendering change
- [x] Empty state — the stale payload *is* the empty case; it no longer clears results
- [x] Error paths — cancelled/superseded request rejects into `catch`; verified it no
      longer clears fresh results or flips the loading flag off early

---

## Checklist

- [ ] `pnpm run build:shared` passes — not run; no shared package touched
- [ ] Backend `typecheck` + `build` pass — not run; no backend files touched
- [ ] Dashboard `lint:errors-only` + `typecheck` + `build` pass — see note below
- [x] New deps added with `pnpm --filter <pkg> add <dep>` — n/a, no new deps
- [x] Commits follow `<type>: <TICKET-ID> <subject>`; branch is `fix/*`
- [x] Docs / guidelines updated where behaviour changed — n/a, no behaviour docs affected

**Note on dashboard checks:** `lint:errors-only` reports 118 errors and `tsc --noEmit`
reports 119 errors in 22 files — **identical counts on an unmodified branch off the
same base**, and none in the two files this PR touches. Both are pre-existing repo
state, not introduced here. `prettier --check` is clean on the changed files.
`build` not run.

[xyne-search-stale-response-race.webm](https://github.com/user-attachments/assets/c39c085a-3dbe-4401-8d7e-4d0e61a460b3)

[xyne-search-race-AFTER-fix.webm](https://github.com/user-attachments/assets/bf5201f6-6e38-4355-80f2-8667fdf020cd)

### DIFF
--- a/apps/dashboard/src/hooks/useSearchMetrics.ts
+++ b/apps/dashboard/src/hooks/useSearchMetrics.ts
@@ -540,6 +540,10 @@ export function useSearchMetrics(options: UseSearchMetricsOptions = {}) {
 
   const pendingSearchCountRef = useRef(0);
   const latestResultsRef = useRef<DisplaySearchResult[]>([]);
+  // The query that produced latestResultsRef, so onComplete always receives a
+  // matching pair. Without it a superseded run — which no longer commits its own
+  // results — would hand the callback fresh results labelled with its stale query.
+  const latestQueryRef = useRef('');
   const sessionFiltersRef = useRef<Set<string>>(new Set());
   // Guards out-of-order search responses. Each performSearch run claims the next
   // sequence number; only the latest run is allowed to commit. A slow/stale response
@@ -1407,6 +1411,7 @@ export function useSearchMetrics(options: UseSearchMetricsOptions = {}) {
 
             setSearchResults(mergedResults);
             latestResultsRef.current = mergedResults;
+            latestQueryRef.current = searchText;
 
             setPaginationState(prev => ({
               ...prev,
@@ -1435,12 +1440,16 @@ export function useSearchMetrics(options: UseSearchMetricsOptions = {}) {
           if (!isStale()) {
             setIsSearching(false);
           }
+          // Fires once every dispatched search has settled — including when a
+          // superseded run is the last to settle — so it must report the results
+          // that were actually committed and the query they came from, not this
+          // run's (possibly stale) searchText.
           if (
             pendingSearchCountRef.current === 0 &&
             latestResultsRef.current.length > 0 &&
             onComplete
           ) {
-            onComplete(latestResultsRef.current, searchText);
+            onComplete(latestResultsRef.current, latestQueryRef.current);
           }
         }
       } else {
--- a/apps/dashboard/src/hooks/useSearchMetrics.ts
+++ b/apps/dashboard/src/hooks/useSearchMetrics.ts
@@ -541,6 +541,13 @@ export function useSearchMetrics(options: UseSearchMetricsOptions = {}) {
   const pendingSearchCountRef = useRef(0);
   const latestResultsRef = useRef<DisplaySearchResult[]>([]);
   const sessionFiltersRef = useRef<Set<string>>(new Set());
+  // Guards out-of-order search responses. Each performSearch run claims the next
+  // sequence number; only the latest run is allowed to commit. A slow/stale response
+  // (e.g. a partial `from` query that resolves seconds after the completed `from:`
+  // filter) is discarded instead of overwriting fresh results with an empty payload.
+  const searchSeqRef = useRef(0);
+  // Cancels the previous in-flight vespaSearch when a newer search is dispatched.
+  const searchAbortRef = useRef<AbortController | null>(null);
 
   /**
    * Generate a unique session ID
@@ -1000,6 +1007,14 @@ export function useSearchMetrics(options: UseSearchMetricsOptions = {}) {
       }>,
       onComplete?: (results: DisplaySearchResult[], query: string) => void,
     ) => {
+      // Claim this run's sequence and abort any previous in-flight request so a slow,
+      // stale response can neither waste the network nor overwrite fresh results.
+      const seq = ++searchSeqRef.current;
+      const isStale = () => seq !== searchSeqRef.current;
+      searchAbortRef.current?.abort();
+      const abortController = new AbortController();
+      searchAbortRef.current = abortController;
+
       const {
         searchText,
         board: boardFilter,
@@ -1315,18 +1330,24 @@ export function useSearchMetrics(options: UseSearchMetricsOptions = {}) {
 
             if (activeTab === TabType.ALL) {
               const currentSessionId = searchSessionId || '';
-              const vespaResponse = await searchService.vespaSearch({
-                ...searchFilters,
-                //unified rank profile filters
-                ...(effectiveRankProfile === 'unified'
-                  ? {
-                      groupBy: '',
-                      apps: `${VespaApps.CHAT},${VespaApps.TICKET},${VespaApps.FILE}`,
-                    }
-                  : options.groupByDocType && { groupBy: 'docType' }),
-                searchId: currentSessionId,
-                presentationSummary: 'lean',
-              });
+              const vespaResponse = await searchService.vespaSearch(
+                {
+                  ...searchFilters,
+                  //unified rank profile filters
+                  ...(effectiveRankProfile === 'unified'
+                    ? {
+                        groupBy: '',
+                        apps: `${VespaApps.CHAT},${VespaApps.TICKET},${VespaApps.FILE}`,
+                      }
+                    : options.groupByDocType && { groupBy: 'docType' }),
+                  searchId: currentSessionId,
+                  presentationSummary: 'lean',
+                },
+                abortController.signal,
+              );
+
+              // A newer search superseded this one — drop this out-of-order response.
+              if (isStale()) return;
 
               // Honor the backend grouping decision: flat response => flat ALL view.
               setIsGrouped(vespaResponse.grouped);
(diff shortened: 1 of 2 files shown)


nasimshk (dismissed):

apps/dashboard/src/hooks/useSearchMetrics.ts
```diff
@@ -905,6 +912,14 @@ export function useSearchMetrics(options: UseSearchMetricsOptions = {}) {
       }>,
       onComplete?: (results: DisplaySearchResult[], query: string) => void,
     ) => {
+      // Claim this run's sequence and abort any previous in-flight request so a slow,
+      // stale response can neither waste the network nor overwrite fresh results.
+      const seq = ++searchSeqRef.current;
+      const isStale = () => seq !== searchSeqRef.current;
+      searchAbortRef.current?.abort();
+      const abortController = new AbortController();
```
XyneSpaces, line 1019: **[High] Same race is left unfixed in `loadMoreResults`.** This `AbortController` + `isStale()` guard only covers `performSearch`. `loadMoreResults` (line ~1637) awaits its own `searchService.vespaSearch(...)` with no `signal` and no `isStale()` check, then appends via `setSearchResults(prev => [...prev, ...results.results])`. A search dispatched while a page load is in flight can still append a stale page onto fresh results. Thread the same `abortController.signal` + a staleness guard into `loadMoreResults`, or explicitly note it as a tracked follow-up in the PR description.

apps/dashboard/src/hooks/useSearchMetrics.ts
```diff
@@ -1297,14 +1325,21 @@ export function useSearchMetrics(options: UseSearchMetricsOptions = {}) {
...
+            // stale results in the previous grouped/flat mode.
+            setIsGrouped(true);
+          }
         } finally {
-          setIsSearching(false);
           pendingSearchCountRef.current -= 1;
+          // Only the latest run may flip the shared loading flag off.
+          if (!isStale()) {
+            setIsSearching(false);
+          }
```
XyneSpaces, line 1442: **[Medium] `onComplete` in the `finally` is not stale-guarded.** Only `setIsSearching(false)` is wrapped in `if (!isStale())`. The `onComplete(latestResultsRef.current, searchText)` call at line ~1350 is not: when a superseded run is the one that drives `pendingSearchCountRef` to 0, it fires `onComplete` with THIS run's stale `searchText` while `latestResultsRef.current` holds the fresh results — a query/results mismatch delivered to the callback. Guard the `onComplete` call with `!isStale()`, or pass the latest committed query instead of this run's `searchText`.

apps/dashboard/src/hooks/useSearchMetrics.ts
```diff
@@ -446,6 +446,13 @@ export function useSearchMetrics(options: UseSearchMetricsOptions = {}) {
   const pendingSearchCountRef = useRef(0);
   const latestResultsRef = useRef<DisplaySearchResult[]>([]);
   const sessionFiltersRef = useRef<Set<string>>(new Set());
+  // Guards out-of-order search responses. Each performSearch run claims the next
+  // sequence number; only the latest run is allowed to commit. A slow/stale response
+  // (e.g. a partial `from` query that resolves seconds after the completed `from:`
+  // filter) is discarded instead of overwriting fresh results with an empty payload.
+  const searchSeqRef = useRef(0);
+  // Cancels the previous in-flight vespaSearch when a newer search is dispatched.
+  const searchAbortRef = useRef<AbortController | null>(null);
```
XyneSpaces, line 554: **[Nit] `searchAbortRef` is never aborted on unmount.** It only cancels the previous search when a new one starts. If the component unmounts mid-request, the resolving `vespaSearch` still runs `setIsGrouped`/`setSearchResults` (nothing bumped `seq`, so `isStale()` is false), i.e. setState on an unmounted component. Add an effect cleanup: `useEffect(() => () => searchAbortRef.current?.abort(), [])`.

apps/dashboard/src/hooks/useSearchMetrics.ts
```diff
@@ -446,6 +446,13 @@ export function useSearchMetrics(options: UseSearchMetricsOptions = {}) {
   const pendingSearchCountRef = useRef(0);
   const latestResultsRef = useRef<DisplaySearchResult[]>([]);
   const sessionFiltersRef = useRef<Set<string>>(new Set());
+  // Guards out-of-order search responses. Each performSearch run claims the next
+  // sequence number; only the latest run is allowed to commit. A slow/stale response
+  // (e.g. a partial `from` query that resolves seconds after the completed `from:`
+  // filter) is discarded instead of overwriting fresh results with an empty payload.
+  const searchSeqRef = useRef(0);
```
XyneSpaces, line 552: **[Nit] No automated test or POT for the race fix.** Per the repo's testing guidelines (add unit tests for new edge cases; include a proof-of-testing link for behavioural fixes), add a test that resolves two `vespaSearch` promises out of order and asserts the stale/empty payload is discarded, or attach a POT screen recording to the PR description.

XyneSpaces (commented):
Reviewed the stale-response fix. The core defect is correctly addressed: `performSearch` now claims a monotonic `seq`, aborts the previous in-flight request, and drops out-of-order responses after every await and in catch/finally — this resolves the `from:<user>` black-screen. Well-commented and typed.

Changes requested (nothing Critical):
- **High** — the same race remains in `loadMoreResults` (separate unguarded `vespaSearch`); either extend the guard or note it as a tracked follow-up.
- **Medium** — `onComplete` in the `finally` isn't stale-guarded and can deliver a stale query with fresh results.
- **Nits** — no unmount abort (setState-after-unmount risk); no test/POT for the fix.

`searchService.ts` change is correct — the conditional `...(signal ? { signal } : {})` spread properly satisfies `exactOptionalPropertyTypes`.

VishalM-23 (dismissed):

mamtha-v (dismissed):

apps/dashboard/src/hooks/useSearchMetrics.ts
MukherjeeSourish: pre-existing, untouched. loadMoreResults never had a signal or any staleness concept; before your PR nothing in the hook did. The PR neither created this race nor made it more reachable. It only made it visible, because now there's a guard right next to it that this path doesn't use. Fixing it is an extension of your change, not a repair of it.

apps/dashboard/src/hooks/useSearchMetrics.ts
MukherjeeSourish: We didn't make anything worse; We just didn't take the extra win while you were there.

apps/dashboard/src/hooks/useSearchMetrics.ts
MukherjeeSourish: harmless

Pradeesh333 (approved):

mamtha-v (approved):

PR #690, opened 2026-08-18
feat: XYNE-56410 add people to a DM with conversation history scoping

Dnyaneshwari-04:
[POT](https://drive.google.com/file/d/1UF-QLWFHsmphCHkahj7NUFyiE6iYsLHw/view?usp=sharing)
Add people to this conversation and choose how much of the conversation history they can see.
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [x] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


apps/backend/src/services/groupDmParticipantService.ts
```diff
@@ -0,0 +1,331 @@
...
+        conversationsCopied,
+        message:
+          conversationsCopied > 0
+            ? `${conversationsCopied} conversation(s) shared with the existing group DM`
+            : 'Navigated to existing group DM',
+      };
+    }
+
+    if (historyScope.mode === 'beginning' && channel.scopeType === ChannelScopeType.GROUP_DM) {
+      return this.addParticipantsInPlace(channelId, currentParticipants, newUsers);
```
harshpreet931: pls correct me if i am wrong but,, only beginning stays in this channel... today/custom/none forks a new one with everyone in it and nobody but the new person gets told can we notify the existing members when this forks?

apps/backend/src/controllers/channelController.ts
```diff
@@ -2438,257 +2443,119 @@ export class ChannelController {
...
 
-      if (!isParticipant) {
-        res.status(403).json({
-          error: 'You must be a participant to add others to this GROUP_DM'
-        });
-        return;
-      }
+      const result = await groupDmParticipantService.addParticipants({
+        channelId,
+        currentUserId,
```
harshpreet931, line 2535: system message only fires when isInPlace\hhistory copied into an existing group dm now goes in silently. can we send one here too

apps/backend/src/services/groupDmParticipantService.ts
```diff
@@ -0,0 +1,331 @@
...
+      await this.channelParticipantRepository.getChannelParticipants(channelId);
+
+    if (!currentParticipants.some(p => p.userId === currentUserId)) {
+      throw new AppError('You must be a participant to add others to this conversation', 403);
+    }
+
+    const newUsers = await this.loadActiveUsers(uniqueUserIds);
+
+    const currentUserIds = currentParticipants.map(p => p.userId);
+    const allParticipantIds = [...currentUserIds, ...uniqueUserIds].sort();
```
harshpreet931: uniqueUserIds isnt deduped against currentUserIds pls correct me if i am wrong

apps/backend/src/database/repositories/conversationRepository.ts
```diff
@@ -268,26 +279,265 @@ export class ConversationRepository extends BaseRepository<Conversation, CreateC
...
+          : null,
+        nudgeCount: null,
+        isSent: true,
+        reactions_md: null,
+        link_preview_md: null,
+        childConversationId: null
+      });
+
+      await this.db.$transaction(async tx => {
+        await tx.conversation.create({
```
harshpreet931: parentMessageId isnt carried over a copied thread reply loses its parent and shows up as a top level post. same for childConversationId at 479

apps/dashboard/src/components/Chat/AddPeopleForm/AddPeopleForm.utils.ts
```diff
@@ -0,0 +1,132 @@
...
+    case 2:
+      return 'nd';
+    case 3:
+      return 'rd';
+    default:
+      return 'th';
+  }
+}
+
+export function formatDayDivider(timestamp: number): string {
```
harshpreet931: always shows the full date even for today, ThreadPannel already has formatDatePill for this. can we reuse it

apps/dashboard/src/services/Chat/channelService.ts
```diff
@@ -83,17 +84,21 @@ export interface CreateDmResponse {
   isExisting: boolean;
 }
 
-export interface AddGroupDmParticipantsRequest {
-  userIds: string[];
-  includeHistory: boolean;
+export type { AddGroupDmParticipantsRequest, AddGroupDmParticipantsResponse } from '@xyne/shared';
+
+export interface HistoryPreviewEntry {
```
harshpreet931: same interface redeclared in groupDmParticipantService.ts,  can we share theM

apps/dashboard/src/components/Chat/AddPeopleForm/AddPeopleForm.types.ts
```diff
@@ -0,0 +1,72 @@
...
+  existingUserIds?: string[] | undefined;
+}
+
+export interface HistoryScopeOption {
+  mode: HistoryScopeMode;
+  label: string;
+  requiresDate?: boolean;
+}
+
+export interface PreviewAttachment {
```
harshpreet931: PreviewAttachment/PreviewMessage/PreviewGroup here are basically HistoryPreviewEntry again third copy of the same shape only wecna just use HistoryPreviewEntry directly

apps/backend/src/services/groupDmParticipantService.ts
```diff
@@ -0,0 +1,331 @@
...
+              content: row.initialMessage.content,
+              senderName: sender?.displayName || sender?.name || 'Unknown',
+            }
+          : null,
+        attachments: row.attachments,
+      };
+    });
+  }
+
+  private async loadActiveUsers(userIds: string[]): Promise<User[]> {
```
harshpreet931, line 240: this is basically the same active user check createNewDM ig

apps/backend/src/database/repositories/conversationRepository.ts
```diff
@@ -268,26 +268,112 @@ export class ConversationRepository extends BaseRepository<Conversation, CreateC
...
+    }
+
+    const conversationIds = conversations.map(c => c.conversationId);
+
+    const [movedConversations] = await this.db.$transaction([
+      this.db.conversation.updateMany({
+        where: { conversationId: { in: conversationIds } },
+        data: { channelId: targetChannelId }
+      }),
+      this.db.conversationParticipant.updateMany({
```
santhoshJus, line 410: **Stale denormalized `channelId` on satellite rows after a move.** This transaction reparents `conversation.channelId` and `conversation_participant.channelId`, but several other tables denormalize `channelId` keyed by the moved `conversationId`/`messageId` and are left pointing at the old channel:

- `ConversationLabelMapping.channelId` — schema comments it as "Denormalized for ACL", and its Zero ACL gates on the channel relation off this column (`conversation-label-mappings-acl.ts:19`). A moved conversation's labels would be ACL-checked against the *old* channel.
- `DraftMessage.channelId`, `SurfaceNudge`/`SurfaceNudgeCount`, `ReleaseEvent` — all carry `channelId` alongside `conversationId`/`messageId` and become orphaned pointers.

Message read-ACL itself is fine (it traverses `message → conversation → channel → participants`, all off the updated `conversation.channelId`). But please either update these satellite rows in the same transaction, or document why leaving them stale is safe for DMs. Low impact for pure DMs today, but this is a generic conversation-move primitive.

apps/backend/src/services/groupDmParticipantService.ts
```diff
@@ -0,0 +1,281 @@
...
+
+    if (!existingGroupDM) {
+      for (const participantId of allParticipantIds) {
+        const role = participantId === currentUserId ? ChannelRole.ADMIN : ChannelRole.MEMBER;
+        await this.channelParticipantRepository.addParticipant(destination.id, participantId, role);
+      }
+    }
+
+    const { moved, remaining } = carriesHistory
+      ? await this.moveHistory(channelId, destination.id, cutoff)
```
santhoshJus, line 154: **The overall operation isn't atomic.** Only the two `updateMany`s inside `moveConversationsToChannel` share a transaction. Here the sequence `channelRepository.create` (l.132) → `addParticipant` loop (l.142-145) → `moveHistory` (l.149) are separate awaits with no outer transaction or rollback. A failure partway through can leave observable partial state — e.g. a freshly-created empty group channel with participants but no history moved, which then also becomes the `getGroupChannelByMembers` match on retry. Consider wrapping create + participant-add + move in a single transaction (or making the whole path idempotent/compensating on failure).

apps/backend/src/services/groupDmParticipantService.ts
```diff
@@ -0,0 +1,281 @@
...
+    const carriesHistory = historyScope.mode !== 'none';
+
+    if (historyScope.mode === 'custom' && (!cutoff || Number.isNaN(cutoff.getTime()))) {
+      throw new AppError('Invalid custom history cutoff date', 400);
+    }
+
+    const existingGroupDM =
+      await this.channelRepository.getGroupChannelByMembers(allParticipantIds);
+
+    const destination =
```
santhoshJus, line 130: **Check-then-create race on the destination channel.** `getGroupChannelByMembers(allParticipantIds)` (l.127-128) followed by `channelRepository.create(...)` (l.132) is a read-then-write with no guarding unique constraint on the member-set `name`. Two concurrent "add people" requests for the same resulting member set can both miss and each create a separate GROUP_DM, splitting history across two channels. Recommend a unique constraint on `(scopeType, name)` (or equivalent) for group DMs and handling the unique-violation by re-reading the canonical row.

Dnyaneshwari-04: this needs a migration to add the unique constraint. Do you want me to include that in this PR, (scopeType, name) constraints?

apps/backend/src/services/groupDmParticipantService.ts
Dnyaneshwari-04: I checked the retry path and it looks idempotent already. addParticipant won’t duplicate participants, and moveHistory still runs if the group DM already exists, so a retry should finish whatever was left.

The remaining gap is that we don’t retry automatically, so a failure can still leave partial state until the user retries. I’d rather avoid adding an outer transaction here unless we want strict all-or-nothing behavior.

apps/backend/src/services/groupDmParticipantService.ts
santhoshJus: We can take it separately

apps/backend/src/database/repositories/conversationRepository.ts
```diff
@@ -268,26 +271,154 @@ export class ConversationRepository extends BaseRepository<Conversation, CreateC
...
+      return {
+        moved: 0,
+        remaining: await this.db.conversation.count({ where: { channelId: sourceChannelId } })
+      };
+    }
+
+    const conversationIds = conversations.map(c => c.conversationId);
+
+    const [movedConversations] = await this.db.$transaction([
+      this.db.conversation.updateMany({
```
santhoshJus, line 406: Are all the data moved to new channel here? Feels like activities, tickets are missing among other tables

apps/backend/src/database/repositories/conversationRepository.ts
```diff
@@ -268,26 +271,192 @@ export class ConversationRepository extends BaseRepository<Conversation, CreateC
...
+    if (conversations.length === 0) {
+      return {
+        moved: 0,
+        remaining: await this.db.conversation.count({ where: { channelId: sourceChannelId } })
+      };
+    }
+
+    const conversationIds = conversations.map(c => c.conversationId);
+
+    const [movedConversations] = await this.db.$transaction([
```
santhoshJus, line 405: **Completeness relies on a hand-maintained list.** It's correct today — I checked against the schema: every table that denormalizes `channelId` against a conversation is here, every `messageId + channelId` table (`PrThreadLink`, `MessageArtifact`, `Activity`, `SurfaceNudgeCount`, `DraftMessage`) also carries `conversationId` and is covered, and `messages` / `message_attachments` / `reactions` have no `channelId` so they follow the conversation automatically.

The risk is the future: any new table that adds a denormalized `channelId` — especially an ACL one like `conversation_participants.channelId` (`// Denormalized from conversation.channelId for direct ACL check`) — will be **silently missed** here. No compile error, no test failure, just rows left pointing at the old channel, which for the ACL columns is a permission leak/loss.

Suggest a guard test that reflects over `Prisma.dmmf.datamodel.models` (already used in `acl-extension.ts` / `stamp.ts`), collects every model with **both** `conversationId` and `channelId`, and asserts each is either in this move set or an explicit `INTENTIONALLY_NOT_MOVED` allowlist (`Ticket` / `Email` / `EmailDraft` / `PrThreadLink` — can't exist on a DM/GROUP_DM). A new such table then fails CI → forces a conscious decision instead of a silent gap.

apps/backend/src/database/repositories/conversationRepository.ts
```diff
@@ -268,26 +271,192 @@ export class ConversationRepository extends BaseRepository<Conversation, CreateC
...
+   * `conversationId` so they follow automatically; `conversationParticipant.channelId` is
+   * denormalized and has to be moved explicitly. Threads move whole, keyed on when the thread
+   * started — one opened before the cutoff stays put even if it has replies after it.
+   */
+  async moveConversationsToChannel(
+    sourceChannelId: string,
+    targetChannelId: string,
+    createdAfter?: Date | null
+  ): Promise<{ moved: number; remaining: number }> {
+    const conversations = await this.db.conversation.findMany({
```
santhoshJus, line 368: **No limit or batching on the move.** This `findMany` has no `take`, so every matching conversation goes into a single `$transaction` of ~15 `updateMany({ where: { conversationId: { in: [...all ids...] } } })` — a large `IN (...)` list across 15 tables held in one transaction (lock contention / statement-timeout / deadlock risk on a big history).

`reindexMovedMessages` (L438) compounds it: an unbounded `message.findMany` then `Promise.all(messages.map(addJob))` (L448) fires **every** Vespa enqueue at once — no chunking, no backpressure. And this whole thing runs **synchronously in the add-participants request path**, so a large move can time out the HTTP request itself.

For 1:1 DMs this is a non-issue, but a long-lived GROUP_DM with a null cutoff ("move everything") is exactly where it bites. Suggest: chunk `conversationIds` (e.g. 500/batch) for both the DB writes and the reindex, enqueue the Vespa reindex in bounded batches (or one bulk job), and/or push large reparents to a background job rather than the request path.

apps/dashboard/src/components/ui/MessageBubble/NonParticipantActions.tsx
```diff
@@ -115,7 +115,7 @@ export const NonParticipantActions: React.FC<NonParticipantActionsProps> = ({
       addGroupDmParticipantsMutation.mutate({
         channelId: activeChannelId,
         userIds,
-        includeHistory: true, // Default to including history for GROUP_DM
+        historyScope: { mode: 'beginning' },
```
santhoshJus: **Quick-add bypasses the entire history-scoping model (privacy).** `handleAddToChannel` hardcodes `historyScope: { mode: 'beginning' }`, so this one-click "Add Them" affordance shares the **full** conversation history — every past message and attachment — with the newly added person, with no scope choice and no confirmation.

That's exactly what the main `AddPeopleForm` flow is built to prevent: it defaults to `today` and gates full history (`beginning`) behind an explicit confirm dialog. This path preserves the old `includeHistory: true` behavior, but it's now inconsistent with the privacy model this PR introduces. Please make it a deliberate choice — route through the scoping step, or default to `today`/`none` here — rather than silently defaulting to full history.

apps/backend/src/database/repositories/conversationRepository.ts
```diff
@@ -268,26 +282,205 @@ export class ConversationRepository extends BaseRepository<Conversation, CreateC
...
+        createdAt: conversation.createdAt,
+        initialMessage: initial
+          ? { senderId: initial.senderId, content: initial.content }
+          : null,
+        attachments: attachmentsByMessage.get(conversation.initialMessageId) ?? []
+      };
+    });
+  }
+
+  /**
```
santhoshJus, line 357: **The move reparents more than the preview showed — the consent surface under-represents the actual share.**

`getHistoryPreview` (L290) filters `OR: [{ doNotPostToChannel: null }, { doNotPostToChannel: false }]`, but this move selects only `{ channelId, createdAt >= cutoff }` with **no** `doNotPostToChannel` filter. So conversations hidden from the preview (`doNotPostToChannel = true`, e.g. `showInChannel` children) are silently moved into the new group and become visible to the newly-added members — even though the person who consented never saw them in the preview. With a cutoff this also splits threads: a hidden child created after the cutoff moves while its parent before the cutoff stays behind.

The preview is the feature's consent surface, so preview and move must use the **same** predicate. Separately: the preview is capped (~20/≤50) while this move is unbounded and there's no "+N more" indicator, so even for visible history the user under-estimates what's being shared.

apps/backend/src/services/groupDmParticipantService.ts
```diff
@@ -0,0 +1,281 @@
...
+      existingGroupDM ??
+      (await this.channelRepository.create({
```
santhoshJus, line 132: **No transaction spans channel-create → add-participants → move-history.** These three steps run as independent commits (only each `reparentChunk` is a `$transaction`, and `reindexMovedMessages` runs after, un-transacted).

If `moveHistory` throws — e.g. chunk 2 of 3 fails — the new GROUP_DM and its participant rows are already committed and chunk 1's conversations already moved, but the caller gets a 500. The result is an orphaned, half-populated group DM with history split across source and destination. Retry only partially heals it: `getGroupChannelByMembers` then finds the orphan, treats it as existing (so participants aren't re-added), and re-moves whatever source rows remain.

Consider wrapping create + participants + move in one transaction, or making the operation idempotent/resumable with cleanup on failure, so a mid-flight error can't leave a durable inconsistent group DM.

santhoshJus (approved):

devesh-juspay (approved):

PR #696, opened 2026-08-18
fix: harden xyne-claw + xyne-claw-auth framework paths

XyneSpaces:
## Summary
Three surgical, low-risk framework fixes from the `/framework` scan of `apps/xyne-claw` and `apps/xyne-claw-auth`. The larger refactors (OAuth-router factory, worker-lifecycle factory, converter registry) are intentionally deferred to follow-up PRs so this PR stays reviewable and safe.

### #1 — S2S routes could hang the caller on a transient error
`routes/goal-judge.ts`, `routes/eval-extract.ts`, and `routes/eval-judge.ts` each `await` an LLM call with **no try/catch**. Express 4 does not auto-forward async rejections, so a thrown error left the S2S request open until the caller's own timeout instead of returning a clean 500.
- Wrapped every handler body in `try/catch` (matching the existing `curator.ts` route envelope), returning `500 { success:false, error }` and guarding with `!res.headersSent`.
- Added a **global Express error handler** in `main.ts` (registered after all routes, 4-arg signature) as a last-resort net for any future sync throw / `next(err)`.

### #4 — Divergent LiteLLM base-URL config
`config.ts` defaulted `LITELLM.url` to `http://localhost:4000` with no trailing-slash strip, while `curator.ts` / `failure-curator.ts` each re-derived their own URL (different default `https://grid.ai.example.com` + a local `.replace(/\/$/,"")`), and other call sites hand-stripped the slash too.
- Made `config.ts` the single source of truth: strip the trailing slash **once** in `LITELLM.url`.
- Both curators now consume `LITELLM.url` and `LITELLM.automationApiKey` instead of re-reading env.
- The deliberately-divergent, team-scoped **model** resolution (documented with prod-incident comments) is left untouched.

### #8 — Workers with no `error` listener
8 BullMQ workers registered `.on("failed")` but **not** `.on("error")`, so a connection-level worker error was swallowed with no log line.
- Added a matching `.on("error")` handler to each of: `agent-backfill`, `digital-twin-backfill`, `entity-extraction`, `eval-generation`, `eval-import`, `eval-judge`, `provider-retry`, `search-eval-run`, using each file's own logger (`logger`/`log`/`console`) and label.

## Explicitly dropped after verifying against live code
The scan's finding #5 (twin-delivery S2S key drift) was **refuted**: `twin-delivery.ts` calls Spaces `/api/internal/*` and correctly uses `INTERNAL_S2S_KEY` with an unconditional `x-s2s-key` header — identical to its sibling `app-callback.ts` that posts to the same `postAsUser` endpoint. `XYNE_CLAW_S2S_KEY` authenticates a **different** direction (calls to claw). Changing it would have broken twin-delivery, so it was left as-is.

## Validation
- `apps/xyne-claw`: `tsc --noEmit` — **clean (exit 0)**.
- `apps/xyne-claw-auth`: remaining `tsc` errors are entirely the pre-existing `@prisma/client not generated` cascade (no Prisma engine network in the CI sandbox) on files this PR does not touch. None of my edited lines produce errors.

## Reviewer note
`eval-import-worker.ts` renders as a **binary diff** because the source contains an intentional NUL-byte separator inside a template literal (`` `${message}\x00${expectedResponse}` ``) — pre-existing in HEAD, unrelated to this change. The only change in that file is the added 3-line `.on("error")` handler.


XyneSpaces (commented):
Xyne Spaces Reviewer inline findings for PR #696.

apps/xyne-claw-auth/backend/src/lib/dcr-oauth-flow.ts
```diff
@@ -0,0 +1,464 @@
...
+          userId,
+          clientId,
+          codeVerifier,
+          redirectUri: callbackUri,
+          scope,
+        });
+
+        const url = new URL(authUrl);
+        url.searchParams.set("client_id", clientId);
+        url.searchParams.set("redirect_uri", callbackUri);
```
XyneSpaces, line 326: **[blocking] R1/S23** — The OAuth factory still accepts a body-supplied `redirectUri` and registers it with DCR, so any caller that can hit authorize can make the provider redirect an authorization code to an arbitrary origin. Please derive this callback from trusted server config (or enforce a strict allowlist matching this auth service) instead of trusting request body input.

apps/xyne-claw/src/main.ts
```diff
@@ -71,6 +71,19 @@ app.use(evalExtractRouter);
...
 
+// Global error handler — last-resort net for a synchronous throw in any route
+// or an error explicitly forwarded via next(err). Async handlers still catch
+// their own rejections (Express 4 does not auto-forward promise rejections),
+// but this guarantees an unexpected throw returns 500 instead of hanging the
+// S2S caller until its timeout. Must be registered AFTER all routes, and must
+// keep the 4-arg signature so Express treats it as error-handling middleware.
+app.use((err: unknown, _req: express.Request, res: express.Response, _next: express.NextFunction) => {
+  log.error("[xyne-claw] Unhandled route error:", err instanceof Error ? err.message : String(err));
+  if (!res.headersSent) {
```
XyneSpaces, line 82: **[major] S43/R2** — The global error handler returns `err.message` to callers, which can leak provider/DB/internal failure details from whatever route forwarded the error. Please keep the detailed message in logs and return a generic `Internal error` response body.

XyneSpaces:
## Xyne Spaces Reviewer — PR #696

**Reviewed:** 14 files, 14 changed file entries · **Findings:** 1 blocking, 1 major, 0 nits

### Findings
| blocking | R1 / S23 | apps/xyne-claw-auth/backend/src/lib/dcr-oauth-flow.ts:326 | OAuth authorize still trusts body-supplied `redirectUri`, then registers it via DCR; derive/allowlist callback URI server-side. |
| major | S43 / R2 | apps/xyne-claw/src/main.ts:82 | Global error handler returns raw `err.message` to callers; log details but return a generic response. |

### Recommended reviewer
**anurag.dwivedi@juspay.in** (review request not assigned: this runner has no GitHub reviewer-request tool, and GitHub login could not be resolved from the email) — best fallback because CODEOWNERS is absent and PR context/request came from Anurag.

| anurag.dwivedi@juspay.in | unknown — GitHub path-author lookup unavailable for this private repo | no CODEOWNERS found | unknown | requester/thread owner for this change |

Reviewer selection caveat: `.github/CODEOWNERS` and `CODEOWNERS` were not found; the GitHub subagent could not access this private repo for path-author lookup, and `search_users` did not resolve `anurag.dwivedi@juspay.in` to a GitHub login. I therefore could not request the reviewer via GitHub API in this run.

Top blocking issue: the new OAuth DCR factory should not accept an arbitrary callback URI from request body before this is expanded to more providers.

PR #711, opened 2026-08-18
fix: ignore URLs inside code blocks for link preview extraction

XyneSpaces:
## Problem

When a message contains a URL inside a `<pre>` or `<code>` block AND a different URL in the message body, the link preview was generated for the code-block URL instead of the body URL.

## Root cause

`extractFirstUrl()` in `apps/backend/src/utils/urlUtils.ts` calls `stripAndDecodeHtml()` which strips ALL HTML tags indiscriminately, including `<pre>`/`<code>`. URLs inside code blocks survive in the text and can be selected as the first URL for link previews.

The frontend renderer (`RenderMessageWithHTML.tsx`) already skips auto-linking inside code blocks, but the backend does not share that logic.

## Fix

Added two regex replacements in `stripAndDecodeHtml()` to strip `<pre>...</pre>` and `<code>...</code>` content BEFORE the generic tag removal step. This makes the backend consistent with the frontend.

## Tests

Added `apps/backend/src/utils/urlUtils.test.ts` with 8 test cases covering:
- Normal body URLs (regression check)
- URLs inside `<pre><code>` blocks (ignored)
- URLs inside inline `<code>` (ignored)
- Code-block URL appearing before body URL (body URL wins)
- Multiple URLs with first in inline code
- Nested/classed code blocks
- Deduplication outside code blocks
- Code-block-only URLs excluded

All 8 tests pass.

### DIFF
--- a/apps/backend/src/utils/urlUtils.test.ts
+++ b/apps/backend/src/utils/urlUtils.test.ts
@@ -1,0 +1,45 @@
+import { extractFirstUrl, extractUrls } from './urlUtils';
+
+describe('extractFirstUrl', () => {
+  it('returns a normal body URL', () => {
+    const input = '<p>See <a href="https://docs.example.com/page">https://docs.example.com/page</a></p>';
+    expect(extractFirstUrl(input)).toBe('https://docs.example.com/page');
+  });
+
+  it('ignores URLs inside <pre><code> blocks', () => {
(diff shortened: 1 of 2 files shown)


XyneSpaces:
Closing in favor of a new PR from branch fix/link-preview-code-block-urls-v2 (test file removed per reviewer request).

PR #732, opened 2026-08-19
feat: XYNE-56495 eta management new

architjuspay:
[POT-1](https://drive.google.com/file/d/11CYcTAiUwbsB2UUCPg9Quy3DGfnG1bUH/view?usp=sharing)

[POT-2](https://drive.google.com/file/d/17j-eQ4c2MD-0zikY5KaFqD2mm4hQkyQp/view?usp=sharing)

Environment-Changes:
  - ETA_MANAGEMENT_ENABLED: Deployment-level kill switch for the ETA risk-detection/auto-recalculation feature

  - backend
  - dashboard

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

- [x] Key added to **all** relevant env files, not just the one you run — `apps/backend/.env.{example,local,test}`, `apps/dashboard/.env.{example,local,test}`, plus the app you touched

## API Contract

- [x] No API changes
- [ ] New endpoint

---

### Automated

- [x] Not covered by tests 

### Manual

**Users**
- [x] Single user
- [x] Two users at once (two accounts / two browsers) — sync lands on both, no cross-user leakage

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist

### DIFF
--- a/apps/backend/src/services/etaManagement/etaRecipients.ts
+++ b/apps/backend/src/services/etaManagement/etaRecipients.ts
@@ -1,0 +1,85 @@
+import { db } from '@/database/client';
+import { getFormFieldUserActors } from '@/utils/ticketActorUtils';
+import { canUserModifyTicketControl } from './etaPermissions';
+
+/**
+ * Same "who might care about this ticket" set every existing
+ * `sendTicket*Notification` method in notificationService.ts already uses
+ * (see `TicketsSideEffectHandler`'s private `fetchTicketActors` in
+ * zero/side-effects/tables/tickets-handler.ts, reimplemented here rather
+ * than exported cross-module for a single shared call).
+ */
+async function fetchTicketActors(ticketId: string): Promise<string[]> {
+  const [roleAssignments, formFieldUserActors] = await Promise.all([
+    db.ticketAssignment.findMany({ where: { ticketId }, select: { userId: true } }),
+    getFormFieldUserActors(ticketId),
+  ]);
+  return roleAssignments
+    .map((a) => a.userId)
+    .filter((id): id is string => Boolean(id))
+    .concat(formFieldUserActors);
+}
+
+/**
+ * "Awareness recipients": assignee, creator, active ticket
+ * assignments, and user-valued ticket fields. Notification-preference
+ * filtering (mute/pause/channel level) happens separately in the actual
+ * send methods (`notificationFilterService`), not here.
+ */
+export async function resolveAwarenessRecipients(
+  ticketId: string,
+  createdBy: string,
+  assignedTo: string | null,
+  excludeUserId?: string,
+): Promise<string[]> {
+  const extraActors = await fetchTicketActors(ticketId);
+  const all = [createdBy, assignedTo, ...extraActors].filter(
+    (id, index, arr): id is string => Boolean(id) && arr.indexOf(id) === index,
+  );
+  return excludeUserId ? all.filter((id) => id !== excludeUserId) : all;
+}
+
+/**
+ * "Action recipients": awareness recipients who also satisfy the
+ * board's ETA-update permission policy. This is a net-new concept - no
+ * existing recipient-resolution function in this codebase filters by
+ * permission, only by notification preference (mute/pause/channel level).
+ */
+export async function resolveActionRecipients(
+  awarenessRecipients: string[],
+  ticketUserGroupId: string | null,
+  boardId: string,
+): Promise<string[]> {
+  if (!ticketUserGroupId || awarenessRecipients.length === 0) {
+    // No user group on the ticket: canUserModifyTicketControl's underlying gate is
+    // unrestricted for everyone in that case, so the full awareness set qualifies.
+    return awarenessRecipients;
+  }
+
+  const board = await db.board.findUnique({ where: { id: boardId }, select: { metadata: true } });
+
+  // Two queries total regardless of recipient count: the board above, and every
+  // relevant membership in one batch below. Notification fan-out can be large, so
+  // this must not scale with the recipient list.
+  const mappings = await db.userGroupMapping.findMany({
+    where: { userGroupId: ticketUserGroupId, userId: { in: awarenessRecipients } },
+    select: { userId: true, roleId: true, responsibility: true },
+  });
+  const mappingByUserId = new Map(
+    mappings.map((m) => [
+      m.userId,
+      { roleId: m.roleId ?? null, responsibility: m.responsibility ?? null },
+    ]),
+  );
+
+  const results = await Promise.all(
+    awarenessRecipients.map(async (userId) => {
+      const permission = await canUserModifyTicketControl(userId, ticketUserGroupId, boardId, {
+        getBoardMetadata: async () => board?.metadata ?? null,
+        getUserGroupMapping: async (uid) => mappingByUserId.get(uid) ?? null,
+      });
+      return permission.allowed ? userId : null;
+    }),
+  );
+  return results.filter((id): id is string => id !== null);
+}
--- a/apps/backend/src/workers/stageEtaDeadlineWorker.ts
+++ b/apps/backend/src/workers/stageEtaDeadlineWorker.ts
@@ -1,11 +1,53 @@
+import { Prisma } from '@prisma/client';
 import { logger } from '@/utils/logger';
 import { db, readReplicaDb } from '@/database/client';
+import { runAsServiceActor } from '@/database/tenant/context';
 import { stageEtaDeadlineQueue } from '@/queues/stageEtaDeadlineQueue';
-import { OPEN_STATUSES } from '@/utils/etaNotificationUtils';
-import { Prisma } from '@prisma/client';
+import {
+  getTicketBotActorId,
+  TicketWithStageInfo,
+  OPEN_STATUSES,
+} from '@/utils/etaNotificationUtils';
+import {
+  parseTicketEtaManagement,
+  mergeTicketEtaManagement,
+  BoardType,
+} from '@xyne/shared';
+import { recordTicketTimelineEvent } from '@/services/ticketTimelineEventService';
+import {
+  evaluatePlanningRisk,
+  buildRiskTransitionActivityIntents,
+  dispatchEtaNotifications,
+  etaSignalsFromResult,
+} from '@/services/etaManagement';
 
+interface TicketForReconciliation extends TicketWithStageInfo {
+  boardId: string;
+  userGroupId: string | null;
+  statusV2: string;
+  metadata: unknown;
+}
+
+// Batching for the bulk `isStageOverdue` flag sync.
 const BATCH_SIZE = parseInt(process.env.STAGE_ETA_DEADLINE_BATCH_SIZE || '15', 10);
 const BATCH_SLEEP_MS = parseInt(process.env.STAGE_ETA_DEADLINE_BATCH_SLEEP_MS || '1000', 10);
+
+// Batching for the planning-risk reconciliation pass, which does per-ticket writes
+// (metadata + timeline + notifications) rather than a bulk `updateMany`.
+const STAGE_ETA_REMINDER_BATCH_SIZE = 15;
+const STAGE_ETA_REMINDER_BATCH_DELAY_MS = 1000;
+// Page size for both cursor-paginated TicketStageEta scans below - shared so the two
+// stay in sync rather than drifting as two separately-tuned magic numbers.
+const QUERY_BATCH_SIZE = 10000;
+
+const chunkArray = <T>(items: T[], chunkSize: number): T[][] => {
+  if (chunkSize <= 0) return [items];
+  const chunks: T[][] = [];
+  for (let index = 0; index < items.length; index += chunkSize) {
+    chunks.push(items.slice(index, index + chunkSize));
+  }
+  return chunks;
+};
 
 const sleep = (ms: number): Promise<void> => new Promise(resolve => setTimeout(resolve, ms));
 
@@ -47,7 +89,13 @@ class StageEtaDeadlineWorker {
       logger.info(
         '[STAGE-ETA-DEADLINE-WORKER] Processing stage ETA deadline check job',
       );
-      await this.syncStageOverdueFlags();
+      const now = new Date();
+      // Two independent passes over the same hourly tick, split by which side of `now`
+      // the active visit's deadline falls on: already-breached visits drive the
+      // denormalized `isStageOverdue` flag, not-yet-breached ones drive planning-risk
+      // reconciliation.
+      await this.syncStageOverdueFlags(now);
+      await this.reconcileOpenTicketPlanningRisk(now);
       logger.info('[STAGE-ETA-DEADLINE-WORKER] Stage ETA deadline check completed');
     } catch (error) {
       logger.error('[STAGE-ETA-DEADLINE-WORKER] Error checking stage ETA deadlines:', error);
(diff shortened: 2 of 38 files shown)


apps/backend/src/services/etaManagement/etaRecipients.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const board = await db.board.findUnique({ where: { id: boardId }, select: { metadata: true } });
+  const results = await Promise.all(
+    awarenessRecipients.map(async (userId) => {
+      const permission = await canUserModifyTicketControl(userId, ticketUserGroupId, boardId, {
+        getBoardMetadata: async () => board?.metadata ?? null,
+        getUserGroupMapping: async (uid, ugId) => {
+          const mapping = await db.userGroupMapping.findFirst({
+            where: { userId: uid, userGroupId: ugId },
+          });
```
5uryan5h: findUnique

apps/backend/src/services/etaManagement/featureFlag.ts
```diff
@@ -0,0 +1,11 @@
+import { config } from '@/config/env';
```
5uryan5h: not needed

packages/shared/src/validation/etaManagementSchema.ts
```diff
@@ -0,0 +1,211 @@
...
+    acknowledgmentReason: null,
+  },
+};
+
+export function validateTicketEtaManagement(data: unknown) {
+  return ticketEtaManagementSchema.safeParse(data);
+}
+
+/** Tolerant parse: absent or invalid `etaManagement` falls back to the documented defaults. */
+export function parseTicketEtaManagement(metadata: unknown): TicketEtaManagement {
```
5uryan5h: it should be defined

packages/shared/src/validation/etaManagementSchema.ts
```diff
@@ -0,0 +1,211 @@
...
+
+/**
+ * Merge a patch into `Ticket.metadata.etaManagement` while preserving every
+ * unrelated key already on `Ticket.metadata` (flow, reporterEmail,
+ * workflowType, ...). `activeVisit`/`planningRisk` are merged one level deep
+ * so a single-field patch (e.g. just `planningRisk.state`) doesn't clobber
+ * untouched sibling fields on the same nested object.
+ */
+export function mergeTicketEtaManagement(
+  existingMetadata: unknown,
```
5uryan5h: +1

apps/backend/src/workers/stageEtaDeadlineWorker.ts
```diff
@@ -1,11 +1,53 @@
...
+  metadata: unknown;
+}
+
+// Batching for the bulk `isStageOverdue` flag sync.
 const BATCH_SIZE = parseInt(process.env.STAGE_ETA_DEADLINE_BATCH_SIZE || '15', 10);
 const BATCH_SLEEP_MS = parseInt(process.env.STAGE_ETA_DEADLINE_BATCH_SLEEP_MS || '1000', 10);
 
+// Batching for the planning-risk reconciliation pass, which does per-ticket writes
+// (metadata + timeline + notifications) rather than a bulk `updateMany`.
+const STAGE_ETA_REMINDER_BATCH_SIZE = 50;
```
PiyushJuspay: Make batch size 15

apps/backend/src/zero/mutators.ts
```diff
@@ -6299,8 +6393,140 @@ export function createMutators(
...
+            }
+          }
+
+          if (isPureManualEtaChange) {
+            // User-attributed (not the system actor) - this is a direct, deliberate manual
+            // edit, distinct from the system-attributed ETA_AUTO_RECOMPUTED/ETA_RISK_* rows
+            // written below. Kept alongside the existing generic ActivityType.ETA row the
+            // `fields` loop already writes (unchanged, for existing timeline consumers) -
+            // this one carries the reason.
+            await tx.mutate.ticket_activities.insert({
```
PiyushJuspay: Move to side effect

apps/backend/src/zero/mutators.ts
```diff
@@ -6575,6 +6801,121 @@ export function createMutators(
...
+          });
+
+          await tx.mutate.tickets.update({
+            id: ticketId,
+            metadata: mergedMetadata as ReadonlyJSONValue,
+            updatedAt: now,
+          });
+
+          // Record + system message only - no additional notification.
+          await tx.mutate.ticket_activities.insert({
```
PiyushJuspay: +1

apps/backend/src/zero/mutators.ts
```diff
@@ -6575,6 +6801,121 @@ export function createMutators(
...
+              acknowledgedBy: authData.sub,
+              acknowledgedAt: now,
+            } satisfies EtaRiskAcknowledgedActivityValue as ReadonlyJSONValue,
+            channelId: ticket.channelId,
+          });
+
+          if (ticket.conversationId) {
+            const user = await tx.run(zql.users.where('id', authData.sub).one());
+            const userName = user?.displayName || user?.name || 'Someone';
+            await tx.mutate.messages.insert({
```
PiyushJuspay: can we do this as well in side effects. its common everywhere

packages/shared/src/validation/etaManagementSchema.ts
```diff
@@ -0,0 +1,211 @@
...
+// no schema migration; every reader/writer must go through the parse/merge
+// helpers below rather than casting the raw column.
+// ============================================================================
+
+export const boardEtaManagementSchema = z.object({
+  schemaVersion: z.literal(1),
+  autoRecomputeEnabled: z.boolean(),
+  standardPathStageIds: z.array(z.string()),
+  configVersion: z.number().int().nonnegative(),
+  updatedAt: z.number(),
```
PiyushJuspay: Don't add so much data in board's metadata column. Check what can be reduced

PiyushJuspay: why we need schemaVersion, configVersion, updatedAt, updatedBy ?

architjuspay: config version we need because we compare against it during risk fingerprint calculation

apps/backend/src/zero/mutators.ts
architjuspay: done

apps/backend/src/zero/mutators.ts
architjuspay: ok

apps/backend/src/zero/mutators.ts
architjuspay: ok

apps/backend/src/workers/stageEtaDeadlineWorker.ts
architjuspay: ok

apps/backend/src/config/env.ts
```diff
@@ -107,6 +107,7 @@ const envSchema = Joi.object({
   ENABLE_SCHEDULED_MESSAGE_WORKER: Joi.boolean().default(false),
   ENABLE_STAGE_ETA_DEADLINE_WORKER: Joi.boolean().default(false),
   ENABLE_ETA_DEADLINE_WORKER: Joi.boolean().default(false),
+  ETA_MANAGEMENT_ENABLED: Joi.boolean().default(false),
```
PiyushJuspay: WE can remove this ENV. feature should be enabled for everyone. board controls auto recompute

PiyushJuspay:
`Breaking Changes — High
H1 New tickets on existing boards no longer get an initial due date
- What: createTicket used to set eta = now + sum(stage etas). Now it only sets eta when the caller passes one, or when the board opted into auto-recompute. Existing boards parse to autoRecomputeEnabled: false, and there is no backfill.
- Where: apps/backend/src/database/repositories/ticketRepository.ts:218 — ...(data.eta && { eta: data.eta }),
- Who affected: every existing board with stage ETAs (REST create, automations, Xyne Apps, flow cascade all use this repo method); old dashboard during skew.
- Impact: new tickets silently have no due date — no due-date display, no overdue tracking, no overdue notifications, empty calendar entries.
- Fix: confirm this product decision; if intended, backfill etaManagement.autoRecomputeEnabled for existing boards or keep the stage-sum initial ETA as the fallback when the board hasn't opted in.
- Issue Summary:
ticketRepository.ts:218 — Creating a ticket no longer sets an initial due date on boards that haven't opted into ETA management. All existing boards default to off, so every new ticket loses its auto-computed ETA. Can we confirm this is intended, or keep the old stage-sum as the default when the board hasn't configured etaManagement?
H2 Editing a stage deadline no longer moves the ticket due date (and board transfers keep a stale one)
- What: ticketStageEta.update always recalculated Ticket.eta from the new stage deadline + future stages. Now it only does when the board opted in. Same for board transfers (Prisma ticketController.ts:420 and Zero mutators.ts:6035 no longer reset eta).
- Where: apps/backend/src/zero/mutators.ts:7229 — if (boardEtaManagement.autoRecomputeEnabled) {
- Who affected: users editing stage deadlines / moving tickets between boards on existing boards.
- Impact: user changes a stage deadline and the ticket's due date silently stays put; a transferred ticket keeps a due date computed against the old board's stages.
- Fix: keep the old recalculation for non-opted-in boards, or make the opt-out explicit in the UI.
- Issue Summary:
mutators.ts:7229 — On boards that haven't opted into ETA management, editing a stage deadline (and moving a ticket to another board) no longer updates the ticket's due date. This silently changes existing behavior for every current board — should the old recalculation stay as the default?
H3 Manual due-date edits no longer create the ETA activity row or activity-feed entries (all boards, ungated)
- What: For a pure manual eta edit, the generic ActivityType.ETA timeline row, its thread message, and the activityService feed entries (tickets-handler.ts:438 filters on 'ETA') are all skipped, replaced by ETA_MANUALLY_UPDATED (which writes a timeline row + thread message, but no feed entries). The Issue Summary at mutators.ts:6564 claims the ETA row still gets written — it doesn't.
- Where: apps/backend/src/zero/mutators.ts:6251 — if (field === 'eta' && isPureManualEtaChange) continue;
- Who affected: every board (not gated by opt-in); old dashboard renders the new activity type generically.
- Impact: manual due-date edits vanish from the user activity feed; timeline shows the new type only in the new dashboard.
- Fix: also emit ETA_MANUALLY_UPDATED through the activity-feed path, or stop skipping the generic ETA row; fix the stale Issue Summary.
- Issue Summary:
mutators.ts:6251 — This continue drops the generic ETA activity row and its activity-feed entries for manual due-date edits, and the Issue Summary further down ("Stays alongside the generic ActivityType.ETA row") says the opposite of what the code does. Either keep writing the ETA row or add feed entries for ETA_MANUALLY_UPDATED, and fix the Issue Summary.
H4 Duplicate planning-risk notifications on board transfer
- What: A board transfer issues two tickets.updates in one Zero transaction. Both side-effect jobs run the handler, and each computes the risk signal from "previous metadata vs freshly-fetched current ticket" — identical inputs — so dispatchEtaNotifications runs twice. The Issue Summary at tickets-handler.ts:258 claims this is idempotent; it isn't.
- Where: apps/backend/src/zero/side-effects/tables/tickets-handler.ts:276 — await dispatchEtaNotifications(
- Who affected: action recipients on opted-in boards during board transfers that detect/reopen a risk.
- Impact: users get two identical "Planning Risk" notifications for one transfer.
- Fix: derive the "current" side from the job's own args (like etaChanged does with args.eta), or only dispatch from the job whose args carried the metadata write.
- Issue Summary:
tickets-handler.ts:276 — For board transfers the mutator emits two tickets.update jobs, and both compute the same prev→current metadata diff here (current is re-fetched from DB), so the risk alert is sent twice. Gate this on the job's own args carrying the metadata change, as the etaChanged branch does.
H5 Lost-update race on Ticket.metadata in the Prisma paths
- What: updateTicketStage (and transferTicketToBoard, ticketStageTransitionService) reads the ticket before the transaction, then merges etaManagement into that stale snapshot and overwrites the whole metadata JSON. A concurrent Zero write (e.g. acknowledgeEtaRisk, or a just-staged pendingActivities outbox) between the read and the commit is silently clobbered. The old code never wrote metadata on these paths.
- Where: apps/backend/src/database/repositories/ticketRepository.ts:657 — const mergedMetadata = mergeTicketEtaManagement(
- Who affected: any ticket where a Prisma stage move/transfer overlaps a Zero metadata write.
- Impact: lost acknowledgments (banner reappears), lost activity outbox (audit rows never written), reverted risk state.
- Fix: re-read the row (or SELECT ... FOR UPDATE, as the worker already does) inside the transaction before merging.
- Issue Summary:
ticketRepository.ts:657 — The metadata merge uses currentTicket.metadata, read before the transaction opened, and overwrites the whole JSON column. A concurrent metadata write (e.g. acknowledgeEtaRisk) in between is lost. Re-read the row inside the transaction (or lock it like stageEtaDeadlineWorker does) before merging.
H6 Worker reconciliation loads every future-deadline visit in one unbounded query on the primary DB
- What: The new pass runs a single findMany with no limit (existing getOverdueTicketIds paginates 10k at a time on the read replica), pulling full ticket rows including metadata JSON.
- Where: apps/backend/src/workers/stageEtaDeadlineWorker.ts:188 — const entries = await db.ticketStageEta.findMany({
- Who affected: large workspaces; the hourly worker job.
- Impact: memory blowup / very long job; the isProcessing guard then skips subsequent hourly ticks.
- Fix: paginate with a cursor like getOverdueTicketIds, and use the read replica for the scan.
- Issue Summary:
stageEtaDeadlineWorker.ts:188 — This loads all matching stage-eta rows (with nested ticket + metadata) in one unbounded query on the primary DB. Please cursor-paginate like getOverdueTicketIds does and read from the replica.
H7 Shared working-day math changed for exact-day estimates — affects all existing deadlines, ungated
- What: calculateETADeadline now pins an estimate that exactly fills N working days to the END of the last working day; previously it landed on the START of the next working day (e.g. 9h on a 9h workday entered at 09:00: 18:00 same day now vs 09:00 next day before).
- Where: packages/shared/src/utils/etaCalculation.ts:201 — if (remainingHours === workingHoursPerDay) {
- Who affected: every caller — stage-deadline creation on all boards, pause-resume ETA push, forecasts; backend and dashboard.
- Impact: some existing deadlines silently move ~1 working day earlier → more overdue flags/notifications than before, with no feature flag.
- Fix: if this fix is intended, call it out in the release notes; otherwise keep old pinning for non-opted-in boards.
- Issue Summary:
etaCalculation.ts:201 — This changes every existing ETA computation for estimates that are exact multiples of the working day (deadlines now land at end-of-day instead of next-day start), for all boards, not just ETA-managed ones. Is this intentional, and can it be flagged/announced?
H8 New DEFAULT boards get auto-recompute ON by default
- What: BoardRepository.createBoard writes autoRecomputeEnabled: true for DEFAULT boards, while every other creation path (Zero boards.insert for release boards) and every existing board starts off.
- Where: apps/backend/src/database/repositories/boardRepository.ts:91 — autoRecomputeEnabled: resolvedBoardType === BoardType.DEFAULT,
- Who affected: all newly created linear boards via POST /boards (the dashboard's board-create flow).
- Impact: automatic due-date extension is enabled without an admin opting in — the opposite of the "opt-in" framing everywhere else, and inconsistent with boards created through other paths.
- Fix: default to false and let admins opt in via the new toggle.
- Issue Summary:
boardRepository.ts:91 — New DEFAULT boards get autoRecomputeEnabled=true while all existing boards (and boards created via Zero inserts) default to off. Should new boards really start with automatic due-date extension enabled without an admin opting in?
H9 Combined board + eta/status update evaluates ETA against the OLD board
- What: In ticket.update, isPureManualEtaChange doesn't exclude isBoardChanging, and the second evaluation resolves effectiveStage from ticket.boardId/ticket.stageName (pre-transfer values). A single call changing boardId + eta (or statusV2) overwrites the transfer branch's correct metadata with an evaluation against the deleted old-board visit.
- Where: apps/backend/src/zero/mutators.ts:6492 — ? await tx.run(zql.stages.where('boardId', ticket.boardId).where('name', effectiveStageName).one())
- Who affected: any caller sending boardId together with eta/status in one mutation (bulk actions spread ...fields).
- Impact: corrupted activeVisit/planningRisk state (risk wrongly resolved, missed alert) until the next evaluation.
- Fix: exclude isBoardChanging from isPureManualEtaChange and skip the second evaluation when the board changed.
- Issue Summary:
mutators.ts:6492 — When boardId and eta/status change in one call, this evaluation looks up the stage on the OLD board and overwrites the board-transfer evaluation just written. Please skip the second evaluation when isBoardChanging, and exclude board changes from isPureManualEtaChange.`

apps/backend/src/zero/mutators.ts
```diff
@@ -5894,50 +5916,71 @@ export function createMutators(
...
+            : null;
+          const boardEtaManagement = parseBoardEtaManagement(etaBoard?.metadata ?? null, etaBoard?.boardType ?? BoardType.DEFAULT);
+
+          // Stage-only moves are gated only once a board has opted into automatic ETA
+          // management (stage moves can now trigger auto-recompute) - scoped the same way as
+          // the nonLinear.transition / ticketStageEta.update gates, so boards not using this
+          // feature see no behavior change.
+          const isStageChangingGated = isStageChanging && boardEtaManagement.autoRecomputeEnabled;
+
+          if ((isAssigneeChanging || isEtaChanging || isBoardChanging || isStageChangingGated) && ticket.userGroupId) {
```
PiyushJuspay, line 5952: remove boardEtaManagement.autoRecomputeEnabled;

PiyushJuspay:
Q2 The FOR UPDATE fix locks metadata but eta is still read outside the lock
- What: All three Prisma paths now SELECT "metadata" ... FOR UPDATE and feed the locked metadata into evaluateEta — but still pass currentTicketEta: currentTicket.eta / ticket.eta, read before the transaction. The fix comment says "feeding this into evaluateEta too keeps the risk decision … based on current state" — only metadata was. eta is both a fingerprint input and the extend-only baseline, so a concurrent due-date edit landing in the window is decided against (and can be clobbered by) a stale value.
- Where: apps/backend/src/apps/controllers/ticketController.ts:463 (lock) + :480 (currentTicketEta: currentTicket.eta); apps/backend/src/database/repositories/ticketRepository.ts:637 + :649; apps/backend/src/services/stageTransition/ticketStageTransitionService.ts:481
- Fix: add "eta" to the locked SELECT and use it for currentTicketEta.
- detail:
ticketController.ts:463 — The lock selects only metadata, but currentTicketEta (line 480) still comes from the pre-transaction read. Please select "eta" in the same FOR UPDATE query and feed that into evaluateEta, otherwise the comment's "decision based on current state" doesn't hold for the due date.

Q1 Board settings screen reads the flag raw — Standard Path save silently turns auto-recompute OFF
- What: The toggle's initial state is parsed by hand from metadata.etaManagement.autoRecomputeEnabled === true. Every board relying on the default (ON, per the accepted defaultAutoRecomputeEnabled) has no etaManagement key, so the toggle shows off while the backend behaves on. Worse: handleSaveStandardPath sends that wrongly-initialized autoRecomputeEnabled: false along with the path — so an admin configuring a Standard Path on a NON_LINEAR board (the primary way to enable forecasting there) silently writes autoRecomputeEnabled: false in the same save, making the path they just saved inert.
- Where: apps/dashboard/src/components/Board/BoardRolesConfigScreen/BoardRolesConfigScreen.tsx:180 — setAutoRecomputeEnabled(Boolean(etaManagement && ... ['autoRecomputeEnabled'] === true)) and :381 — autoRecomputeEnabled, inside handleSaveStandardPath's mutator args
- Impact: feature setup defeats itself on NON_LINEAR; toggle misrepresents effective state on all default-on boards.
- Fix: initialize from the shared parseBoardEtaManagement(board.metadata, boardType) (exists, exported, boardType-aware — this drift is exactly why it exists), and omit autoRecomputeEnabled from the Standard Path save unless the user touched the toggle.
- detail:
BoardRolesConfigScreen.tsx:180 — The toggle parses etaManagement by hand, so boards relying on the default-on (no etaManagement key) show "off", and handleSaveStandardPath sends that false along — saving a Standard Path silently disables auto-recompute. Please initialize via parseBoardEtaManagement(board.metadata, boardType) and only send autoRecomputeEnabled when the user actually toggled it.



Q3 etaExpectedFingerprint / etaChangeReason are shipped dead — the "reason" is always empty
- What: No dashboard code ever sends either param (verified across apps/dashboard/src). Consequences: (a) the optimistic-concurrency guard for manual due-date edits can never fire; (b) every ETA_MANUALLY_UPDATED row — the type that replaced the generic ActivityType.ETA row specifically to "carry the reason" — records reason: ''. The stated design ("the dashboard's dedicated due-date editor is expected to always pass both") has no editor in this PR.
- Where: apps/backend/src/zero/mutators.ts:5844 (schema fields); writer at :6596 — reason: params.etaChangeReason ?? ''
- Impact: the manual-edit timeline rows are a strict downgrade in practice (same info as the old ETA row, plus an always-empty reason field), and two params + server checks + client schema support are dead weight until an editor ships.
- Fix: either wire the due-date editor to send both (it's the feature's own spec), or drop the params and the ETA_MANUALLY_UPDATED replacement until that editor exists.
- Detail:
mutators.ts:5844 — Nothing in the dashboard sends etaExpectedFingerprint/etaChangeReason, so the concurrency guard never fires and every ETA_MANUALLY_UPDATED row has an empty reason — the whole point of replacing the generic ETA row. Can we wire the due-date editor to send these, or hold the replacement until that editor ships?


Q4 standardPathUsed is hardcoded false — the "via the Standard Path" trail can never appear
- What: buildEtaActivityIntents always writes standardPathUsed: false, even when the forecast route was built from a configured Standard Path. The matching UI branch is dead code.
- Where: apps/backend/src/services/etaManagement/activityIntents.ts:174 — standardPathUsed: false, (UI: TicketActivity.tsx:294 — ' via the Standard Path')
- Fix: thread "route came from the Standard Path" out of resolveForecastRoute/evaluateEta into the intent context, or drop the field.
- detail:
activityIntents.ts:174 — standardPathUsed is hardcoded false, so the timeline can never say "via the Standard Path" even when one was used (and the UI branch for it is dead). Either thread the route source through or remove the field.

PiyushJuspay (approved):

mamtha-v (approved):

santhoshJus (approved):

PR #203, opened 2026-08-05
feat: custom form field conditions in ticket created & updated triggers

Himanshvarma:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Port `formFieldChanges`/`formFieldConditions` logic from the TICKET_UPDATED automation trigger to TICKET_CREATED, using the oneteam branch's value-matching schema (`{fieldId, match: 'changed'|'contains', value?}`) for both triggers.

**Backend changes:**
- `ticket-created.trigger.ts`: Rewritten to support `formFieldConditions` in trigger config. Fixed `matchFilters` which was firing unconditionally when no form fields were set (inverted guard bug).
- `ticket-updated.trigger.ts`: Ported oneteam branch's `FormFieldConditionSchema` (with superRefine validation), `resolveFormFieldConditions`, `matchesFormFieldCondition`, and 4-branch `matchFilters`.
- `automation-events.ts`: `TicketCreatedEventPayload` extended with `formFieldChanges?`, `performedBy?`, `boardId`, `channelId`, `projectId`.
- `ticketCustomFieldService.ts`: Exported `toTicketChangeValue`, added `buildCreationFormFieldChanges` (triple-keyed: fieldId/fieldName/lowercase for resilient matching).
- `ticketRepository.ts`: `createTicket` accepts `formFieldChanges?`, emits in TICKET_CREATED payload.
- `ticketutils.ts`: Builds `formFieldChanges` pre-transaction, passes to `createTicket`.
- `ticketController.ts`: Emits companion TICKET_CREATED event with `formFieldChanges` after `formEntityValues.createMany`.

**Dashboard changes:**
- `AutomationBuilder.utils.ts`: `buildVariableSources` now reads `formFieldConditions` (extracts `fieldId` from conditions array) for variable-picker injection.
- `TriggerCard.tsx`: Wired to `formFieldConditions` UI, dual-writes `formFieldConditions` + `formFieldIds` on change for TICKET_UPDATED backward compat.
- `TicketUpdatedFormFieldsSection.tsx`: Replaced with oneteam version (contains-match UI, `conditionsFromTriggerConfig`).

## Areas Touched

- [x] `apps/backend` (API / worker)
- [x] `apps/dashboard`
- [ ] `apps/electron`
- [ ] `apps/xyne-claw` / `apps/xyne-claw-auth`
[...]

PR #749, opened 2026-08-19
feat: XYNE-56504 Implement release manager v4 multi-repo support

sumantop:
## What
  Adds multi-repo support to the release manager (v4). A single release ticket can now span multiple repositories, each with its own commit range, instead of one
  repo per release.

  ## How
  - **New table `release_ticket_repos` (ReleaseTicketRepo) in the `non_zero` schema (server-only, not Zero-replicated):** one row per (release ticket × repo), 
  holding `mainReleaseBoardId`, `branch`, `deployedCommit`, `newCommit`. Absence of rows = single-repo legacy mode.
  - **Write path:** CreateTicketModal serializes per-repo ranges into `releaseRepos`; ticketController persists them via 
  `ReleaseRepository.createReleaseTicketRepos`. Analysis fires when `hasRepoRows || hasScalarRange` (backward compatible).
  - **Analysis:** `commitAnalysisController.deriveReleaseContexts` reads the rows and fans out one `ReleaseRepoContext` per repo; each repo gets its own VCS diff
  + release.
  - **UI:** ReleaseDetailScreen reads the rows via a backend HTTP endpoint (`GET /api/commits/analyze/repos/:releaseId`) and groups the Dev Tickets view per 
  repo; new ReleaseRepoConfigModal and repoVisual helpers.
  - **Access:** server-only table — writes via Prisma (`ReleaseRepository`), reads via the workspace-scoped HTTP endpoint above and `db.releaseTicketRepo` in 
  analysis. Intentionally **not** Zero-replicated (no shared/backend Zero schema, query, or ACL entry), keeping this table off Zero.

  ## Notes
  - Single-repo releases keep working via the scalar-range fallback.
  - Migration: `20260814120000_add_release_ticket_repos` (creates the table in the `non_zero` schema).


sumantop:
Screenshot 2026-08-19 at 6 29 50 PM

sumantop:
connecting a repo creates 1 main release board + 1 board per service: 
So 2 repos × 2 boards = 4 
image

apps/backend/scripts/seed.ts
```diff
@@ -44,6 +44,7 @@ const ESSENTIAL_RESOURCES = [
   { name: 'AUTH', description: 'Authentication endpoints (/api/auth/*)' },
   { name: 'XYNE-APPS', description: 'Xyne Apps management endpoints (/api/apps/*)' },
   { name: 'VESPA', description: 'Vespa backfill / reindex admin endpoints (/api/admin/vespa-backfill/*, /api/migration/vespa-workspace-backfill/*)' },
+  { name: 'RELEASEMANAGER', description: 'Release Manager tab and repository commit-analysis endpoints (/api/commits/analyze/*)' },
```
devesh-juspay: RELEASE-MANAGER

apps/backend/prisma/schema.prisma
```diff
@@ -3733,6 +3734,24 @@ model ApplicationReleaseTicket {
   @@schema("public")
 }
 
+model ReleaseTicketRepo {
```
devesh-juspay: Can we have table name as ReleaseRepository ?

apps/backend/src/routes/commitAnalysis.ts
```diff
@@ -89,12 +90,41 @@ router.post('/re-run/:ticketId', async (req: Request, res: Response): Promise<vo
...
+    logger.error(`[CommitAnalysis] re-run failed for ticket=${safeTicketId}: ${msg}`);
     res.status(500).json({ error: msg });
   }
 });
 
-router.post('/test-connection', authMiddleware.requireAdminOrOwner, async (req: Request, res: Response): Promise<void> => {
+// Per (release ticket × repo) commit ranges for a release ticket. Read from the
+// non_zero schema (server-only, not Zero-replicated) and served over HTTP — the
+// Release Detail screen fetches this instead of syncing the table via Zero.
+router.get('/repos/:releaseId', async (req: Request, res: Response): Promise<void> => {
```
devesh-juspay, line 104: we are not using any auth middleware here?

devesh-juspay (commented):
Review of the multi-repo release manager change — thanks, the architecture (server-only `non_zero.release_ticket_repos`, per-repo fan-out, scalar fallback, `RELEASEMANAGER` resource replacing `requireAdminOrOwner`) is the right shape, and typecheck is clean on the branch.

Inline comments below, grouped by severity:
- **Must fix:** aggregate `success` flag when all repos fail; inverted `releaseTrackingMode` guard; no `RELEASEMANAGER` grant backfill for existing admins (deploy will hide the tab for everyone).
- **Should fix:** dev-ticket rows dropped on config drift; missing `orderBy` for the primary repo; CodeQL log-injection remainder; `re-run` / `repos` routes unguarded; modal opened on a release board lands in task mode (suggest auto-switching to release mode rather than loosening the picker filter).
- **Minor:** silently dropped branch-less repos; silent fetch fallback.

Also: the branch conflicts with `main` only in `scripts/seed.ts` / `seed-acl.ts` (both sides appended a resource — `RELEASEMANAGER` vs `ROLES`); resolution is keep both.

Verified fine: per-repo error isolation, single-repo compat on both ends, `workspaceId` + ACL on the new table (incl. `createMany`), `/repos` workspace filtering, `non_zero` exclusion from the client Zero schema + validator, `CONCURRENTLY` index migration, canvas side-effect try/catch, HTML escaping in banners, no internal URLs/secrets.

apps/backend/src/controllers/commitAnalysisController.ts
```diff
@@ -289,164 +290,198 @@ export class CommitAnalysisController {
...
+          appMatchSummary.push(...slice.appMatchSummary);
+          repoSlices.push({
+            workspace: ctx.projectKey,
+            repoSlug: ctx.repoSlug,
+            deployedCommitId: ctx.deployedCommitId,
+            newCommitId: ctx.newCommitId,
+            results: slice.viewResults,
+          });
+        } catch (repoError) {
+          failedRepos.push(label);
```
devesh-juspay, line 512: **[Must fix] `analyzeCommits` still returns `success: true` when every repo fails.**

Per-repo isolation here is good (one repo failing no longer aborts the others), but the error never reaches the outer `catch`, and the function falls through to the unconditional `return { success: true, data: results, canvasUrl }` at the end of the loop — even when `failedRepos.length === contexts.length` and `results` is empty. On `main` a failure propagated as `success: false`.

Consumers act on the flag: `ticketController` logs "Commit analysis completed" and `/re-run/:ticketId` returns `{ success: true, canvasUrl: undefined }`, so the UI shows a success toast with no canvas.

Suggested: after the loop
```ts
if (contexts.length > 0 && results.length === 0 && failedRepos.length > 0) {
  return { success: false, error: `Analysis failed for: ${failedRepos.join(', ')}` };
}
```
(keep the partial-success + warning banner behaviour when at least one repo succeeded).

apps/backend/src/controllers/commitAnalysisController.ts
```diff
@@ -215,45 +231,30 @@ export class CommitAnalysisController {
...
+      select: { boardType: true, vcsProvider: true, releaseTrackingMode: true },
     });
+    if (!board) return null;
 
-    if (!ticket?.boardId || !ticket.board) {
-      return null;
-    }
-
-    if (ticket.board.releaseTrackingMode !== ReleaseTrackingMode.COMMIT_RANGE) {
+    if (board.releaseTrackingMode === ReleaseTrackingMode.VERSION) {
```
devesh-juspay: **[Must fix] Tracking-mode guard is inverted.**

`main` had `if (releaseTrackingMode !== ReleaseTrackingMode.COMMIT_RANGE) skip`; this now only skips `VERSION`. `Board.releaseTrackingMode` is nullable, so unconfigured boards (`null`) now pass into commit-range analysis — and in the multi-repo path `mainReleaseBoardId` comes straight from client-supplied `dynamicFields.releaseRepos`.

Please restore the positive check: `if (board.releaseTrackingMode !== ReleaseTrackingMode.COMMIT_RANGE) { … skip … }`.

apps/backend/src/controllers/commitAnalysisController.ts
```diff
@@ -289,164 +290,198 @@ export class CommitAnalysisController {
...
+    releaseTicketId: string,
+    fallback: { deployedCommitId: string; newCommitId: string; branch: string },
+  ): Promise<{ contexts: ReleaseRepoContext[]; skipped: string[] }> {
+    const ticket = await db.ticket.findUnique({
+      where: { id: releaseTicketId },
+      select: { projectId: true, boardId: true, project: { select: { code: true } } },
+    });
+    if (!ticket?.boardId) return { contexts: [], skipped: [] };
+
+    const rows = await db.releaseTicketRepo.findMany({ where: { releaseId: releaseTicketId } });
```
devesh-juspay: **[Should fix] No `orderBy` — `contexts[0]` ("primary") is non-deterministic.**

`contexts[0]` drives the canvas title / workspace / repoSlug metadata and the parent-ticket labels. Postgres doesn't guarantee row order without `ORDER BY`, so the "primary" repo can differ between the create-time run and a later re-run. The new `GET /repos/:releaseId` already orders by `createdAt desc, id desc` (and the migration adds the matching composite index) — use the same here so both agree.

apps/backend/src/services/permissionMatrix.ts
```diff
@@ -95,6 +96,7 @@ export const PERMISSION_MATRIX: Record<WorkspaceRole, readonly PermissionEntry[]
     { resourceName: 'PROJECTS', accessType: AccessType.ADMIN },
     { resourceName: 'PRODUCT-INSIGHTS', accessType: AccessType.ADMIN },
     { resourceName: 'LISTPROJECTS', accessType: AccessType.ADMIN },
+    { resourceName: 'RELEASEMANAGER', accessType: AccessType.ADMIN },
```
devesh-juspay: **[Must fix — rollout] No backfill for existing ADMIN/OWNER users.**

The resource row is created at boot (`configSyncService`) and this grants `RELEASEMANAGER` ADMIN to the ADMIN/OWNER roles — but the matrix is only applied by `grantPermissionsForRole` on user creation / role change. Nothing re-applies it to *existing* users, so on deploy no current admin has a `ResourceAccess` row: the Release Manager tab disappears for everyone and `test-connection` 403s until grants are added by hand.

Needs a backfill (migration or script granting `RELEASEMANAGER` ADMIN to existing ADMIN/OWNER users — same shape as `migration/roleFrameworkBackfill.ts`). Also worth stating in the PR body that MEMBER gets no default grant (intentional — per-user grants), so ops know to grant release folks explicitly.

apps/dashboard/src/routes/ReleaseDetailScreen/releaseReport.utils.ts
```diff
@@ -193,6 +201,91 @@ export function buildReleaseDetailDevTicketRows(
...
+      const bucket = rowsByRepo.get(repoId);
+      if (bucket) bucket.push(row);
+      else rowsByRepo.set(repoId, [row]);
+    } else {
+      unmapped.push(row);
+    }
+  }
+
+  const groups: DevTicketRepoGroup[] = [];
+  for (const repo of repos ?? []) {
```
devesh-juspay, line 289: **[Should fix] Dev-ticket rows can silently disappear when live config drifts from the release snapshot.**

`boardToRepo` is built from the project's *current* `applications` (live config), while `repos` is the `release_ticket_repos` snapshot written at release creation. A row whose app now maps to a `mainReleaseBoardId` that isn't in `repos` lands in `rowsByRepo` but is never iterated here and isn't pushed to `unmapped` — it vanishes from the Dev Tickets table with no indication. Reachable via the new `ReleaseRepoConfigModal` (moving an app to another main board after the release exists), hotfix sync adding ART rows, or a partially-persisted `releaseRepos`.

Suggested: iterate the union of `repos` and `rowsByRepo.keys()` (label unknown keys via `repoMeta`), or route unknown keys into `unmapped` — so no row is ever lost.

apps/backend/src/routes/commitAnalysis.ts
```diff
@@ -89,12 +90,41 @@ router.post('/re-run/:ticketId', async (req: Request, res: Response): Promise<vo
...
+  try {
+    const repos = await db.releaseTicketRepo.findMany({
+      where: { releaseId, workspaceId },
+      orderBy: [{ createdAt: 'desc' }, { id: 'desc' }],
+    });
+    res.json({ repos });
+  } catch (e) {
+    const msg = e instanceof Error ? e.message : String(e);
+    const safeReleaseId = String(releaseId).replace(/[\r\n]/g, '');
+    logger.error(`[CommitAnalysis] repos fetch failed for release=${safeReleaseId}: ${msg}`);
```
devesh-juspay: **[Should fix] CodeQL log injection — `msg` is still interpolated unsanitised.**

`safeReleaseId` fixes the path param, but `msg` (the exception message) goes into the same template string; if a DB/Prisma error echoes the raw input, the vector reopens. Prefer structured logging: `logger.error('[CommitAnalysis] repos fetch failed', { releaseId: safeReleaseId, error: msg })` — this is what closed the equivalent CodeQL alert in #889.

apps/backend/src/routes/commitAnalysis.ts
```diff
@@ -89,12 +90,41 @@ router.post('/re-run/:ticketId', async (req: Request, res: Response): Promise<vo
     res.json({ success: result.success, canvasUrl: result.canvasUrl, error: result.error });
   } catch (e) {
     const msg = e instanceof Error ? e.message : String(e);
-    logger.error(`[CommitAnalysis] re-run failed for ticket=${ticketId}: ${msg}`);
+    const safeTicketId = String(ticketId).replace(/[\r\n]/g, '');
```
devesh-juspay: Same as below (line 122): sanitise / structure-log `msg` here too, not just the ticket id.

apps/backend/src/routes/commitAnalysis.ts
devesh-juspay, line 104: **[Should fix] Guard inconsistency.**

`test-connection` is now `authorize('RELEASEMANAGER', AccessType.READ)` 👍 — but this new `GET /repos/:releaseId` and the sibling `POST /re-run/:ticketId` have no `authorize` at all (only the global `authenticate`). Any workspace member can enumerate a release's repo ranges or trigger a re-run (which spends the VCS token). Apply the same `authorize('RELEASEMANAGER', READ)` to both.

apps/dashboard/src/components/Tickets/CreateTicketModal/CreateTicketModal.tsx
```diff
@@ -766,6 +862,7 @@ export const CreateTicketModal: React.FC<CreateTicketModalProps> = ({
   useEffect(() => {
     if (isOpen) {
       form.reset();
+      setTicketKind(initialTicketKind);
```
devesh-juspay, line 916: **[Should fix] Opening the modal on a release board lands in task mode.**

`KanbanBoardScreen` / `BoardsTable` open this modal with `selectedBoardId` = a release board but `ticketKind` defaults to `'task'`; since `boardOptions` now filters out release boards, the picker shows "Select board" and the board can't be re-picked. Rather than loosening the filter, auto-switch to release mode when the preselected board is a release board:

```ts
const preselected = selectedBoardId ? boards?.find(b => b.id === selectedBoardId) : undefined;
setTicketKind(preselected && isReleaseBoard(preselected.boardType) ? 'release' : initialTicketKind);
```
(and seed the release-mode primary board from `selectedBoardId`) — so the Kanban entry point behaves like `ProjectDetailScreen`'s explicit `initialTicketKind='release'`.

apps/dashboard/src/components/Tickets/CreateTicketModal/CreateTicketModal.tsx
```diff
@@ -1546,18 +1719,103 @@ export const CreateTicketModal: React.FC<CreateTicketModalProps> = ({
...
-        label: board.name,
-        value: board.id,
-        icon: (
-          <span className='bg-primary text-primary-foreground text-xs aspect-square size-4 rounded text-center'>
-            {board.name.charAt(0)}
-          </span>
-        ),
-      })) ?? [],
+      boards
+        ?.filter(board => !isReleaseBoard(board.boardType))
```
devesh-juspay, line 1884: Related to the comment at the open/reset effect above: keep this filter (it correctly keeps release boards out of the *task* picker), but make sure the preselected-release-board entry path switches to release mode instead of showing an empty picker.

apps/backend/src/controllers/ticketController.ts
```diff
@@ -1343,19 +1344,48 @@ export class TicketController {
...
+            completeRepoRows = parsed
+              .map((r: any) => ({
+                workspaceId: ticket.workspaceId,
+                releaseId: ticket.id,
+                mainReleaseBoardId: String(r?.mainReleaseBoardId ?? '').trim(),
+                branch: String(r?.branch ?? '').trim(),
+                deployedCommit: String(r?.deployedCommit ?? '').trim(),
+                newCommit: String(r?.newCommit ?? '').trim(),
+              }))
+              .filter((r) => r.mainReleaseBoardId && r.branch && r.deployedCommit && r.newCommit);
```
devesh-juspay, line 1579: **[Minor] Entries missing `branch` are silently dropped.**

The modal only gates `deployedCommit`/`newCommit` for additional repos (`branch` isn't required client-side, unlike the primary board's), so a repo whose branch was cleared is dropped here with only a warn log — the user thinks the release covers N repos, it covers N-1. Either enforce `branch` in the modal for additional repos, or reject the request with a clear error instead of filtering.

apps/dashboard/src/routes/ReleaseDetailScreen/ReleaseDetailScreen.tsx
```diff
@@ -333,6 +349,31 @@ const ReleaseDetailScreen = (): ReactElement => {
     [artPage, artHasMore],
   );
 
+  // release_ticket_repos lives in the non_zero schema (server-only, not
+  // Zero-replicated), so it's fetched over HTTP instead of synced via Zero.
+  // Refetched after a re-run — the only in-screen action that changes these rows.
+  const [releaseTicketRepos, setReleaseTicketRepos] = useState<ReleaseTicketRepoRow[]>([]);
+  const fetchReleaseTicketRepos = useCallback(async (): Promise<void> => {
+    if (!releaseTicketId) {
+      setReleaseTicketRepos([]);
```
devesh-juspay: **[Minor] Silent fallback.** A failed `fetchReleaseTicketRepos` becomes `[]`, which renders exactly like a genuine single-repo release. Fine if intentional (graceful degradation), but a toast/inline note on error would avoid "where did my repo grouping go?" confusion.

apps/backend/src/routes/commitAnalysis.ts
sumantop: Auth is applied at the mount (authMiddleware.authenticate on the whole /api/commits/analyze router), and the handler is workspace-scoped -> it's authenticated and tenant-isolated. It intentionally doesn't add the RELEASE-MANAGER RBAC check because all sibling release data is workspace-read-only by the same model. Happy to add authorize('RELEASE-MANAGER', READ) if we want to tighten that.

sumantop: Gated /re-run with authorize('RELEASE-MANAGER', READ) since it burns the VCS token.
 and hid the button for non-RM folks so no ugly 403
 
/repos is open on purpose. The release screen isn’t RM-only — users can reach it from a release ticket (“Open release view” in TicketDetails) or the project’s releases table.

Also, the other release data on the page (events, changes, attributions) is already visible to all members. So only blocking /repos would partly lock the page and cause valid users to get a 403.

apps/backend/src/controllers/commitAnalysisController.ts
```diff
@@ -289,164 +290,206 @@ export class CommitAnalysisController {
...
+      select: { projectId: true, boardId: true, project: { select: { code: true } } },
+    });
+    if (!ticket?.boardId) return { contexts: [], skipped: [] };
+
+    const rows = await db.releaseRepository.findMany({
+      where: { releaseId: releaseTicketId },
+      orderBy: [{ createdAt: 'desc' }, { id: 'desc' }],
+    });
+
+    const entries =
```
devesh-juspay: **Hotfix-delta sync analyzes the wrong range.** When `release_repositories` rows exist (every modal-created release has ≥1), the caller-supplied `deployedCommitId/newCommitId/branch` are replaced by the stored row range, and rows are never updated after creation. So the webhook path (`releaseWebhookSync.ts:159-169`, `hotfixSync: true`, range = frozen head → merge SHA) ends up re-running `release()` over the whole original range with `isHotFix=true`: original PRs land in the 🔥 Hotfix PRs section, original dev-ticket ART rows get tagged hotfix, and the actual delta is never analyzed.

Fix: when `hotfixSync` is set, honor the caller range for the matching board (or update `release_repositories.newCommit` on sync).

apps/backend/src/controllers/commitAnalysisController.ts
```diff
@@ -289,164 +290,206 @@ export class CommitAnalysisController {
...
+  ): Promise<{ contexts: ReleaseRepoContext[]; skipped: string[] }> {
+    const ticket = await db.ticket.findUnique({
+      where: { id: releaseTicketId },
+      select: { projectId: true, boardId: true, project: { select: { code: true } } },
+    });
+    if (!ticket?.boardId) return { contexts: [], skipped: [] };
+
+    const rows = await db.releaseRepository.findMany({
+      where: { releaseId: releaseTicketId },
+      orderBy: [{ createdAt: 'desc' }, { id: 'desc' }],
```
devesh-juspay, line 314: **Wrong primary repo.** Rows are ordered `createdAt desc, id desc`, but the ticket's own board is inserted *first* in the same `createMany` (monotonic cuids), so the last-added repo sorts first and `const primary = contexts[0]` (line ~472) picks it. Canvas title, persisted canvas metadata and every argument to `postToParentTicket` then describe the additional repo instead of the release board's repo.

Fix: order asc, or select the row whose `mainReleaseBoardId === ticket.boardId` as primary.

apps/backend/src/controllers/commitAnalysisController.ts
```diff
@@ -289,164 +290,206 @@ export class CommitAnalysisController {
...
+
+      const failedRepos: string[] = [];
+      for (const ctx of contexts) {
+        const label = `${ctx.projectKey}/${ctx.repoSlug}`;
+        try {
+          const slice = await this.analyzeReleaseContext(ctx, params);
+          results.push(...slice.results);
+          viewResults.push(...slice.viewResults);
+          affectedApplications.push(...slice.affectedApplications);
+          migrationLinks.push(...slice.migrationLinks);
```
devesh-juspay, line 501: **Change summary is not repo-scoped.** `release()` → `buildReleaseChangeSummary(projectKey, repoSlug, releaseId)` reads *all* `release_change_types` for the release with no repo filter and builds URLs / re-fetches diffs with the *current* repo's coordinates. With two repos, repo 2's call returns repo 1's rows too: `diffUrl` points at `KEY2/slug2/commits/<repo1-sha>`, `getFileDiff` against the wrong repo fails → `newValue ''`, migration count doubles (`Yes (4 files)`), and `indexMigrationLinksByCommit` (last-write-wins) overwrites repo 1's links with wrong-repo URLs.

Fix: scope `buildReleaseChangeSummary` to the current repo's `applicationIds`, or call it once after the loop with per-application coordinates.

apps/dashboard/src/components/Tickets/CreateTicketModal/CreateTicketModal.tsx
```diff
@@ -1182,10 +1362,11 @@ export const CreateTicketModal: React.FC<CreateTicketModalProps> = ({
         formDataPayload.append('title', formData.title.trim());
         formDataPayload.append('description', formData.description.trim());
         formDataPayload.append('boardId', formData.boardId);
-        // For subtickets or AI-initiated tickets, use the selected channel from form; otherwise use the prop
+        // For subtickets, AI-initiated tickets, or a release-only launch, use the
+        // channel picked in the form; otherwise use the prop (the current channel).
         formDataPayload.append(
           'channelId',
-          isFromSubTicket || isFromAI ? formData.channelId : channelId,
+          isFromSubTicket || isFromAI || releaseOnly ? formData.channelId : channelId,
```
devesh-juspay, line 1431: **JSON submit path still ignores the selected release channel.** This multipart branch was updated to use `formData.channelId` in `releaseOnly` mode, but the no-attachment JSON path at line 1472 still has `isFromSubTicket || isFromAI ? formData.channelId : channelId`. With `releaseChannelIds=[A,B]` and the user picking B, a submit without attachments creates the release ticket and its analysis conversation in A.

Fix: add `|| releaseOnly` at line 1472 to match this line.

apps/dashboard/src/components/Tickets/CreateTicketModal/CreateTicketModal.tsx
```diff
@@ -1067,32 +1181,75 @@ export const CreateTicketModal: React.FC<CreateTicketModalProps> = ({
     ],
   );
 
+  const releaseGateMessage = useMemo(() => {
+    if (ticketKind !== 'release') return null;
+    if (!formValues?.boardId) return 'Select at least one repository';
+    const df = formValues?.dynamicFields ?? {};
+    const filled = (v: string | string[] | undefined): boolean =>
+      Array.isArray(v) ? Boolean(v[0]?.trim()) : Boolean(v?.trim());
+    const primaryComplete = filled(df['deployedCommitId']) && filled(df['newCommitId']);
```
devesh-juspay: **Gate omits `branch` for the primary repo.** Additional repos (1191-1196) and the new backend validation (`ticketController.ts:560-568`, which checks `branch` for every `releaseRepos` entry including index 0) both require it. If the user clears the auto-filled Branch input and fills both commits, the gate returns null, *Create Release* enables, and the server replies 400 `Each release repo needs mainReleaseBoardId, branch, deployedCommit and newCommit…` surfaced as a generic error.

Fix: `const primaryComplete = filled(df['branch']) && filled(df['deployedCommitId']) && filled(df['newCommitId']);`

apps/dashboard/src/routes/ReleaseDetailScreen/ReleaseDetailScreen.tsx
```diff
@@ -566,13 +646,130 @@ const ReleaseDetailScreen = (): ReactElement => {
...
+
   return (
     <div className='h-full bg-muted flex flex-col'>
       <div className='flex-1 overflow-auto p-8'>
         <div className='max-w-none'>
           <button
             onClick={() =>
-              void navigate(`/listProjects/${projectId}`, { state: { tab: 'release' } })
+              void navigate(`/listProjects/${projectId}`, {
+                state: { tab: 'release', from: 'releaseManager' },
```
devesh-juspay: **Lands on the Repositories tab.** In release-manager mode `ProjectDetailScreen` now maps `'release'` → Repositories and the Releases list → `'releases'` (`ProjectDetailScreen.tsx:519-521, 757`), and `initialTab` takes `location.state.tab` verbatim. So *Back to Releases* opens the repo-config tab. Correct on main (single `'release'` tab); this PR introduced `'releases'` without updating this caller.

Fix: `state: { tab: 'releases', from: 'releaseManager' }`.

apps/backend/src/zero/mutators.ts
```diff
@@ -7021,6 +7042,8 @@ export function createMutators(
             throw new Error('All applications in a release group must use the same repository URL');
           }
 
+          const vcsProvider = inferVcsProvider([...normalizedRepoUrls][0]);
```
devesh-juspay: **`vcsProvider` arg is accepted but never read.** `saveReleaseBoardConfig` takes an optional `vcsProvider` (line ~6938) but always overwrites the stored provider with `inferVcsProvider(repoUrl)` on both update and insert, and `inferVcsProvider` only returns GITHUB for `github.com` / `github.<domain>` hosts. An existing board stored as GITHUB with a GitHub Enterprise URL flips to `BITBUCKET_SERVER` on the next *Edit repository* / *Add service* save, silently.

Either honor and validate the incoming arg (reject on mismatch), or drop it from the mutator schema so the API surface isn't misleading.

apps/dashboard/src/routes/ProjectDetailScreen/ProjectDetailScreen.tsx
```diff
@@ -76,7 +96,27 @@ const ProjectDetailScreen = (): ReactElement => {
   // Initial tab can be overridden via `navigate(..., { state: { tab } })` —
   // ReleaseDetailScreen's "Back" button uses this to return to the Releases tab
   // instead of the default Boards tab.
-  const initialTab = (location.state as { tab?: TabValue } | null)?.tab ?? 'boards';
+  const navState = location.state as { tab?: TabValue; from?: string } | null;
+  const initialTab = navState?.tab ?? 'boards';
+  // Entry point gates the tab set: Release Manager shows release-repo config,
+  // List Projects shows the SDLC repositories view.
+  const fromReleaseManager = navState?.from === 'releaseManager';
```
devesh-juspay, line 101: **Release-Manager mode is keyed off transient navigation state.** The tab set, *Connect Repository* / *Create Release* buttons, back-link and the sidebar highlight (`AppSidebar.tsx:196-198`) all depend on `location.state.from === 'releaseManager'`. A bookmarked or pasted `/listProjects/<id>` URL for a release project opens with `state = null` → List-Projects mode: no way to connect another repo or create a release, and *Create Board* (which this PR stripped of the Release option) offers only Standard/Flow. Reload is fine (history.state survives); fresh tab is not.

Proper fix is a dedicated route family for Release Manager so mode derives from the URL. If that's out of scope here, worth a follow-up ticket.

devesh-juspay (commented):
Review of the multi-repo release changes. Inline comments cover the confirmed issues; two more that can't be anchored inline:

**`apps/backend/src/services/release/releaseWebhookSync.ts:97` — PR merges into non-primary repos never sync.** `syncReleaseOnPRMerge()` resolves candidate tickets by `ticket.boardId` only (the primary board) and never consults `release_repositories`. A PR merging into repo B's release branch resolves `boardIds=[Q]`, finds no ticket with `boardId in [Q]`, and takes the *No active release tickets on matched boards* early return — no hotfix section, no ART rows, no re-run for repo B. Fix: also match via `release_repositories.mainReleaseBoardId`.

**Lower-severity, worth a look:**
- `apps/dashboard/src/utils/boardUtils.ts:10` — `isMainReleaseBoard` keys on truthy `vcsProvider`; legacy boards (NULL until the backfill SQL runs) disappear from the Create Release picker while every other screen uses `Application.mainReleaseBoardId`.
- `analyzeCommits` runs the VCS fan-out twice per repo (pre-existing, now × repo count) — Bitbucket rate-limit risk; passing precomputed results into `release()` is the cheap win.
- N+1 board lookups in `deriveReleaseContexts` (incl. a second `findUnique` on the skip path just for `name`); sequential independent reads in `GET /repos/:releaseId`.
- `releaseRepos` is parsed twice in `ticketController.ts` with different semantics (~549 rejects with 400, ~1389 silently filters).
- Dead wizard-step machinery left in `useReleaseConfigForm.ts` after step 1 was removed; `repoVisual.tsx` reimplements `inferRepositoryNameFromUrl` with a `/bitbucket/i` substring test that can disagree with the backend badge.

Not flagged: the general ticket modal no longer lists release boards, so Hotfix-*type* release tickets can't be created from the UI. That's the intended model (Hotfix is a dev-ticket type on normal boards; release boards carry Release only). The backend still branches on `ticketType === Hotfix` as a release (`isReleaseTicket()` and friends) — cleanup in a separate PR.

apps/backend/src/controllers/commitAnalysisController.ts
sumantop: hotfix sync now threads a hotfixOverride { boardIds, mergeCommitSha } from the webhook into deriveReleaseContexts, which scopes analysis to the merged repo's board and uses (that repo's frozen head → mergeSHA] instead of the stored frozen range. Sibling repos are untouched; legacy scalar releases unchanged.

apps/backend/src/controllers/commitAnalysisController.ts
sumantop: deriveReleaseContexts now promotes the entry matching ticket.boardId to the front, so primary = contexts[0] is the release board's repo rather than the last-added one. Canvas title/metadata and the postToParentTicket headline now describe the release board's repo. Falls back to the first row if the ticket's board isn't among the repos (e.g. hotfix override).

apps/backend/src/controllers/commitAnalysisController.ts
sumantop: buildReleaseChangeSummary now takes mainReleaseBoardId, resolves that board's applications, and scopes the read to where: { releaseId,  applicationId: { in: applicationIds } }. Each repo's call returns only its own rows, so diff URLs use the right coordinates and the aggregated env/migration counts no longer double-count (which also removes the last-write-wins overwrite). Single-repo behavior is unchanged.

apps/dashboard/src/components/Tickets/CreateTicketModal/CreateTicketModal.tsx
sumantop: added || releaseOnly to the JSON path's channelId so it matches the multipart branch; a release ticket created without attachments now uses the picked release channel (formData.channelId) instead of the prop.

apps/dashboard/src/components/Tickets/CreateTicketModal/CreateTicketModal.tsx
sumantop: primaryComplete now checks filled(df['branch']) too, so clearing the branch keeps Create Release disabled (matching the additional-repos gate and the backend validation). Verified in the running app: blanking the primary branch disables the button.

apps/dashboard/src/routes/ReleaseDetailScreen/ReleaseDetailScreen.tsx
sumantop: the Back button now navigates with tab: 'releases'. In release-manager mode ProjectDetailScreen maps 'release' → Repositories and 'releases' → the Releases list, so "Back to Releases" now lands on the list instead of the repo-config tab. (The project-card entry intentionally keeps tab: 'release' to open Repositories for configuration.)

apps/backend/src/zero/mutators.ts
sumantop: both handled:

  1. Arg never read — dropped vcsProvider from saveReleaseBoardConfig (backend + shared mutator schema) as part of dead-code cleanup, so the API surface no longer accepts it (your option "drop it from the schema").
  2. Silent GHE flip — the update path no longer re-infers the provider; it preserves existingMainBoard.vcsProvider. A repo's provider is fixed once at creation, so an existing GITHUB board with a GHE host no longer flips to BITBUCKET_SERVER on an unrelated regex/path edit. (Full GHE-on-custom-domain  creation still needs a GitHub-host allowlist config — out of scope here; flagging separately.)

apps/dashboard/src/routes/ProjectDetailScreen/ProjectDetailScreen.tsx
sumantop: this is a known limitation of the state-based mode

devesh-juspay (commented):
Second pass on `756d3cc8`. Fixed since last round: hotfix-delta range (#1), primary-repo ordering, repo-scoped change summary, release-channel JSON path, branch gate, *Back to Releases* tab, `vcsProvider` arg removed. 👍

Inline comments cover what's still pending. Summary:

**Blocks merge**
- `ci / Dashboard typecheck` fails at `prettier --check` — 4 files, see inline on `CreateTicketModal.tsx`.

**PR-introduced CodeQL alerts (small, worth fixing here)**
- High: ReDoS in `parseBitbucketRepoUrl` (`repoUrlParser.ts:69`) — see inline.
- Medium: log injection via `err.message` (`ticketController.ts:1166`) — see inline.

**Still open from round one (multi-repo / UX only, no single-repo regression)**
- Non-primary-repo webhook sync (`releaseWebhookSync.ts`) — see inline.
- Release-Manager mode keyed off `location.state` (`ProjectDetailScreen.tsx:105`) — see inline.

**Pre-existing CodeQL alerts flagged as "new" only because `main` was never analyzed — not this PR, suggest a separate small PR**
- 2 × Critical `js/request-forgery` in `apps/backend/src/services/release/repoInspector.ts:43, :86` (from #398). Host comes from `config.github.apiUrl` / `config.bitbucket.baseUrl`, only the parsed `owner/repo` segments are user-supplied and the endpoint is admin-gated — real exposure is path traversal into the same API with the server token (e.g. `owner = ..`). `encodeURIComponent` the segments and reject `..`.
- 2 × High `js/remote-property-injection` in `apps/backend/src/services/ticketCustomFieldService.ts:394-395` (from #203): `changes[field.fieldName] = …` with a DB-sourced name; a field named `__proto__` would pollute. Use `Object.create(null)` / a `Map`, or skip `__proto__`/`constructor`/`prototype`.

CodeQL is not in the required-checks list, so it doesn't gate the merge.

apps/backend/src/services/release/releaseWebhookSync.ts
```diff
@@ -160,13 +160,16 @@ export async function syncReleaseOnPRMerge(params: ReleaseMergeSyncParams): Prom
           conversationId: ticket.conversationId,
           userId: bot?.id || ticket.createdBy,
           channelId: ticket.channelId || undefined,
+          // Scalar range — used only when the release has no release_repositories
```
devesh-juspay, line 187: **Still open: merges into a non-primary repo never sync.** `hotfixOverride` fixes *which range* is analyzed, but only once a ticket is found — and the candidate lookup above (`ticketWhere` at ~98-100) is still `boardId: { in: boardIds }`, i.e. the release ticket's own board only. For a release on board A with `release_repositories` rows for board B, a PR merging into repo B resolves `boardIds=[Q]`, finds no ticket, and hits the *No active release tickets on matched boards* early return. `release_repositories` is never consulted here.

Fix: also match tickets via `release_repositories.mainReleaseBoardId in boardIds`, then use the matching row's `branch` for the base-branch comparison instead of the scalar form value.

apps/backend/src/utils/repoUrlParser.ts
```diff
@@ -103,3 +105,15 @@ export function parseGitHubRepoUrl(url: string): { owner: string; repo: string }
...
   return { owner, repo };
 }
+
+export function inferVcsProvider(repoUrl: string): VCSProviderType {
+  return parseGitHubRepoUrl(repoUrl) ? VCSProviderType.GITHUB : VCSProviderType.BITBUCKET_SERVER;
+}
+
+// Like inferVcsProvider but returns null for an unrecognized URL, so callers can
+// reject it instead of defaulting to a misleading Bitbucket-flavored error.
+export function detectVcsProvider(repoUrl: string): VCSProviderType | null {
```
devesh-juspay, line 121: **CodeQL (high) — polynomial ReDoS in `parseBitbucketRepoUrl`** (`js/polynomial-redos`, line 69): the clone-URL regex `/(?:\/scm)?\/([^/]+)\/([^/?#]+?)(?:\.git)?\/?(?:[?#].*)?$/` has a lazy quantifier followed by several optionals and can go polynomial on adversarial input. Input is an admin-gated repo URL so practical risk is low, but the fix is also cleaner: drop the regex — strip `?…`/`#…`, trailing `/` and `.git`, split on `/`, take the last two segments.

Same rewrite lets you tighten the over-match: today `detectVcsProvider('https://ghe.corp.com/org/repo')` and `gitlab.com/a/b.git` both return `BITBUCKET_SERVER` because any `host/x/y` matches the clone pattern, so the *Unsupported URL* error added here never fires for those. Requiring `/scm/` or a `/projects/…/repos/…` path (or a `bitbucket` host) for the Bitbucket branch would make the new error message reachable.

apps/backend/src/controllers/ticketController.ts
```diff
@@ -1120,7 +1159,13 @@ export class TicketController {
 
       const ticketChannelId = sourceConversationId ? validatedConversation.channelId : channelId;
       if (ticketChannelId) {
-        await this.channelRepository.updateLastActivity(ticketChannelId);
+        try {
+          await this.channelRepository.updateLastActivity(ticketChannelId);
+        } catch (err) {
+          logger.warn(
+            `[createTicket] Skipped last-activity update for channel ${String(ticketChannelId).replace(/[\r\n]/g, '')}: ${err instanceof Error ? err.message : String(err)}`,
```
devesh-juspay: **CodeQL (medium) — log injection** (`js/log-injection`): the channel id is `\r\n`-stripped but `err.message` is interpolated raw. Apply the same `.replace(/[\r\n]/g, '')` to the message (or log `err` as a structured field rather than in the template string).

apps/dashboard/src/routes/ProjectDetailScreen/ProjectDetailScreen.tsx
```diff
@@ -82,7 +98,27 @@ const ProjectDetailScreen = (): ReactElement => {
...
```
devesh-juspay, line 101: **Still open: Release-Manager mode is keyed off transient navigation state.** The tab set, *Connect Repository* / *Create Release* buttons, back-link and the sidebar highlight all depend on `location.state.from === 'releaseManager'`. A bookmarked/pasted `/listProjects/<id>` for a release project opens with `state = null` → List-Projects mode: no way to add a repo or create a release from that page (reload is fine; fresh tab is not).

Not blocking for merge. Cheap interim: also treat the project as release-managed when it already has a main release board (`isMainReleaseBoard` over `boards`). Proper fix: a dedicated `/releaseManager/:projectId` route so mode derives from the URL — follow-up ticket.

apps/dashboard/src/components/Tickets/CreateTicketModal/CreateTicketModal.tsx
```diff
@@ -1217,10 +1404,11 @@ export const CreateTicketModal: React.FC<CreateTicketModalProps> = ({
         formDataPayload.append('title', formData.title.trim());
         formDataPayload.append('description', formData.description.trim());
         formDataPayload.append('boardId', formData.boardId);
-        // For subtickets or AI-initiated tickets, use the selected channel from form; otherwise use the prop
+        // For subtickets, AI-initiated tickets, or a release-only launch, use the
+        // channel picked in the form; otherwise use the prop (the current channel).
         formDataPayload.append(
           'channelId',
-          isFromSubTicket || isFromAI ? formData.channelId : channelId,
+          isFromSubTicket || isFromAI || releaseOnly ? formData.channelId : channelId,
```
devesh-juspay, line 1431: **Blocks merge — `ci / Dashboard typecheck` is red on `prettier --check`, not tsc.** Four PR files need formatting:
- `src/components/Release/ReleaseConfigWizard/ReleaseConfigWizard.tsx`
- `src/components/Tickets/CreateTicketModal/createTicket.utils.ts`
- `src/components/Tickets/CreateTicketModal/CreateTicketModal.tsx`
- `src/routes/ReleaseDetailScreen/ReleaseDetailScreen.tsx`

`cd apps/dashboard && npx prettier --write <the four files>` and push. That check is required by the main-branch ruleset; everything else in `ci /` is green.

apps/backend/src/services/release/releaseWebhookSync.ts
sumantop: The lookup keyed only on boardId in boardIds, and a multi-repo release ticket lives only on its primary board — so a non-primary merge found no ticket and returned early. Now it also matches tickets via release_repositories.mainReleaseBoardId in boardIds, and uses the matched row's branch/newCommit for the base-branch check and hotfix boundary instead of the scalar (primary-only) form values. Single-repo releases fall back to the form values unchanged.

apps/backend/src/utils/repoUrlParser.ts
sumantop: ReDoS is fixed as suggested — dropped the regex, strip ?…/#…/.git/trailing /, split on /, take the last two segments.

On the over-match: left it permissive on purpose. Bitbucket Server SSH clones are host/KEY/slug.git with no /scm/ or /projects/ marker — structurally dentical to gitlab.com/a/b.git — so requiring those paths would break a supported Bitbucket form. Since detectVcsProvider tries GitHub first and only falls back to Bitbucket, the over-match only affects the error message for unsupported inputs (GitLab, or GHE on a non-github.* host). Can add a bitbucket-host heuristic if you'd prefer, but it'd still miss generic-host SSH clones.

apps/dashboard/src/routes/ProjectDetailScreen/ProjectDetailScreen.tsx
sumantop: Agreed, and not blocking. Holding this for the follow-up with the proper fix — a dedicated /releaseManager/:projectId route so mode derives from the URL. Skipping the isMainReleaseBoard interim on purpose: the same project-detail page intentionally shows the SDLC repositories view from List Projects vs the  release config from Release Manager, and a project can have both — so a board-based fallback would hijack the SDLC view for release-managed projects.

devesh-juspay (dismissed):

devesh-juspay (approved):

ameer030 (approved):

abhishekverma-droid (approved):

PR #760, opened 2026-08-19
fix: remove PII (email, OAuth codes, search payloads) from xyne-spaces backend logs

XyneSpaces:
## What
Application-log PII remediation for `apps/backend`, from the `/experiment` PII-in-logs audit. Each change removes cleartext PII from a log sink while preserving the operational signal (ids / counts / shape). No behavioural change beyond what is written to the log. 8 files, +69/-15.

## Root fix
- **F50 — `utils/redact.ts`**: `redactSensitiveUrl` now scrubs sensitive query-string **values** (`code`, `state`, `token` / `*_token`, `email`, `q` / `query`, `sig` / `signature`, `api_key`, ...) in addition to the existing webhook path secret. This protects every logged URL at once — `requestLogger` REQUEST_START / REQUEST_END, `errorHandler`, and `app.ts`. OAuth authorization codes and user email that arrive in query strings no longer reach the logs.

## Systemic context field
- **F2 — `utils/logger.ts` + `middleware/requestLogger.ts`**: the per-request `LogContext` no longer carries the user's email (`x-user-email`); it now carries `userId` (`x-user-id`), which `injectContext` stamps onto every log line. This removes email from effectively all request-scoped logs. `database/middleware/userSessionLogging.ts` updated for the rename — its session-audit log still resolves the actor email from the DB/result and is kept intentionally as an access-controlled audit trail.

## Call sites
| Finding | File | Change |
|---|---|---|
| F51 | `middleware/auth.ts` | Drop `email` from the two silent `refreshTokenBySession` log lines (keep `userId`). JWT claim untouched. |
| F56 / F14 | `services/websocketService.ts` | Typing start/stop logs drop `userName` + `userEmail` (keep `[ID]`); the two "Typing users" name lists reduced to a count. |
| F53 | `controllers/emailController.ts` | Drop `userEmail` from the two `email_sent` logs (keep `channelName` / `ticketId`). |
| F35 | `vespa/src/services/searchService.ts` | Reduce the two full-payload dumps (query text + bound permission filters) to a byte-length shape. |

## Deliberately NOT changed here (follow-ups)
- **Client IP** in `auth.ts` security logs (`clientIp: req.ip`) and request-start `ip` — kept for incident response / abuse investigation. IP-retention is a policy call, not a clear PII bug; flagging for the team to decide separately.
- **Interactive Google-login email lines** in `authV2Controller.ts` (~11 `logger.info(... ${...email})`) — lower-volume interactive OAuth path, and several have no `userId` available pre-provisioning. Worth a dedicated pass.
- **Uploaded filenames** in upload/document/collection controllers (can embed PII) — broad, separate PR.
- **Third-party `superposition-provider` `console.log`** (~28.6M email lines/wk in the audit) — lives in `node_modules`, needs a `patch-package` patch + build-tooling change.

## Validation
- Backend `tsc --noEmit` diagnostic pass reported **0 type errors**; the only cross-file ripple from the `emailId`→`userId` rename (`userSessionLogging.ts`) is fixed in this PR.
- All edits are anchored, mechanical, and preserve control flow.

## Relationship to PR #733
Companion to #733 (Claw / claw-auth PII logs, base `feature/deploy-xyneclaw`). This PR is the `apps/backend` half and targets `main`.

### DIFF
--- a/apps/backend/src/controllers/emailController.ts
+++ b/apps/backend/src/controllers/emailController.ts
@@ -369,7 +369,6 @@ export class EmailController {
         event: 'email_sent',
         type,
         channelName: channel.name,
-        userEmail: req.user?.email,
       });
 
       // 6. Save reply in database
@@ -1113,7 +1112,6 @@ export class EmailController {
           type: 'COMPOSE',
           channelName: channel.name,
           ticketId: ticket.id,
-          userEmail: req.user?.email,
         });
 
         return res.status(200).json({
--- a/apps/backend/src/database/middleware/userSessionLogging.ts
+++ b/apps/backend/src/database/middleware/userSessionLogging.ts
@@ -52,7 +52,7 @@ export function setupUserSessionLogging(prisma: PrismaClient, enabled: boolean =
     const operation = params.action.toUpperCase();
     const result = await next(params);
     const affectedRows = getAffectedRows(params.action, result);
-    const identifiers = await resolveLogIdentifiers(prisma, params.args?.data, params.args?.where, result, context?.emailId);
+    const identifiers = await resolveLogIdentifiers(prisma, params.args?.data, params.args?.where, result);
 
     if (params.action === 'delete' || params.action === 'deleteMany') {
       logDeleteOperation(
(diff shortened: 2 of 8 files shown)


devesh-juspay:
not needed

PR #764, opened 2026-08-19
feat: add ordered encryption key rotation

siddhantagrawal-ops:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`
- [x] `apps/xyne-claw` / `apps/xyne-claw-auth`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [x] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


apps/xyne-claw-auth/backend/src/config.ts
```diff
@@ -38,14 +39,21 @@ export const CONFIG = {
...
-    : Buffer.alloc(0),
+  // signing-secret backfill.
+  //
+  // A ring rather than one key, because Spaces stamps the key id into the
+  // ciphertext and can hold several at once while a rotation drains. Whatever
+  // ids Spaces has in ENCRYPTION_KEYS must appear here too, under the same
+  // names, or rows written under a missing id cannot be read. SPACES_ENCRYPTION_KEY
+  // is registered as "legacy" to match the id Spaces gives unversioned rows.
+  // Empty ring when unset; the backfill short-circuits with a clear error.
+  spacesEncryptionKeys: parseSpacesEncryptionKeyRing(
```
devesh-juspay: `CONFIG` is a module-level const, so `parseSpacesEncryptionKeyRing` runs at import time. That makes a malformed `SPACES_ENCRYPTION_KEY` fatal to **claw-auth startup** — OAuth, sessions and agent auth all fail to come up over a key that is only used by the temporary Spaces signing-secret backfill.

Before this PR the value was `Buffer.from(hex, "hex")` with no validation, so a bad value was accepted silently and surfaced only when the backfill ran.

The validation itself is a clear improvement — the previous code would happily build a short buffer and fail later at `createDecipheriv`. It is the placement I would reconsider: parsing lazily on first use, or tolerating a bad legacy key while still rejecting bad `SPACES_ENCRYPTION_KEYS` entries, keeps the strictness without tying a whole auth service's boot to a key that one route needs.

Unset is unaffected — that returns an empty ring and no throw. It is only set-but-malformed that becomes an outage.

devesh-juspay:
Three documentation items, all in `apps/backend/.env.example`, grouped because one edit covers them.

**1. `ENCRYPTION_KEYS` is not documented anywhere.** It is absent from `.env.example` and from the Joi schema in `src/config/env.ts`, so the only description of the ordered-array format lives in the header comment of `encryptionService.ts`.

That comment carries the rule that decides the rollout — *the final array entry is the active writer* — and whoever performs a rotation will be editing a deployment secret, not reading the service source. Worth an entry showing the shape, with the ordering rule stated:

```
# Ordered JSON array. The LAST entry is the active write key; every entry stays
# available for decryption. Absent or empty means writes stay on the legacy format.
# ENCRYPTION_KEYS=[{"id":"k1","key":"<64-hex>"}]
```

**2. Dangling documentation reference.** Line 369 reads `# See docs/ENCRYPTION_IMPLEMENTATION_PLAN.md for full setup guide.` — that file is not present in `docs/` on this branch. Since this PR is what changes encryption setup, it is the natural point to either add the guide or drop the pointer.

**3. Duplicate `ENCRYPTION_KEY` entry.** It appears twice under two separate `# Security` headings — active at line 106 (`ENCRYPTION_KEY=set-me`) and commented at line 308 (`# ENCRYPTION_KEY=`). Pre-existing and harmless at runtime, but this is the file an operator opens to find where to add `ENCRYPTION_KEYS`, and two entries for one variable invites editing the wrong one.

apps/xyne-claw-auth/backend/src/config.ts
siddhantagrawal-ops: Addressed. Spaces key-ring parsing is now deferred through a lazy loader and the `CONFIG.spacesEncryptionKeys` getter, so constructing/importing `CONFIG` no longer validates the optional Spaces keys during Claw-auth startup.

Strict validation is still preserved: a malformed `SPACES_ENCRYPTION_KEY` or `SPACES_ENCRYPTION_KEYS` value throws when the signing-secret functionality first requests the ring, without preventing unrelated OAuth, session, and agent-auth functionality from starting.

I also added focused tests verifying that malformed configuration is not evaluated during loader creation, validation occurs on first use, valid legacy keys load correctly, and successful parsing is cached. The existing Spaces key-ring and CBC decryption tests, Claw-auth typecheck, and backend typecheck pass.

siddhantagrawal-ops: Addressed all three documentation/configuration items:

1. Added `ENCRYPTION_KEY` and `ENCRYPTION_KEYS` explicitly to the backend Joi environment schema.
2. Documented the ordered-array format in `apps/backend/.env.example`, including that the final entry is the active write key, all entries remain available for decryption, and an unset or empty array keeps writes in legacy format.
3. Removed the dangling `docs/ENCRYPTION_IMPLEMENTATION_PLAN.md` reference and consolidated the duplicate `ENCRYPTION_KEY` entries into one canonical entry.

The example contains placeholders only; no key material was committed. Backend typecheck and the targeted encryption tests pass.

devesh-juspay (dismissed):

apps/xyne-claw-auth/backend/src/lib/spaces-app-secret.ts
```diff
@@ -59,7 +59,7 @@ export async function backfillSigningSecretFromSpacesDb(args: {
   spacesAppId: string;
 }): Promise<boolean> {
   const { agentId, spacesAppId } = args;
-  if (CONFIG.spacesEncryptionKey.length === 0) {
+  if (CONFIG.spacesEncryptionKeys.size === 0) {
```
satvikbatra: it calls the lazy parser, which throws on a malformed SPACES_ENCRYPTION_KEYS (bad JSON, dup id, bad hex, etc). so this throws instead of returning false like the doc comment says. The backfill loop doesn't catch it either, so one bad env var 500s the whole run instead of failing per agent.

apps/xyne-claw-auth/backend/src/routes/admin-backfill-signing-secrets.ts
```diff
@@ -107,15 +107,15 @@ router.get("/diagnose-signing-secret/:slug", async (req: Request<{ slug: string
   let db: { ok: boolean; fp?: string; len?: number; error?: string };
   if (!agent.spacesAppId) {
     db = { ok: false, error: "agent has no spacesAppId" };
-  } else if (CONFIG.spacesEncryptionKey.length === 0) {
+  } else if (CONFIG.spacesEncryptionKeys.size === 0) {
```
satvikbatra: same here

apps/backend/src/services/encryptionService.ts
```diff
@@ -1,74 +1,220 @@
...
 
 /**
- * Get encryption key from environment
- * Must be 32 bytes (64 hex characters) for AES-256
+ * Parse a 32-byte (64 hex char) key. Rejects anything that is not AES-256 sized.
  */
-function getEncryptionKey(): Buffer {
-  const key = process.env.ENCRYPTION_KEY;
+function parseHexKey(raw: unknown, label: string): Buffer {
+  if (typeof raw !== 'string' || !/^[0-9a-fA-F]{64}$/.test(raw)) {
```
satvikbatra: If the deployed value has any trailing whitespace, every integration credential read breaks. Can we .trim() before the regex?

apps/xyne-claw-auth/backend/src/spaces-encryption-key-ring.ts
```diff
@@ -0,0 +1,103 @@
+const VERSION_TAG = "v2";
+const LEGACY_KEY_ID = "legacy";
+
+interface OrderedEncryptionKey {
+  id: string;
+  key: string;
+}
+
+function parseHexKey(raw: unknown, label: string): Buffer {
+  if (typeof raw !== "string" || !/^[0-9a-fA-F]{64}$/.test(raw)) {
```
satvikbatra: same trim here

siddhantagrawal-ops: Addressed. 
Updated the Spaces backend and Claw-auth parseHexKey() implementations to trim leading and trailing whitespace before validation and decoding. Keys must still contain exactly 64 hexadecimal characters after normalization; internal whitespace, invalid hex, and incorrect lengths remain rejected.

Protected the lazy CONFIG.spacesEncryptionKeys access in backfillSigningSecretFromSpacesDb(). Malformed SPACES_ENCRYPTION_KEY or SPACES_ENCRYPTION_KEYS configuration is now logged safely and returns false for the affected operation instead of escaping and interrupting the entire backfill run.

Protected the same lazy access in the admin signing-secret diagnostic route. Configuration parsing, database lookup, and decryption errors are now converted into the existing structured db.error result instead of producing an unhandled HTTP 500.

Strict validation remains in place for malformed JSON, duplicate key IDs, invalid IDs, invalid hex, and incorrectly sized keys. No key material, ciphertext, or plaintext is included in the new logs or error responses.

Verified the changes locally with the existing backend and Claw-auth suites, whitespace smoke checks for legacy and ordered keys, malformed lazy-configuration checks, both package typechecks, and the backend build.

Removed the test-file changes from the final PR diff : encryptionService.test.ts, crypto.test.ts, spaces-encryption-key-ring.test.ts, and spaces-encryption-key-ring.lazy.test.ts.

satvikbatra (approved):

devesh-juspay (approved):

PR #819, opened 2026-08-20
Feature/add existing subticket picker release

rahul1841:


apps/backend/src/zero/mutators.ts
```diff
@@ -6816,6 +6819,258 @@ export function createMutators(
...
+                nextAncestorId = parentMapping.ticketId;
+                break;
+              }
+            }
+            ancestorCursor = nextAncestorId;
+          }
+
+          // One parent per linked ticket, so the `.one()`/`findFirst` lookups on
+          // mappedTicketId elsewhere stay unambiguous.
+          const existingSubTickets = await tx.run(
```
PiyushJuspay, line 6912: what happens if two concurrent call happens with two different parent ticketIds, will it create two parent for single ticket?

PR #849, opened 2026-08-21
feat: add recording participants

harshpreet931:
POT - https://drive.google.com/file/d/11xKOczEc36TYrV9nIbpjMlizv_ljnpeP/view?usp=sharing
updated positioning - 
Screenshot 2026-08-21 at 12 47 35
default state - 
Screenshot 2026-08-21 at 12 47 51

## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Recordings were only ever attributed to the person who pressed record. If you recorded a call *about* someone else — a 1:1, an interview, a customer call someone else led — the list pill showed your name, and the People filter could only ever find it under you. There was no way to say who a recording is actually about.

This adds an explicit, owner-curated participant set on headless recordings, and threads it through the list, the filter, and the detail header.

**Data model** — new `Call.recordingParticipants` (`TEXT[]`, default `[]`) with its migration, mirrored into both Zero schemas (`packages/shared/src/zero/schema.ts` and the generated backend copy) so the dashboard reads it straight off the synced row with no extra fetch. Note-taker calls seed the array with their creator, so "the owner is a participant" holds from row one; older rows fall back to the creator client-side via `getRecordingParticipantIds`.

**Backend** — one new endpoint, `POST /api/calls/recordings/:callId/participants`, taking an `{ action: 'add' | 'remove', userId }` command. It resolves the call by external id and 404s for anything that isn't a headless recording in the caller's workspace, 403s for anyone who isn't the recording owner, and — on `add` — verifies the target is an active member of the same workspace before touching the row. Writes are `array_append` / `array_remove` in raw SQL, with the append guarded by `NOT (... = ANY(...))`: two clients adding the same person concurrently can't produce a duplicate, and there's no read-modify-write window where a concurrent edit gets lost.

**Recording detail** — a `RecordingParticipants` popover in the header. Search over active workspace users by name or email, keyboard-navigable (arrows / enter, `role="combobox"` + `aria-activedescendant`), owner pinned and non-removable. Mutations are optimistic: a pending add/remove is layered over the synced Zero row, reconciled when the server value catches up, and rolled back with a toast on failure.

Being a participant is *not* the same as having access, so the popover also reports how many participants can actually open the recording and offers a one-click share for those who can't. That count is deliberately hidden when the recording is shared with a group or channel — membership isn't resolvable client-side there, and a wrong number is worse than no number. Non-owners get the read-only view: the list renders, the search box and add/remove controls don't.

**Recordings list** — pills show a participant line ("Alice & you", "Alice, Bob & 2 others") instead of the bare creator name, falling back to the old behaviour when nothing resolves. The People filter is populated from participants rather than creators and matches a recording if *any* participant is selected, so filtering by a colleague surfaces calls they were in but didn't start. The selected user is kept in the option list even when the current result set doesn't contain them, so the filter can always be cleared. Picking someone no longer force-switches the Created / Shared tab — the two dimensions are independent now. The filter trigger is also width-capped and truncates long names so it stops stretching the toolbar.

## Areas Touched

- [x] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

- [x] New tables exported in `createSchema()` and given a Query ACL registered in `query-acl-factory.ts`

The mutator change is a single added field (`recordingParticipants: []`) on the existing `calls.insert` inside the channel-call mutator. There is no corresponding `calls.insert` in the client mutators, so nothing to mirror; no new table, so no new ACL. Participant writes intentionally go through the REST endpoint rather than a mutator — they need an owner check and a workspace-membership check that don't belong in an optimistic client mutation.

## Schema & Data Changes

- [x] Migration added

If any of the above:

- [x] Clients regenerated (`db:generate`, `db:common:generate`)
- [x] Backfill is idempotent

**Rollout / rollback notes:** additive column with a default, so old code ignores it and new code tolerates its absence. No backfill needed — existing rows get `[]` and the UI falls back to the creator, which is exactly the pre-change behaviour. Rollback is a column drop; nothing else reads it.

## Config, Environment & URLs

## API Contract

- [ ] No API changes
- [x] New endpoint

`POST /api/calls/recordings/:callId/participants` → `{ action: 'add' | 'remove', userId }`, returns `{ success: true }`. `401` unauthenticated, `403` non-owner, `404` unknown / non-headless / out-of-workspace recording, `400` invalid body or a target who isn't an active member of the workspace. Purely additive — no existing route or response shape changed.

---

## How did you test it?

Manual verification is still pending — flagging that honestly rather than ticking boxes. Reviewers should focus on:

- Owner vs. non-owner in the detail popover (search + add/remove must be owner-only; the 403 path).
- Two users at once: A adds B, B's list should re-render off the synced Zero row without a flash.
- Filtering by a colleague in the People filter — a recording they were added to but didn't create should appear, and the filter should still clear after the results change.
- Recording shared with a group/channel: the "N of M have access" count should disappear rather than show a wrong number.
- Legacy rows with an empty `recordingParticipants` — pill and popover should fall back to the creator.

### Automated

- [x] Not covered by tests 

---

## Checklist

### DIFF
--- a/apps/backend/src/controllers/callController.ts
+++ b/apps/backend/src/controllers/callController.ts
@@ -51,6 +51,11 @@ import { canvasAuthService } from '@/services/canvasAuthService';
 import { buildCallInviteUrl } from '@/utils/urlUtils';
 import { readRecordingGoogleDocLinks } from '@/utils/recordingGoogleDocs';
 
+const RecordingParticipantsCommandSchema = z.object({
+  action: z.enum(['add', 'remove']),
+  userId: z.string().min(1).max(64),
+});
+
 const UpdateHeadlessRecordingSchema = z
   .object({
     title: z.string().trim().min(1).max(255).optional(),
@@ -1457,6 +1462,55 @@ export class CallController {
     } catch (error) {
       logger.error('Failed to fetch recording detail:', error);
       res.status(500).json({ success: false, error: 'Failed to fetch recording' });
+    }
+  };
+
+  manageRecordingParticipants = async (req: Request, res: Response): Promise<void> => {
+    const userId = req.user?.id;
+    const { callId } = req.params;
+
+    if (!userId) {
+      res.status(401).json({ success: false, error: 'Unauthorized' });
+      return;
+    }
+
+    try {
+      const input = RecordingParticipantsCommandSchema.parse(req.body);
+      const call = await repositories.calls.findByExternalId(callId);
+
+      if (
+        !call ||
+        call.callType !== CallType.HEADLESS ||
+        (call.workspaceId !== null && call.workspaceId !== req.user!.workspaceId)
+      ) {
+        res.status(404).json({ success: false, error: 'Recording not found' });
+        return;
+      }
+
+      if (call.createdByUserId !== userId) {
+        res.status(403).json({ success: false, error: 'Access denied' });
+        return;
+      }
+
+      if (input.action === 'add') {
+        const participant = await repositories.users.findById(input.userId);
+        if (
+          !participant ||
+          participant.workspaceId !== req.user!.workspaceId ||
+          participant.leftAt !== null
+        ) {
+          res.status(400).json({ success: false, error: 'User not found in this workspace' });
+          return;
+        }
+        await repositories.calls.addRecordingParticipant(callId, input.userId);
+      } else {
+        await repositories.calls.removeRecordingParticipant(callId, input.userId);
+      }
+
+      res.json({ success: true });
+    } catch (error) {
+      logger.error('Failed to manage recording participants', { error, callId });
+      res.status(400).json({ success: false, error: 'Failed to update participants' });
     }
   };
 
--- a/apps/dashboard/src/routes/RecordingDetailV2Screen/components/RecordingParticipants.tsx
+++ b/apps/dashboard/src/routes/RecordingDetailV2Screen/components/RecordingParticipants.tsx
@@ -1,0 +1,379 @@
+import { useEffect, useMemo, useRef, useState, type ReactElement } from 'react';
+import { AnimatePresence, motion, useReducedMotion } from 'framer-motion';
+import { Search, Share2, UserPlus, Users, X } from 'lucide-react';
+import { toast } from 'sonner';
+import { queries } from '../../../zero/queries';
+import { useCachedQuery } from '../../../hooks/useCachedQuery';
+import { useActiveUsers, useUsersById, useSelf, searchUsers } from '../../../hooks/useUsers';
+import { recordingService } from '../../../services/Recording/recordingService';
+import { getUserDisplayName } from '../../../utils/userDisplayName';
+import { getRecordingParticipantIds, logRecordingError } from '../../../utils/recordingUtils';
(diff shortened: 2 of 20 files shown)


samitbarai (dismissed):

apps/dashboard/src/routes/RecordingDetailV2Screen/components/RecordingParticipants.tsx
```diff
@@ -0,0 +1,367 @@
...
+  const [highlighted, setHighlighted] = useState(0);
+  const [busyIds, setBusyIds] = useState<ReadonlySet<string>>(new Set());
+  const [pending, setPending] = useState<ReadonlyMap<string, 'add' | 'remove'>>(new Map());
+  const searchRef = useRef<HTMLInputElement>(null);
+  const reduceMotion = useReducedMotion();
+
+  const users = useActiveUsers();
+  const usersById = useUsersById();
+  const self = useSelf();
+  const [recordingRow] = useCachedQuery(
```
Mayankbansal-juspay: do we need this, can't we pass this data as a prop, otherwise this component will keep running it's own query

apps/dashboard/src/routes/RecordingDetailV2Screen/components/RecordingParticipants.tsx
```diff
@@ -0,0 +1,367 @@
...
+    queries.oatsRecordingByExternalId({ callId: recordingExternalId }),
+  );
+
+  const canManage = Boolean(self?.id && createdByUserId && self.id === createdByUserId);
+
```
Mayankbansal-juspay: inn sab helper functions ko kisi hook ya utils mein daal do, cleaner lagega

apps/backend/src/controllers/callController.ts
```diff
@@ -1420,6 +1422,55 @@ export class CallController {
     }
   };
```
Mayankbansal-juspay, line 1467: i Was thinking ki if we can make a generic mutator for this like, call.update, in that basically we won't have keep making new api's as per requirements, it will generic and will do the work quickly as well. can u check this once

apps/dashboard/src/routes/RecordingDetailV2Screen/components/RecordingParticipants.tsx
harshpreet931: move these to props

apps/dashboard/src/routes/RecordingDetailV2Screen/components/RecordingParticipants.tsx
harshpreet931: moved to useRecordingParticipants hook, the hook now has the participant list, optimistic add/remove state, access lookup, and search

apps/backend/src/controllers/callController.ts
harshpreet931: i think this makes sense but for title and metadata generally.. we will need this for this usecase.

Mayankbansal-juspay (approved):

samitbarai (approved):

PR #854, opened 2026-08-21
feat: XYNE-56580 canvas suggestion mode - agent edits require human approval (non-claw)

adarshr001-stack:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
POT : https://drive.google.com/drive/u/0/folders/1jGw_JA_VGi_JUWFfm2AHLKq87v57_6rW

## Areas Touched

- [x] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [x] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


apps/backend/src/services/canvas/suggestions.ts
```diff
@@ -0,0 +1,213 @@
...
+        if (!ids.length) return;
+        await prisma.canvasSuggestionChange.updateMany({ where: { id: { in: ids } }, data: { status } });
+      };
+
+      if (!outcome.applied.length) {
+        await markStatuses(outcome.stale, 'STALE');
+        return { applied: 0, stale: outcome.stale.length };
+      }
+
+      const ok = await syncToYSweet(canvasId, outcome.blocks, userId);
```
ameer030, line 218: 🔴 **B01 — accepted suggestions never reach `canvas.content`; SDLC reads go stale forever**

Accepting a suggestion now writes only Y-Sweet — the Postgres `canvas.content` is never updated. The SDLC artifact read (`/current/read`) and its `contentHash` read `canvas.content`, so after the first accepted batch the claw re-reads the pre-accept document and a stale hash, permanently.

- `applySuggestionChanges` writes the document via `syncToYSweet` and then writes only row **statuses** — no `prisma.canvas.update`: `apps/backend/src/services/canvas/suggestions.ts:162-169`.
- The new SDLC update path parks suggestions and also no longer writes `content`/`lastEditedBy`/`lastEditedAt`: `apps/backend/src/sdlc/SdlcHubService.ts:1468-1529` (the removed `$transaction` did write them).
- `readArtifact` selects `canvas.content` and hashes its rendered markdown: `apps/backend/src/sdlc/SdlcArtifactVersionStore.ts:196-209`; the claw advertises this as live content + hash (`xyne-spaces-tools.ts:5739-5752`).
- Checked for another writer: dashboard `performSave` is the non-collaborative editor only; the exit snapshot skips collaborative documents (`apps/dashboard/src/components/Canvas/canvasVersioning.ts:275-278`); version-restore also skips collaborative; no Y-Sweet→DB webhook exists (`routes/ysweet.ts` is auth only). WIKI/BASELINE still refresh via their own transactions — PRD/Tech Doc (`DEFAULT`) artifacts stale on first accepted suggestion.

Checklist rule: BUG-C32 (stored denormalized columns must be re-derived at every write path).
</details>

In `applySuggestionChanges`, after a successful `syncToYSweet`, also persist the result:

```ts
await prisma.canvas.update({
  where: { id: canvasId },
  data: { content: outcome.blocks as unknown as Prisma.InputJsonValue, lastEditedBy: userId, lastEditedAt: new Date() },
});
```

(or make `readArtifact` read Y-Sweet-first with the stored column as fallback). While here, re-wrap the title update + `sdlcArtifact.upsert` at `SdlcHubService.ts:1501-1519` in one `$transaction` — they are two halves of one provenance write.
</details>

rules: BUG-C32, BUG-C22
agent: omp
run: n/a
</details>

packages/shared/src/canvas/suggestionApply.ts
```diff
@@ -0,0 +1,214 @@
...
+      continue;
+    }
+    const md = markdownOf(row);
+    const parsed = md ? await toBlocks(md) : [];
+    if (!parsed[0]) {
+      stale.push(row.id);
+      continue;
+    }
+    // Replaced block keeps its id; extra blocks follow it under the row's id.
+    const [first, ...rest] = withIds(parsed, row.id, row.orderIndex);
```
ameer030, line 120: 🟡 **W02 — re-applying a replace/move duplicates follower blocks with duplicate ids**

Only inserts guard against double-apply. If a replace or a move-with-content is applied twice (a retry after the document was written but the status commit failed), its `rowId__n` follower blocks are spliced in again — with identical ids, which corrupts the Yjs/BlockNote document.

- Replace phase splices followers unconditionally: `packages/shared/src/canvas/suggestionApply.ts:98-117`; move-with-content the same at `:178-190`; the only re-apply guard is `findIdx(row.id)` for inserts at `:191`.
- The retry window is real: `syncToYSweet` writes the document *before* statuses are written (`apps/backend/src/services/canvas/suggestions.ts:162-169`); the no-outcome server path catches and only logs, leaving rows `PENDING` with the doc already applied (`apps/backend/src/zero/mutators.ts` resolveChange asyncTask). The client-apply outcome CAS gates status flips, never the document write.
- Verified by replaying `applyOps` twice on the same rows: `["blk1","rowR1__1","rowR1__1","blk2"]` after a double replace; insert twice stays a no-op.

Checklist rule: BUG-C04 (retried operations safe to run twice).
</details>

Before splicing followers, drop ones whose deterministic id already exists:

```ts
const rest = rawRest.filter(b => !b.id || working.findIndex(x => x.id === b.id) < 0);
```

One guard in `withIds` (or at both splice sites) covers replace, move, and insert.
</details>

rules: BUG-C04
agent: omp
run: n/a
</details>

apps/backend/src/services/canvas/suggestions.ts
```diff
@@ -0,0 +1,213 @@
...
+  return JSON.stringify(normalize(value));
+}
+
+const LOCK_TTL_SECONDS = 30;
+
+/** Serialises concurrent accepts on one canvas; the loser gets onBusy(). */
+async function withCanvasLock<T>(canvasId: string, fn: () => Promise<T>, onBusy: () => T): Promise<T> {
+  const key = `canvas-apply-lock:${canvasId}`;
+  const acquired = await redisService
+    .set(key, '1', LOCK_TTL_SECONDS, true)
```
ameer030, line 46: 🟡 **W03 — 30s accept-lock can expire mid-apply (and silently vanishes when Redis blips)**

The per-canvas apply lock is a fixed `SET NX EX 30`. A big canvas with an accept-all batch — per-block renders, two Y-Sweet round-trips, then sequential status updates — can run past 30s, letting a concurrent accept slip in and double-apply. On any Redis error the `.catch(() => true)` also turns the mutex off without a log.

- `LOCK_TTL_SECONDS = 30`, acquire at `apps/backend/src/services/canvas/suggestions.ts:44` (NX+EX semantics confirmed at `redisService.ts:872-884`), no renewal.
- Under the lock: sequential `blocksToMarkdownLossy` awaits per block (`blockRender.ts:27-34`), `readFromYSweetOrNull`, `syncToYSweet`, then one `updateMany` per placed row + per deletion event (`suggestions.ts:178-201`).
- Fail-open: `.catch(() => true)` also swallows "Redis not initialized", so an outage quietly removes the only guard against concurrent accepts (which is exactly what W02's duplication makes dangerous).

Checklist rule: BUG-C04 (distributed lock TTL must outlive the job).
</details>

Raise the TTL to cover the worst-case batch (e.g. 120s) or renew it on a short interval for the duration of the apply; log at least a warning on the fail-open branch so a Redis outage is visible.
</details>

rules: BUG-C04
agent: omp
run: n/a
</details>

apps/backend/src/services/canvas/suggestions.ts
```diff
@@ -0,0 +1,213 @@
...
+    ...(op.beforeContent !== undefined ? { beforeContent: toJsonSafe(op.beforeContent) as never } : {}),
+    ...(op.afterMarkdown !== undefined ? { afterContent: { markdown: op.afterMarkdown } as never } : {}),
+    status: 'PENDING',
+  }));
+
+  const targetIds = ops.map(op => op.blockId).filter((id): id is string => Boolean(id));
+  await prisma.$transaction(async tx => {
+    if (targetIds.length) {
+      // A newer proposal supersedes older pending rows for the same blocks.
+      await tx.canvasSuggestionChange.updateMany({
```
ameer030, line 118: 🟡 **W04 — insert rows are never superseded, so re-proposals stack and accept-all double-inserts**

Supersede matches only rows whose `blockId` is in the new ops' block ids; inserts carry `blockId: null`, so when the agent proposes the same paragraph again in a later batch, both generations stay PENDING. Accept-all then applies both — each batch mints fresh row ids, so the insert idempotency guard can't catch it and the paragraph lands twice.

- Supersede filter: `apps/backend/src/services/canvas/suggestions.ts:92-98` (`blockId: { in: targetIds }` — inserts never match); row ids are `randomUUID()` per batch: `suggestions.ts:72-89`.
- The panel shows every batch that still has a PENDING row, with no content-based dedup (`CanvasSuggestionsPanel.tsx` `visibleRowsOf`), so both cards stay visible and clickable.

Checklist rule: BUG-C01 (logical-singleton/duplicate creation needs a stable key guard).
</details>

Give inserts a stable natural key — e.g. `(canvasId, resolved anchor, hash of normalized afterContent markdown)` — and supersede earlier PENDING inserts on it at batch-creation time, the same way replace/delete rows supersede by target block.
</details>

rules: BUG-C01
agent: omp
run: n/a
</details>

packages/shared/src/zero/queries.ts
```diff
@@ -3165,6 +3165,36 @@ export const queries = defineQueries({
...
+        );
+    },
+  ),
+
+  canvasSuggestionChanges: defineQuery(
+    z.object({ canvasId: z.string() }),
+    ({ ctx, args: { canvasId } }) => {
+      return zql.canvas_suggestion_changes
+        .where('canvasId', canvasId)
+        .where('status', 'IN', ['PENDING', 'STALE'])
```
ameer030, line 3292: 🟡 **W06 — `beforeContent`/`afterContent` JSONB is Zero-synced to every canvas viewer**

Zero syncs whole rows (no column projection), so every client that can see the canvas receives the full `beforeContent` and `afterContent` payloads of every suggestion row. `beforeContent` in particular is never read by any client — it's write-only — yet it fans out to all subscribers.

- New synced table with JSONB payload columns: `apps/backend/prisma/schema.prisma` `CanvasSuggestionChange`; Zero mirror without projection: `packages/shared/src/zero/schema.ts`.
- `beforeContent` consumers: only the writer (`apps/backend/src/services/canvas/suggestions.ts:85`); `applyOps`' `SuggestionRowLike` has no `beforeContent` field ("no conflict detection"); the dashboard references it only in a type declaration (`CanvasSuggestionsPanel.tsx:36`).
- The two new queries return all columns: `packages/shared/src/zero/queries.ts:3171-3195`.

Checklist rules: SEC-C14 (sync surfaces expose minimal fields), ARCH-C11 (unbounded JSON columns off hot synced tables unless justified).
</details>

Keep `beforeContent` (and possibly `afterContent`) out of the synced table — e.g. store snapshots in a non-synced sibling table or column and have the server query it only where needed — or document the measured payload justification for co-locating them on a Zero-synced table.
</details>

rules: SEC-C14, ARCH-C11
agent: omp
run: n/a
</details>

apps/dashboard/src/components/Canvas/CanvasScreen/CanvasScreen.tsx
```diff
@@ -954,13 +969,96 @@ const CanvasScreen: React.FC<CanvasScreenProps> = ({
...
+      prevBlockIdsRef.current = nowIds;
+      if (!prevIds || !hasPendingSuggestionsRef.current) return;
+
+      const events = computeDeletionEvents(prevIds, nowIds);
+      if (!events.length) return;
+      deletionEventsRef.current.push(...events);
+      if (deletionFlushRef.current !== null) window.clearTimeout(deletionFlushRef.current);
+      deletionFlushRef.current = window.setTimeout(() => {
+        deletionFlushRef.current = null;
+        const batch = deletionEventsRef.current.splice(0, 100);
```
ameer030: 🔵 **N01 — beyond 100 deletions in one flush, the rest are stranded**

The debounce flush ships `splice(0, 100)` once and clears the only timer; anything beyond 100 deletions in a 500ms window stays buffered with no timer armed — it only goes out when the user deletes something else later.

`apps/dashboard/src/components/Canvas/CanvasScreen/CanvasScreen.tsx:1023-1033`: after the splice, `deletionFlushRef` is null and `deletionEventsRef` may still hold events; nothing re-arms. Affected anchors then forward late (or never, on canvas switch — the documented degradation).

Checklist rule: BUG-C29 (loops/batching must not lose the tail).
</details>

After the splice, if `deletionEventsRef.current.length > 0`, re-arm the timeout (drain in 100-chunks); also clear + null the handle in the canvas-switch cleanup.
</details>

rules: BUG-C29
agent: omp
run: n/a
</details>

apps/backend/src/services/canvas/blockLabels.ts
```diff
@@ -0,0 +1,249 @@
...
+  blockId?: string;
+  anchor?: string | null;
+  beforeContent?: BlockNoteBlock;
+  afterMarkdown?: string;
+  orderIndex: number;
+}
+
+/** Longest common subsequence of two id sequences — the blocks that did NOT move. */
+export function lcsStationary(a: string[], b: string[]): Set<string> {
+  const dp: number[][] = Array.from({ length: a.length + 1 }, () => new Array(b.length + 1).fill(0));
```
ameer030: 🔵 **N02 — unbounded O(n·m) LCS matrices on the agent write path**

Both LCS implementations allocate `(n+1)×(m+1)` arrays per write request. The edit tool accepts up to ~5MB of markdown, so an adversarial or just very large document pair can pin hundreds of MB of heap on one request.

`lcsStationary` here (`apps/backend/src/services/canvas/blockLabels.ts:147-148`) and `lcsPairs` (`apps/backend/src/services/canvas/blockDiff.ts:14-21`) build full DP matrices with no size guard; inputs are the live canvas block list and the agent's whole-document reply.

Checklist rule: perf guardrail class (PERF-C01-family) — unbounded work per request.
</details>

Cheap guard before the LCS: if `a.length * b.length` exceeds a cap (e.g. 2M), skip the LCS — treat nothing as stationary (ops degrade to anchored inserts/moves) or fall back to an index-map intersection. Same cap in `deriveDiffOps`.
</details>

rules: PERF-C01
agent: omp
run: n/a
</details>

apps/backend/src/services/canvas/suggestions.ts
```diff
@@ -0,0 +1,213 @@
...
+      await markStatuses(outcome.stale, 'STALE');
+
+      // An accepted placement row hands its still-pending same-anchor
+      // followers (same batch, higher orderIndex) to the block it placed.
+      // Descending orderIndex so simultaneous accepts chain: d→c, c→b, b→a.
+      const appliedSet = new Set(outcome.applied);
+      const placedRows = rows
+        .filter(r => appliedSet.has(r.id) && (r.op === 'insert' || r.op === 'move'))
+        .sort((a, b) => b.orderIndex - a.orderIndex);
+      for (const row of placedRows) {
```
ameer030, line 234: 🔵 **N03 — sequential per-row DB writes inside the accept lock**

After the document write, the accept path issues one `updateMany` per placed row and one per deletion event — up to ~2×batch-size sequential round-trips, all inside the 30s Redis lock window.

`apps/backend/src/services/canvas/suggestions.ts:178-201`: `for (const row of placedRows) await updateMany(...)` and `for (const event of events) await updateMany(...)`. Compounds W03's lock-TTL risk on accept-all.

Checklist rule: PERF-C12 (no per-item query fan-out).
</details>

Collapse to one `updateMany` per distinct `(batchId, currentAnchorId)` anchor pair (or a single CASE-based statement), preserving the descending-`orderIndex` chaining semantics.
</details>

rules: PERF-C12
agent: omp
run: n/a
</details>

packages/shared/src/zero/queries.ts
```diff
@@ -3165,6 +3165,36 @@ export const queries = defineQueries({
...
+  /** Unresolved suggestion changes for a canvas; the panel groups them by batchId and shows batches with a pending row. */
+  /** Placement rows (insert/move) of a canvas, ACCEPTED included — the client-side
+   *  siblingOrder source for ordered placement. Never rendered as cards. */
+  canvasSuggestionPlacementOrder: defineQuery(
+    z.object({ canvasId: z.string() }),
+    ({ ctx, args: { canvasId } }) => {
+      return zql.canvas_suggestion_changes
+        .where('canvasId', canvasId)
+        .where('op', 'IN', ['insert', 'move'])
+        .where('status', 'IN', ['PENDING', 'ACCEPTED'])
```
ameer030, line 3279: 🔵 **N04 — client and server compute sibling order from different row sets**

The server seeds `suggestionSiblingOrder` from **all** insert/move rows of the involved batches regardless of status; the dashboard hook uses `canvasSuggestionPlacementOrder`, which is filtered to `PENDING` + `ACCEPTED`. After rejects/supersedes, a browser-applied accept and a server-applied accept can place the same sibling at different positions.

Server: `apps/backend/src/services/canvas/suggestions.ts:149-154` (no status filter). Client: `packages/shared/src/zero/queries.ts:3171-3179` (`.where('status', 'IN', ['PENDING','ACCEPTED'])`). Same shared algorithm — divergent inputs.

Checklist rule: BUG-C03 (one source of truth for shared derivations).
</details>

Align the filters (exclude `REJECTED`/`STALE`/`SUPERSEDED` placement rows server-side too) or share one derivation helper so both paths see the same sibling set.
</details>

rules: BUG-C03
agent: omp
run: n/a
</details>

packages/shared/src/zero/mutators.ts
```diff
@@ -425,6 +425,54 @@ async function assertCanvasCommentEditAccess(
...
+async function applySuggestionOutcome(
+  tx: Transaction<Schema>,
+  outcome: { applied: string[]; stale: string[] },
+  timestamp: number,
+): Promise<void> {
+  const appliedRows = outcome.applied.length
+    ? await tx.run(zql.canvas_suggestion_changes.where('id', 'IN', outcome.applied))
+    : [];
+  for (const id of outcome.applied) {
+    await tx.mutate.canvas_suggestion_changes.update({ id, status: 'ACCEPTED', updatedAt: timestamp });
```
ameer030, line 438: 🔵 **N05 — optimistic status flip is unconditional while the server compare-and-sets**

The client-mirrored mutator writes `ACCEPTED`/`STALE` for every id in the outcome with no status predicate; the server filters to `status = 'PENDING'` first. If another writer rejected or superseded a row in between, the client briefly paints it accepted, then silently reverts on rebase.

Optimistic: `packages/shared/src/zero/mutators.ts:434-439` (unconditional updates). Server: `apps/backend/src/zero/mutators.ts:689-696` (CAS on `PENDING`). End state is correct; the flash is cosmetic.

Checklist rule: ARCH-C10 (mirrored mutators behave the same).
</details>

Add the same `PENDING` predicate to the optimistic updates (read status before flipping), or note the deliberate divergence in a comment.
</details>

rules: ARCH-C10
agent: omp
run: n/a
</details>

apps/backend/src/services/canvas/blockLabels.ts
```diff
@@ -0,0 +1,249 @@
...
+        return {
+          handle: handleMatch[1] as string,
+          isNew: false,
+          markdown: chunk.replace(HANDLE_PATTERN, '').trim(),
+        };
+      }
+      if (NEW_PATTERN.test(chunk)) {
+        return { handle: null, isNew: true, markdown: chunk.replace(NEW_PATTERN, '').trim() };
+      }
+      // Bare: label dropped — recoverable later by similarity matching.
```
ameer030: 🔵 **N06 — comments that no longer match behavior (3 spots)**

Three new comments state behavior the code doesn't have: the "similarity matching" fallback for label-less paragraphs doesn't exist, the `CONFLICT` status is never written by any path, and the docstring describing the panel query sits above the placement-order query in both mirrored query files.

- (a) `apps/backend/src/services/canvas/blockLabels.ts:123` — bare entries go straight to `insert`; no similarity matcher.
- (b) `apps/backend/prisma/schema.prisma` CanvasSuggestionChange `status` comment lists `CONFLICT` — nothing writes it.
- (c) `apps/backend/src/zero/queries.ts:3633-3635` and `packages/shared/src/zero/queries.ts:3167-3169` — the "Unresolved suggestion changes … panel groups them by batchId" docstring is attached to `canvasSuggestionPlacementOrder`, not `canvasSuggestionChanges`.

Checklist rule: DOCS-C04 (comments carry the invariant spec and must match behavior).
</details>

Fix or move all three comments.
</details>

rules: DOCS-C04
agent: omp
run: n/a
</details>

apps/dashboard/src/components/Canvas/CanvasSuggestionsPanel/CanvasSuggestionsPanel.tsx
```diff
@@ -0,0 +1,305 @@
...
+  onToggleReview,
+}: Props): ReactElement | null => {
+  const z = useZero();
+  const [busy, setBusy] = useState<string | null>(null);
+  const [rows = []] = useCachedQuery(queries.canvasSuggestionChanges({ canvasId }), {
+    enabled: Boolean(canvasId),
+  });
+
+  const { visible, batches } = useMemo(
+    () => visibleRowsOf(rows as unknown as SuggestionRow[]),
```
ameer030, line 86: 🔵 **N07 — shape-hiding casts at the new suggestion surfaces**

`z.mutate(mutation as never)` (forcing a second `result as { server: … }` cast) and several `as unknown as SuggestionRow[] / InlineSuggestionRow[] / AcceptableRow[]` casts hide the Zero row types from the compiler, so field drift (e.g. `createdAt` Date|number) won't be caught by typecheck.

`CanvasSuggestionsPanel.tsx:85` and `:96-99`; `CanvasScreen.tsx:984-985`; `useCanvasSuggestionAccept.ts:74`. The pre-existing BlockNote `as unknown as` pattern is out of scope — these five are new.

Checklist rule: STYLE-C11 (no `as` escape hatches that defeat narrowing).
</details>

Derive the row types from the query results (e.g. `ZeroQueryResult<typeof queries.canvasSuggestionChanges>` or an inferred zod shape) and type `run()` with the mutator-request parameter type instead of `unknown`/`never`.
</details>

rules: STYLE-C11
agent: omp
run: n/a
</details>

apps/backend/src/services/canvas/suggestions.ts
```diff
@@ -0,0 +1,213 @@
...
+    status: 'PENDING',
+  }));
+
+  const targetIds = ops.map(op => op.blockId).filter((id): id is string => Boolean(id));
+  await prisma.$transaction(async tx => {
+    if (targetIds.length) {
+      // A newer proposal supersedes older pending rows for the same blocks.
+      await tx.canvasSuggestionChange.updateMany({
+        where: { canvasId, status: 'PENDING', blockId: { in: targetIds } },
+        data: { status: 'SUPERSEDED' },
```
ameer030, line 120: 🔵 **N08 — suggestion statuses are raw string literals across ~6 files**

`PENDING / ACCEPTED / REJECTED / STALE / SUPERSEDED` appear as string literals in the backend service, both mutator copies, both query files, and the dashboard, while the shared package already hosts the codebase's status enums.

Literals at e.g. `suggestions.ts:96/169`, `packages/shared/src/zero/mutators.ts:434-439`, both `queries.ts`, `CanvasSuggestionsPanel.tsx:88-89`; the shared enum home is `packages/shared/src/zero/types.ts` (`TicketStatus`, `CanvasCommentThreadStatus`).

Checklist rule: STYLE-C17 / STYLE-C01 (constants centralized in the shared source of truth).
</details>

Export a `CanvasSuggestionChangeStatus` const enum from `packages/shared/src/zero/types.ts` and reference it from all layers.
</details>

rules: STYLE-C17, STYLE-C01
agent: omp
run: n/a
</details>

apps/backend/src/controllers/canvasController.ts
```diff
@@ -1034,6 +1038,26 @@ export class CanvasController {
...
+          id: canvas.id,
+          title: canvas.title,
+          markdown: labelled,
+          labelInstruction: LABEL_INSTRUCTION,
+          url: getCanvasUrl(canvas.id, req.user?.workspaceId),
+        });
+        return;
+      }
+
       const markdown = blocks.length > 0 ? await convertBlockNoteToMarkdown(blocks) : '';
```
ameer030: 🔵 **N09 — unreachable branch in `readCanvas`**

After the labelled early return for non-empty documents, the non-empty arm of this ternary can never run — the markdown here is always `''`.

`apps/backend/src/controllers/canvasController.ts:1043-1057` returns early when `blocks.length > 0`; line 1061 still computes `blocks.length > 0 ? await convertBlockNoteToMarkdown(blocks) : ''`.

Checklist rule: STYLE-C14 (dead code removed, not parked).
</details>

Collapse the line to `const markdown = '';` (or drop the variable entirely in the cutover).
</details>

rules: STYLE-C14
agent: omp
run: n/a
</details>

ameer030 (commented):
🔴 **B02 — claw reports "Canvas updated." while nothing was applied**

`PATCH /api/internal/canvas/view/:id` now returns `202 pending-review` — no document change — but the unchanged claw `spaces-edit-canvas` treats any 2xx as success and tells the user "Canvas updated." Every agent edit now claims completion while the work sits parked waiting for a human click the user never learns about.

- Backend returns `202 {status:'pending-review', batchId, changeCount, message, url}`: `apps/backend/src/controllers/canvasController.ts:1164-1176`; `spacesFetch` only fails on non-2xx: `apps/xyne-claw-auth/backend/src/mcp/servers/xyne-spaces-client.ts:133-136`.
- The claw handler unconditionally prints "Canvas updated." + title/URL, still described as "Replace the contents of an existing canvas": `apps/xyne-claw-auth/backend/src/mcp/servers/xyne-spaces-tools.ts:4930-4997`. A grep for `pending-review|pendingChanges|changeCount|labelInstruction` across `apps/xyne-claw*` finds zero handling.
- Same class on the SDLC path: `updateArtifact` now returns `{parked: true, pendingChanges}` with no claw-side handling, so checkpoint flows treat parked output as landed.
- The behavior flip ships with no feature flag and no companion claw PR; backend and claw deploy independently, so the mendacity goes live the moment either side deploys. The `409` (no read receipt) also surfaces to the agent only as a raw JSON error.

Checklist rules: BUG-C34 (side-effect outcomes reported truthfully), ARCH-C01 (cross-version tolerant rollout), PROC-C05 (new surfaces default-OFF).
</details>

Companion change on the claw side (this PR or a paired PR with the merge order named in both bodies): surface `status: 'pending-review'` with its `message`/`changeCount` ("N changes proposed — a human must approve them"), keep "Canvas updated." only for real writes, and update the tool descriptions for `spaces-edit-canvas` and `spaces-sdlc-mutate-artifact`. Alternatively, flag-gate the S2S suggestion mode OFF until the claw handling ships.
</details>

rules: BUG-C34, ARCH-C01, PROC-C05
agent: omp
run: n/a
</details>

ameer030 (commented):
🟡 **W01 — the label protocol never reaches the agent: `labelInstruction` is dropped**

The read route returns the explanation of what `[b2]`-labels mean in a separate `labelInstruction` field, but the claw read handler maps only `{title, markdown, url}`. A model that never sees the rules strips the unknown `[bN]` prefixes on write-back — every paragraph becomes insert+delete duplicates for the human to review.

- Read response now carries `labelInstruction: LABEL_INSTRUCTION`: `apps/backend/src/controllers/canvasController.ts:1048-1058`; the claw destructures only `{title, markdown, url, error}`: `apps/xyne-claw-auth/backend/src/mcp/servers/xyne-spaces-tools.ts:4906-4931`.
- Unlabelled paragraphs parse as bare entries and `deriveOps` routes them straight to **insert** ops, while every seen-but-unclaimed block becomes a **delete** op: `apps/backend/src/services/canvas/blockLabels.ts` (`parseLabelledMarkdown` + `deriveOps`). One accept-all then rewrites the whole document.
- The comment at `blockLabels.ts:123` advertises "Bare: label dropped — recoverable later by similarity matching" — no similarity matching exists anywhere.

Checklist rule: BUG-C08 (diff-internal wiring: the consumer must know the contract the producer ships).
</details>

Append `labelInstruction` to the `spaces-read-canvas` tool output (and mention the labels in the tool description). Separately, implement the advertised similarity fallback in `deriveOps` or correct the comment at `blockLabels.ts:123`.
</details>

rules: BUG-C08, DOCS-C04
agent: omp
run: n/a
</details>

ameer030 (commented):
🟡 **W05 — PR body template statements don't match the diff**

Three template sections mis-declare the change: Areas Touched lists the claw apps (no claw files changed), "Env variable added/renamed/removed" is ticked (no env var in the diff, keys section still placeholder), and the API Contract section is entirely blank although the S2S contract changed in shape and status codes (202 / 409 / labelled read).

- `gh pr diff 854 --name-only`: 38 files, zero under `apps/xyne-claw*`, yet `apps/xyne-claw / apps/xyne-claw-auth` is checked under Areas Touched.
- Config section: `Env variable added / renamed / removed` is `[x]`; the diff contains no env change and `<!-- Keys added/changed: -->` is untouched placeholder.
- API Contract: neither "No API changes" nor "Existing contract modified" is checked; all five Zero-section sub-checkboxes (mutator mirroring, no `Date.now()`, optimistic parity, registration, regen) are unchecked even though schema + mutators + ACLs all changed.
- The unchecked "Backward compatible with deployed code, or rollout order noted" line is exactly the question B02 raises.

Checklist rules: PROC-C15 (template sections filled truthfully), PROC-C02 (declared scope honesty), PROC-C04 (verification statements precise).
</details>

Correct the four sections: untick the claw area (or land the claw companion and keep it ticked), untick the env-var box, declare the S2S contract change under API Contract with the rollout order for the claw, and run/fill the Zero sub-checklist.
</details>

rules: PROC-C15, PROC-C02, PROC-C04
agent: omp
run: n/a
</details>

ameer030 (commented):
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 2 |
| 🟡 Better to fix | 5 |
| 🔵 Nit | 9 |

### 🔁 Re-check
- 💬 **Existing threads:** none — first review pass on this PR
- ✅ **Verified OK:** S2S-only route wiring with `validateS2SKey` + `attachInternalUser` · tenant-pinned ACLs registered in all three layers with mutate ⊆ read · server-side accept CAS on `PENDING` rows · Redis lock is a genuine `SET NX EX` with finally-release · delete-ops gated on a read receipt and limited to blocks the agent saw · decoration plugin is stateless/memoized (undo stack intact) · dashboard listeners/intervals/RAF all cleaned up · anchor resolution order and insertion idempotency are correct for the single-apply case

### 📌 Verdict
2 blockers must be fixed before merge ❌ — accepted suggestions never refresh `canvas.content` (B01: SDLC artifact reads + hashes go permanently stale), and the unmodified claw reports a false "Canvas updated." for parked changes (B02: contract flip without a companion claw change or a flag).

apps/backend/src/services/canvas/suggestions.ts
adarshr001-stack: Fixed. readArtifact now reads Y-Sweet first: its document is the truth (empty included), canvas.content stands in only when Y-Sweet has no document (404), and any other failure is a 503 rather than a stale copy. Writing content on accept would have missed browser-applied accepts and human edits, which never touch canvas.content either.

packages/shared/src/canvas/suggestionApply.ts
adarshr001-stack: Fixed. Followers whose deterministic id is already present are skipped before splicing, at all three sites — the guard inserts already had. Verified by applying the same rows twice: replace, move, insert and delete are each idempotent, no duplicate ids.

apps/backend/src/services/canvas/suggestions.ts
adarshr001-stack: Fixed. TTL raised to 120s so an accept-all cannot outlive it, and the fail-open branch logs a warning instead of silently dropping the only guard.

apps/backend/src/services/canvas/suggestions.ts
adarshr001-stack: Fixed. A new batch now supersedes older PENDING inserts whose normalized text lands at the same anchor, so a re-proposal collapses to the newest. Matched on text+anchor rather than wholesale, so the same paragraph proposed at a different place stays its own suggestion.

apps/dashboard/src/components/Canvas/CanvasScreen/CanvasScreen.tsx
adarshr001-stack: Fixed. The flush re-arms while events remain, so a large deletion drains in 100-event chunks instead of stranding the tail until the next edit, and a queued flush is cancelled on canvas switch.

packages/shared/src/zero/queries.ts
adarshr001-stack: Fixed. The server now filters sibling rows to PENDING/ACCEPTED like canvasSuggestionPlacementOrder. Reproduced first: a rejected move claimed an order for a block that never moved, so the same accept placed a block differently depending on which runtime applied it.

apps/backend/src/services/canvas/blockLabels.ts
adarshr001-stack: Fixed all three: the similarity-matching comment is gone, CONFLICT removed from the status list, and the canvasSuggestionChanges docstring moved off canvasSuggestionPlacementOrder in both mirrored query files.

apps/backend/src/controllers/canvasController.ts
adarshr001-stack: Fixed. The empty string is inlined and convertBlockNoteToMarkdown dropped — it was the last user of that import.

adarshr001-stack:
> 🔴 **B02 — claw reports "Canvas updated." while nothing was applied**
> 
> `PATCH /api/internal/canvas/view/:id` now returns `202 pending-review` — no document change — but the unchanged claw `spaces-edit-canvas` treats any 2xx as success and tells the user "Canvas updated." Every agent edit now claims completion while the work sits parked waiting for a human click the user never learns about.
> 
> 🔍 Details
> * Backend returns `202 {status:'pending-review', batchId, changeCount, message, url}`: `apps/backend/src/controllers/canvasController.ts:1164-1176`; `spacesFetch` only fails on non-2xx: `apps/xyne-claw-auth/backend/src/mcp/servers/xyne-spaces-client.ts:133-136`.
> * The claw handler unconditionally prints "Canvas updated." + title/URL, still described as "Replace the contents of an existing canvas": `apps/xyne-claw-auth/backend/src/mcp/servers/xyne-spaces-tools.ts:4930-4997`. A grep for `pending-review|pendingChanges|changeCount|labelInstruction` across `apps/xyne-claw*` finds zero handling.
> * Same class on the SDLC path: `updateArtifact` now returns `{parked: true, pendingChanges}` with no claw-side handling, so checkpoint flows treat parked output as landed.
> * The behavior flip ships with no feature flag and no companion claw PR; backend and claw deploy independently, so the mendacity goes live the moment either side deploys. The `409` (no read receipt) also surfaces to the agent only as a raw JSON error.
> 
> Checklist rules: BUG-C34 (side-effect outcomes reported truthfully), ARCH-C01 (cross-version tolerant rollout), PROC-C05 (new surfaces default-OFF).
> 
> 💡 Recommended fix
> Companion change on the claw side (this PR or a paired PR with the merge order named in both bodies): surface `status: 'pending-review'` with its `message`/`changeCount` ("N changes proposed — a human must approve them"), keep "Canvas updated." only for real writes, and update the tool descriptions for `spaces-edit-canvas` and `spaces-sdlc-mutate-artifact`. Alternatively, flag-gate the S2S suggestion mode OFF until the claw handling ships.
> 
> 🛠 Debug

fixed in the companion claw PR #1801: spaces-edit-canvas now reports pending-review / no-changes truthfully and tells the model reads won't show pending changes, so it doesn't resend. #1801 merges first.

apps/backend/src/services/canvas/blockLabels.ts
adarshr001-stack: Fixed, and the labelled path no longer needs a matrix at all: both id lists are the same claimed blocks in two orders, so the LCS is the longest increasing run of positions — O(n log n), verified identical to the matrix over 4,500 randomised permutations, subsets and malformed inputs. The text diff keeps a matrix (duplicates are real there) but trims the common prefix/suffix first and caps the remaining window at 16M cells (64MB, flat Int32Array), degrading to positional matching past that and reporting it to the caller.

packages/shared/src/zero/queries.ts
adarshr001-stack: Kept on the synced row deliberately.

`afterContent` is read by the client — it is the proposal card's content, and the browser-side accept applies it — so it has to sync. `beforeContent` is server-only, but small: measured over real batches it averages ~440 B per row and ~1 KB per batch. Moving it to a non-synced sibling table would add a table, a migration and a server round-trip per card to save about a kilobyte per proposal.

It also exposes nothing new: it is a snapshot of blocks the viewer can already read on the canvas they have access to.

apps/backend/src/services/canvas/suggestions.ts
adarshr001-stack: Leaving this as is — the per-row writes are what implement the chaining.

Placed rows are processed in descending `orderIndex` so that simultaneous accepts forward d→c, c→b, b→a, each step seeing the previous one's result. Collapsing them into one statement would have to reproduce that ordering dependency in SQL. The count is bounded by the batch's placement rows, and with the lock now at 120s (W03) the TTL risk this compounded no longer applies.

packages/shared/src/zero/mutators.ts
adarshr001-stack: Intentional. The optimistic copy only ever flips ids this client itself just applied, and the server copy is authoritative: it filters to `status = 'PENDING'` (scoped by `canvasId`) before writing, and Zero rebases the client onto that result. The end state is always the server's CAS outcome.

The only observable difference is a momentary flash if another user rejected the same row in the same instant. A client-side predicate would not close that window either — the client's local view of the row can itself be stale — so it would add a read per row without changing the guarantee.

apps/dashboard/src/components/Canvas/CanvasSuggestionsPanel/CanvasSuggestionsPanel.tsx
adarshr001-stack: These are boundary adapters, not casts over unknown shapes.

`rows` and `placementRows` arrive from Zero as readonly arrays with `ReadonlyJSONValue` for the JSON columns, while the local row types and the shared `applyOps` input take plain arrays and `unknown` content. The cast bridges readonly-vs-mutable; every field the code actually reads is declared on those interfaces and typechecked from there.

`mutation as never` exists because `run()` accepts any of three mutator requests and Zero's `mutate` is generic per mutator, so no single parameter type unifies them. The one thing `run()` needs from the result is `.server`, and that is asserted where it is awaited.

apps/backend/src/services/canvas/suggestions.ts
adarshr001-stack: `status` is a plain `String` column, not a Prisma enum, and its five values are only compared inside the suggestion module and its mirrored query/mutator files — nothing outside the feature reads them.

A shared constant would not constrain the stored value any further; it would add the same import across both mutator mirrors and both query files for the same five strings. The values are exercised directly by the panel and the query filters, so a mistyped one shows up immediately as rows that stop filtering, not as silent drift.

packages/shared/src/zero/queries.ts
adarshr001-stack: The browser needs afterContent to draw the card and to apply an accept. beforeContent is small and only holds text the viewer can already read. A separate table would cost a migration and an extra request per card.

apps/backend/src/services/canvas/suggestions.ts
adarshr001-stack: The row-by-row writes are how the chaining works: rows are processed in descending order so each step sees the previous one's result. W03's 120s lock removed the timeout risk this added to.

ameer030:
🔴 **B01 — Claw canvas tools misreport parked edits as "Canvas updated." and drop the label instructions**

PATCH /api/internal/canvas/view now parks every edit (202, status 'pending-review', batchId, message) and readCanvas returns labelled markdown plus a labelInstruction field, but the route's only consumer ignores all of it: spaces-edit-canvas hardcodes `prefixChunk(1, "Canvas updated.", ...)` so the agent is told the write landed when nothing changed in the document, and spaces-read-canvas builds its reply from title+markdown only, dropping labelInstruction — the agent receives [bXXXX] labels it was never taught to echo back. <details><summary>🔍 Details</summary> xyne-spaces-tools.ts:4965 hardcodes "Canvas updated." reading only {url,title,error} off the response; :4895-4916 (spaces-read-canvas handler) never references labelInstruction (zero grep hits in the file). Backend side of the contract: canvasController.ts:1163-1174 (202 + status/batchId/message) and :1046-1059 (labelled markdown + labelInstruction field). Verified via grep that `/api/internal/canvas/view` has exactly two call sites, both in this file; no other consumer. </details> <details><summary>💡 Recommended fix</summary> Minimal: in spaces-edit-canvas forward `result.message` (or `status`) when present instead of the hardcoded sentence — e.g. `result.message ?? "Canvas updated."`; in spaces-read-canvas append `result.labelInstruction` to the returned text when present. Both fall back to today's behavior, so any rollout order stays truthful. </details>

rules: ARCH-C01, BUG-C34
agent: omp
run: pr-854/20260918T100856Z
</details>

ameer030:
🟡 **W04 — PR template left blank — no type ticked, no areas, checklist/rollback sections all placeholders**

The body keeps every template checkbox unchecked: no Type of Change, no Areas Touched, the Zero (Sync) and Schema sections are unmarked even though this PR adds a synced table, mutators, query ACLs and a migration, and the per-surface test-plan and self-review checklists are untouched. Reviewers have to reconstruct scope and risk from the diff alone, which is exactly what the template exists to remove. <details><summary>🔍 Details</summary> PR body: Description is only a POT drive link; every `##` section's checkboxes are `[ ]`. This PR does exercise the template's Zero and Schema sections (new canvas_suggestion_changes table + migration + mirrored mutators). </details> <details><summary>💡 Recommended fix</summary> Tick Type of Change (New feature), the touched areas, and fill the Zero + Schema sections' required checkboxes (mutator mirror, ACL registration, regeneration, backwards-compatible rollout order). One minute of metadata, big review-burden win. </details>

rules: PROC-C15
agent: omp
run: pr-854/20260918T100856Z
</details>

apps/backend/src/zero/mutators.ts
```diff
@@ -10551,6 +10623,131 @@ export function createMutators(
...
+          const change = await tx.run(zql.canvas_suggestion_changes.where('id', changeId).one());
+          if (!change) {
+            throw new Error('Suggestion change not found');
+          }
+          await requireCanvasEditAccess(tx, change.canvasId, authData.sub);
+
+          if (!accept) {
+            await tx.mutate.canvas_suggestion_changes.update({
+              id: change.id,
+              status: 'REJECTED',
```
ameer030, line 10651: 🟡 **W01 — Reject status transitions are not compare-and-set — a stale reject clobbers a committed ACCEPTED (and the service's accept clobbers a committed REJECTED)**

The accept path is compare-and-set (applySuggestionOutcome only touches rows still PENDING) but the reject transition has no state predicate anywhere, so a user rejecting a card from a stale view flips an already-ACCEPTED row back to REJECTED while its applied content stays in the document — and symmetrically, applySuggestionChanges' markStatuses updateMany carries no status predicate, so an in-flight legacy accept (the per-canvas lock only serializes applies, not rejects) overwrites a REJECTED with ACCEPTED after the user explicitly rejected the change. <details><summary>🔍 Details</summary> apps/backend/src/zero/mutators.ts:10651 writes status REJECTED immediately after the fresh read without testing change.status — contrast the CAS filter built into applySuggestionOutcome (same file, `.where('status', 'PENDING')`). services/canvas/suggestions.ts:208-225: markStatuses does updateMany({ where: { id: { in: ids } } }) — adding status to the where is the missing predicate; between the initial PENDING read (suggestions.ts:148-150) and these writes sit two Y-Sweet round-trips under the lock, a genuinely wide race window for a concurrent reject mutator. Shared-optimistic mirror has the same guardless reject: packages/shared/src/zero/mutators.ts:6926. Callers checked: inline ✓/✗ and Dismiss decorations (suggestionDecorations.ts) and the panel's Reject-all both route here; resolveAll's reject is fetch+update inside one mutator tx so it is consistent by itself. </details> <details><summary>💡 Recommended fix</summary> In resolveChange, after the read, return early unless change.status is PENDING or STALE (same-tx read, so this is the CAS); in suggestions.ts markStatuses, add `status: 'PENDING'` to the updateMany where for both ACCEPTED and STALE writes; mirror the reject guard in packages/shared/src/zero/mutators.ts resolveChange. </details>

rules: BUG-C01
agent: omp
run: pr-854/20260918T100856Z
</details>

apps/dashboard/src/components/Canvas/CollaborativeCanvasEditor/suggestionDecorations.ts
```diff
@@ -0,0 +1,262 @@
...
+  // preview must sit there too. Same rule as suggestionSiblingOrder: the group
+  // exists only while its head is an insert row of the SAME batch as the row
+  // being placed, and only rowless followers belong to it.
+  const insertHeads = new Set(
+    rows.filter(r => r.op === 'insert').map(r => `${r.batchId ?? ''}:${r.id}`),
+  );
+  const addressed = new Set(rows.map(r => r.blockId).filter((b): b is string => Boolean(b)));
+  const groupEnd = (id: string, batchId: string | undefined): number => {
+    let i = order.indexOf(id);
+    let end = blocks.get(id)!.end;
```
ameer030, line 127: 🟡 **W02 — groupEnd does an Array.indexOf per anchored row — O(placements × blocks) per decoration rebuild**

buildDecorations planes the doc into `order: string[]`, then `resolveAnchor`→`groupEnd` runs `order.indexOf(id)` — an O(n) scan — for every pending insert/move row, so one rebuild costs O(k·n). Rebuilds fire on every doc change while review mode is open (the plugin memo only dedupes selection-only updates), i.e. per keystroke: a whole-document proposal (which the code comments say can hold thousands of rows) on a large canvas is a quadratic hot path, which is exactly what the head commit claims to have eliminated. <details><summary>🔍 Details</summary> suggestionDecorations.ts:105-138 — `order.push(id)` linear array built in `doc.descendants`; `groupEnd` at :126 starts with `let i = order.indexOf(id)`; called per row via `resolveAnchor` at :160-170 from the placement loop at :195-229. Memo at :245-259 keys on (doc, rows, canEdit) identity, so any keystroke in review mode re-enters the full build. Callers checked: only this file.</details> <details><summary>💡 Recommended fix</summary> Build a `Map<string, number>` id→index in the same descendants pass (`indexOf` → `positions.get(id)`); groupEnd then costs O(group length). One extra Map, no other changes.</details>

rules: PERF-C01
agent: omp
run: pr-854/20260918T100856Z
</details>

apps/dashboard/src/components/Canvas/CanvasSuggestionsPanel/CanvasSuggestionsPanel.tsx
```diff
@@ -0,0 +1,332 @@
...
+  className?: string;
+  reviewMode?: boolean;
+  onToggleReview?: () => void;
+}
+
+/** A batch stays visible (rows painted, counted) while it has a PENDING row. */
+const visibleRowsOf = (rows: SuggestionRow[]): { visible: SuggestionRow[]; batches: number } => {
+  const byBatch = new Map<string, SuggestionRow[]>();
+  for (const row of rows) {
+    byBatch.set(row.batchId, [...(byBatch.get(row.batchId) ?? []), row]);
```
ameer030, line 58: 🟡 **W03 — visibleRowsOf rebuilds each batch bucket with a spread per row — O(n²) on the panel's hot grouping**

`byBatch.set(row.batchId, [...(byBatch.get(row.batchId) ?? []), row])` copies the existing bucket for every row, so grouping one batch is O(n²) in its size — the exact idiom flagged in mined PERF-47 ('the exact same Map pattern'). The memo recomputes whenever the Zero query delivers, which happens repeatedly while suggestions are pending: every blockDeleted flush bumps rows, every accept chain rewrites statuses. <details><summary>🔍 Details</summary> CanvasSuggestionsPanel.tsx:56-59. `rows` comes from `useCachedQuery(queries.canvasSuggestionChanges(...))` (:83); blockDeleted mutator updates currentAnchorId/updatedAt on rows (packages/shared/src/zero/mutators.ts:6980+), each delivered update changes row identity and re-runs the memo. With the 'thousands of rows per whole-doc proposal' sizing the author's own OUTCOME_CHUNK comment anticipates, one batch of 2k rows is ~2M element copies per delivery.</details> <details><summary>💡 Recommended fix</summary> `const bucket = byBatch.get(row.batchId); if (bucket) bucket.push(row); else byBatch.set(row.batchId, [row]);` — one push, O(n) total.</details>

rules: PERF-C01
agent: omp
run: pr-854/20260918T100856Z
</details>

packages/shared/src/zero/mutators.ts
ameer030, line 438: 🔵 **N01 — Client (optimistic) applySuggestionOutcome does not mirror the server: no PENDING compare-and-set, no canvasId scoping**

The shared/client copy flips every id in the outcome to ACCEPTED/STALE unconditionally, while the server copy only flips rows it re-read as `status == 'PENDING'` for the given canvasId. A row another device just REJECTED (or already ACCEPTED — e.g. a racing accept) is optimistically flipped to ACCEPTED on this client until the server push corrects it, and the un-scoped `where('id','IN', ...)` read feeds already-resolved rows into follower forwarding. The Zero contract depends on optimistic and server mutators agreeing, so this breaks the 'client mutators mirror the server ones' invariant of the feature. <details><summary>🔍 Details</summary> packages/shared/src/zero/mutators.ts:434-442 (unconditional `update({id, status})` loops) vs apps/backend/src/zero/mutators.ts:703-724 (`pending = tx.run(... where canvasId, id IN ids, status PENDING); for (const row of pending) update`). Callers checked: shared resolveChange/resolveAll (:6921, :6947) and backend equivalents (:10639+, :10730+).</details> <details><summary>💡 Recommended fix</summary> Mirror the server read in the shared copy: `tx.run(zql.canvas_suggestion_changes.where('canvasId', canvasId).where('id','IN',ids).where('status','PENDING'))` and update only those rows (also gives the client the same no-op behavior for racing accepts).</details>

rules: BUG-C28
agent: omp
run: pr-854/20260918T100856Z
</details>

apps/dashboard/src/components/Canvas/CanvasSuggestionsPanel/CanvasSuggestionsPanel.tsx
```diff
@@ -0,0 +1,332 @@
...
+  );
+  const pending = visible.filter(r => r.status === 'PENDING');
+  const attention = visible.filter(r => r.status === 'STALE');
+
+  // Fires one mutator with busy-state + error toast handling.
+  const run = useCallback(
+    async (key: string, mutation: unknown, failure: string): Promise<boolean> => {
+      setBusy(key);
+      try {
+        const result = z.mutate(mutation as never);
```
ameer030, line 97: 🔵 **N02 — run() smuggles the mutator through `as never` and re-casts the result to an inline server-result shape**

`z.mutate(mutation as never)` followed by `result as { server: Promise<...> }` discards the types the compiler already tracks (`z.mutate` is generic and returns the typed MutationResult), so a wrongly-shaped mutation would only surface at runtime. <details><summary>🔍 Details</summary> CanvasSuggestionsPanel.tsx:96-103. Callers pass values from `mutators.canvasSuggestion.resolveChange(...)`/`resolveAll(...)` (:166-170, :230-238, :274-282), which are fully typed — no cast is needed.</details> <details><summary>💡 Recommended fix</summary> Type the parameter instead of casting: `async (key: string, mutation: Parameters<Z['mutate']>[0], failure: string)` (or the Zero `MutatorRequest` equivalent), and read `const server = await z.mutate(mutation).server` directly.</details>

rules: STYLE-C11
agent: omp
run: pr-854/20260918T100856Z
</details>

apps/dashboard/src/components/Canvas/CollaborativeCanvasEditor/suggestionDecorations.ts
```diff
@@ -0,0 +1,262 @@
...
+        }),
+      );
+    }
+  }
+
+  // Inserts + moves in proposal order, so stacked widgets at one anchor keep it.
+  const placement = pending
+    .filter(r => r.op === 'insert' || r.op === 'move')
+    .sort((a, b) => a.createdAt - b.createdAt || a.orderIndex - b.orderIndex);
+  placement.forEach((row, i) => {
```
ameer030, line 195: 🔵 **N03 — Review mode mounts one ProseMirror widget tree per pending row with no bound**

Every pending/stale row becomes at least one widget (card + action bar) plus a node decoration, mounted simultaneously into the editor DOM; the author's own comments anticipate thousands of rows for a whole-document proposal, and there is no cap, grouping, or see-more affordance, so review mode on such a batch mounts thousands of widget subtrees. <details><summary>🔍 Details</summary> suggestionDecorations.ts:154-229 — `decos.push` per row in the stale loop (:158-167), the replace/delete loop (:170-186) and the placement loop (:195-229); all returned in one DecorationSet.create. The panel bar itself renders O(1) aggregates (CanvasSuggestionsPanel.tsx:246+), so only the inline path is affected. Verified callers: suggestions prop is NO_SUGGESTIONS unless reviewMode (CanvasScreen.tsx:1941).</details> <details><summary>💡 Recommended fix</summary> Cap the painted rows per open session (e.g. first N pending rows + a '+k more — accept/reject in the bar' aggregate widget), or collapse same-batch cards past a threshold; batchId is already available for grouping.</details>

rules: PERF-C06
agent: omp
run: pr-854/20260918T100856Z
</details>

ameer030 (commented):
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 1 |
| 🟡 Better to fix | 4 |
| 🔵 Nit | 3 |

### 🔁 Re-check
- 💬 **Existing threads:** 8 fixed ✅ · 3 satisfied by reply · 3 still open — "unconditional optimistic flip" (→ nit below), "CONFLICT in status comment" (schema.prisma) & "shape-hiding casts" (→ nit below)
- ✅ **Verified OK:** canvas_suggestion_changes ACL registered fail-closed + tenant-pinned in all three layers (backend zero, shared, prisma) · Prisma model ↔ shared zero schema ↔ generated mirror ↔ migration in sync · server/client mutators mirrored and deterministic (no Date.now/uuid, ids as args) · accept path CAS-guarded and apply idempotent · new queries tenant-pinned via canvas visibility · controller keeps requireView/EditAccess on resolved canvas · matrix path now patience LIS + 16M-cell-capped trimmed window with positional fallback · readReceipt fails closed (409 on Redis miss)

### 📌 Verdict
Request changes: the claw S2S consumer now misreports every parked canvas edit as "Canvas updated." (blocker), plus two correctness/perf shoulds and 3 still-open threads.

✍️ Replies on the still-open threads follow.

## 📋 Findings

The per-finding comments carry the details; blocks omitted from the table per template.

- 🔴 **Claw canvas tools misreport parked edits as "Canvas updated." and drop the label instructions** (apps/xyne-claw-auth/backend/src/mcp/servers/xyne-spaces-tools.ts)
- 🟡 **Reject status transitions are not compare-and-set** (apps/backend/src/zero/mutators.ts)
- 🟡 **groupEnd does an Array.indexOf per anchored row — O(placements × blocks)** (suggestionDecorations.ts)
- 🟡 **visibleRowsOf rebuilds each batch bucket with a spread per row — O(n²)** (CanvasSuggestionsPanel.tsx)
- 🟡 **PR template left blank** (body)
- 🔵 **Client optimistic applySuggestionOutcome does not mirror the server** (packages/shared/src/zero/mutators.ts)
- 🔵 **run() smuggles the mutator through `as never`** (CanvasSuggestionsPanel.tsx)
- 🔵 **Review mode mounts one widget tree per pending row with no bound** (suggestionDecorations.ts)

packages/shared/src/zero/mutators.ts
ameer030: Re-checked at head 2740c35: the optimistic flip is still unconditional — `packages/shared/src/zero/mutators.ts:437-443` updates every outcome id to ACCEPTED/STALE with no PENDING predicate, and the un-scoped `where('id','IN',...)` read at :434-436 still feeds follower forwarding. Neither of the two remedies offered above (PENDING predicate, or a divergence comment) landed. Re-raised in the new review as 🔵 N01.

apps/backend/src/services/canvas/blockLabels.ts
ameer030: Re-checked at head 2740c35: (a) the similarity-matching comment and (c) the misplaced docstring are fixed ✅ (docstring now sits on `canvasSuggestionChanges` in both queries.ts copies). But (b) is not: the `CanvasSuggestionChange.status` comment still lists `CONFLICT` at `apps/backend/prisma/schema.prisma:3176`, and no code path writes that value — the "CONFLICT removed" claim doesn't hold. Leaving this thread open on (b) only; deleting "CONFLICT" from that comment closes it.

apps/dashboard/src/components/Canvas/CanvasSuggestionsPanel/CanvasSuggestionsPanel.tsx
ameer030: Re-checked at head 2740c35: all four cast sites are unchanged — `CanvasSuggestionsPanel.tsx:86` (`rows as unknown as SuggestionRow[]`), `:96-99` (`z.mutate(mutation as never)` + inline result re-cast), `CanvasScreen.tsx:989-991` (`as unknown as InlineSuggestionRow[]` ×2), `useCanvasSuggestionAccept.ts:75` (`as unknown as AcceptableRow[]`). The hand-written interfaces remain the unchecked drift surface. The readonly↔mutable bridge covers mutability, not shape drift; the `as never` core is re-raised in the new review as 🔵 N02.

PR #863, opened 2026-08-21
feat: XYNE-55505 Spaces SDK and Spaces MCP

PrajwalPrasad2212:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates
- [x] Documentation
- [ ] CI/CD

## Description

Adds `packages/xyne-spaces-mcp`, an MCP server that gives coding agents read/write access to Spaces over the public API. XYNE-55505.

`packages/xyne-spaces-mcp` was deleted in `e6e43322c` to be rebuilt once `/api/sdk` was ready. It is ready, so this is a fresh build rather than a restore — the old package never touched `/api/sdk` at all. It drove the Electron app's local `/interact` endpoint with raw Prisma ASTs behind a consent-dialog token, and the routes it depended on (`/auth/me`, ticket PATCH, message reactions) were removed from `apps/electron/src/services/agent-auth.ts` in `6c88c823a`.

**Auth is unchanged and shared with the SDK.** The server speaks plain HTTP to `/api/sdk` carrying the same `xyne_sk_` API key, minted in Dashboard → Apps → API keys. It does not import `@xyne/spaces-sdk`; a catalog operation is a string in a request body, not a function call:

```
POST /api/sdk/catalog/query    {"name":"ticketDetailsByIdV2","args":{"ticketId":"…"}}
POST /api/sdk/catalog/mutate   {"name":"messages.react","args":{…}}
GET  /api/sdk/me
```

A key acts as its user, so Zero's per-table ACL remains the only authorization boundary — the MCP adds no second permission model.

**33 tools, not 508.** The API exposes 508 catalog operations; this exposes the few dozen an agent actually reaches for, one tool per operation rather than one per resource. The deciding factor is schema precision: a single `spaces_tickets` tool with an `action` enum has to flatten eight disjoint parameter sets into one schema where nothing is required, so wrong guesses (`status` for `statusV2`) are only caught server-side. Per-operation tools carry exact `required` arrays and enums, so the MCP client rejects a malformed call before it is sent. It also makes permissions expressible — a user approves `spaces_message_send` separately from `spaces_messages_list`.

| Area | Tools |
|---|---|
| Identity & search | `spaces_whoami`, `spaces_search` |
| Channels | `spaces_channels_list`, `spaces_channel_participants`, `spaces_channel_create` |
| Threads | `spaces_threads_list`, `spaces_thread_get`, `spaces_thread_create` |
| Messages | `spaces_messages_list`, `spaces_user_messages`, `spaces_message_send`, `spaces_message_update`, `spaces_message_react` |
| Tickets | `spaces_tickets_list`, `spaces_tickets_search`, `spaces_ticket_get`, `spaces_ticket_activities`, `spaces_ticket_create`, `spaces_ticket_update`, `spaces_ticket_transition` |
| Lookups | `spaces_projects_list`, `spaces_board_stages`, `spaces_users_list` |
| Comms | `spaces_notifications_list`, `spaces_notifications_mark_read`, `spaces_emails_list`, `spaces_calls_list`, `spaces_drafts_list`, `spaces_canvases_list`, `spaces_canvas_get` |
| Claw | `spaces_claw_list_agents`, `spaces_claw_run`, `spaces_claw_get_run` |

`spaces_user_messages` is built on `messagesBySenderPaginated`, the query added in `3fe7a9a16` specifically to remove an MCP limitation: "everything person X wrote, in order" was previously only reachable by fanning relevance-ranked searches across date windows, where an empty window was indistinguishable from a truncated one.

Claw is included because `/api/sdk/claw/*` exists precisely so one credential covers both — the backend relays with the deployment's own service credential. **Note the overlap with `packages/xyne-claw-mcp`**, which ships its own Claw tools behind a separate device login; installing both means the model sees two Claw tool sets and may pick the one that is not logged in. The README says to install one or the other.

**Output is deliberately unstructured.** Read tools return everything a row carries rather than a fixed projection, following `apps/xyne-claw-auth/backend/src/mcp/servers/xyne-spaces-tools.ts`. Descriptions and message bodies are uncapped, every id is labelled so the next call can chain on it, users render as `Name <email> (id: …)`, timestamps as IST, and every list ends with a pagination footer. What is bounded is list length, never field content. Canvas bodies are flattened from BlockNote blocks to markdown — headings, nested lists, checklists, numbered runs, code blocks and links all survive.

**A build-time guard, because nothing here is type-checked against the server.** Operation names are strings, so a backend rename compiles fine and fails at runtime with a 400. `scripts/check-operations.mjs` parses the real catalog and route table and asserts: every operation name exists; every argument the operation *requires* is one the tool sends (`.nullable()` counts as required, `.optional()` and `.default()` do not); every direct route is still mounted; and the search enums match the contract. It runs as part of `build`.

That second check is the one that earns its keep. Zero's optimistic-write model makes operations demand arguments no human caller would think of — `channelLatestMultipleConversationsV3` requires an `isMember` its body never reads, `userActivitiesPaginatedV2` requires `start` as an explicit `null`, `ticket.update` requires a caller-supplied `updatedAt`. Getting one wrong produces `"Validation failed: Required, Required"`, naming neither the operation nor the argument — the exact failure a client already hit on `kanbanTicketsPageV2`. This turns it into a build failure that names both.

**Refactoring:** the catalog parser was extracted from `packages/xyne-spaces-sdk/scripts/coverage.mjs` into a new `scripts/catalog.mjs`, imported by both gates. It could not simply be imported from `coverage.mjs`, which runs its whole check on import; and copying ~400 lines of parser that has already had four bugs found in it would re-earn all of them. `coverage.mjs` output is byte-identical to its baseline.

**No backend changes.** Every tool maps to an endpoint that already exists.

### Bugs found by testing this against a live server

Four, all fixed in this PR:

- **`spaces_ticket_transition` only worked on `NON_LINEAR` boards.** `nonLinear.transition` throws `"only valid for NON_LINEAR boards"`, and its own comment says DEFAULT/RELEASE boards must use the standard stage-update path. All 8 boards in the dev DB are `DEFAULT`, so the tool failed for the common case. It now reads the board type and routes accordingly.
- **…and the two paths disagreed.** `nonLinear.transition` applies the target stage's `defaultTicketStatusV2`; `ticket.update` does not derive it. Moving a ticket to Done left its status on `TODO`. The linear path now resolves the stage and passes the status explicitly, and rejects an unknown stage name with the list of valid ones.
- **`spaces_channels_list` reported "0 members" for every channel.** `Channel.participantCount` is unmaintained and reads 0 for all 27 rows; the accurate count is on `channel_stats`, which the query already joins. Free fix.
- **Numbered lists in canvases restarted at 1** wherever a blank paragraph interrupted a run.

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

## Schema & Data Changes

- [ ] Migration added

## Config, Environment & URLs

If any of the above:

## API Contract

- [x] No API changes
- [ ] New endpoint

---

## How did you test it?

**End to end against a live backend on `localhost:3001`, with a real API key and the seeded dev database** (27 channels, 302 tickets, 1145 messages, 6 projects). The key was minted through the backend's own `mintApiKey` + `apiKeyExpiryFrom`, resolving the org member by `memberId` exactly as `routes/sdk-keys.ts` does, so the credential path is the production one.

Reads, driven through the built server over stdio JSON-RPC: `spaces_whoami` named the acting user, workspace, org and key expiry. `spaces_projects_list`, `spaces_channels_list` (with real member counts and a `[Showing 1-3 of 26 …]` footer), `spaces_channel_participants`, `spaces_threads_list`, `spaces_board_stages` (5 stages in workflow order), `spaces_tickets_search`, `spaces_tickets_list`, `spaces_ticket_get`, `spaces_ticket_activities`, `spaces_user_messages`, `spaces_notifications_list`, `spaces_drafts_list`, `spaces_calls_list` (recording + transcript + AI summary), `spaces_canvases_list`, and `spaces_canvas_get` — which rendered a 11.7 KB Incident Runbook into clean markdown with nested bullets and correctly numbered procedure steps.

Writes, confirmed both through the tools and directly in Postgres:

```
spaces_thread_create   → conversation ede4474b… + message 1ad6f491…
spaces_message_send    → message 2b66dbf2…
spaces_message_react   → reaction row (tada, correct userId)
spaces_ticket_create   → JSDK-0002, key allocated by the project sequence
spaces_ticket_update   → priority CRITICAL, assigned via EMAIL → resolved to a userId
spaces_ticket_transition → Backlog → In Progress → Done, status COMPLETED
```

The stage moves also produced the SYSTEM messages the post-commit side effects emit, which then showed up in `spaces_user_messages` — so the full mutate path including the async drain ran.

**ACL boundary, with two users' keys.** A private DM channel with 5 threads: the member's key returned them; a second user's key returned "No thread(s) found" for the same channel id, and the channel did not appear in that key's `spaces_channels_list` at all.

**Auth failure paths.** A revoked key (row deleted) → *"Not authorized… Set XYNE_SPACES_API_KEY to a key minted in the Spaces dashboard, under Apps → API keys."* An expired key (`expires_at` backdated) → *"Your Xyne Spaces API key has expired — keys last 30 days…"* A forged key → `unauthenticated` with a `request_id`.

**Search** reached Vespa and returned cleanly, but the local index is not backfilled, so it answered 0 hits. Parameter validation was still proven both ways: `type: ["message"]` (singular) was rejected with `validation_failed`, which is the mistake the JSON Schema enum prevents a model from making in the first place.

Build gate, against the real catalog and route table:

```
tools:      33
catalog:    36 operation call(s) checked against 508 catalog operations
direct:     8 route call(s) checked against 12 mounted routes
enums:      2 search value set(s) compared with the contract
```

Each of the four checks was deliberately broken and confirmed to report, then restored — including the two failures that already reached a client of the SDK:

```
- "channelLatestMultipleConversationsV3" requires `isMember`, `limit`, which the tool does not send
- spaces_search.type offers message, which the server rejects
```

Regression on the extracted parser: `coverage.mjs` output diffed byte-identical to its pre-refactor baseline, and `pnpm --filter @xyne/spaces-sdk verify` reports 508/508 with contract OK.

### Automated

- [x] Not covered by tests 

### Manual

**Users**
- [x] Single user
- [x] Two users at once (two accounts / two browsers) — sync lands on both, no cross-user leakage 

**Login**
- [ ] Google
- [ ] Microsoft
- [x] Session expiry, re-login, logout 

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop 

**Surface & data**
- [ ] Web browser 
- [ ] Electron desktop 
- [x] Empty state 
- [x] Large / realistic data volume 

**Not verified: the three `spaces_claw_*` tools.** `claw-auth` was not running locally, so `spaces_claw_list_agents` returned `upstream_unavailable` — which does confirm the route, the auth and the error mapping, but not a real agent run. Worth exercising once against an environment with Claw up.

---

## Checklist


apps/backend/src/api/sdk/auth.ts
```diff
@@ -0,0 +1,230 @@
...
+  readonly authData: AuthData;
+  readonly ctx: Context;
+  readonly keyId: string;
+  /** When the presented key stops working, so a caller can rotate before it does. */
+  readonly keyExpiresAt: Date;
+}
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace Express {
```
utkarsh-611: this we are already doing in the auth middleware you dont need to add it again

apps/backend/src/api/sdk/auth.ts
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * Mint a key string for an identity.
+ *
+ * Exported for the key-management routes; nothing else should call it. The
+ * caller is responsible for persisting the returned string as `token`, because
+ * it cannot be re-derived — `encrypt` uses a random IV, so encrypting the same
+ * identity twice yields two different keys.
+ */
+export function mintApiKey(identity: KeyIdentity): string {
```
utkarsh-611: why is this not JWT ?

apps/backend/src/api/sdk/auth.ts
```diff
@@ -0,0 +1,230 @@
...
+  const sealed = encrypt(JSON.stringify(identity));
+  return `${API_KEY_PREFIX}${Buffer.from(sealed, 'utf8').toString('base64url')}`;
+}
+
+/** When a key minted now should stop working. */
+export function apiKeyExpiryFrom(now: Date = new Date()): Date {
```
utkarsh-611: why users can't set the ttl ? for the keys

apps/backend/src/api/sdk/auth.ts
```diff
@@ -0,0 +1,230 @@
...
+  return new Date(now.getTime() + sdkConfig.apiKey.ttlDays * 24 * 60 * 60 * 1000);
+}
+
+/** Recover the identity from a key string, or undefined if it is not readable. */
+function openApiKey(key: string): KeyIdentity | undefined {
```
utkarsh-611: when will this function be used ?

apps/backend/src/api/sdk/auth.ts
```diff
@@ -0,0 +1,230 @@
...
+}
+
+/**
+ * Authenticate an API key and attach the acting principal.
+ *
+ * An expired key is reported as such, since the caller can act on that. Every
+ * other failure returns the same `unauthenticated` error: distinguishing "no such
+ * key" from "key for a deleted user" would let a caller probe which keys exist.
+ */
+export async function apiKeyAuth(
```
utkarsh-611: why this is not in middleware ?

apps/backend/src/api/sdk/config.ts
```diff
@@ -0,0 +1,35 @@
...
+  get enabled(): boolean {
+    return process.env['SDK_API_ENABLED'] === 'true';
+  },
+
+  /**
+   * Development escape hatch: run reads against the primary pool when no read
+   * replica is configured. Never enable in production — the replica exists to
+   * keep SDK read traffic off the write path.
+   */
+  get allowPrimaryForReads(): boolean {
```
utkarsh-611, line 22: why can't we just use the environment development for this

apps/backend/src/api/sdk/config.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * /api/sdk configuration.
+ *
+ * Read straight from the environment rather than threaded through the Joi
+ * schema in `config/env.ts`, so enabling the SDK surface in a deployment is
+ * additive and cannot break boot for deployments that do not run it.
+ */
+
+export const sdkConfig = {
```
utkarsh-611, line 9: why unecessary envs are getting added ?

apps/backend/src/api/sdk/errors.ts
```diff
@@ -0,0 +1,117 @@
...
+  return new ApiError('internal', 'Unexpected error.', { cause: err });
+}
+
+interface ApplicationErrorLike {
+  readonly name: string;
+  readonly message: string;
+  readonly details: unknown;
+}
+
+function isApplicationError(err: unknown): err is ApplicationErrorLike {
```
utkarsh-611, line 168: 400, 404, 500 error ideally can only happen here after auth is done why so much handling is happening here ?

apps/backend/src/api/sdk/errors.ts
```diff
@@ -0,0 +1,117 @@
...
+import {
+  ERROR_CATALOG,
+  isErrorCode,
+  type ErrorCode,
+  type ErrorDetail,
+} from '@xyne/spaces-contract';
+import { MutationACLError } from '@/zero/acl/core/types';
+import { ZodError } from 'zod';
+
+export class ApiError extends Error {
```
utkarsh-611: if you making your own api error class for sdk then name is accordingly sdk errors or something

apps/backend/src/api/sdk/handler.ts
```diff
@@ -0,0 +1,110 @@
...
+ * Wrap an async endpoint so a rejected promise reaches `errorHandler` instead of
+ * becoming an unhandled rejection, and so schema failures arrive as validation
+ * errors rather than as a generic 500.
+ */
+export function handle(
+  fn: (req: Request, res: Response) => Promise<void>,
+): RequestHandler {
+  return (req, res, next) => {
+    fn(req, res).catch((err: unknown) => {
+      next(err instanceof ZodError ? ApiError.validation(err) : err);
```
utkarsh-611: I thought these error handling would be happening in the Api Error ?

apps/backend/src/zero/queries.ts
```diff
@@ -5156,6 +5124,46 @@ dmChannelsLatestMessagesPaginated: defineQuery(
...
+   *
+   * The arbitrary-user counterpart to `userSentMessagesPaginated`, which is fixed to
+   * the caller. Built for "everything person X wrote, in order" — previously only
+   * reachable by fanning out relevance-ranked searches across date windows, which
+   * could not distinguish a genuinely quiet window from a truncated one.
+   *
+   * `MessagesACL` still applies, so this reveals nothing the caller could not
+   * already read; it only makes the ordering and paging reliable.
+   */
+  messagesBySenderPaginated: defineQuery(
```
utkarsh-611: why is this required ? Vespa search is the place to do this right? from:user and the message ?

apps/backend/src/app.ts
```diff
@@ -326,6 +329,24 @@ export class App {
     this.app.use(decryptRequestBodyMiddleware);
     this.app.use(encryptResponseBodyMiddleware);
 
+    // Public SDK API. It authenticates its own API keys, so it must be mounted
+    // before the legacy catch-all `/api` session middleware.
+    if (sdkConfig.enabled) {
+      this.app.use('/api/sdk', createSdkRouter());
```
utkarsh-611: without any middleware for authentication ?

apps/backend/src/app.ts
```diff
@@ -326,6 +329,24 @@ export class App {
...
+      logger.info('Public SDK API mounted at /api/sdk');
+
+      // Where those keys are minted. Session-authenticated, not key-authenticated:
+      // you cannot use an API key to mint another one.
+      this.app.use('/api/sdk-keys', authMiddleware.authenticate, sdkKeyRoutes);
+    } else {
+      this.app.use('/api/sdk', (_req, res) => {
```
utkarsh-611: this will happen automatically you dont neeed to explicitly write 404

apps/backend/.env.example
```diff
@@ -6,8 +6,8 @@ HOST=0.0.0.0
 # ORG_MEMBER_LIMIT=10
 
 # CORS Configuration
-CORS_ORIGIN=http://localhost:5173,http://localhost:5174,xyne-spaces://.
-ALLOWED_REDIRECT_ORIGINS=http://localhost:5173,http://localhost:5174
+CORS_ORIGIN=http://localhost:5173,http://localhost:5174,http://localhost:4173,xyne-spaces://.
```
utkarsh-611: development changes leaking in

apps/backend/package.json
```diff
@@ -89,6 +89,7 @@
     "@types/unzipper": "^0.10.11",
     "@types/yauzl": "^2.10.3",
     "@xyne/shared": "workspace:*",
+    "@xyne/spaces-contract": "workspace:*",
```
utkarsh-611: this library we would need to expose to people name is spaces-sdk

apps/backend/prisma/migrations/20260820120000_sdk_api_keys/migration.sql
```diff
@@ -0,0 +1,25 @@
+-- CreateTable
+CREATE TABLE "non_zero"."sdk_api_keys" (
+    "workspaceId" TEXT NOT NULL,
+    "id" TEXT NOT NULL,
+    "name" TEXT NOT NULL,
+    "userId" TEXT NOT NULL,
+    "token" TEXT NOT NULL,
+    "expires_at" TIMESTAMP(3) NOT NULL,
```
utkarsh-611: status column also will be required

packages/xyne-spaces-sdk/src/resources/messages.ts
```diff
@@ -0,0 +1,298 @@
...
+import type { Message, MessageType } from '../types/index.js';
+
+export class MessagesResource extends Resource {
+  /**
+   * List the messages in a thread, oldest first.
+   *
+   * @example
+   * const messages = await sdk.messages.listByConversation('conv-123');
+   */
+  listByConversation(conversationId: string): Promise<Message[]> {
```
utkarsh-611: no pagination ? you fill fetch all the messages of a conversations ?

utkarsh-611: this pattern is in lot of places, no paginations

packages/xyne-spaces-sdk/README.md
```diff
@@ -0,0 +1,406 @@
...
+and write the Spaces product itself performs — plus Xyne Claw remote agents.
+
+Zero runtime dependencies. Runs in Node 18+ and in the browser.
+
+```typescript
+import { createClient } from '@xyne/spaces-sdk';
+
+const sdk = createClient({
+  baseUrl: 'https://spaces.example.com',
+  apiKey: process.env.XYNE_SPACES_API_KEY,
```
JamesGeorg: Do we have a token equivalent for this? Cannot have client side apps push API key

apps/backend/package.json
PrajwalPrasad2212: This is @xyne/spaces-contract, not spaces-sdk — it's an internal shared-types package (error codes, the search schema) consumed only by the backend and by the SDK/MCP's own build gates, never published or imported externally. @xyne/spaces-sdk is the package people actually install.

apps/backend/src/api/sdk/config.ts
PrajwalPrasad2212: Done.

apps/backend/src/api/sdk/config.ts
PrajwalPrasad2212: Removed.

apps/backend/src/api/sdk/errors.ts
PrajwalPrasad2212: Because the mutator catalog throws plain Errors at ~485 sites, plus ApplicationError and MutationACLError. Those have to be mapped to codes somewhere.

apps/backend/prisma/migrations/20260820120000_sdk_api_keys/migration.sql
PrajwalPrasad2212: Done.

packages/xyne-spaces-mcp/src/render.ts
```diff
@@ -0,0 +1,268 @@
...
+		.replace(/<h3[^>]*>/gi, "\n\n### ")
+		.replace(/<h[4-6][^>]*>/gi, "\n\n#### ")
+		.replace(/<li[^>]*>/gi, "\n- ")
+		.replace(/<\/?(strong|b)\b[^>]*>/gi, "**")
+		.replace(/<\/?(em|i)\b[^>]*>/gi, "*")
+		.replace(/<br\s*\/?>/gi, "\n")
+		// `li` is deliberately absent: `<li>` already opens its own line, so
+		// closing it too would put a blank line between every bullet.
+		.replace(/<\/(p|div|h[1-6]|ul|ol|blockquote|tr)>/gi, "\n")
+		.replace(/<[^>]+>/g, "");
```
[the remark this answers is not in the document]
PrajwalPrasad2212: not a sanitizer — output is plain text, never HTML

packages/xyne-spaces-sdk/README.md
PrajwalPrasad2212: @utkarsh-611

apps/backend/src/zero/queries.ts
PrajwalPrasad2212: Sure, will route it through vespa search.

packages/xyne-spaces-sdk/README.md
utkarsh-611: @PrajwalPrasad2212 for people using sdk directly in frontend, we need to have a xyne sso kind of thing where we open xyne space in the tab or in the browser ( if this is running in node), keep on polling from initiator to weather the sso is completed on xyne spaces. If xyne spaces was not logged in the user has to login to xyne spaces first, once done, or if xyne spaces was already logged in, the user has to approve the sso, and then the next poll of the user will get success response and along with the set cookies, from now on whatever apis are called will have this cookies.

Only issue in this approch is if some is using this in there parent app, then these cookie will also go to there backend in header. This cookie will be short lived

We need to make this little more solid so please check this implementation.

utkarsh-611: Do not remove api key, keep api key and add this as another method of authentication and update your middleware to handle both.

packages/xyne-spaces-mcp/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "scripts": {
+    "check": "node scripts/check-operations.mjs",
+    "build": "tsc && npm run check",
+    "typecheck": "tsc --noEmit"
+  },
+  "engines": {
+    "node": ">=22"
+  },
+  "dependencies": {
+    "@modelcontextprotocol/sdk": "^1.29.0"
```
utkarsh-611: where is the dependency for the xyne-sdk here ?

apps/backend/package.json
utkarsh-611: I didn't see this dependency in xyne mcp or anywhere else

packages/xyne-spaces-sdk/README.md
PrajwalPrasad2212: Sure

packages/xyne-spaces-mcp/package.json
PrajwalPrasad2212: Spaces MCP use api/sdk routes, not the sdk functions. Hence, no dependency.

packages/xyne-spaces-sdk/README.md
PrajwalPrasad2212: This is done.

PR #869, opened 2026-08-21
fix: global always-on default rate limiter with per-endpoint overrides

XyneSpaces:
## Summary
Mounts a single **always-on** rate limiter at the app level so **every** route gets a protective default — even when the route author never added a limiter. This is the root cause behind the CodeQL `js/missing-rate-limiting` alerts: limiting was opt-in per router, so new endpoints shipped unprotected. To raise or tighten a specific endpoint you now **pass its value** through an override registry instead of adding another limiter.

New file `apps/backend/src/middleware/globalRateLimit.ts`; wired once in `apps/backend/src/app.ts` (replaces the long-commented-out limiter block).

## PR Checklist
1. **Problem Description:** Rate limiting was per-router and opt-in, so any endpoint whose author forgot to add a limiter shipped with no protection (the CodeQL `js/missing-rate-limiting` finding, the single largest recurring alert class). There was no enforced default.
2. **How the issue was identified:** CodeQL code-scanning backlog review + 7-day production traffic analysis of the auth/session/webhook route classes.
3. **Solution implemented:** A single always-on `globalRateLimiter` dispatcher mounted once in `app.ts`. It resolves a per-endpoint override by longest-prefix match on each request and applies the matching limiter, falling back to a protective default (`DEFAULT_LIMIT`/`DEFAULT_WINDOW_MS`, seeded from existing `RATE_LIMIT_*` env). Endpoints raise/tighten their ceiling with a one-line `registerRateLimit('/api/x', { limit })`; webhook/S2S routes opt out with `{ skip: true }`. Buckets are keyed per-endpoint-group AND per-caller — since this runs before per-route auth, the caller key is derived from auth material already on the raw request (hashed Bearer token or session cookie, client-IP fallback via X-Forwarded-For). Built-in overrides sized from prod traffic: chatty session reads generous (240/min), credential endpoints tight (login 20/min), MCP moderate (60/min), webhooks/S2S skipped (they keep their dedicated limiters).
4. **Documentation:** Inline module docstring documents the override API and the keying/store trade-offs.
5. **DB Alter Details:** N/A
6. **Data fix Details:** N/A
7. **Environment variable Changes:** None new. Reuses existing `RATE_LIMIT_WINDOW_MS` / `RATE_LIMIT_MAX_REQUESTS` (already in `config/env.ts`) for the default.
8. **Testing:** `tsc --noEmit` clean; `eslint` clean on changed files; longest-prefix resolution smoke-tested (`/api/auth/me`->240, `/api/auth/login`->20, `/api/query/claw`->60, `/api/webhooks/...`->skip, everything else->default).
9. **Services to which this change is to be deployed:** backend (apps/backend).
10. **Main PR link (if against a release branch):** N/A (targets `main`).

## Known follow-ups (called out, not blocking)
- **Store is in-process.** With N backend replicas the effective ceiling is `limit x N`. A shared Redis store (`rate-limit-redis` over the existing ioredis client / `redisFactory`) is the tracked follow-up to make the limit exact fleet-wide.
- Recommend shipping in shadow/log-only observation for 1–2 weeks (the limiter already logs every block via `[rate-limit] blocked`) before treating the numbers as final, then tune per-endpoint overrides from real `blocked` rates.

### DIFF
--- a/apps/backend/package.json
+++ b/apps/backend/package.json
@@ -143,6 +143,7 @@
     "pg": "^8.16.3",
     "postgres": "^3.4.7",
     "pptxgenjs": "^4.0.1",
+    "rate-limit-redis": "^4.3.1",
     "reflect-metadata": "^0.2.2",
     "remark-frontmatter": "^5.0.0",
     "remark-parse": "^11.0.0",
--- a/apps/backend/src/app.ts
+++ b/apps/backend/src/app.ts
@@ -5,6 +5,7 @@ import helmet from 'helmet';
 import compression from 'compression';
 import cookieParser from 'cookie-parser';
 import { webhookLimiter } from '@/middleware/rateLimiters';
+import { globalRateLimiter } from '@/middleware/globalRateLimit';
 import morgan from 'morgan';
 
 import { config } from '@/config/env';
@@ -220,17 +221,11 @@ export class App {
       })
     );
 
-    // Rate limiting
-    // const limiter = rateLimit({
-    //   windowMs: config.rateLimit.windowMs,
-    //   max: config.rateLimit.max,
-    //   message: {
-    //     success: false,
-    //     error: 'Too many requests from this IP, please try again later.',
-    //     timestamp: new Date().toISOString(),
-    //   },
-    // });
-    // this.app.use(limiter);
+    // Global rate limiting — always-on default for EVERY route. Endpoints that
+    // need a different ceiling register an override via registerRateLimit()
+    // (see @/middleware/globalRateLimit). Mounted here so a route can never ship
+    // unprotected just because its author forgot to add a limiter.
+    this.app.use(globalRateLimiter);
 
     // Compression - skip SSE endpoints to allow streaming
     this.app.use(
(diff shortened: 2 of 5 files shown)


XyneSpaces:
**Update — Redis-backed store added (commit b88f38c)**

Addressing review feedback ("let's use Redis else it's not a proper solution"): the global rate limiter is now backed by a **shared Redis store** instead of the in-process one, so the ceiling is exact across all backend replicas rather than `limit x replicaCount`.

- `rate-limit-redis@^4.3.1` over a dedicated `ioredis` client from `redisFactory` (`createRedisClient('rate-limit')`), one client shared by all limiter instances, per-window key prefix `rl:<windowMs>:`.
- **Fail-open:** `passOnStoreError: true` — a Redis outage lets traffic through instead of 500-ing every route.
- **Kill-switch:** new optional env `RATE_LIMIT_STORE` (`redis` | `memory`, default `redis`) forces the per-pod in-memory store if ever needed.
- `tsc --noEmit` and `eslint` clean on changed files.

PR #886, opened 2026-08-21
fix: XYNE-54658 add artifacts for agent discovery, mcp suggest

12-aryan:
POT for suggest connectors : https://drive.google.com/file/d/15S95M4YjXmSnqUChDTrY6NU8CvyZgfGh/view?usp=drive_link
POT for sidebar quickaccess : https://drive.google.com/file/d/1JQ7iB1-PnugJoXGo4nj7gFgWrdVlUFax/view?usp=sharing
POT for OAuth fix of Anthropic : https://drive.google.com/file/d/1qf_5ZGIIP7vEJHwFH2LW8bSrXdKtwQ7g/view?usp=sharing

## Type of Change

- [x] New feature
- [x] Enhancement
- [x] Bug fix

## Description

Agent and MCP discovery: Ask AI answers "what agents/connectors do we have?" with cards instead of prose, those sections are reachable straight from the app rail, and signing in to Anthropic Claude works end to end. XYNE-54658 / XYNE-56602.

**Connector suggestion cards in chat** — new `mcp_suggest` flow component, with two triggers:
- Ask AI can call a new `suggest-connectors` tool (available to every agent, no per-agent config).
- claw-auth also scans the user's message for connector signals (a GitHub link, "figma", "grafana"…), checks which of those they haven't connected, and posts the card itself.

The model only names connector types — every name, description and connected state comes from `mcp_servers` / `user_mcp_connections`, so it can't invent a connector or claim something is connected when it isn't. "List all the MCPs" shows 5 plus a **Browse MCPs** link.

**Agent cards and roster summary** — `describe-agent` gains two modes beyond the single card:
- `summary: true` → roster card: total count with an org-wide / personal breakdown, a sample of agents each with a **View** button, and a **Browse agents** footer. For "list all my agents".
- `slugs: [...]` → a stack of profile cards, best match first. For "which agents can review PRs?".

Counts and rows come from the server's own query and the tool tells the model not to state a number or list agents in prose, so the roster can't be misreported. Card visuals match the connector card and the existing created-agent card.

**Connect / manage from the MCP library** — `ai/library/mcp/:type` gains Connect (OAuth, credential modal, or direct), Disconnect, Edit credentials and Edit definition. Definition editing follows the existing server rule (connector author or CLAW_ADMIN) and says up front when a change goes to the admin review queue instead of applying. OAuth consent now always opens in the system browser rather than an in-app tab.

**Sidebar quick access** — hovering the Xyne AI or Chat icon in the app rail opens a hover card with that section's own links, so a section is one hover away instead of two navigations. Xyne AI reads the same `NAV_ITEMS` the AI sidebar renders (with the same admin / org-manager filtering, so it can't reveal a section the real sidebar hides); Chat mirrors the rows at the top of `ChatDirectory`. Once you're already inside a section the rail falls back to a plain tooltip, since its sidebar lists the same links a few pixels away.

**Anthropic Claude sign-in** — the provider settings tab gets "Sign in with Claude": a user-scoped PKCE flow that opens claude.ai, then takes the pasted code or callback URL. Anthropic's Claude Code flow redirects to a loopback port we don't own, hence the paste step. This captures a **refreshable** token, unlike a pasted `claude setup-token` value.

Two fixes alongside it:
- Model list returned 400 — the stored credential is a JSON bundle but the models route was passing it to Anthropic as a raw bearer. Now goes through `extractClaudeBearer`, which handles both a bundle and a bare pasted key.
- The model select rendered blank when nothing was chosen; it now shows **Use default** (Radix Select can't hold an empty value, so `__default__` stands in and maps back to `''`).

## Areas Touched

- [x] `apps/dashboard`
- [x] `apps/xyne-claw` / `apps/xyne-claw-auth`

## Zero (Sync) Changes

## Schema & Data Changes

## Config, Environment & URLs

Claude sign-in reuses the existing credential encryption key and Redis connection; PKCE verifiers are stored under `claude-pkce-user:` with a 10-minute TTL.

## API Contract

- [x] New endpoints (additive, claw-auth)

- `POST /provider-credentials/claude/oauth/start` — returns the claude.ai authorize URL + state.
- `POST /provider-credentials/claude/oauth/exchange` — accepts the pasted code or callback URL, stores the encrypted token bundle.

No existing contract changed. `GET /settings/claude/models` is unchanged in shape — it was returning 400 and now resolves the bearer correctly. The dashboard also now calls existing claw-auth routes (`/users/:id/oauth/:type/authorize`, `/users/:id/connections`, `/servers`) that were previously only used by claw's own UI.

---

## Rollout note

`@xyne/shared` needs to ship together with claw-auth and the dashboard. The backend validates every posted flow against that schema, so a build that doesn't know `mcp_suggest` rejects the whole message instead of rendering the card.

## How did you test it?

Manually against a local stack, confirmed from `agent_runs` and the posted messages:

- "list all the MCPs" → connector card with 5 rows + Browse MCPs
- pasted a Google Doc link → Google connector card
- "list all my agents" → roster card with counts, View buttons and Browse agents
- MCP detail page: Connect (OAuth + credential), Disconnect, Edit credentials
- Rail hover: Xyne AI and Chat quick menus navigate and dismiss on click; tooltip-only once inside the section
- Settings: Sign in with Claude → paste callback URL → model list loads; blank selection shows "Use default"

### Automated

- [x] Not covered by tests — no existing harness for flow components or claw tool wiring

### Manual

**Users**
- [x] Single user

**Login**

**Run mode**


apps/xyne-claw-auth/backend/src/routes/settings.ts
```diff
@@ -710,4 +715,156 @@ function decodeJwtChatgptAccountId(jwt: string): string | undefined {
...
 
+// ── Claude browser sign-in (user-scoped) ──────────────────────────────────────
+// Mirrors the agent-scoped flow in agents.ts, but writes the user's own
+// credential. Anthropic's Claude Code flow is PKCE with the verifier doubling
+// as `state`, and its redirect lands on localhost:53692 — a port only the
+// Claude Code CLI listens on — so the browser cannot post back to us. The user
+// copies the code (or the whole redirect URL) and pastes it into /exchange.
+//
+// Captures a REFRESHABLE token, unlike a pasted `claude setup-token` value.
+const USER_CLAUDE_CLIENT_ID = "9d1c250a-e61b-44d9-88ed-5944d1962f5e";
```
Pradeesh333: do we need these hardcoded values?

12-aryan: Yes,  they are needed and none of them are secrets. The
client ID, URLs and scopes are Anthropic's own fixed values, identical in
every environment. REDIRECT_URI is the loopback Anthropic registered against that client ID, so any other value just fails the token exchange. PKCE_PREFIX is only a Redis key namespace

harsh-1923 (dismissed):

Mohan0544 (dismissed):

harsh-1923 (approved):

devesh-juspay (approved):

PR #901, opened 2026-08-21
fix: twin why popover

Pradeesh333:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

UI-only update to the digital twin's reply tray. Two changes:

**1. "Why this reply" is now a popover, not a drawer.**
`TwinReasoningDrawer` is replaced by `TwinReasoningPopover`, anchored above its trigger in the tray. It is deliberately non-modal — no backdrop or dimming — so the thread stays readable and clickable behind it. Escape, click-outside and re-clicking the trigger all dismiss it.

The trigger is now a dedicated **"Why?"** button. Previously the whole footer provenance line was clickable, which made a full-width line of prose easy to hit by accident; that line is now plain text.

**2. Real avatars for the person being replied to.**
`SourceAvatar` previously hashed the display *name* into a private 8-colour palette, so the same person could appear in one colour in the tray and a different one in the thread directly above. It now uses the shared `Avatar`, so the photo and identity colour match the rest of the app. `senderId` is plumbed through the draft view to resolve the real profile picture.

An unresolved sender falls back to a neutral initial rather than a colour — `Avatar` given an unknown id renders a colour hashed from that id, and given none renders the *current* user, both of which are confidently wrong identities.

image
image
image
image

### Also included

- **Tray no longer auto-expands** when a draft arrives (`useThreadAssist`). The collapsed bar already announces "N AI replies ready", and auto-expanding shoved the composer down mid-read. Behaviour change, worth a look.
- `AskAIDebugPanel` gains a `data-debug-json-modal` marker so its maximised JSON viewer — portalled to `<body>`, i.e. a sibling of the popover — is not treated as a click-outside.
- `apps/backend/scripts/seed-thread-recap.ts`: dev-only seed script for the thread recap. Not wired into any package script; no runtime impact.

## Areas Touched

- [x] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

## Schema & Data Changes

## Config, Environment & URLs

## API Contract

- [x] No API changes

---

## How did you test

Manually in the dashboard: opened the reply tray on a thread with a twin draft, toggled the **Why?** popover (Escape / click-outside / re-click), switched between the Reasoning and Debug tabs, and confirmed the thread stays interactive behind the panel. Verified the source avatar renders the sender's real profile picture and falls back to a neutral initial for an unresolved sender.


apps/backend/scripts/seed-thread-recap.ts
```diff
@@ -0,0 +1,567 @@
+#!/usr/bin/env tsx
```
harsh-1923: can we remove this?

harsh-1923 (approved):

mamtha-v (approved):

PR #906, opened 2026-08-22
fix: keep the recordings list on the tab you left it on

harshpreet931:
POT - https://github.com/user-attachments/assets/073cf7d6-dfa8-47ce-9106-9930befc6c5a

## Problem

Open a recording from the **Shared with me** tab, come back, and the list is on **Created by me**.

`recordings` and `recordings/:recordingId` are sibling routes in `AppRoot.tsx`, not nested — so opening a recording unmounts the list component entirely. `activeListTab` was plain `useState<RecordingOwnershipTab>('created')`, so it wasn't being reset, it was being destroyed and rebuilt from its default.

## Fix

**The tab lives in the URL.** `activeListTab` is derived from a `tab` search param, written through a `setActiveListTab` helper with `replace: true` so toggling doesn't stack history entries. This matches the dominant convention in the dashboard (`useSearchParams` + `replace: true`), and this screen already used `useSearchParams` for the templates modal.

**The detail view remembers where it came from.** The URL alone isn't enough here, because none of the ways out of the detail view pop history — they all navigate to a bare `/recordings`:

- the breadcrumb in `RecordingDetailV2Header`
- `handleMinimize`
- the error-state back button

Both navigation handlers now carry `from: <pathname><search>` in navigation state alongside the existing `recordingIds`, and all three exits route through it, falling back to `/recordings` when a recording is opened directly (deep link, notification). This mirrors the existing `returnTo` / `returnToUrl` pattern used by `TicketDetails` and `ClawAgentDetailV2`.

Prev/next paging already forwards the whole navigation state object, so the origin survives paging between recordings with j/k.

## Notes

- `handleCreatorChange` also switches tab when you pick a person in the People filter; it goes through the same URL-backed setter, so that behaviour is unchanged and now also survives the round trip.
- v1 `RecordingsScreen` has no ownership tabs, so this is v2-only.
- Filters (`selectedCreatorId`, `selectedDatePreset`, `selectedLabels`, `selectedSharerIds`), scroll position and the pagination cursor are still reset by the same unmount. Left alone deliberately — `handleTabChange` already clears those filters on every tab switch, so they're tab-scoped and ephemeral by design. Restoring them is a larger change (array params, plus Virtuoso `initialTopMostItemIndex` wiring that has no restore path today).

## Testing

`tsc --noEmit`, `eslint` (0 errors) and `prettier --check` pass for the three changed files.

### DIFF
--- a/apps/dashboard/src/routes/RecordingDetailV2Screen/RecordingDetailV2Screen.tsx
+++ b/apps/dashboard/src/routes/RecordingDetailV2Screen/RecordingDetailV2Screen.tsx
@@ -95,6 +95,7 @@ import { useSummaryTemplates } from '../../hooks/useSummaryTemplates';
 
 interface RecordingNavState {
   recordingIds?: string[];
+  from?: string;
 }
 
 const AUDIO_POLL_INTERVAL_MS = 10_000;
@@ -319,6 +320,7 @@ export default function RecordingDetailV2Screen(): ReactElement {
   // j/k keyboard navigation between recordings
   const navState = location.state as RecordingNavState | null;
   const recordingIds = navState?.recordingIds;
+  const backTo = navState?.from ?? '/recordings';
   const currentIndex = useMemo(
     () => (recordingId ? (recordingIds?.indexOf(recordingId) ?? -1) : -1),
     [recordingId, recordingIds],
@@ -537,8 +539,8 @@ export default function RecordingDetailV2Screen(): ReactElement {
 
   const handleMinimize = useCallback((): void => {
     sendRecordingEvent({ type: 'setTranscriptMinimized', isMinimized: false });
-    void navigate('/recordings');
-  }, [navigate]);
+    void navigate(backTo);
+  }, [backTo, navigate]);
 
   /**
    * The panels differ wildly in height — a long transcript against a short notes
@@ -775,7 +777,7 @@ export default function RecordingDetailV2Screen(): ReactElement {
       <RecordingLoadError
         failure={resolvedFailure}
         viewerEmail={currentUser?.email}
-        onBack={() => void navigate('/recordings')}
+        onBack={() => void navigate(backTo)}
       />
     );
   }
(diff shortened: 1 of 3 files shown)


harshpreet931:
## Proof of testing

Recorded against the running app on `localhost:5173` (seeded workspace, real sign-in), driven with Playwright and composed with Remotion. Same script run twice — once with the list file reverted to its pre-fix state, once with the fix — so the two clips are the identical flow.

**Flow:** Recordings → *Shared with me* → open a recording → breadcrumb back.

| | URL after going back | `show_shared_with_me` `aria-pressed` |
|---|---|---|
| before | `/recordings` | `false` |
| after | `/recordings?tab=shared` | `true` |

Captured run output:

```
before:
  clicking Shared with me
  url after tab switch: /cmt2e9wpf.../recordings          <- no tab param
  opening first shared recording
  detail url:  /cmt2e9wpf.../recordings/seed-rec-0106
  clicking Recordings breadcrumb
  back url:    /cmt2e9wpf.../recordings
  shared tab aria-pressed: false
  RESULT: FAIL - tab not preserved

after:
  clicking Shared with me
  url after tab switch: /cmt2e9wpf.../recordings?tab=shared
  opening first shared recording
  detail url:  /cmt2e9wpf.../recordings/seed-rec-0106
  clicking Recordings breadcrumb
  back url:    /cmt2e9wpf.../recordings?tab=shared
  shared tab aria-pressed: true
  RESULT: PASS - returned to Shared with me
```

The "before" clip runs against the same commit with only `RecordingsV2Screen.tsx` reverted, so the detail screen's `backTo` falls through to its `/recordings` default — which is exactly the old behaviour.

Note the breadcrumb is what's exercised here, not browser back: it navigates to a bare path rather than popping history, which is why the fix needs both the `tab` param *and* the `from` value carried in navigation state.

samitbarai (approved):

Mayankbansal-juspay (approved):

PR #908, opened 2026-08-22
fix: XYNE-56726 tear SDLC hub channels back down to ordinary channels

ameer030:
## What

Adds `POST /api/sdlc/cleanup/channels/:channelId/teardown`, a one-off endpoint that converts an SDLC hub channel back into an ordinary channel and deletes the SDLC bookkeeping behind it.

## Why

Production created its SDLC hubs before ca3b73606 landed, so those channels carry the pre-fix shape:

| | pre-fix (what prod has) | post-fix |
|---|---|---|
| `name` | `` `<Repo> · SDLC` `` — U+00B7, uppercase, spaces | `normalizeChannelName(repo)` |
| `type` | `DEFAULT` | `SDLC` |
| `metadata` | `{ surface, hiddenFromChat: true, repoId }` | `{ surface, repoId }` |

Those channels are stuck three ways: the name fails `validateChannelName` so it can never be corrected through the rename dialog, `hiddenFromChat` keeps them out of the chat directory so there is no route to them in the UI at all, and `archiveChannel` only accepts `type: DEFAULT`.

## What it deletes, and what it keeps

**Deleted** — the `repos` row, its `sdlc_entity_links`, its `SDLC_SETUP`/`SDLC_WORK`/`SDLC_WIKI` workflows and executions, the knowledge documents its baselines produced, and the repo's queued jobs plus admission permits.

**Kept** — every canvas, folder, conversation, message and ticket. SDLC metadata is cleared from the channel and from each canvas carrying `surface: 'SDLC'`. Because nothing with content is destroyed, there is no Vespa or Y-Sweet reconciliation to do.

Archiving is deliberately left to the `archiveChannel` mutator so the channel still gets its system message.

## Notes for review

- **No cascades exist.** `relationMode = "prisma"` and `20260227120000_drop_all_foreign_keys` removed the rest, so every child row is deleted explicitly or through a Prisma model call. `knowledge_documents` declares no relation to workflows and is deleted by hand.
- **Canvases are selected by `metadata->>'surface'`, not `artifactKind`** — wiki pages carry `documentKind` instead, and `artifactKind` alone silently misses every one of them.
- **`repos` rows are deleted, not `channelId`-nulled** — the repos ACL reads a null `channelId` as "legacy IDE repository" and would expose the row workspace-wide.
- **Runs under `runAsSystem`.** The ACL extension scopes channel reads to participants, so a hub the caller is not a member of would read as missing and the deletes would match no rows. Authorization is done by hand instead; every query carries an explicit `workspaceId`.
- **Auth**: channel `ADMIN`, or `ADMIN` on `LISTPROJECTS`. The second is needed in practice — a pre-fix hub has exactly one participant and no way for anyone else to join.
- **`dryRun` defaults to `true`** and the returned plan is exactly what an apply performs, so the preview cannot drift from the write. The full restore payload is logged *before* the transaction.
- **Refuses while a run holds an admission permit, and `force` does not override that.** The bull job completes the moment the run is handed to Claw, so an in-flight run has no active job and the permit is the only liveness signal. An unreachable queue fails closed with a 503.
- Transaction takes an explicit 60s timeout; Prisma's 5s default can roll a hub with a long run history back mid-cascade.

## Removal

Everything is in `src/sdlc/cleanup/` and mounted with one line in `routes/sdlc.ts`. Delete both once no channel carries `metadata->>'surface' = 'SDLC'`.

## Testing

Driven against a local Postgres through the real HTTP endpoint — 27 assertions, all passing:

- dry run returns the correct plan **and leaves the DB byte-identical**
- apply deletes the bookkeeping; 5 canvases, 5 folders, conversation, message and ticket survive; the 4 SDLC canvases strip to `NULL` while a plain canvas keeps its metadata
- re-running returns 400 rather than half-applying
- live permit + `force: true` → 409; pending execution without force → 409; non-member → 403; unknown channel → 404; non-SDLC channel → 400 and never converted
- `Acme Repo · SDLC` → `acme-repo`, an explicit name normalises, and the collision suffix (`acme-repo-2`) fires

Not covered: real wiki canvases with Y-Sweet documents, and archiving through the UI.

  - backend

### DIFF
--- a/apps/backend/src/sdlc/cleanup/hubTeardown.ts
+++ b/apps/backend/src/sdlc/cleanup/hubTeardown.ts
@@ -1,0 +1,522 @@
+import { Prisma, PrismaClient } from '@prisma/client';
+import { z } from 'zod';
+import {
+  AccessType,
+  CHANNEL_NAME_MAX_LENGTH,
+  ChannelRole,
+  ChannelType,
+  normalizeChannelName,
+  validateChannelName,
+} from '@xyne/shared';
+import { ChannelRepository } from '@/database/repositories/channelRepository';
+import { DatabaseClient } from '@/database/client';
+import { runAsSystem } from '@/database/tenant/context';
+import { AppError } from '@/middleware/errorHandler';
+import { sdlcAdmission } from '@/queues/sdlcAdmission';
+import { sdlcQueue } from '@/queues/sdlcQueue';
+import { logger } from '@/utils/logger';
+import type { SdlcActor } from '../types';
+
+const TAG = '[SDLC-TEARDOWN]';
+const SDLC_WORKFLOW_TYPES = ['SDLC_SETUP', 'SDLC_WORK', 'SDLC_WIKI'] as const;
+const ACTIVE_EXECUTION_STATUSES = ['NEW', 'PENDING', 'RUNNING', 'SCHEDULED'];
+const channelRepository = new ChannelRepository();
+
+export const sdlcHubTeardownSchema = z.object({
+  dryRun: z.boolean().default(true),
+  name: z.string().trim().min(1).max(120).optional(),
+  /**
+   * Tear down despite a PENDING/SCHEDULED execution. Never bypasses a live run:
+   * an execution still holding an admission permit is refused either way.
+   */
+  force: z.boolean().default(false),
+});
+export type SdlcHubTeardownInput = z.infer<typeof sdlcHubTeardownSchema>;
+
+/** The bull job id sdlcQueue minted for a run: `wiki:<executionId>`. */
+const jobIdFor = (workflowType: string, executionId: string): string =>
+  `${workflowType.replace('SDLC_', '').toLowerCase()}:${executionId}`;
+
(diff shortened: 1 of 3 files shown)


apps/backend/src/sdlc/cleanup/hubTeardown.ts
```diff
@@ -0,0 +1,522 @@
...
+      await tx.workflow.deleteMany({ where: { id: { in: workflowIds } } });
+    }
+    // Executions whose workflow was already gone miss the cascade above.
+    if (executionIds.length > 0) {
+      await tx.workflowExecution.deleteMany({ where: { id: { in: executionIds } } });
+    }
+    if (canvasIds.length > 0) {
+      await tx.canvas.updateMany({
+        where: { id: { in: canvasIds } },
+        data: { metadata: Prisma.DbNull },
```
utkarsh-611, line 386: no need of touching metadata at most only remove your stuff, I dont know what esle we put in this metadat and if we remove everythign what will happen

PR #914, opened 2026-08-23
fix: switch full-screen attachment viewer when clicking another video in the side thread panel

XyneSpaces:
[POT](https://drive.google.com/file/d/1UuRZQibvX3UNJ32VEropeluREIN6ekNu/view?usp=sharing)
## Summary
When multiple attachments exist in a thread and the full-screen attachment viewer is already open with the side thread panel visible, clicking another video (or text/code file) in the side panel did nothing — the viewer did not switch to the clicked attachment.

Root cause: the attachment viewer XState machine (`attachmentViewerMachine.ts`) only handles the `OPEN` event from the `closed` state. Three inline call sites in `MessageAttachment.tsx` (video, large-text file, code file) unconditionally dispatched `OPEN`, so when the viewer was already open the event was dropped and the click was a no-op. Switching only worked for the gallery card path, which already used the `isOpen ? 'UPDATE' : 'OPEN'` pattern.

## Fix
Guard all three inline call sites to send `UPDATE` when the viewer is already open and `OPEN` when it is closed (mirroring the existing gallery pattern). For the video path, also start playback on switch (`viewerOpen`) so clicking a second video switches the viewer and plays it, which is the reported expectation.

Only `apps/dashboard/src/components/Chat/MessageAttachment/MessageAttachment.tsx` changes (+25 / −4). No schema, env, or unrelated viewer behavior changes.

## Testing
Reproduced and verified in a local sandbox (dashboard + backend in test mode, seeded thread with 1 image + 2 videos) using a standalone Playwright script:
- Video 1 → Video 2: viewer title switches `one.mp4 → two.mp4`, `<video>` src switches to the second attachment and plays (`paused:false`).
- Image → Video 2: viewer switches `image0.jpg → two.mp4` and plays.

## Checklist (repo template)
1. Problem Description: Clicking a second video in the side thread panel while the full-screen viewer is open did not switch the viewer.
2. How the issue was identified: Live repro in sandbox + tracing the XState machine — `OPEN` only handled in `closed` state, inline call sites always sent `OPEN`.
3. Solution: Send `UPDATE` when viewer already open (else `OPEN`) at all three inline `MessageAttachment` call sites; autoplay on video switch.
4. Documentation: N/A
5. DB Alter Details: N/A
6. Data fix Details: N/A
7. Environment variable Changes: N/A
8. Testing: Playwright end-to-end in sandbox; both Video→Video and Image→Video switches confirmed with playback. Screen recording attached in the originating thread.
9. Services to which this change is to be deployed: dashboard (frontend only)
10. Main PR link (if against a release branch): N/A


apps/dashboard/src/components/Chat/MessageAttachment/MessageAttachment.tsx
```diff
@@ -1046,7 +1166,22 @@ const InlineVideoPlayer: React.FC<{
...
+            <div
+              className='relative h-full cursor-pointer'
+              role='button'
+              tabIndex={0}
+              aria-label={`Open ${fileName} preview`}
+              data-track-category='MESSAGE'
+              data-track-name='OPEN_ATTACHMENT_PREVIEW'
+              data-track-metadata={JSON.stringify({ fileName, attachmentId })}
+              onClick={() => openModal({ startPlayback: true })}
+              onKeyDown={e => {
```
harshpreet931, line 1127: no target check here

apps/dashboard/src/components/Chat/MessageAttachment/MessageAttachment.tsx
```diff
@@ -903,20 +1008,35 @@ const InlineVideoPlayer: React.FC<{
     // Capture current video time before opening modal
     const currentTime = videoRef.current?.currentTime;
     const isPlayingInline = videoRef.current ? !videoRef.current.paused : false;
+    // Autoplay when switching into an already-open viewer (reported expectation).
+    const viewerOpen = attachmentViewerActor.getSnapshot().value !== 'closed';
```
harshpreet931: viewerOpen checks if anything is open not "a" video specifically

apps/dashboard/src/components/Chat/MessageAttachment/MessageAttachment.tsx
harshpreet931, line 1265: basically the same function duplicated as the one at the top line 86 and this one's missing the empty-array fallback the new one has. so we can just call the new one now here too

apps/dashboard/src/components/Chat/MessageAttachment/MessageAttachment.tsx
```diff
@@ -76,6 +76,81 @@ interface MessageAttachmentProps {
...
 }
 
+type ViewerRawAttachments = QueryResultType<
+  typeof queries.conversationMessagesV2
+>[number]['attachments'];
+
+// Open or switch the attachment viewer to `targetId`, carrying the full sibling
+// set so the in-viewer next/prev navigation keeps working. UPDATE when the
+// viewer is already open (incl. error state), OPEN when it is closed.
+const openOrSwitchViewer = (params: {
```
harshpreet931: this function should be in utils

apps/dashboard/src/components/Chat/MessageAttachment/MessageAttachment.tsx
```diff
@@ -76,6 +76,81 @@ interface MessageAttachmentProps {
   parentMessage?: AttachmentRef['parentMessage'];
 }
 
+type ViewerRawAttachments = QueryResultType<
```
harshpreet931: same type already inlined a few lines up on the interface

harshpreet931 (approved):

devesh-juspay (approved):

PR #920, opened 2026-08-23
fix: preserve attachment order with explicit position column

XyneSpaces:
[POT](https://drive.google.com/file/d/15tfM_3yV-i5gjPBl1jwpsLGaM80tAPiw/view?usp=sharing)
## What & why

Attachment display order was reconstructed at render time from `createdAt` (with `id` as a tiebreak). The message **forward** path cloned every attachment with an identical `createdAt` (`Date.now()`), so the `id` tiebreak decided the order and forwarded galleries rendered in a **different order than the source message** — e.g. source `1, 2, 3` came out as `2, 1, 3`, mismatching the message caption. This is a user-visible data-integrity bug (verified end-to-end against the DB and the rendered UI).

Root cause: order was an *emergent* property of a non-unique sort key, not a *stored* one.

## Fix

Make attachment order an explicit, stored property (the approach Mattermost/Rocket.Chat/Slack/Signal all use — order lives on the message, it is never re-derived from a timestamp):

- **Schema**: add nullable `MessageAttachment.position` (Prisma + Zero schema) and index `message_attachments(entityId, position)`.
- **Migration**: `20260823120000_add_message_attachment_position` adds the column and **backfills** existing rows from the currently displayed order (`createdAt ASC, id ASC`), so there is **no visible change to existing data**.
- **Writes**: stamp `position` on every insert path — send/upload (backend + shared), forward-clone (backend), and call-message clones (backend + shared). The forward-clone now sorts the source rows by the same comparator the UI uses **before** assigning positions, so the copy mirrors what the sender saw.
- **Render**: `MessageBubble` now orders by `position` first, falling back to `createdAt`/`id` for legacy rows (backward compatible).

## Testing

- `@xyne/shared`, `apps/backend`, `apps/dashboard` all typecheck clean (`tsc --noEmit`).
- Migration applied to a local DB; `position` column + index confirmed present.

## Rollout notes

- Backward compatible: `position` is nullable and the render sort falls back to the old comparator for any null-position rows, so old and new clients interoperate during rollout.
- The migration is additive (add column + backfill + index); no data is dropped.


apps/backend/prisma/migrations/20260823120000_add_message_attachment_position/migration.sql
```diff
@@ -0,0 +1,26 @@
+-- Add an explicit display-order column for message attachments.
+-- Ordering was previously reconstructed from `createdAt` (with `id` as a tiebreak),
+-- which collides when multiple attachments share a timestamp (e.g. forward-clones
+-- inserted with the same `now`), producing a different order than the source message.
+-- AlterTable
+ALTER TABLE "public"."message_attachments" ADD COLUMN "position" INTEGER;
+
+-- Backfill existing rows: freeze the currently displayed order (createdAt asc, id asc)
+-- as an explicit per-message position so ordering stays stable going forward.
+WITH ordered AS (
```
santhoshJus: no need to backfill

santhoshJus (approved):

Mohan0544 (approved):

PR #930, opened 2026-08-24
feat: warn before a workspace switch or refresh stops an active recording

harshpreet931:
POT -  https://github.com/user-attachments/assets/2429b80c-e229-4ae2-9d92-ff1a7a7b1557

## Problem

Switching workspace and refreshing both tear down the page. That destroys the in-memory LiveKit `Room` held in `recordingStore`, so an in-progress recording ends with no warning at all. On the backend, `noteTakerCallRepository.handleParticipantLeave` ends the call the instant the participant disconnects — there is no grace period — so the recording is genuinely over, not merely interrupted.

None of these paths go through the React Router history stack (`window.location.href` for workspace switch, main-process `loadApp()` for Electron reload), so `useBlocker` would not guard any of them.

## What this does

Adds a confirmation guard on the **user-initiated** paths:

| Path | Desktop | Web |
| --- | --- | --- |
| Workspace switch (sidebar switcher, workspace create, cross-workspace notification click, SOS alert banner) | modal | modal |
| `Cmd/Ctrl+R`, `Cmd+Shift+R`, `F5` | native dialog | modal |
| Reload button, tab close, address bar | n/a | browser `beforeunload` prompt |

Electron's reload is confirmed with `dialog.showMessageBox` in the main process because `before-input-event` intercepts the keystroke before the renderer ever sees it, then asks the renderer to stop cleanly over a new `recording:stop-for-teardown` IPC channel and waits for confirmation before navigating. Web intercepts the keystroke in the renderer so browser users get the same modal; `beforeunload` stays as the backstop for the paths no keystroke handler can catch.

## Also fixed: a false "could not save this recording" toast

livekit-client defaults to `disconnectOnPageLeave: true`, so it disconnects the room itself on `beforeunload` — before the app's own `pagehide` teardown gets a chance to mark the disconnect intentional. `recordingStore` then reported an ordinary navigation as a server-side save failure, even though the backend finalizes the recording correctly on any dropped connection (`participant_left` → call ended → egress stopped → stitched → uploaded). Added a page-unload flag with a 5s grace window that suppresses both failure toasts while a real navigation is in flight, without changing any disconnect/network behavior.

## Notes

- Stopping now awaits the room disconnect (3s cap) before navigating, so the recording finalizes cleanly instead of going through the backend's crash-reconciliation path and yielding a truncated transcript.
- No pause option in the modal. Pausing would only cancel the navigation, leaving the user to reopen the switcher and stop anyway — "Keep recording" already covers "I changed my mind," and the recording overlay/tray already has its own pause control for that.
- The guard short-circuits to a no-op when nothing is recording.

## Not covered

Automatic reloads still end recordings silently and are out of scope here — most notably `ui-updater.ts`, which applies a staged UI update on window **blur**, so alt-tabbing during a recording can reload the app.

## Testing

`tsc --noEmit`, `eslint` (0 errors) and `prettier --check` pass for `apps/dashboard` and `apps/electron`.

Proof-of-testing video against the real running app (real login, real LiveKit recording) to follow as a comment.

### DIFF
--- a/apps/dashboard/src/App.tsx
+++ b/apps/dashboard/src/App.tsx
@@ -24,6 +24,7 @@ import { initializeTelemetry } from './services/otel/init';
 import { KeyboardProvider } from './contexts/KeyboardContext';
 import { SwitchLoadingOverlay } from './components/SwitchLoadingOverlay/SwitchLoadingOverlay';
 import { RecordingInterruptGuard } from './components/Recording/RecordingInterruptGuard/RecordingInterruptGuard';
+import { WorkspaceSwitchToastListener } from './components/WorkspaceSwitchToastListener';
 import { TRUSTED_ORIGINS } from '@xyne/shared';
 import { DEFAULT_WORKSPACE_ID } from './config';
 import {
@@ -112,6 +113,7 @@ const App = (): ReactElement => {
                     </main>
                     <SwitchLoadingOverlay />
                     <RecordingInterruptGuard />
+                    <WorkspaceSwitchToastListener />
                     <Toaster
                       position='top-right'
                       richColors
--- a/apps/dashboard/src/App.tsx
+++ b/apps/dashboard/src/App.tsx
@@ -23,6 +23,7 @@ import Wallpaper from './components/Wallpaper/Wallpaper';
 import { initializeTelemetry } from './services/otel/init';
 import { KeyboardProvider } from './contexts/KeyboardContext';
 import { SwitchLoadingOverlay } from './components/SwitchLoadingOverlay/SwitchLoadingOverlay';
+import { RecordingInterruptGuard } from './components/Recording/RecordingInterruptGuard/RecordingInterruptGuard';
 import { TRUSTED_ORIGINS } from '@xyne/shared';
 import { DEFAULT_WORKSPACE_ID } from './config';
 import {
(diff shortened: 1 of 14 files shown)


harshpreet931:
## Proof of testing

Recorded against the running app on `localhost:5173` with real sign-in, a real LiveKit room, and real egress (`xyne-spaces-test-livekit` + `-livekit-egress` containers) — driven with Playwright, composed with Remotion. Two runs of the same flow: once with `WorkspaceSwitcher.handleSwitch`'s guard call temporarily reverted (bug repro), once against this branch (fix).

**Flow:** Recordings → Start recording → open workspace switcher → switch to the other workspace mid-recording.

**Before (no guard):** click the other workspace → immediate, silent navigation. The recording is gone with zero warning — this is the exact bug the PR fixes.

**After (this branch):**
1. Click the other workspace → confirm modal appears: *"Switching workspace will stop your recording"* with **Keep recording** / **Stop and switch**.
2. **Keep recording** → modal closes, URL unchanged, recording keeps running.
3. Reopen the switcher, switch again → modal appears again → **Stop and switch** → recording stops, a "Recording stopped" toast fires, and the app lands on the target workspace.

The "before" clip was captured by removing just the `confirmRecordingInterrupt('workspaceSwitch')` call from `handleSwitch` on the same commit — restored immediately after capture, verified byte-identical to the committed file afterward.

Note: this only demonstrates the workspace-switch path, since that's scriptable end-to-end through the browser. The Electron-side Cmd+R dialog and the web keystroke intercept aren't exercised here — same limitation as noted on #860, they need a manual pass.

samitbarai (dismissed):

samitbarai (approved):

5uryan5h (approved):

Mayankbansal-juspay (approved):

PR #944, opened 2026-08-24
feat: add encryption key rotation backfill

siddhantagrawal-ops:
  - backend

## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Backfill defaults to dry-run.
--apply requires explicit active-key confirmation.
Updates use compare-and-swap.
The command is idempotent.
Local canary and second idempotency run passed.
Key retirement is a later operational configuration step, not automatic deletion by this PR.

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added
- [x] Backfill / data migration included

If any of the above:

- [x] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [x] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [x] Empty state

---

## Checklist

- [x] No debug logging or commented-out code left behind\n\n## Backfill scope

This command rotates only AES-256-CBC ciphertext written by the Spaces `encryptionService` in `iv:ciphertext` or `v2:keyId:iv:ciphertext` format. AES-256-GCM credential envelopes and Claw-auth MCP gateway ciphertext are outside this PR's scope. A successful run therefore does not, by itself, make the legacy `ENCRYPTION_KEY` safe to remove.\n

### DIFF
--- a/apps/backend/src/migration/scripts/backfillEncryptionKeyRotation.ts
+++ b/apps/backend/src/migration/scripts/backfillEncryptionKeyRotation.ts
@@ -1,0 +1,749 @@
+import { PrismaClient } from '@prisma/client';
+import {
+  rotateCiphertext,
+  rotateEncryptedJsonStrings,
+  type CiphertextRotationResult,
+  type JsonRotationStats,
+} from '../../services/encryptionRotation';
+import { decrypt, encrypt, getActiveEncryptionKeyId } from '../../services/encryptionService';
+
+interface Options {
+  apply: boolean;
+  batchSize: number;
+  confirmedKeyId: string | null;
+  targets: Set<string> | null;
+  help: boolean;
+  listTargets: boolean;
+}
+
+interface ScalarRow {
+  id: string;
+  value: string;
+}
+
+interface TargetStats {
+  rowsScanned: number;
+  valuesScanned: number;
+  legacy: number;
+  active: number;
+  otherKey: number;
+  malformed: number;
+  failed: number;
+  wouldRotate: number;
+  rotated: number;
+  updated: number;
+  casSkipped: number;
+  updateFailed: number;
+  invalidJson: number;
+  blockedRows: number;
+  byKeyId: Record<string, number>;
+}
+
+interface ScalarTarget {
+  name: string;
+  fetch: (afterId: string | null, batchSize: number) => Promise<ScalarRow[]>;
+  update: (id: string, originalValue: string, rotatedValue: string) => Promise<number>;
+}
+
+const DEFAULT_BATCH_SIZE = 100;
+const MAX_BATCH_SIZE = 1000;
+const WORKFLOW_TARGET = 'workflows.context';
+
+const TARGET_NAMES = [
+  'workflow-mappings.entitySecret',
+  'org-llm-credentials.credentials',
+  'user-external-tokens.encryptedToken',
+  'user-external-tokens.refreshToken',
+  'external-sources.credentials',
+  'apps.signingSecret',
+  'installed-apps.signingSecret',
+  'incoming-webhooks.secret',
+  'data-sources.credentials',
+  WORKFLOW_TARGET,
+] as const;
+
+function writeStdout(...parts: unknown[]): void {
+  process.stdout.write(parts.map(String).join(' ') + '\n');
+}
+
+function writeStderr(...parts: unknown[]): void {
+  process.stderr.write(parts.map(String).join(' ') + '\n');
+}
+
+function printHelp(): void {
+  writeStdout(`
+Encryption key-rotation backfill
+
+Dry-run is the default:
+
+  pnpm run encryption:backfill
+
+Apply changes:
+
+  pnpm run encryption:backfill -- \\
+    --apply \\
+    --confirm-active-key=<keyId>
+
+Options:
+
+  --apply
+      Write re-encrypted values. Without this flag, no database rows change.
+
+  --confirm-active-key=<keyId>
+      Required with --apply. Must exactly match the final key ID in
+      the ordered ENCRYPTION_KEYS array.
+
+  --batch-size=<1-${MAX_BATCH_SIZE}>
+      Number of rows fetched per batch. Default: ${DEFAULT_BATCH_SIZE}.
+
+  --target=<target>
+      Process one target. May be supplied more than once.
+
+  --list-targets
+      Print available target names.
+
+  --help
+      Print this help.
+`);
+}
+
+function parseOptions(argv: string[]): Options {
+  const options: Options = {
+    apply: false,
+    batchSize: DEFAULT_BATCH_SIZE,
+    confirmedKeyId: null,
+    targets: null,
+    help: false,
+    listTargets: false,
+  };
+
+  const selectedTargets = new Set<string>();
+
+  for (const argument of argv) {
+    // pnpm may forward its argument separator to the script.
+    if (argument === '--') {
+      continue;
+    }
+
+    if (argument === '--apply') {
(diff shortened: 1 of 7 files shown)


apps/backend/src/services/encryptionRotation.ts
```diff
@@ -0,0 +1,247 @@
...
+    if (!rotated.startsWith(expectedPrefix)) {
+      throw new Error(`encrypt() did not write with active key "${normalizedActiveKeyId}"`);
+    }
+
+    return {
+      value: rotated,
+      classification,
+      outcome: 'rotated',
+    };
+  } catch {
```
devesh-juspay: This discards the reason the value could not be rotated, and `outcome: 'failed'` conflates three cases that need different responses:

- the row is encrypted under a key id that is not in this process's ring
- the value is genuinely corrupt
- the string matched `classifyCiphertext` but is not actually our ciphertext

That matters more here than it usually would, because the workflow loop treats any failure as a reason to block the whole row (`result.stats.failed > 0 → blockedRows`). An operator seeing a non-zero `blockedRows` count currently has no way to tell whether they are missing a key from `ENCRYPTION_KEYS` — recoverable, re-run after fixing the ring — or looking at damaged data, without going back and decrypting by hand.

Capturing `err instanceof Error ? err.message : String(err)` onto `CiphertextRotationResult` and surfacing a small sample in the summary would make the failure actionable. Worth doing the same for the `catch` around `prisma.workflow.updateMany` (`updateFailed`), for the same reason.

apps/backend/src/migration/scripts/backfillEncryptionKeyRotation.ts
```diff
@@ -0,0 +1,749 @@
...
+  name: string;
+  fetch: (afterId: string | null, batchSize: number) => Promise<ScalarRow[]>;
+  update: (id: string, originalValue: string, rotatedValue: string) => Promise<number>;
+}
+
+const DEFAULT_BATCH_SIZE = 100;
+const MAX_BATCH_SIZE = 1000;
+const WORKFLOW_TARGET = 'workflows.context';
+
+const TARGET_NAMES = [
```
devesh-juspay, line 65: The target list covers the `encryptionService` AES-256-CBC surface thoroughly, and matches the `encrypt()` call sites across the backend.

Worth stating explicitly in the PR description that it is **CBC-only**, because `ENCRYPTION_KEY` has more than one consumer:

- `src/sdlc/vcs/credentialEnvelope.ts` reads the same `ENCRYPTION_KEY` but writes AES-256-GCM with its own envelope (`encryptionVersion`, `keyVersion: 'workspace-v1'`, AAD bound to workspace + provider). Its ciphertext is not in the `iv:ct` or `v2:keyId:iv:ct` shapes this script classifies, so those rows are invisible to it.
- claw-auth's MCP gateway (`mcpgateway/crypto`) is a third format again, keyed on `BACKEND_CLIENT_SECRET_ENCRYPTION_KEY` with a fallback to `ENCRYPTION_KEY`.

The consequence is the part worth writing down: **a clean run of this backfill does not make `ENCRYPTION_KEY` removable.** The SDLC envelope path still needs it. That is fine under the current decision to cut over without retiring the old key, but "backfill completed" naturally reads as "the old key can be dropped", and here it cannot.

A line in `--help` and in the summary output — something like *"covers AES-256-CBC values written by encryptionService; GCM envelopes are out of scope"* — would stop that being rediscovered later.

devesh-juspay:
**Branch is behind #764.** This branch contains `1d4ace98e` but not `a75210ccb` ("address encryption key rotation review feedback"), so it is missing:

- the lazy key-ring loading in claw-auth, which stops a malformed `SPACES_ENCRYPTION_KEY` from taking the whole service down at startup
- the `ENCRYPTION_KEYS` entry in `.env.example`
- the `ENCRYPTION_KEY` / `ENCRYPTION_KEYS` declarations in the Joi schema

Needs a rebase before merge, and #764 has to land first since this stacks on it.

---

Otherwise this is well built, and it is the piece that makes the rotation worth doing. Three things worth calling out as deliberately good rather than incidental:

**Compare-and-swap on every write.** `updateMany({ where: { id, <column>: originalValue } })` means a row the application rewrote between read and write is skipped (`casSkipped`) rather than clobbered with staler ciphertext. That is what makes this safe to run against a live system.

**The preflight actually proves the configuration.** Requiring `--confirm-active-key` to match the ring's final entry, then encrypting a probe value and asserting the `v2:<id>:` prefix, catches a mismatch between what the operator believes is active and what the process will really write. Given that the active writer is implied by array position rather than named explicitly, that check is doing real work.

**All-or-nothing per JSON row.** Blocking a workflow whose context has any failed or malformed value, instead of writing back a partially rotated config, is the right call — a half-rotated automation would be considerably worse than an unrotated one.

The strict classification is also worth keeping: requiring a 32-hex IV and a whole number of AES blocks means an arbitrary `foo:bar` string inside a config will not be mistaken for ciphertext while walking JSON.

apps/backend/src/services/encryptionRotation.ts
siddhantagrawal-ops: Addressed. `rotateCiphertext()` now preserves the safe error message when rotation fails instead of returning only `outcome: "failed"`.

The backfill summary now includes bounded failure samples with the affected row ID and reason, which makes missing-key failures distinguishable from decryption/corruption failures. Database update exceptions, including the `prisma.workflow.updateMany()` path, are also captured separately as update-failure samples.

Samples are capped and do not include plaintext, ciphertext, workflow JSON, or key material. Added focused tests for missing-key error propagation and nested workflow JSON failure reporting; the targeted backend tests, typecheck, and build pass.

apps/backend/src/migration/scripts/backfillEncryptionKeyRotation.ts
siddhantagrawal-ops: Addressed. The CLI help, runtime start/completion summaries, and PR description now explicitly state that this backfill covers only AES-256-CBC values written by `encryptionService` in `iv:ciphertext` or `v2:keyId:iv:ciphertext` format.

AES-256-GCM credential envelopes and the Claw-auth MCP gateway encryption format remain outside this PR’s scope. The completion summary also explicitly reports that successful completion does not authorize retirement of the legacy `ENCRYPTION_KEY`.

This PR intentionally does not modify or migrate the separate GCM encryption domains.

siddhantagrawal-ops: Addressed. I rebased `feature/encryption-key-rotation-backfill` onto the latest `feature/key-rotation-last-key-wins`, including commit `a75210ccb`.

The backfill branch now includes the lazy Claw-auth key-ring loading, the `ENCRYPTION_KEYS` `.env.example` documentation, and the `ENCRYPTION_KEY` / `ENCRYPTION_KEYS` Joi declarations from PR #764.

This remains a stacked PR and will only be merged after PR #764 lands.

devesh-juspay (dismissed):

PR #974, opened 2026-08-24
feat: XYNE-60922 add Topics Explorer to the desk support screen

Revanthvenkatpasupuleti:
https://github.com/user-attachments/assets/42fbeaa4-329d-4c32-a3b3-fa5e55a824bd

## Type of Change

- [x] New feature

## Description

Adds **Topics Explorer** to the desk support screen — a per-desk panel that breaks ticket volume down by any stack of columns, opened from the desk toolbar (`?topics=open`, alongside `?metrics=open`).

A treemap on the left and one daily-volume trend row per group on the right share order and colour, so the two panels read as a single view. Clicking a tile drills a level deeper, or — at the last level — opens the ticket list filtered to exactly that group.

Group by up to 4 levels from: AI Category, AI Sub-category, Tag, Priority, AI Priority, Stage, Assignee, plus one `AI Tag · <category>` dimension per LLM tag category present in the data. Filters reuse Desk's own submenus, so field lists and value formats match the ticket list exactly.

**Why the shape it has:** Zero has no aggregation, so the rollup runs client-side over synced rows — hence one channel at a time and a bounded created-at window with a 20,000-row cap that is surfaced to the user when hit. LLM tag categories (sentiment, priority) live in `non_zero.tags`, which Zero does not mirror, so they are read through a new date-scoped endpoint rather than the sync.

Branched from `main`; a single commit containing only the Topics Explorer work.

## Areas Touched

- [x] `apps/dashboard`

---

## Zero (Sync) Changes

Adds `topicsExplorerTickets` to both registries (`packages/shared/src/zero/queries.ts` and `apps/backend/src/zero/queries.ts`). No mutator, schema or ACL changes.

- [x] The two registry copies are byte-identical for this query (verified by extracting and diffing both definitions)
- [x] `channelId` + `isMember` are forwarded to `TicketsACL.canSelect`, matching every sibling ticket query — the `isMember: true` branch re-verifies participation via `whereExists`, so a client asserting it gains nothing
- [x] Arg schema carries a `.refine` guard (`createdAtStart <= createdAtEnd`), matching `supportTicketsFilteredV3`

## Schema & Data Changes

## Config, Environment & URLs

- [x] No config changes — skip this section

## API Contract

- [x] New endpoint

`GET /api/channels/:channelId/tags-config/generated-tags-by-conversation?startMs=&endMs=`

Returns LLM tag values per conversation for a channel over a created-at window. Parameterised throughout (no interpolation), same channel-participant ACL as its siblings in the controller, deterministically ordered, and row-capped with a `truncated` flag so the client warns rather than silently under-reporting.

---

## How did you test it?

**Static verification only — I did not run the app.** Please exercise the manual paths below before merging.

Ran and passing:
- `tsc --noEmit` clean in `apps/backend`, `apps/dashboard` and `packages/shared`
- `eslint` — 0 errors on all touched files (one pre-existing error in `deskTagsConfigController.ts:353` is at HEAD and untouched)
- `prettier --check` clean on all new and modified files
- The new SQL executed against `xyne_dev_db`, plus `EXPLAIN` on its plan
- The 8 layout templates verified programmatically: each tiles all 48 grid cells exactly, no overlap, non-increasing areas, max:min ratio 1.0–3.33

### Automated

- [x] Not covered by tests — no test suite was added for this feature

### Manual

- [ ] Nothing manually tested — see above

---

## Known limitations / follow-ups

Deliberately not addressed here, each verified rather than assumed:

1. **Missing composite indexes.** `EXPLAIN` on the new query shows `createdAt` applied as a heap `Filter`, not an `Index Cond` — every email in the channel is read then discarded by date. `emails(channelId, createdAt)` and `tickets(channelId, isArchived, createdAt DESC)` would fix it. Not added here because index DDL on two shared hot tables is a deploy decision.
2. **Sync payload.** The synced query pulls whole ticket rows (Zero cannot project columns). Measured on dev data, the average row is ~3.6 KB of which ~91% — `description`, `classificationData`, `metadata`, `title` — is unused by the rollup. At the 20k cap that is roughly 73 MB average / 129 MB p95 resident client-side. Serving the rollup from a backend aggregate endpoint would remove this entirely.
3. **AI Sub-category cannot drill through.** The ticket list has no filter for that column, so opening from it widens to the parent AI Category and says so. It is not a default group-by for that reason.
4. **`metricsEnabled` gates both panels.** A desk cannot enable Topics without the metrics dashboard. Flagging in case that coupling should be its own preference.
5. **AI tag filter caps at 1,000 conversations** server-side (pre-existing `LIMIT 1000`, no date bound). The panel now warns when it hits the cap, since that whitelist is the denominator of every box.

## Checklist

- [ ] `pnpm run build:shared` passes — not run
- [x] Backend `typecheck` passes — `build` not run
- [x] Dashboard `lint:errors-only` + `typecheck` pass — `build` not run


apps/backend/src/zero/queries.ts
```diff
@@ -1429,6 +1434,30 @@ export const queries: AnyQueryRegistry = defineQueries({
     }
   ),
 
+  // Topics Explorer: tickets for one desk in a created-at window, rolled up
+  // client-side. The row cap keeps that sync bounded; no relations are pulled,
+  // since every grouping reads a column on the ticket itself.
+  // channelId + isMember are forwarded to TicketsACL for membership gating.
+  topicsExplorerTickets: defineQuery(
```
Revanthvenkatpasupuleti: make it generic

apps/backend/src/database/repositories/tagRepository.ts
```diff
@@ -237,6 +240,74 @@ export class TagRepository {
...
+   * so it can bucket tickets by category.
+   */
+  async findGeneratedTagsByConversation(
+    channelId: string,
+    configKey: string,
+    start: Date,
+    end: Date,
+  ): Promise<{ conversationId: string; tagCategory: string; tag: string }[]> {
+    const emails = await this.client().email.findMany({
+      where: { channelId, createdAt: { gte: start, lte: end } },
```
Himanshvarma, line 258: is createdAt indexed can you check this once

apps/backend/src/database/repositories/tagRepository.ts
```diff
@@ -237,6 +240,74 @@ export class TagRepository {
...
+
+    // Chunked: every id is a bind parameter, and Postgres' extended protocol
+    // caps those at 65,535 — one busy desk over a wide window clears that, and
+    // the driver rejects the query rather than returning a short answer.
+    const emailIds = [...conversationByEmailId.keys()];
+    const tags: { sourceId: string; tagCategory: string; tag: string }[] = [];
+    for (let i = 0; i < emailIds.length; i += EMAIL_ID_CHUNK) {
+      const chunk = await this.client().tag.findMany({
+        where: {
+          sourceId: { in: emailIds.slice(i, i + EMAIL_ID_CHUNK) },
```
Himanshvarma: instead of doing in array can we use that map which you have created to know if this sourceId key exists in that map  which will be like O(logn) instead of O(n)

Himanshvarma (commented):
added comments

Navviiii (approved):

knaman2609 (approved):

PR #1001, opened 2026-08-25
feat: XYNE-57204 slack supported ephemeral msg

HarimohanSharma11:
  - backend
  - dashboard

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist

### DIFF
--- a/apps/backend/src/apps/controllers/chatController.ts
+++ b/apps/backend/src/apps/controllers/chatController.ts
@@ -373,7 +373,7 @@ export class ChatController {
       if (!bodyResult.success) {
         res.status(400).json({
           error: `Validation error`,
-          code: 'VALIDATION_ERROR',
+          code: 'VALIDATION__ERROR',
           details: bodyResult.error.errors,
         });
         return;
--- a/apps/backend/src/apps/controllers/chatController.ts
+++ b/apps/backend/src/apps/controllers/chatController.ts
@@ -231,6 +231,24 @@ export class ChatController {
   }
 
   /**
+   * Content for a non-flow message: markdown is stored as-is after sanitizing, plain
+   * text goes through attachment and mention processing. Shared by postMessage and
+   * postEphemeral so the two routes can't drift apart.
+   */
+  private async buildTextContent(
+    fields: Pick<
+      z.infer<typeof PostMessageFieldsSchema>,
+      'text' | 'markdownText' | 'contentFormat' | 'attachments'
+    >,
+    workspaceId: string | undefined,
+  ): Promise<string> {
+    const { text, markdownText, contentFormat, attachments } = fields;
+    if (markdownText) return sanitizeMessageContent(markdownText);
+    if (contentFormat === ContentFormat.MARKDOWN) return sanitizeMessageContent(text || '');
+    return await this.processMessageContent(text, attachments, workspaceId);
+  }
+
+  /**
    * Post a message to a channel or conversation
    * POST /api/external-event/chat/postMessage
    *
@@ -301,12 +319,11 @@ export class ChatController {
         }
         content = built.content;
         isMarkdown = false;
-      } else if (markdownText) {
-        content = sanitizeMessageContent(markdownText);
-      } else if (contentFormat === ContentFormat.MARKDOWN) {
-        content = sanitizeMessageContent(text || '');
       } else {
-          content = await this.processMessageContent(text, attachments, req.user?.workspaceId);
+        content = await this.buildTextContent(
+          { text, markdownText, contentFormat, attachments },
+          req.user?.workspaceId,
+        );
       }
 
       // Messages posted via the internal /api/internal/postAsUser route are authored
@@ -365,6 +382,10 @@ export class ChatController {
    *
    * The message vanishes on reload. A `flow` stays interactive for 30 minutes via
    * ephemeralFlowStore, which is what lets executeAction resolve the owning app.
+   *
+   * Delivery is best-effort and is NOT guaranteed by the 201. Nothing is persisted, so
+   * a recipient with no live socket never receives it and there is nothing to catch up
+   * on when they reconnect. The response reports `delivery: 'best-effort'` to say so.
    */
   postEphemeral = async (req: Request, res: Response): Promise<void> => {
     try {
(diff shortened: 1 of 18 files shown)


Mohan0544 (approved):

Mohan0544 (dismissed):

ameer030:
## Review

**TLDR:** Design is sound on the native path, but the Slack path skips the authz the native path added, Slack ephemeral *flows* can never dispatch, the client listener dies on the first websocket reconnect, and thread panels show every channel ephemeral. Plus an `[EPHEMERAL-FILTER]` debug log left in a per-socket loop.

---

### First, a question worth answering before the rest

We already have "only you can see this": persist the message with `visibleTo`, and `messages-acl.ts:24` filters it out for everyone else. `commandController.ts:56` does exactly that today, and gets flows, threads, avatars, virtualization and reload-safety for free from the normal pipeline.

This PR adds a second, parallel delivery path (redis pub/sub + a bespoke client store + a bespoke row renderer + a redis flow side-store + a websocket filter) to buy one property: *vanishes on reload*. That's ~500 lines and three new failure modes for that one property, and every blocker below is downstream of the choice.

If "vanishes on reload" is a hard product requirement, fair enough — worth stating in the PR description. If it isn't, `visibleTo` + a TTL sweep is a fraction of this.

---

### Blockers

**1. Debug log in a hot loop** — `apps/backend/src/services/websocketService.ts:1351`

`logger.info('[EPHEMERAL-FILTER] …')` fires per socket per message. The filter itself is a correct fix — keep it, drop the log, and fix the 11-space indent on the line above. Worth noting it only guards the *persisted* `visibleTo` path; nothing in this feature routes through `session_activity`.

**2. Client listener is lost on the first reconnect** — `apps/dashboard/src/hooks/useEphemeralMessages.ts:54-60`

`websocketService.on()` is a no-op when `this.socket` is null, and `connect()` does `removeAllListeners()` + recreates the socket (`socketClient.ts:70-76`). `initialized = true` is set once and never reset, so after any reconnect ephemerals stop arriving **permanently and silently**. The pattern elsewhere is a per-mount `useEffect` + `removeListener` (`DynamicDashboardPanel.tsx:86-100`), plus a `'connect'` re-subscribe (`useKanbanCounts.ts:637`).

**3. Every channel ephemeral renders in every thread** — `apps/dashboard/src/components/Chat/ThreadPannel.tsx:293-317`

`useEphemeralMessages(derivedChannelId)` is channel-scoped, then merged into thread messages with no `conversationId` filter — open any thread in the channel and the ephemeral is there too. Mirror image in `ChatListV4.tsx:415-434`: a thread-targeted ephemeral (`conversationId ?? resolvedChannelId`) also renders in the main channel feed. Needs a `conversationId` filter at both call sites.

**4. Slack path drops the authz the native path added** — `apps/backend/src/apps/platform-adapters/slack/controller.ts:443-446`

Native `postEphemeral` checks `channelParticipants.isParticipant` (`chatController.ts:404`). The Slack route does not — `slackChannelValidation` only validates the *bot's* channel access. And `findByMetadataField("slackId", targetUserId)` omits the optional `workspaceId` arg, so a slackId resolves across workspaces. Blast radius is bounded (the client keys by channelId, so a non-participant never renders it), but the two routes to the same feature shouldn't disagree about who may receive a message. Mirror the participant check and pass `context.workspaceId`.

**5. Slack ephemeral flows are dead on arrival**

`transformPostMessage` emits `<div data-flow-json="…">` with **no `data-flow-appid`** (`request-transformers/chat.ts:52`), and the Slack path never calls `storeEphemeralFlow`. So a Block Kit ephemeral with buttons → `executeAction` → 400 `missing appId`. Interactive ephemerals are the main reason apps call `chat.postEphemeral`. Either wire up appId + the flow store, or reject blocks-with-actions on this route so it fails loudly instead of rendering a dead button.

---

### Should fix

- **`buildFlowContent(flow, (req as any).auth?.appId)`** — `chatController.ts:420` drops postMessage's `?? req.body.appId` fallback (line 292). Non-app-token callers get `data-flow-appid=""` → every action 400s, with nothing at post time to say why.
- **Ephemerals are unbadged in threads** — the converted rows (`ThreadPannel.tsx:296-317`) render as ordinary bubbles, and nothing renders `visibleTo` anywhere in the dashboard. Labelled "Only visible to you" in the channel, indistinguishable from a public message in the thread. Easy to reply assuming it was public.
- **`ThreadPannel.tsx:290`** — dropping the `useMemo` on `dbMessages` makes the `[]` fallback a new ref every render, so the `messages` memo recomputes every render and churns everything downstream. That memo was there for a reason.
- **`code: 'VALIDATION__ERROR'`** — `chatController.ts:376`, double underscore. Clients switch on these.
- **Client store never evicts** — module-level `Map`, no clear on logout, survives a user switch in the same tab, and outlives the 30-min redis TTL: the flow stays clickable and the click 404s with no UI feedback.
- **Offline recipient = silent drop, but the API returns 201.** Either document it on the route or return something honest.
- **Duplication** — `buildFlowContent` was extracted, but the markdown/text branch (`chatController.ts:436-440` vs `297-301`) was left copied. Same helper, both routes.

---

### Nits

- `senderName: parsed.data.as_user ? "" : context.appId` (`slack/controller.ts:474`) — appId as a display name, the condition reads inverted, and nothing renders `senderName` anyway.
- Response returns `conversationId: conversationId ?? null` but broadcasts `conversationId ?? resolvedChannelId`. Pick one.
- `error: 'userId is required'` when `req.user` is missing — there's no such field on this route.
- Double blank line after `chatPostEphemeral`.
- PR body is the untouched template — no description, no ticket context, no test evidence, nothing ticked.
- The `flowController.ts:77` `visibleTo` authz branch is the one thing here that really warrants a test.

PR #1046, opened 2026-08-25
fix: channel forward now gates success on server ack (stops silent first-attempt failure)

XyneSpaces:
## 1. Problem Description
Forwarding a message to a channel shows the "Message forwarded" toast but the message is not actually forwarded on the first attempt; a second identical attempt works. Reported in-app by reetam.dey@juspay.in.

## 2. If this is a bug fix or an enhancement, how did you identify the issue?
RCA from logs + code. The channel-forward path in `apps/dashboard/src/components/Chat/ForwardMessageModal/ForwardMessageModal.tsx` called `zero.mutate(mutators.conversations.forwardMessage(...))` **without awaiting the server result**, then fired `toast.success` and navigated away. Zero applies the mutation optimistically on the client; when the server mutator later rejects (production logs showed Postgres `canceling statement due to lock timeout` / `PushProcessor Error … retrying without mutator` around the report), Zero rolls the optimistic write back **after** the user was already told it succeeded — so the message never forwards and only a retry works. Backend logs showed exactly one successful `conversations.forwardMessage` for the user (the retry); the first attempt was rolled back. The DM branch of the same modal already `await`s its network call, which is why only channel forwards were affected.

## 3. Explain the solution implemented in the PR
Route the channel-forward mutation through the existing `surfaceMutationError()` helper (`apps/dashboard/src/utils/zeroMutationToast.ts`), which awaits and inspects the `.server` result (Zero resolves — not rejects — application errors). The success toast / form reset / navigation now run only when the server acknowledges success; on failure the real error toast is shown and the modal stays open for retry. The Forward button already binds to `form.state.isSubmitting`, so it stays in the "Forwarding…" state until the server resolves. One file changed, channel branch only; DM branch untouched.

## 4. Documentation
N/A

## 5. DB Alter Details (if any)
N/A

## 6. Data fix Details (if any)
N/A

## 7. Environment variable Changes (if any)
N/A

## 8. Testing
Manual: with a forced server rejection / throttled network, the channel forward now shows "Failed to forward message" and keeps the modal open (no false success, no navigation). A normal forward shows success only after the mutation is durably persisted. DM/group forwards unchanged. Note: this is a client-side success-signalling fix; the underlying DB lock contention that triggered the rejections is a separate follow-up.

## 9. Services to which this change is to be deployed
dashboard (frontend)

## 10. Main PR link (if this PR is raised against a release branch)
N/A

### DIFF
--- a/apps/dashboard/src/components/Chat/ForwardMessageModal/ForwardMessageModal.tsx
+++ b/apps/dashboard/src/components/Chat/ForwardMessageModal/ForwardMessageModal.tsx
@@ -122,13 +122,7 @@ export const ForwardMessageForm: React.FC<ForwardMessageFormProps> = ({
         const timestamp = Date.now();
 
         try {
-          // Local-first, exactly like sendMessage(): fire the mutator WITHOUT
-          // awaiting the round-trip, report success optimistically, and observe
-          // the real outcome in the background. Awaiting `.server` here would
-          // freeze the modal for the full statement-lock-timeout window (~30s)
-          // on the very failure this targets, and would also fire on transient
-          // `zero`-type errors that Replicache auto-retries — a false "Failed"
-          // that leads the user to re-forward and create a duplicate.
+          // Fire without awaiting (like sendMessage) and observe the outcome below.
           const mutation = zero.mutate(
             mutators.conversations.forwardMessage({
               targetChannelId: firstTarget.id,
@@ -143,7 +137,7 @@ export const ForwardMessageForm: React.FC<ForwardMessageFormProps> = ({
             }),
           );
 
-          // Optimistic success — instant, as before.
+          // Show success message
           logger.info(Event.MESSAGE_FORWARDED, {
             originalMessageId: message.messageId,
             targetType: 'channel',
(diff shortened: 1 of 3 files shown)


apps/dashboard/src/components/Chat/ForwardMessageModal/ForwardMessageModal.tsx
```diff
@@ -121,21 +122,42 @@ export const ForwardMessageForm: React.FC<ForwardMessageFormProps> = ({
...
-              conversationParticipantId: uuidv4(),
-            }),
+          // Await the SERVER result before reporting success. A bare/unawaited
+          // zero.mutate resolves optimistically on the client, so a server-side
+          // rejection (lock timeout, participant guard, "channel not found")
+          // silently rolls the write back AFTER the success toast has fired and
+          // the user has navigated away — the message never actually forwards
+          // and only a retry succeeds. surfaceMutationError awaits `.server`
+          // and inspects it (Zero resolves, not rejects, application errors).
+          const forwarded = await surfaceMutationError(
```
satvikbatra: Our send path is deliberately local-first: ChatInput calls sendMessage (packages/shared/src/messages/send.ts:37) without awaiting — it fires the mutator and observes the outcome in the background via subscribeSendLifecycle, rendering a failed-message UI with retry. Awaiting the round-trip in the submit handler is the opposite, and it's worst in the case this targets: a lock timeout means ~30s stuck on "Forwarding…".

It also introduces the inverse bug. surfaceMutationError toasts on any result.type === 'error', but Zero has two kinds — and mutationLifecycle.ts:15 spells out why that matters: zero-type errors are transient and auto-retried, so "treating these as failures would restore the composer while Zero is still going to persist the write." A brief connection blip → "Failed to forward" → user retries → original replays on reconnect → two forwards. That helper is safe at its current call sites because those are idempotent updates; a forward is a create.

Suggested instead — keep it optimistic, just stop dropping the result:

const mutation = zero.mutate(mutators.conversations.forwardMessage({...}));
toast.success('Message forwarded'); navigate(...);   // instant, as today
subscribeSendLifecycle(mutation, () => { /* surface failure + retry */ });

Still fixes the reported bug (the lock-timeout rejection is an app error and gets surfaced), without the spinner or the false alarm.

devesh-juspay (approved):

satvikbatra (approved):

PR #1054, opened 2026-08-26
feat: XYNE-61810 share an agent DM conversation to a channel

XyneSpaces:
[POT-1](https://drive.google.com/file/d/1-P60Nbbz1nwi-UM6sa23CiG9EIGOJ3C_/view?usp=sharing)
[POT-2](https://drive.google.com/file/d/1ES90tFTAByC1ZWUHxAgllKiC-u9g-aTY/view?usp=sharing)

## What

Lets you take a whole conversation you had with an agent in a DM and put it into a channel or group DM, so teammates can read the thread and carry it forward with the agent.

The original DM is untouched. The destination gets one immutable snapshot message, and the agent can be added so the thread stays interactive — but it only ever sees channel context, never the private source conversation.

## Entry point

No new menu item. The existing **Forward message** action gains a **This message / Whole conversation** toggle, and the toggle appears only in a 1:1 agent DM.

Destinations are channels and group DMs. 1:1 DMs are excluded: delivering a transcript to one named person is unsolicited, and consent there needs its own design.

## Why it is built this way

**The transcript is not Xyne data.** It lives in Claw (`xyne-claw-auth`) as a branching tree, so a Zero mutator cannot read it. The share is a Spaces backend endpoint that fetches the visible transcript over the existing S2S bridge under the sharer's own identity — Claw enforces who may read it, and a guessed conversation id returns an empty slice.

**Only rendered content crosses the boundary.** Reasoning and raw tool results are never read, so they cannot leak. Structured citations referenced by visible text are carried so they still render as chips.

**The message deliberately skips the generic side-effect handler.** Running mention, bot and automation side effects over a 200-message snapshot would fan out a notification per turn. Delivery is explicit instead: websocket, Redis, unread counts, Vespa.

**Source provenance is server-only.** `AgentConversationShare` is excluded from the Zero generator, so the private conversation and message ids are never replicated to members of the destination. The destination message carries display-safe metadata only.

**Group DMs qualify** because they are members-only — a narrower audience than the public channels this already allows — and agents already work there (`APP_MENTION` reaches installed apps in group DMs). They have no `addUserPolicy` and no admins, so the add-agent permission check branches on scope rather than rejecting them.

## Guards

- **Idempotent** — a transaction-scoped advisory lock plus a unique `(workspace, channel, sharer, operationId)` key, so a double click cannot double-post.
- **Nothing new** — re-sharing is blocked when the transcript tip is unchanged since the last share.
- **Immutable snapshots** — re-sharing creates a new, separate thread; it never edits the previous one.
- **Size** — over 60,000 rendered characters is rejected with `TRANSCRIPT_TOO_LARGE` / HTTP 413 before any destination rows are created, so a share never silently loses early context.
- **Membership** — verified before any transcript or share history is read, so a non-member learns nothing about the channel.

## Config

Gated by the `share_agent_conversation_enabled` CAC flag, following `project_recap_enabled`. Enforced on the share, status and preview endpoints — turning it off closes the API, not just the button. Defaults on: it is a kill switch, not a rollout gate.

## API

| | |
|---|---|
| `POST /api/channels/:channelId/share-agent-conversation` | perform the share |
| `GET /api/channels/:channelId/share-agent-conversation/status` | advisory pre-share state |
| `GET /api/xyne-ai/agent-conversation-preview` | channel-independent snapshot preview |

## Persistence

New Prisma model `AgentConversationShare` and migration `20260828120000_add_agent_conversation_shares`. Additive: one table, four indexes, no foreign keys, no backfill.

## Testing

**There are no automated tests in this PR.** They were removed deliberately, so the paths below need manual verification:

- Toggle appears in an agent DM and nowhere else — not in a human DM, not in a channel
- Switching modes clears people and keeps only a valid destination
- Re-share creates a second separate thread; "nothing new" blocks when the DM has not moved
- Agent added to the destination by default; `@`-mentioning it in the shared thread gets a reply
- Setting the CAC flag to `false` hides the toggle and makes all three endpoints return 403

Verified locally with the commands CI runs: backend `tsc --noEmit`, dashboard `tsc --noEmit --project tsconfig.app.json`, and `eslint . --quiet`.

### DIFF
--- a/apps/backend/src/services/shareAgentConversationService.ts
+++ b/apps/backend/src/services/shareAgentConversationService.ts
@@ -1,0 +1,813 @@
+import { randomUUID } from 'crypto';
+import { Prisma } from '@prisma/client';
+import { MessageType, ChannelRole, ChannelAddUserPolicy, ChannelScopeType } from '@xyne/shared';
+import { db } from '@/database/client';
+import { logger } from '@/utils/logger';
+import {
+  agentSlugFromWebhookUrl,
+  getConversationTranscript,
+  listS2SClawAgents,
+} from './clawAgentService';
+import { messageMetadataService } from './messageMetadataService';
+import { ChannelParticipantRepository } from '@/database/repositories/channelParticipantRepository';
+import { websocketService } from '@/services/websocketService';
+import { redisService } from '@/services/redisService';
+import { handleUnreadCount } from '@/zero/utils/unreadCountUtlis';
+import { CacConfigService } from '@/services/cacConfigService';
+import { vespaQueue } from '@/queues/vespaQueue';
+import { messageSchema } from '@/vespa/src/types';
+import { NAMESPACE } from '@/vespa/vespaConfig';
+
+const channelParticipantRepository = new ChannelParticipantRepository();
+
+export const MAX_SHARED_TRANSCRIPT_CHARS = 60_000;
+
+async function isAgentConversationSharingEnabled(workspaceId: string): Promise<boolean> {
+  const enabled = await CacConfigService.fetch('share_agent_conversation_enabled', { workspaceId });
+  return enabled !== false;
+}
+
+async function assertSharingEnabled(workspaceId: string): Promise<void> {
+  if (await isAgentConversationSharingEnabled(workspaceId)) return;
+  throw new ShareAgentConversationError(
+    'SHARING_DISABLED',
+    'Sharing agent conversations is turned off for this workspace'
+  );
+}
+
+export type ShareErrorCode =
+  | 'CHANNEL_NOT_FOUND'
+  | 'NOT_A_CHANNEL_MEMBER'
+  | 'INVALID_TARGET_CHANNEL'
+  | 'CHANNEL_ARCHIVED'
+  | 'RESHARE_CONFIRMATION_REQUIRED'
+  | 'EMPTY_TRANSCRIPT'
+  | 'TRANSCRIPT_TOO_LARGE'
+  | 'NO_NEW_MESSAGES'
+  | 'ADD_AGENT_FORBIDDEN'
+  | 'AGENT_NOT_INSTALLED'
+  | 'SHARING_DISABLED';
+
+export class ShareAgentConversationError extends Error {
+  constructor(
+    public code: ShareErrorCode,
+    message: string
+  ) {
+    super(message);
+    this.name = 'ShareAgentConversationError';
+  }
+}
+
+interface ClawChatMessageLike {
+  id: string;
+  role: string;
+  content?: string | null;
+  reasoning?: string | null;
+}
+
+type SharedCitationInvocation = Prisma.InputJsonObject & {
+  toolCallId: string;
+  citations: Prisma.InputJsonObject[];
+};
+
+const MAX_SHARED_TRANSCRIPT_CITATIONS = 200;
+
+function buildSharedTranscriptCitationMetadata(
+  messages: ClawChatMessageLike[],
+  invocationsByMsgId: Record<string, unknown[]> | undefined,
+  icons: Record<string, string> | undefined
+): {
+  clawCitations: SharedCitationInvocation[];
+  clawCitationIcons?: Record<string, string>;
+} | null {
+  if (!invocationsByMsgId) return null;
+  const byToolCallId = new Map<string, SharedCitationInvocation>();
+  let totalCitations = 0;
+  for (const message of messages) {
+    if (typeof message.content !== 'string' || !message.content.includes('clf-')) continue;
+    const citedIds = new Set<string>();
+    const tokenPattern = /\[clf-([^\][#]+)#\d+\]/g;
+    let match: RegExpExecArray | null;
+    while ((match = tokenPattern.exec(message.content)) !== null) {
+      if (match[1]) citedIds.add(match[1]);
+    }
+    for (const invocation of invocationsByMsgId[message.id] ?? []) {
+      if (!invocation || typeof invocation !== 'object') continue;
+      const record = invocation as Record<string, unknown>;
+      const toolCallId = typeof record['toolCallId'] === 'string' ? record['toolCallId'] : null;
+      const citations = Array.isArray(record['citations']) ? record['citations'] : [];
+      if (!toolCallId || !citedIds.has(toolCallId) || citations.length === 0) continue;
+      if (byToolCallId.has(toolCallId)) continue;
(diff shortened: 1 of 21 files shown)


apps/backend/src/services/shareAgentConversationService.ts
```diff
@@ -0,0 +1,605 @@
...
+
+  const channel = await db.channel.findUnique({
+    where: { id: targetChannelId },
+    select: { id: true, workspaceId: true, visibility: true, addUserPolicy: true },
+  });
+  if (!channel || channel.workspaceId !== workspaceId) {
+    throw new ShareAgentConversationError('CHANNEL_NOT_FOUND', 'Channel not found in this workspace');
+  }
+
+  const actor = await db.channelParticipant.findFirst({
```
XyneSpaces, line 314: Reject callers who are not channel participants before returning status. The current code only uses `actor` to compute `canAddAgent`, so any workspace member can enumerate private-channel visibility, agent membership, and prior-share timestamps.

apps/backend/src/services/shareAgentConversationService.ts
```diff
@@ -0,0 +1,605 @@
...
+      'Nothing new has been added to this conversation since it was last shared'
+    );
+  }
+
+  const header = last
+    ? `> **Updated shared transcript** from ${agentName} — ${visibleCount} message(s)`
+    : `> **Shared transcript** from ${agentName} — ${visibleCount} message(s)`;
+  const body = `${header}\n\n${markdown}`;
+
+  const provenance: ShareProvenance = {
```
XyneSpaces: Do not store private source-DM identifiers in client-visible conversation metadata. Move `sourceConversationId`, `sourceTipMessageId`, and `shareOperationId` to server-only provenance storage because the serializer syncs this metadata to every authorized target-channel client.

apps/dashboard/src/components/Chat/ShareAgentConversationModal/ShareAgentConversationModal.tsx
```diff
@@ -0,0 +1,255 @@
...
+ *
+ * States it renders (all driven by the advisory status endpoint, all
+ * re-enforced by the backend on submit):
+ *  - pick a target channel
+ *  - already-shared / re-share (previouslyShared + hasNewSinceLastShare)
+ *  - no-new-messages (blocks submit)
+ *  - add-agent-to-channel (optional, permission-gated)
+ *  - privacy notice for a private target channel
+ */
+export const ShareAgentConversationModal: React.FC<ShareAgentConversationModalProps> = ({
```
XyneSpaces: Mount this modal from the agent-DM UI and provide a reachable Share action. This component currently has no imports or consumers, so users cannot invoke the feature.

apps/dashboard/src/components/Chat/ShareAgentConversationModal/SharedTranscriptCard.tsx
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/**
+ * Presentational wrapper for a channel message that is a shared agent
+ * transcript (message.metadata.sharedAgentTranscript === true). It gives the
+ * snapshot a distinct, collapsible treatment so it reads as an embedded
+ * artifact rather than a normal chat message. It renders ONLY content the
+ * caller passes; it never fetches the private source conversation.
+ */
+export const SharedTranscriptCard: React.FC<SharedTranscriptCardProps> = ({
```
XyneSpaces, line 19: Wire this card into the message renderer using `metadata.sharedAgentTranscript`. No dashboard code reads that flag today, so shared messages fall through to the normal USER HTML renderer and the generated Markdown labels render literally.

XyneSpaces (commented):
7 blockers, 1 nit. Changes requested — the privacy boundary and backend snapshot design are close, but the status leak, client-visible source provenance, unreachable frontend, concurrency/partial-failure behavior, and missing integration coverage must be fixed before merge. The existing CodeQL log-injection comment also remains unresolved.

apps/backend/src/services/shareAgentConversationService.ts
```diff
@@ -0,0 +1,605 @@
...
+  });
+  if (!actor) {
+    throw new ShareAgentConversationError(
+      'NOT_A_CHANNEL_MEMBER',
+      'You must be a member of the target channel to share into it'
+    );
+  }
+
+  // 2. Idempotency: a prior share under the same operation id wins.
+  const existing = await findShareByOperationId(shareOperationId, targetChannelId);
```
XyneSpaces: Serialize the operation-id and no-new-tip checks with creation. They currently run before the transaction, so concurrent requests with different operation IDs can both pass and create duplicate snapshots for the same source tip.

apps/backend/src/services/shareAgentConversationService.ts
```diff
@@ -0,0 +1,605 @@
...
+        'ADD_AGENT_FORBIDDEN',
+        'You do not have permission to add members to this channel'
+      );
+    }
+    const already = await db.channelParticipant.findFirst({
+      where: { channelId: targetChannelId, userId: agentAppUser.userId },
+      select: { userId: true },
+    });
+    if (!already) {
+      await channelParticipantRepository.addParticipant(
```
XyneSpaces: Make post-commit agent addition recoverable and return the existing message on idempotent retry. If `addParticipant` fails, the snapshot is already durable, while retry returns `messageId: ''` and exits before adding the agent.

apps/backend/src/services/shareAgentConversationService.test.ts
```diff
@@ -0,0 +1,127 @@
+// Bypass the env-validation guard that the global test setup triggers via logger.
```
XyneSpaces: Add service/controller integration coverage for non-member status access, workspace isolation, re-share races, idempotent retry, post-commit failure, and delivery side effects. The current tests exercise only pure transcript/slug helpers, and this PR adds no frontend or E2E coverage.

apps/backend/prisma/generated/zero/schema.ts
```diff
@@ -441,6 +441,7 @@ export const userGroupTable = table("user_groups")
     rotationInterval: string().optional(),
```
XyneSpaces, line 442: Remove these unrelated generated schema changes from this PR. `maxWorkload`, `startOffset`, and `isNotified` are outside the sharing feature and have no source-schema or migration changes here.

anuragdvd (dismissed):

santhoshJus (approved):

aravindsjp (approved):

PR #1059, opened 2026-08-26
fix: XYNE-60938 delete all matching keychain certificates during identity deletion

sithaarth:
## Summary
- `security delete-identity -c <name>` deletes only the first matching identity, so machines enrolled multiple times keep stale duplicate certificates in the keychain.
- `deleteIdentity` now lists all certificates for the common name via `security find-certificate -a -Z`, filters exact label matches, and deletes each by SHA-256 hash.
- Falls back to `security delete-certificate -Z` for orphan certificates left behind by partial enrollment (cert imported but private key missing).
- Real deletion failures now throw instead of being logged as warnings; "not found" remains a no-op. Private key memory is cleared in a `finally`.

## Notes
- Extracted from `fix/electron-whitescreen` — only the certificate deletion change; loader/theme work stays on that branch.

## Test plan
- [ ] Enroll same device twice, verify duplicate identities exist (`security find-certificate -a -c <CN> -Z`)
- [ ] Trigger identity deletion, verify all matching certificates are removed
- [ ] Verify no-op when identity absent

### DIFF
--- a/apps/electron/src/keychain/macKeychainService.ts
+++ b/apps/electron/src/keychain/macKeychainService.ts
@@ -22,6 +22,12 @@ const unlinkAsync = promisify(fs.unlink);
 const SECURITY = '/usr/bin/security';
 const OPENSSL = '/usr/bin/openssl';
 
+// `security` reports a missing keychain item with a "not be found" message on stderr.
+// Single source of truth so the delete-identity paths classify it identically.
+function isNotFoundError(e: any): boolean {
+    return Boolean(e?.stderr?.includes("not be found"));
+}
+
 class MacKeychainService implements IKeychain {
     // Store private key PEM in memory for the duration of the session
     // In a real app, you might want to persist these securely until enrollment is complete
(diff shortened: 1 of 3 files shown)


apps/electron/src/keychain/macKeychainService.ts
```diff
@@ -254,27 +254,52 @@ class MacKeychainService implements IKeychain {
...
-            // It might fail if not found, which is fine
-            if (e.stderr && e.stderr.includes("not be found")) {
+            if (e.stderr?.includes("not be found")) {
                 log.info("Identity not found, nothing to delete.");
                 Logger.info(EnrollmentEvent.IDENTITY_NOT_FOUND, { common_name: commonName });
             } else {
-                log.warn("Delete identity warning:", e.stderr || e.message);
+                log.error("Delete identity failed:", e.stderr || e.message);
                 Logger.logError(EnrollmentEvent.IDENTITY_DELETE_FAILED, e);
+                throw e;
```
devesh-juspay, line 338: **[CONFIRMED] The new `throw` breaks the self-service re-enrollment recovery path.**

`deleteIdentity` previously warn-and-continued on failure; it now throws, which diverts `handleCertificateError` (`apps/electron/src/services/certificate-error-handler.ts:47`) into its catch block, so `invalid-certificate.html` — the only page with the "Re-enroll Device" button — is never loaded.

Scenario: cert is revoked (`ERR_BAD_SSL_CLIENT_AUTH_CERT`) and the keychain delete fails for a real reason (locked keychain, user cancels the keychain auth prompt, ACL denial). The user gets the dead-end "Please restart the application to re-enroll" dialog, and restarting reproduces the identical state — a permanent loop with no path to enrollment. Suggest catching the `deleteIdentity` error in `certificate-error-handler.ts` and still loading the re-enrollment page.

apps/electron/src/keychain/macKeychainService.ts
```diff
@@ -254,27 +254,52 @@ class MacKeychainService implements IKeychain {
...
+                try {
+                    // Hash uniquely identifies one identity even when common names are duplicated.
+                    await execFileAsync(SECURITY, ['delete-identity', '-Z', hash]);
+                } catch (identityError: any) {
+                    if (!identityError.stderr?.includes("not be found")) {
+                        throw identityError;
+                    }
+
+                    // Partial enrollment may leave a certificate without its private key.
+                    await execFileAsync(SECURITY, ['delete-certificate', '-Z', hash]);
```
devesh-juspay: **[PLAUSIBLE] A mid-loop "not be found" from this fallback aborts deletion of the remaining hashes — while reporting success.**

This `delete-certificate` call has no per-iteration catch. If it also fails with "not be found" (item removed concurrently between find and delete, or it lives in a keychain outside the delete search list), the error escapes the `for` loop into the outer catch — which classifies it as benign `IDENTITY_NOT_FOUND` and returns cleanly. The remaining hashes are never attempted, their identities survive, and telemetry says nothing existed — the exact multi-cert bug this PR fixes silently persists.

Fix: catch "not be found" from the fallback per-iteration and `continue`.

apps/electron/src/keychain/macKeychainService.ts
```diff
@@ -254,27 +254,52 @@ class MacKeychainService implements IKeychain {
...
+                }
+            }
+
+            log.info(`Deleted ${hashes.length} identity certificate(s) successfully.`);
```
devesh-juspay: **[CONFIRMED] Logged/telemetry count uses pre-dedup `hashes.length`, and same-hash copies in multiple keychains leave survivors.**

The loop iterates `new Set(hashes)` but this log line and the `IDENTITY_DELETED` event report `hashes.length`. When the same cert (same SHA-256) exists in both the login keychain and a secondary keychain in the search list, `find-certificate -a` (no `-k`) emits two blocks with one hash; the Set collapses them into a single `delete-identity -Z`, which removes only the *first* matching item per the man page. The second copy survives and keeps being offered as the client cert while the log says "Deleted 2 identity certificate(s)" and telemetry emits `deleted_count: 2`.

At minimum report the deduped count; ideally loop the delete per hash until it returns "not be found".

apps/electron/src/keychain/macKeychainService.ts
```diff
@@ -254,27 +254,52 @@ class MacKeychainService implements IKeychain {
...
-        const cmd = `${SECURITY} delete-identity -c "${commonName}"`;
-
         try {
-            await execAsync(cmd);
-            log.info("Identity deleted successfully.");
-            Logger.info(EnrollmentEvent.IDENTITY_DELETED, { common_name: commonName });
+            const { stdout } = await execFileAsync(SECURITY, ['find-certificate', '-a', '-c', commonName, '-Z']);
+            const hashes = stdout
+                .split(/(?=SHA-256 hash:)/)
+                .filter(block => block.match(/"labl"<blob>="([^"]*)"/)?.[1] === commonName)
```
devesh-juspay: **[PLAUSIBLE] The exact-label filter silently skips identities the old `delete-identity -c` would have deleted.**

`security` prints non-ASCII labels as `"labl"<blob>=0x... "escaped"` (verified locally), which this regex mis-captures, and an item renamed in Keychain Access no longer label-matches the CN. Such items are found by `-c` but dropped here; `hashes` ends up empty, the function logs "Identity not found" success, and the stale cert stays installed — `certificate-error-handler` then loops on the same TLS error.

In the normal app flow `label == CN == config.MTLS_IDENTITY_NAME`, and the exact match deliberately protects "Web Simulation Client" from also deleting "...Sandbox" (since `-c` substring-matches), so this only bites externally-mutated or legacy items — but no warning is logged when blocks are dropped. Suggest logging a warning when candidate blocks are filtered out.

apps/electron/src/keychain/macKeychainService.ts
```diff
@@ -254,27 +254,52 @@ class MacKeychainService implements IKeychain {
...
+                .map(block => block.match(/SHA-256 hash:\s*([0-9A-F]+)/i)?.[1])
+                .filter((hash): hash is string => Boolean(hash));
+
+            if (hashes.length === 0) {
+                log.info("Identity not found, nothing to delete.");
```
devesh-juspay: **[CONFIRMED] Cleanup: not-found classification and the not-found log/telemetry pair are each duplicated.**

`stderr?.includes("not be found")` appears at lines 276 and 291, and this "Identity not found" log + `IDENTITY_NOT_FOUND` event pair is repeated at lines 292–293. Two copies of the same branch will drift — a wording/payload change (or a fix to the fragile "not be found" substring) applied to one site but not the other makes `deleteIdentity` swallow an error in one path while rethrowing it in the other.

Suggest extracting an `isNotFoundError(e)` predicate and a single not-found log/event helper. The split/filter/map/filter parse chain could also be a single-pass loop building a `Set`, giving one source of truth for the count.

devesh-juspay (commented):
Automated review (Claude Code) — 8 findings survived adversarial verification. Five are attached as inline comments on the diff; the remaining three are in files this PR doesn't touch, so they're listed here:

**1. [CONFIRMED] The same multiple-certificate bug remains unfixed on Linux** — `apps/electron/src/keychain/linuxKeychainService.ts:276`
`deleteIdentity` runs `certutil -D -n <nickname>` once per NSS DB, but `certutil -D` deletes only **one** certificate per invocation while duplicates accumulate under the same nickname on every re-enrollment (Linux `importCertificate` does no pre-delete; `pk12util` adds a second cert under the same nickname). A Linux user who re-enrolls twice keeps presenting the stale cert and reproduces the `ERR_BAD_SSL_CLIENT_AUTH_CERT` loop this PR fixes on macOS. (Windows' `Get-ChildItem | Where | Remove-Item` pipeline does delete all matches.) The `IKeychain` contract is also now inconsistent: mac throws on failure, linux/win swallow all errors.

**2. [PLAUSIBLE] `delete-keys` IPC can now reject where it never did** — `apps/electron/src/ipc/mtls-handlers.ts:165`
The handler returns `keychain.deleteIdentity(...)` directly, so the renderer's `deleteKeys()` (typed `Promise<void>`, historically never rejecting) can now reject on a locked-keychain/auth-denied failure. The dashboard UI is loaded from a remote `FRONTEND_URL`, so no renderer-side handling ships with this commit — a remote flow doing `await deleteKeys(); await generateKeys(); ...` written against the old contract now aborts.

**3. [CONFIRMED] Root cause is import-time, this loop is mitigation** — `apps/electron/src/keychain/macKeychainService.ts:140`
`importCertificate` imports the p12 with no pre-delete or existence check (contrast `installRootCA` in the same file, which checks for an existing cert first), and `store-certificate` can run repeatedly with a fresh key each time — so duplicates keep accumulating between deletes. One `deleteIdentity` (or `checkIdentity` gate) at the top of the import path would keep the keychain at exactly one identity and make this delete loop a safety net.

devesh-juspay (approved):

ameer030 (approved):
Approved. The exact-label filter is the right call here: `security find-certificate -c` is a prefix match on the keychain label, so "Web Simulation Client" would otherwise also match "Web Simulation Client Sandbox". Since the label is set by us at import (`openssl pkcs12 -name`), equality on it is a real ownership check rather than a heuristic, and deletion itself is already exact via `-Z <sha256>`.

A fully deterministic alternative would be a dedicated keychain file removed with `security delete-keychain`, but that needs search-list and unlock handling plus migration for existing installs — out of scope for a bugfix. Worth filing separately if the orphan-on-rename case (currently logged as `skippedBlocks`) turns out to matter.

PR #1104, opened 2026-08-26
feat: XYNE-59714 integrated workflow-sdk

Navviiii:
**Proof Of Testing**:

https://github.com/user-attachments/assets/d22f9a50-4f21-4b50-9e95-9e6b190bff4c

Migration-Changes:
  - 20260825000000_workflow_sdk_engine

Environment-Changes:
  - ENABLE_WORKFLOW_STUDIO: gates the workers only (default false)

  - backend
  - runner
  - dashboard

## Type of Change

- [x] New feature

## Description

Integrates `@xyne/workflow-sdk@3.2.32` + `@xyne/workflow-ui@1.3.30` as **Workflow Studio** — a
second, visual workflow engine at `/api/workflow-studio`, rendered at
`/:workspaceId/workflow-studio`. XYNE-59714.

It runs **alongside** the existing engines, not as a replacement. `/api/workflows` (legacy) and
`/api/automations` are untouched and keep their own tables, ACL resources and semantics.

**The SDK owns the engine; we own the infrastructure.** It ships the execution state machine,
`{{...}}` template resolution, config validation, retries, pause/resume, branching, 43
framework-agnostic route definitions and the built-in steps (CODE, HTTP_REQUEST, CONDITIONAL,
LOOP, MAP, PARALLEL, SWITCH, WAIT, DEDUP). It owns no infrastructure at all — everything it needs
is a port the host implements. `src/workflowSdk/` is those ports plus Express glue:

| port | ours |
|---|---|
| `PersistenceAdapter` | `persistence.ts` (Prisma) |
| `QueueAdapter` / `SchedulerAdapter` | `queue.ts` / `scheduler.ts` (Bull + Redis) |
| `StorageAdapter` | `storage.ts` (GCS/S3, prefix-validated) |
| `WorkflowAuthorizer` | `authorizer.ts` |
| `SandboxAdapter` | not implemented — CODE fails closed |

**Storage.** The SDK's four core stores reuse the existing tables, discriminated by
`workflowType='SDK'` on every row and every query, and excluded from the legacy generic poller via
`GENERIC_RECOVERY_EXCLUDED_WORKFLOW_TYPES`. Everything without a legacy counterpart is new and
`sdk_`-prefixed, so each Prisma model maps 1:1 (`db.sdkFolder` → `workflow.sdk_folders`).

**Access** is the `WORKFLOW-STUDIO` ACL resource, administered from User Management like any other
gated screen: READ browses, WRITE creates/edits/runs, ADMIN adds approving paused executions and
managing credentials. Within a workspace everything is visible to everyone holding the grant — the
same flat model automations use. Cross-workspace is a hard deny.

**Domain surface** is deliberately small: one trigger (`MESSAGE_RECEIVED`) and one step
(`REPLY_TO_MESSAGE`, which posts as the Automations bot and reuses the same side-effect pipeline so
mentions and unread counts behave identically).

Phase-1 limits: no sandbox, no AgentStep/ai-builder, no per-resource sharing. Credentials are
AES-encrypted at the adapter boundary; `resolveCredential` is the only path that decrypts.

## Areas Touched

- [x] `apps/dashboard`

---

## Zero (Sync) Changes

Seven nullable columns on `workflows` / `workflow_executions` / `workflow_execution_states` appear
in the generated Zero schema. The eight new `sdk_*` tables are deliberately **excluded** via the
generator's `excludeTables` — they are backend-only and three of them hold secrets
(`sdk_credentials.data`, `sdk_webhooks.secret`, `sdk_workflow_callbacks.secret`), so they must never
reach a client. No mutators, no queries, no Query ACL needed.

## Schema & Data Changes

- [x] Prisma schema modified
- [x] Migration added

- [x] Clients regenerated (`db:generate`)
- [x] No new Postgres enum or enum value
- [x] Backward compatible with deployed code

All added columns are nullable and null on legacy rows; all new tables are new. Nothing existing is
altered or backfilled, so deploy order does not matter and rollback is a plain `DROP`.

**Two things a reviewer should know about the shared tables:**

1. Every future query against `workflows` / `workflow_executions` must filter `workflowType`. That
   table already mixes Automations, SDLC_* and legacy rows, so this is one more type rather than a
   new problem — but it is a standing tax of reusing them.
2. The SDK writes `COMPLETED` / `FAILED` / `SKIPPED` where the other engines write `SUCCESS` /
   `FAILURE`. The generic poller already excludes SDK rows, so the risky path is covered; the
   residue is admin/debug listings showing unfamiliar statuses.

## Config, Environment & URLs

- [x] Env variable added

`ENABLE_WORKFLOW_STUDIO` (default **false**) gates **only the execution + cron workers**. The API is
gated by the `WORKFLOW-STUDIO` ACL resource instead.

> **Rollout order matters:** with the flag off but the ACL granted, workflows can be created and
> activated while nothing consumes the queue — executions sit in `PENDING` with no error surfaced.
> Turn the workers on wherever the resource is granted.

No new base URL: webhook and wait-callback URLs render from the existing `BACKEND_URL`.

- [x] Key added to `apps/backend/.env.example`

**Any database that already ran an earlier revision of this migration** (a local that pulled the
branch mid-review) needs, in place so grants survive: rename the `WORKFLOW-BUILDER` resource row to
`WORKFLOW-STUDIO`, rename the seven `workflow.*` tables to their `sdk_` names with their indexes,
and update the `_prisma_migrations` checksum. Fresh databases need none of this.

## API Contract

- [x] New endpoint

43 routes under `/api/workflow-studio`, all new. Two are public (webhook trigger, wait callback) and
authenticate by a secret in the path; the rest sit behind `authMiddleware` + the ACL gate. Public
routes are an explicit allow-list, so a new unauthenticated route in a future SDK version is
withheld and logged at boot rather than silently exposed.

`POST /credentials/:name/test` is additionally restricted to ADMIN: the SDK handler consults no
authorizer and would otherwise let any WRITE-level user send a decrypted credential to a
caller-supplied URL. Raised upstream — see Known Gaps.

Nothing existing changed shape. `workflowRepository.createWorkflowExecution` had one type
adjustment, forced by adding a nullable column to a table whose Prisma **row** type it derives from.

---

## How did you test it?

Backend and dashboard typecheck clean; dashboard `lint:errors-only` clean; `src/workflowSdk` lints
clean. The migration was applied to a scratch database from empty and verified to produce exactly
the tables, columns, nullability, index names and seeded ACL resource the Prisma schema declares.
Adapter wiring verified against the live dev database through the regenerated client.

Manual exercise in local dev: workflows and folders created and listed through the UI, ACL gate
verified (403 without the grant), worker consuming the queue.

### Automated

- [x] Not covered by tests

No tests were added. The reference integration covers credential encryption round-trips, credential
authorization and storage path validation — the three riskiest spots here — and that is the gap
worth closing first. Tracked as follow-up rather than blocking.

### Manual

- [x] Single user
- [x] Different roles or permission levels (member vs admin, ACL boundaries)
- [x] Web browser

---

## Known gaps / follow-ups

**In the SDK, not this PR** — all three affect xyne-search too:

1. `POST /credentials/:name/test` performs no authorization check before decrypting a credential and
   sending it to a caller-supplied URL. Gated host-side here; the fix belongs upstream.
2. `@xyne/workflow-ui` references eight `--wui-*` CSS tokens it never defines and gives no `var()`
   fallbacks, so consumers must declare them (`workflow-ui-vars.css`). It already does this
   correctly for `--wui-danger`.
3. Workflow deletion is inert: `workflow:delete` is in the action vocabulary and the UI renders the
   button, but no `DELETE /workflows/:id` route exists. Deactivate works.

**Ours:** no tests; `sdk_resource_permissions` records owners but is not yet read for authorization
(kept so the per-resource phase needs no backfill); the workflow UI renders light-only under
`[data-theme="midnight"]`, which needs a palette override pass.

## Checklist

- [x] Backend `typecheck` passes
- [x] Dashboard `lint:errors-only` + `typecheck` pass
- [x] New deps added with `pnpm --filter <pkg> add <dep>`
- [x] Commits follow `<type>: <TICKET-ID> <subject>`; branch is `feature/*`
- [x] Docs updated where behaviour changed (`src/workflowSdk/README.md`)


apps/backend/prisma/schema.prisma
```diff
@@ -470,6 +483,10 @@ model WorkflowExecution {
   ignoreDuration            Int      @default(0) // ~22 days max (PAUSED or Waiting for external input)
   mode                      String                @default("AUTOMATIC") // for agent step continuation control
   createdBy                 String?  // User ID of who created/triggered this execution
+  // v2 workflow-sdk: the execution this one was rerun from. Set only on rows
+  // with workflowType='SDK' (excluded from the generic poller via
+  // GENERIC_RECOVERY_EXCLUDED_WORKFLOW_TYPES).
+  sourceExecutionId         String?
```
Shailendra-03: parentWorkflowExecutionId is already there

PR #1127, opened 2026-08-27
feat: XYNE-59762 export canvas as Word document (.docx)

adarshr001-stack:
  - dashboard

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
POT: https://drive.google.com/file/d/1Ff5HT70FO3hj19hdFDE-BFSnbBtf7snY/view?usp=sharing

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


utkarsh-611 (dismissed):

utkarsh-611:
## Review notes

One blocker, then a security item and a couple of smaller things.

### 1. Every image is silently dropped from the exported .docx

The browser build of `@turbodocx/html-to-docx` builds image parts with Node's `Buffer`, which the dashboard never polyfills. `buildImage` executes `Buffer.from(i.fileContent, "base64")`, throws `ReferenceError: Buffer is not defined`, the library **swallows it internally** (`[ERROR] buildImage: Error during XML generation`) and resolves normally.

Verified against `@turbodocx/html-to-docx@1.22.2`'s own `dist/html-to-docx.browser.esm.js` — the file Vite resolves via the package's `browser` export condition — run under jsdom with `Buffer` removed, matching a real browser. The resulting zip has **no `word/media/` entries at all**; the same HTML through the Node build (Buffer present) produces `word/media/image-*.png`.

So: export a canvas containing a whiteboard, a spreadsheet, or any image block → `normalizeBlocksForExport` spends real time rasterizing each one with html2canvas at `scale: 2` and inlining every attachment as base64 → all of it is discarded → the user gets a .docx with the images missing and a `toast.success('Document download started')`.

Electron is affected identically (`apps/electron/src/app/main.ts:277` sets `nodeIntegration: false`, so the renderer has no `Buffer`), and there's no Buffer polyfill anywhere in `apps/dashboard`.

### 2. `canvas:export-docx` IPC handler has no sender validation

`apps/electron/src/ipc/handlers.ts:440` — `ipcMain.handle('canvas:export-docx', async (_event, {fileName, data}) => ...)` discards the event and never calls `isMainWindowSender(event)`.

The only other handler in this file that writes an attacker-controllable `ArrayBuffer` to disk, `error-report:save-file`, calls `assertTrustedErrorReportSender(event)`. The file's own comment at ~line 127 says privileged handlers must *"only honor them from the trusted main window's top-level frame, so a sub-frame, webview, or untrusted origin cannot force a one-call logout/DoS"*.

Any sub-frame or window reaching the bridge can invoke this repeatedly to pop native save dialogs and write arbitrary bytes under any filename the user accepts. There's also no size cap on `data`, so a multi-hundred-MB buffer gets structured-cloned into the main process and copied again by `Buffer.from(new Uint8Array(data))`.

### 3. `define: { global: 'globalThis' }` is applied app-wide

`vite.config.ts:188`. Vite's `define` is global: in the production build `vite:define` rewrites the `global` identifier in **every** JS module in the graph — all app code and all other dependencies — not just `@turbodocx`.

Dev works by a different mechanism entirely (`definePlugin.transform` early-returns for the client; the value reaches the page via `vite/dist/client/env.mjs` doing `context['global'] = globalThis`), so dev creates an enumerable `window.global` self-reference while the build creates no such property. Any dependency branching on `'global' in window` behaves differently between dev and prod.

Narrower fix: `optimizeDeps.esbuildOptions.define` plus a targeted shim, or `globalThis.global ??= globalThis` immediately before the dynamic import.

### 4. `buildPdfHtml` reuse — the CSS is discarded, the PDF footer isn't

Confirmed by unzipping a generated .docx: `word/document.xml` contains no `box-sizing`, i.e. the entire ~1.5 KB `PRINT_CSS` block is stripped along with `<head>`, so none of the declared fonts, colors, or code/blockquote/table styling reaches the document. What *does* survive is the PDF-specific chrome — a run reading `Exported from Xyne Canvas · <date>` appears as a stray paragraph in the Word file.

The discarded CSS isn't free either: html-to-docx runs html-minifier-terser (bundling terser and clean-css) over the input on every export by default. `preprocessing: { skipHTMLMinify: true }` would drop that pass.

### Smaller

- **No pending state or re-entrancy guard** on `handleExportDocx` (`CanvasScreen.tsx:1152`). The dropdown closes and nothing visible happens while html2canvas, a 1.7 MB library chunk, terser and jszip run on the main thread. A user who thinks nothing happened clicks again and starts N concurrent exports, each with its own save dialog. `ReleaseDetailScreen.tsx:767` has the pattern (`{isExportingArt ? 'Preparing…' : ...}`).
- **Errors logged under `Event.CANVAS_MENTION_DEBUG`** (`CanvasScreen.tsx:1167`) — the third unrelated failure class funnelled into the mention-debug event, so docx failures can't be distinguished from mention bugs.
- **Missing `data-track-category`/`data-track-name`** on the new menu item, unlike Markdown/PDF/Present beside it, so the Docx branch is invisible in the export funnel. The eslint rule can't catch it because `DropdownMenuItem` is a component, not a native element.
- `exportCanvasAsDocx` lines 341-346 are byte-identical to `exportCanvasAsPDF` 302-310 including the DOMPurify config — but only the PDF copy carries the comment explaining why data URIs must stay allowed. A shared `buildSanitizedBodyHtml(editor, domRoot)` would stop the two drifting.

PR #1131, opened 2026-08-27
fix: XYNE-59907 close MAPT/SCR retest findings — Zero tenant scoping, CSWSH, SSRF pinning, upload screening, enum hardening, non-root

devesh-juspay:
Closes security-assessment findings across three Payatu reports. The fixes are backend and cross-cutting, so one change often closes the same issue in more than one report.

## Findings addressed

| Report | Finding | Change |
|---|---|---|
| Mobile MAPT retest (iOS & Android) | iOS **PY-JP-003** / Android **PY-JP-005** — Intra-org IDOR | `createNewDM` / `addGroupDmParticipants` / `createChannel` all scope DM target + participants to the session workspace (`findByIdInWorkspace`; DM target 404s before the channel is created) |
| Mobile MAPT retest (iOS & Android) | iOS **PY-JP-006** / Android **PY-JP-007** — Blind SSRF (link preview) | IP/DNS-pinning guard + internal-host-suffix denylist (primary control is the egress policy) |
| Mobile MAPT (iOS & Android) · Thick-client DAPT | iOS **PY-JP-008** / Android **PY-JP-010** · thick-client **PY-JP-018** — File upload | server-side allow-list + magic-byte screen + `.zip` inspection (shadow) |
| Thick-client DAPT | **PY-JP-006** — BFLA in Zero queries | Zero per-query tenant scoping (`zero/tenant-scope.ts`, `zero/server.ts`) |
| Thick-client DAPT | **PY-JP-012** — Cross-Site WebSocket Hijacking | WebSocket Origin gate |
| Thick-client DAPT | **PY-JP-017** — User Enumeration | email-login enumeration hardening |
| SCR retest | **PY-JP-009** — Unenforced Non-Root Container Execution | `tools/xyne-automation-old` runs as non-root |

The claw-auth frontend half of **PY-JP-009** is split into **#1264** (→ `feature/deploy-xyneclaw`). The Zero BFLA also underlies the mobile/SCR Zero findings (already *fixed* at their retests); this central scoping is the durable fix.

## Notable changes & risk

- **Zero per-query tenant scoping** (highest-risk — it scopes *every* read query): the root table is scoped to the caller's workspace, or org-membership / parent-canvas for the few tables above workspaces. Verified **all 127 tables** are covered (a cross-check test fails CI if any table lacks a rule), `.where()` chains correctly after `.one()` / `.related()`, and singular results are preserved.
- **Upload:** deny-by-default in the shared middleware — extension allow-list + declared-MIME block-list + magic-byte screen (PE/ELF/Mach-O/EICAR, incl. extensionless). `.zip` entries inspected in **shadow** mode by default (`UPLOAD_ARCHIVE_SCREENING`). Detection **fails open** (logs, never errors the upload).
- **WebSocket Origin gate:** matches `config.cors.origin`, which now includes the app's own `FRONTEND_URL` — so a same-origin dashboard is accepted even if `CORS_ORIGIN` doesn't list it.

**Tests/checks:** tenant-scope **9/9**, upload **25/25**; typecheck adds no new errors (the `MessageAttachment.position` ones are pre-existing Prisma-client drift); gitleaks clean; comments trimmed for open-sourcing.

### 1. Zero tenant scoping — BFLA / cross-tenant reads
**Must still work**
- [ ] Your own workspace data loads — Channels, DMs, Tickets, Boards, Canvases, Dashboards
- [ ] Canvas comments load on a canvas you can access
**Must be blocked**
- [ ] No other tenant's rows ever appear (a query never returns another workspace/org's data)
- [ ] `zero_query_unscoped_table` never fires in logs

### 2. Link preview — SSRF (always strict)
**Must still work**
- [ ] Public URL → preview renders (title / image / favicon)
- [ ] Public URL that redirects to another **public** host → still previews
**Must be blocked**
- [ ] `http://127.0.0.1`, `http://169.254.169.254`, `http://10.x`, an internal hostname → no preview, no server-side call
- [ ] A public URL that 3xx-redirects to an internal address → not followed

### 3. Outbound webhooks — SSRF + allow-internal
**Must still work**
- [ ] App / automation webhook to a **public** URL → fires, response returned
- [ ] With `WEBHOOK_ALLOW_INTERNAL_HOSTS=true` (default): webhook to an **internal** service (`10.x` / `192.168.x` / internal `*.svc.k8s…`) → fires
- [ ] App-action button callback (internal S2S) → dispatches, message flips to "actioned"
**Must be blocked**
- [ ] Webhook to `169.254.169.254` (cloud metadata) → refused **even with the flag on**
- [ ] Webhook to `127.0.0.1` / loopback → refused
- [ ] Webhook target that 3xx-redirects → not followed
- [ ] With `WEBHOOK_ALLOW_INTERNAL_HOSTS=false`: any private / internal target → refused

### 4. File upload — content screening
**Must still work**
- [ ] An image / PDF / `.docx` uploads and renders
**Must be blocked**
- [ ] `.exe`, a PE renamed to `.png`, and the EICAR test file → each refused

### 5. Email login — enumeration hardening
**Must still work**
- [ ] Valid credentials log in; wrong password rejected; 5 fails → 429
**Must be blocked (indistinguishable)**
- [ ] Response **and timing** for a KNOWN email vs an UNKNOWN email are the same (no enumeration)
- [ ] Same status / body for wrong-password vs no-such-account

### 6. WebSocket — CSWSH
**Must still work**
- [ ] Real-time chat — a message appears live in a second session
- [ ] Native / electron client (sends no Origin) connects
**Must be blocked**
- [ ] WS handshake from an origin not in `config.cors.origin` → rejected

### 7. DM / channel scoping — IDOR
**Must still work**
- [ ] DM / channel with a **same-workspace** user succeeds
**Must be blocked**
- [ ] Cross-workspace DM via `POST /api/users/me/dms` → "Participant not found"

### 8. Non-root container — SCR-009
- [ ] `tools/xyne-automation-old` container runs as a **non-root** UID (`id` / `whoami` inside the pod)

## Deploy / environment notes

- **Deploy this PR** so the fixes are in the build Payatu retests.
- **SSRF:** the primary control is the **network egress policy** (infra) — the app-side guard is defense-in-depth.
- **`CORS_ORIGIN`** (checked via kubectl on the `xyne-backend` deployments). WS is now handled by the `FRONTEND_URL` folding so there is no WS breakage, but sandbox should be corrected to its own origins for HTTP consistency — exact values to set:

  **prod** — already correct, no change:
  ```
  CORS_ORIGIN=https://spaces.xyne.juspay.net,https://app.spaces.xyne.juspay.net,https://auth.spaces.xyne.juspay.net,xyne-spaces://.
  ```
  **sandbox** — currently `https://spaces.xyne.juspay.net,xyne-spaces://.` (points at the prod host); change to:
  ```
  CORS_ORIGIN=https://spaces.sandbox.xyne.juspay.net,https://app.spaces.sandbox.xyne.juspay.net,https://auth.spaces.sandbox.xyne.juspay.net,xyne-spaces://.
  ```

## Logs to monitor

- `[UPLOAD] Archive content violation` (shadow) — flip `UPLOAD_ARCHIVE_SCREENING=enforce` once only real payloads appear.
- `[UPLOAD] content screening skipped after error` — spike = detection bug or probing (file still served as an opaque download).
- `zero_query_unscoped_table` (error) — should **never** fire; if it does, add the table to `CUSTOM_SCOPES` / `GLOBAL_REFERENCE_TABLES`.
- `WebSocket handshake rejected from disallowed origin` — a legit origin here → add it to `CORS_ORIGIN`.


Mohan0544 (dismissed):

gyash1512 (dismissed):

apps/backend/src/zero/tenant-scope.ts
```diff
@@ -0,0 +1,77 @@
...
+      t.whereExists('canvas', (c: ScopableQuery) => c.where('workspaceId', ctx.workspaceId)),
+    ),
+};
+
+/** Reference data that is the same for every tenant. */
+const GLOBAL_REFERENCE_TABLES = new Set(['lookup_values', 'merchants', 'resources']);
+
+const zeroTables = schema.tables as Record<string, { columns: Record<string, unknown> }>;
+
+export function scopeQueryToTenant<T>(query: T, ctx: Context, queryName: string): T {
```
XyneSpaces, line 60: **[major] R1/S11 — fail-open if `ctx.workspaceId` is ever undefined.** The workspace branch calls `scopable.where('workspaceId', ctx.workspaceId)` with no guard that `ctx.workspaceId` (nor `ctx.memberId`, required by the `CUSTOM_SCOPES` rules) is present. If any Zero auth path mints a context without `workspaceId`, an `undefined` value may be dropped from the predicate rather than matching nothing — the silent fail-open pattern S11 warns about — and the whole table would be served. *Assumption to confirm:* whether Zero's `.where()` drops `undefined` values and whether every auth path populates `workspaceId` (neither is visible in this diff). Suggested fix: at the top of `scopeQueryToTenant`, throw when `ctx.workspaceId` (or `memberId` for org rules) is missing, so a missing tenant id fails closed exactly like an unlisted table does.

apps/backend/src/zero/tenant-scope.node-test.ts
```diff
@@ -0,0 +1,92 @@
...
+    assert.match(threads, /"table":"canvases"/);
+    assert.match(threads, /"name":"workspaceId"/);
+    assert.match(threads, /"value":"ws-1"/);
+    const comments = whereOf(scopeQueryToTenant(zql.canvas_comments.where('threadId', 'th'), ctx, 'canvasThreadComments'));
+    assert.match(comments, /"table":"canvas_comment_threads"/);
+    assert.match(comments, /"table":"canvases"/);
+    assert.match(comments, /"value":"ws-1"/);
+  });
+
+  it('covers every Zero table — none is served unscoped or throws (guards against a new table missing a scope rule)', () => {
```
XyneSpaces: **[major] R7 — the PR's riskiest claim has no test.** The description states "`.where()` chains correctly after `.one()` / `.related()`, and singular results are preserved", but the suite has no case for a `.one()` query or a `.related()` subquery: the coverage test only proves every table has *some* rule, and every other case is a list query. If applying `.where()` after `.one()` changed the result shape, every single-row Zero query would silently break at runtime. Please add: (a) a query ending in `.one()` asserting the singular shape survives `scopeQueryToTenant`, and (b) a query using `.related(...)` asserting related rows still come back and are still workspace-scoped.

apps/backend/src/middleware/upload.ts
```diff
@@ -45,14 +48,434 @@ export function isBlockedUpload(mimetype?: string, originalName?: string): boole
...
+          }
+          budget.entries += 1;
+          if (budget.entries > MAX_ARCHIVE_ENTRIES) {
+            throw new ArchiveViolation('has more entries than can be inspected', entry.path);
+          }
+          if (entry.type === 'Directory') {
+            await entry.autodrain().promise();
+            return;
+          }
+          if (classifyUpload(undefined, entry.path) === 'blocked') {
```
XyneSpaces: **[major] R5/R1 — archive entries bypass the deny-by-default extension policy.** Inside a zip, entries are refused only when `classifyUpload(...) === 'blocked'`, while a direct upload refuses `'not-allowlisted'` as well (`uploadFileFilter` skips anything `!== 'allowed'`). So `evil.jar` / `evil.hta` / `evil.vbs` (non-executable bytes, so `executableKind` won't catch them) pass inside `bundle.zip` while the same file is refused as a direct upload — the control can be sidestepped by zipping. Suggested fix: refuse `'not-allowlisted'` entries too, mirroring the top-level verdict; if the relaxation is intentional (to avoid false positives on archive-internals naming), document that decision in the comment block above `MAX_ARCHIVE_DEPTH`.

apps/backend/src/controllers/channelController.ts
```diff
@@ -1024,8 +1024,18 @@ export class ChannelController {
         ChannelRole.ADMIN
       );
 
-      // For DM channels, add the other user as participant
+      // For DM channels, add the other user as participant.
+      // Observability only: record when the DM target resolves to a different
+      // workspace than the caller so this can be measured for this path.
       if (scopeType === 'DM' && scopeId) {
+        const dmTarget = await this.userRepository.findById(scopeId);
+        if (dmTarget && dmTarget.workspaceId !== req.user!.workspaceId) {
```
XyneSpaces: **[major] R1/S7 — `createChannel` still adds cross-workspace participants (log-only).** The IDOR fix is enforced in `createNewDM` / `addGroupDmParticipants` via `findByIdInWorkspace`, but this path resolves the DM target with the installation-wide `findById` and only logs `[workspace-scope]` before adding the participant — PY-JP-003/005 remains exploitable through `createChannel`. The description marks this as deliberate observation-first, so I'm not blocking *this* PR on it, but the residual hole needs a tracked follow-up ticket that flips this to the same `findByIdInWorkspace` + 404 behaviour, with the observed log rate as the exit criterion. Please link that ticket here so it doesn't get lost.

apps/backend/src/utils/ssrfGuard.ts
```diff
@@ -143,6 +160,136 @@ export async function assertHostIsExternal(
...
+/**
+ * Validate a host and return the addresses that passed, so the caller can connect
+ * to exactly those instead of resolving the name again. `assertHostIsExternal`
+ * proves the name resolved to public addresses when asked; it cannot stop the HTTP
+ * client resolving again and getting a different (internal) answer — DNS rebinding.
+ * Returns null when the bypass is active and there is nothing to pin.
+ */
+export async function resolveExternalHostPinned(
+  host: string,
+  allowPrivate: boolean = isAllowPrivate(),
```
XyneSpaces: **[minor] R2/S29 — the `allowPrivate` default is not environment-gated.** `resolveExternalHostPinned(host, allowPrivate = isAllowPrivate())` returns `null` — no validation, no pinning — whenever `DATA_SOURCE_ALLOW_PRIVATE_HOSTS` is true, including outside development. Both current callers pass the dev-gated value explicitly (`isAllowPrivate() && config.env === 'development'`), but the default diverges from that policy and is a footgun for the next caller: a staging-only enable of the env var would silently bypass the guard entirely. Suggested fix: gate the default the same way (`isAllowPrivate() && config.env === 'development'`) or drop the default so each caller must decide.

XyneSpaces (commented):
**Xyne Spaces Reviewer** — automated hunk-by-hunk review against the [Xyne Spaces review guidelines](https://github.com/juspay/xyne-spaces) (25 files, all hunks checked). Findings are anchored inline below; the verdict, findings table, and reviewer recommendation follow in a top-level summary comment. This is a COMMENT-type review — no approval or rejection is implied.

apps/backend/src/middleware/upload.ts
```diff
@@ -45,14 +48,434 @@ export function isBlockedUpload(mimetype?: string, originalName?: string): boole
...
+  // images
+  'png', 'jpg', 'jpeg', 'gif', 'webp', 'svg', 'heic', 'heif', 'bmp', 'tiff', 'tif', 'ico', 'jfif', 'avif',
+  // video / audio
+  'mp4', 'mov', 'webm', 'm4v', 'avi', 'mkv', 'mp3', 'm4a', 'wav', 'ogg', 'opus', 'aac', 'amr',
+  // documents
+  'pdf', 'doc', 'docx', 'xls', 'xlsx', 'xlsm', 'xlsb', 'ppt', 'pptx', 'odt', 'ods', 'odp',
+  'rtf', 'pages', 'numbers', 'key', 'eml', 'msg', 'ics', 'vcf',
+  // text / data
+  'txt', 'md', 'csv', 'tsv', 'log', 'json', 'jsonl', 'xml', 'yaml', 'yml', 'toml',
+  'ini', 'conf', 'env', 'sql', 'har', 'html', 'htm', 'xhtml',
```
XyneSpaces, line 68: **[nit] R8 — secrets-bearing formats are shareable.** The allow-list includes `env` and `har` — HAR files carry cookies/tokens and `.env` files credentials; these are the classic accidental-leak formats. Understood that the list mirrors production traffic, but consider warning the uploader (or excluding) these two. Not blocking.

apps/backend/src/zero/tenant-scope.ts
```diff
@@ -0,0 +1,77 @@
...
+  // @ts-ignore - asQueryInternals works with any Query type at runtime
```
XyneSpaces, line 61: **[nit] R8 (guideline 242) — `@ts-ignore` + `any` casts in new code guarding tenant isolation.** `// @ts-ignore` over `asQueryInternals(...)` and the `({ or, exists }: any)` callbacks silence the type system in the very file that enforces the tenant boundary. A minimal `.d.ts` (or a typed wrapper) for the Zero internals would keep that boundary explicit. Low priority — the node-test exercises the runtime behaviour.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1131

**Verdict:** Changes requested — 4 major findings to address before merge (no blocking issues).
**Reviewed:** 25 files, +1490/−100 at head `10ba5f3` · **Findings:** 0 blocking, 4 major, 1 minor, 3 nits
**Inline review:** https://github.com/juspay/xyne-spaces/pull/1131#pullrequestreview-5101295008 (7 comments on the head commit)

### Findings
| major | R1 | `apps/backend/src/zero/tenant-scope.ts:60` | No fail-closed guard for missing `ctx.workspaceId`/`memberId` — if Zero drops a `where('workspaceId', undefined)` clause, the table is served unscoped (assumption, see inline comment). Fix: throw explicitly before building the query. |
| major | R7 | `apps/backend/src/zero/tenant-scope.node-test.ts:72` | No test that `.one()` / `.related()` singular-result semantics survive the scope rewrite — this is the PR's riskiest behavioural claim. Add a singular-result + related-model test. |
| major | R5 | `apps/backend/src/middleware/upload.ts:338` | Archive screening only refuses entries whose reason is `'blocked'`; `'not-allowlisted'` types (.jar/.hta/.vbs) still pass inside zips, so deny-by-default is bypassed in archives. Treat any non-allowlisted entry as refused. |
| major | R1 | `apps/backend/src/controllers/channelController.ts:1032` | Cross-workspace DM participant add in `createChannel` remains log-only (documented as intentional) — residual of PY-JP-003/005. Needs a tracked follow-up ticket to flip to `findByIdInWorkspace` + 404. |
| minor | R2 | `apps/backend/src/utils/ssrfGuard.ts:178` | `resolveExternalHostPinned` defaults `allowPrivate = isAllowPrivate()` without a dev gate, unlike `checkHostAccess` — latent guard bypass if the flag is ever enabled in prod. Gate on `isDev` as well. |
| nit | R8 | `apps/backend/src/middleware/upload.ts:68` | Upload allow-list includes secrets-bearing formats (.env, .har). Consider dropping or scrubbing them. |
| nit | R8 | `apps/backend/src/zero/tenant-scope.ts:61` | `@ts-ignore` + `any` casts inside the tenant-isolation file — type it properly so scope regressions surface at compile time. |
| nit | — | PR scope | 25 files exceeds the ~20-file guideline; consider splitting future retest-fix PRs by domain (Zero / uploads / SSRF / auth). |

### Recommended reviewer
**@gyash1512** — Yash Gupta (`yash.gupta@juspay.in`)

No CODEOWNERS pattern matches any of the 25 changed files (`@juspay/xyne-acl-guardians` covers only `acl/`, `tenant/`, and mutator paths), so candidates were ranked on prior engagement with this PR, review capacity, and Spaces activity:

| @gyash1512 | — (reviewed this PR on 1 Sep) | no (no CODEOWNERS match) | 0 | most recent reviewer of this PR; 40 authored PRs in the repo; most recently active in Spaces (28 Aug) |
| @Mohan0544 | — (reviewed this PR on 27 Aug) | no | 0 | — |
| @devesh-juspay | all 25 (PR author) | no | — | excluded: PR author |

@gyash1512 — you reviewed the latest head of this branch on 1 Sep and have no pending review requests; please pick this up again. (Reviewer-request tooling note: my toolset has no `requested_reviewers` API capability, so this @mention is the formal ask — it notifies you on GitHub.)

### Notes
- Solid work on: deny-by-default upload allow-list with EICAR + zip-bomb budgets, dev-gated private-host bypass in `checkHostAccess`, fail-closed pinned HTTP dispatcher on redirects, constant-time password comparison with timing-parity tests, non-root Dockerfile, and centralizing Zero query scoping in `tenant-scope.ts`.
- The CodeQL SSRF alert on `chatController` should auto-resolve now that external fetches route through the pinned dispatcher — re-check the alert after merge.

apps/backend/src/zero/tenant-scope.ts
devesh-juspay: Verified this is not a fail-open: `.where('workspaceId', undefined)` compiles to `workspaceId = null` (checked the query AST), which matches no rows — it returns empty, not all rows, so the table is never served unscoped. Added a `logger.warn` for a missing `workspaceId`/`memberId` for observability rather than a throw (a throw would 500 any flow that ever sends an empty context, for no security gain). — f0b676b9d

apps/backend/src/zero/tenant-scope.node-test.ts
devesh-juspay: The behaviour was verified manually: `.where()` chains after `.one()` with the singular result format preserved, `.related()` keeps its subquery while the root is scoped, and `.where('workspaceId', undefined)` compiles to `workspaceId = null` (empty). Note: test files were removed from this PR per maintainer decision, so no test is committed here.

apps/backend/src/middleware/upload.ts
devesh-juspay: Fixed — archive entries now refuse anything `!== 'allowed'`, matching the direct-upload deny-by-default (`uploadFileFilter`), so a non-allowlisted-but-dangerous type (`.jar`/`.hta`/`.vbs`, no executable magic bytes) can no longer slip through inside a `.zip`. — f0b676b9d

apps/backend/src/controllers/channelController.ts
devesh-juspay: Fixed — flipped from log-only to enforced: the DM target is validated in-workspace via `findByIdInWorkspace` (404 before the channel is created) and group participants resolve via `findByIdInWorkspace`, so a caller-supplied id in another workspace is refused rather than added — matching `createNewDM` / `addGroupDmParticipants`. — e41639329

apps/backend/src/utils/ssrfGuard.ts
devesh-juspay: Fixed — gated the default `allowPrivate` on `config.env === 'development'` for both `assertHostIsExternal` and `resolveExternalHostPinned`, so the default fails safe if a future caller relies on it (both current callers already pass the dev-gated value explicitly). — f0b676b9d

apps/backend/src/middleware/upload.ts
devesh-juspay: Acknowledged — the allow-list mirrors real production upload traffic, so leaving `.env`/`.har` in for now; noted as a candidate for a future tightening (warn or exclude).

apps/backend/src/zero/tenant-scope.ts
devesh-juspay: Acknowledged — the `@ts-ignore` / `any` sit over the Zero query internals, which aren't publicly typed; a minimal typed wrapper is a reasonable follow-up. Low priority, leaving for now.

Mohan0544 (dismissed):

Mohan0544 (approved):

sithaarth (approved):

PR #1133, opened 2026-08-27
Fix: XYNE-59762 spaces collaboration fixes

adarshr001-stack:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Two access-control fixes in Spaces:
### 1. Channel members — honor EVERYONE add-user policy
When a channel's add-user setting is EVERYONE, any member should be able to add
new members, but the add still failed for non-admins.
POT: https://drive.google.com/file/d/1ggh5RvPo0tIhVW7nHaPHjj_4Yul9LQgd/view?usp=sharing
### 2. SDLC artifacts — channel members get edit access by default
POT: https://drive.google.com/file/d/1ueLIIqVTwTo8eo9mvwuGiMVx4wh0ApID/view?usp=sharing

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


utkarsh-611:
## Review notes

The diff is small but there's an access-control regression in it. Flagging the blockers first.

### 1. `ADMINS_ONLY` silently degrades to `EVERYONE` — needs a backfill

`channel-user-status-acl.ts` switches the policy read from `channels.addUserPolicy` to `channel_stats.addUserPolicy`. But the `channel_stats` split migration (`20260318210304_channel_stats_split`) only does `CREATE TABLE` with `addUserPolicy DEFAULT 'EVERYONE'` — there's no `INSERT ... SELECT` from `channels`, and every lazy `channelStats.upsert` create block (`channelRepository.updateLastActivity`, `callRepository`, `channelParticipantRepository`, `channelController`) omits the column.

So any channel whose admin set `ADMINS_ONLY` before 2026-03-18 still carries that **only** on `channels.addUserPolicy`. After this change both ACLs read `channel_stats`, get NULL or a missing row, fall through to `?? 'EVERYONE'` — and any MEMBER can add users to a channel that is supposed to be admin-only.

This PR removes the last reader that still honoured the old column, so the restriction is lost with nothing carrying it over. Needs a backfill migration alongside the change.

### 2. Missing self-exemption blocks self-service joins

`joinChannel` (`packages/shared/src/zero/mutators.ts:534-590`) is self-service: the user inserts their own participant row, then restores their own soft-deleted `channel_user_status` (line 566) or inserts a fresh one (575). Every `tx.mutate` inside a custom mutator routes through this ACL.

For a PUBLIC DEFAULT channel with `addUserPolicy=ADMINS_ONLY`, a non-admin self-join now throws *"only channel admins can restore users to this channel"*. The soft-delete branch at line 71 explicitly exempts `status.userId === this.ctx.userID`; the restore branch and `canInsert` don't.

Worth noting this is a **live** regression rather than latent — before this diff the check was inert, because nothing writes `channels.addUserPolicy` post-split. Making it resolve a real policy is what activates it. Same gap in the Prisma-side `canCreate` (`database/acl/tables/channel-user-status-acl.ts:43`).

### 3. The policy is enforced at the wrong layer

`ChannelParticipantsACL.canInsert` never reads `addUserPolicy` — it returns as soon as `channel.visibility === PUBLIC || existingParticipant`. A member of an `ADMINS_ONLY` channel can issue a bare `tx.mutate.channel_participants.insert({ channelId, userId: <someone else> })` and the target becomes a real participant (visible in the member list, receives messages) with no status row at all.

Hardening the status table while the participant table stays open means the gate can be walked around.

### 4. The canvas VIEWER → EDITOR change grants more than editing

`CanvasParticipantsACL.getRequesterEffectiveRole` (`canvas-participants-acl.ts:21-56`) derives effective role from any `canvas_participants` row whose `channelId` matches one of the user's channels. With that row now EDITOR:

- `canUpdate` (line 119) lets any channel member change other participants' roles — it only blocks EDITOR from *granting* OWNER, so a member can demote the canvas OWNER to VIEWER
- `canDelete` (line 147) lets any member remove participants, including the channel-share row itself, cutting off the whole channel

VIEWER made all of that impossible. The intent was "edit the document"; this also hands out access control over it.

### 5. Existing artifacts don't get the fix

The role is baked into the `canvas_participants` row at creation time, so only PRD/Tech Docs created after deploy get EDITOR. Existing ones keep VIEWER and `canvasAuthService.checkAccess` leaves `canEdit` false. Users will see the fix apply to new documents but not the ones they already have open, with no migration updating rows for canvases whose `sdlc_artifacts.artifactType` is PRD or TECH_DOC.

### 6. ACL and mutator disagree on GROUP_DM

`mutators.ts:1284` guards the same check with `if (channel.scopeType !== ChannelScopeType.GROUP_DM)`, and `mutators.ts:869` does likewise. Both ACL sites (lines 32 and 90) omit it, so a GROUP_DM carrying `ADMINS_ONLY` passes the mutator and is then rejected by the ACL with a confusing "only channel admins" error.

### Worth doing while you're here

`channels.addUserPolicy` now has **zero readers**, but `SdlcHubService.ts:120` still writes it and `packages/shared/src/zero/schema.ts:787` still ships it to every client. The two columns will drift permanently, and the next person to grep `addUserPolicy` picks the dead one — which is exactly the bug this PR fixes. Either drop the column (with the backfill from #1) or remove the SDLC write and the schema field.

The `?? EVERYONE` resolution is now copy-pasted across five sites that already disagree about GROUP_DM and self-service. A single `canUserAddToChannel(channelId, userId, tx)` helper would make this one decision, and would have made the `channels` → `channel_stats` move a one-line change rather than a four-file hunt.

PR #1134, opened 2026-08-27
feat: XYNE-17048 app contributor

HarimohanSharma11:
Migration-Changes:
  - added app collaborator table

  - backend
  - dashboard

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


ameer030 (commented):
Structure looks solid — table + Prisma model + Zero schema/relationships, both mutator files mirrored, all three ACL layers (Prisma mutate, Zero mutate, Zero read) registered, UI gated consistently. Notes below, roughly by severity.

## Security

**1. Collaborators can be added across workspaces.** `AppCollaboratorsACL.canInsert` (`apps/backend/src/zero/acl/tables/app-collaborators-acl.ts:49`) validates that the target user exists, but not that they belong to the caller's workspace or org — and neither does the `appCollaborators.add` mutator. This is reachable rather than theoretical: `wrapTransactionWithACL` (`apps/backend/src/zero/acl/wrappers/transaction-wrapper.ts:51`) proxies only `mutate`, so `tx.run()` reads inside a mutator are not read-ACL filtered, and `zql.users.where('id', args.userId)` resolves any user in any workspace. The client-side picker is workspace-scoped so the UI never offers it, but a crafted mutation grants an out-of-org user template-edit rights plus signing-secret access. Suggest checking the target user's `workspaceId` against `ctx.workspaceId` (or the app's org) in `canInsert`.

**2. `canInsert` does not pin `workspaceId`.** Same file. `AppsACL.canInsert` (`apps-acl.ts:33`) throws on `args.workspaceId !== this.ctx.workspaceId`; the collaborator ACL has no equivalent. The shipped mutator stamps `authData.workspaceId` server-side, so this is not exploitable today — but the ACL layer exists precisely as the backstop for that, and it should mirror its sibling.

**3. Signing secret is now readable by `CONTRIBUTOR`.** `apps/backend/src/apps/core/appUtils.ts:277`. That key signs the webhook HMAC and the per-install JWT for *every* workspace that installed the app, so it is the strongest privilege in the feature and it is being granted at the lower of the two roles. The inline comment ("they need the secret to develop") suggests this is deliberate — if so, worth stating explicitly in the PR description; otherwise gate secret reads to `ADMIN`.

**4. Cross-org read of collaborator lists.** `packages/shared/src/zero/acl/tables/app-collaborators-acl.ts` mirrors the `apps` read ACL, which makes `GLOBAL` apps readable everywhere. Consequence: any workspace can enumerate the full collaborator user-ID list of any marketplace app. The `apps` ACL only exposes `createdBy`, so this widens the surface. Consider restricting collaborator reads to the app's own org even for `GLOBAL` apps.

**5. No guest write check.** `apps-acl.ts` calls `assertGuestWriteBlocked` on all four operations; the new `app-collaborators-acl.ts` calls it on none. A guest who holds an ADMIN collaborator row can add and remove collaborators.

**6. Last-admin invariant is read-then-write.** Enforced in three places (mutator, ACL, UI) with no DB constraint behind it. Two concurrent `remove` calls against the last two admins each read `admins.length === 2` and both proceed. Low frequency, but the rule is stated as a hard invariant.

## Simplification

**7. The backfill may not need to exist.** `routes/appCollaboratorsBackfill.ts`, the 124-line `services/appCollaboratorsBackfillService.ts`, both mounts in `app.ts:421-422`, and the seed insert in `appsRepository.createApp` together write rows that no *authorization* path reads — every consumer short-circuits on the creator first: `getAppEditorRole` returns `ADMIN` for `createdBy`, all three Zero mutators check `app.createdBy !== userID` before querying rows, `AppCollaboratorsACL.assertCanManage` does the same, and so do `AppsTable.canEditApp` and `CollaboratorsSection.canManage`.

The two things that genuinely read the seeded row are the Collaborators list display and the last-admin count. Both look cheaper to solve directly — render the creator as a synthetic implicit-admin row in the list, and count the implicit creator in the admin check. That removes roughly 190 lines, one route, and the migration note from the PR body. If the seed is kept for another reason, worth a comment saying which.

**8. `createApp` is not transactional.** `appsRepository.createApp` inserts the app and the collaborator row as two separate statements. A failure between them leaves an app with no ADMIN row. Recoverable via the implicit-creator fallback, but `$transaction` is cheap — or moot if item 7 is taken.

**9. Dead code.** `assertCanEditApp` and `assertCanManageCollaborators` (`apps/backend/src/apps/core/appCollaboratorUtils.ts:29,37`) are exported with zero callers.

**10. `getAppEditorRole` re-fetches the app.** All six call sites (`appController.ts:389`, `appUtils.ts:277`, `commandController.ts:304,339,411`, `permissionController.ts:68`) already hold the app row, and the helper queries `apps` again before touching `app_collaborators`. Taking `createdBy` as a parameter removes one query per request on the template-edit paths. (The `app.createdBy !== userId &&` short-circuit in `appUtils.ts:277` is fine as written — it avoids the call entirely for the creator.)

## Frontend

**11. `runMutation` has no `catch`.** `CollaboratorsSection.tsx:100` handles `res.type === 'error'`, but a transport-level rejection from `zero.mutate(...).server` escapes `void handleAdd()` as an unhandled rejection with no toast; the `finally` only resets `saving`.

**12. Collaborators tab can render empty.** `EditAppForm.tsx:1258` adds the nav item whenever `editMode === 'template'`, but the panel body also requires `appCreatedBy && currentUserId`. If either is falsy the tab is clickable and blank — worth gating both on the same condition.

## Hygiene

**13. `apps/backend/prisma/generated/zero/schema.ts` was not regenerated.** It is a tracked file and has been updated alongside previous table additions (`3a2efcebd`, `9148f1d98`); it currently has no `app_collaborators` entry.

**14. No tests.** The last-admin rule, the implicit-creator fallback, and the mutate ACL are all branching logic with no coverage.

**15. PR body.** Description and "How did you test it?" are both empty and the checklists are untouched, though the code does satisfy most of the Zero and schema items.

PR #1180, opened 2026-08-28
fix: XYNE-60937 harden agent-auth against process spoofing and validate ask-ai d…

sithaarth:
…eep-link params

POT: https://drive.google.com/file/d/1vP85T3jUHF-zgo9IJyGMdT4RCm5y5dq0/view?usp=sharing

Bind agent-auth tokens to the requesting process and validate externally triggerable deep-link input (Payatu #1/#2/#3, PY-JP-019).

agent-auth:
- Resolve the real requesting process from the OS TCP table (client port -> PID -> exe path) via new peer-process module; surface it in the consent dialog so users see the true binary, not just a self-declared agentName.
- Bind issued tokens to that exe path; every token-gated route re-resolves the caller and rejects a path mismatch.
- Reject any /auth/request carrying an Origin header, killing the browser dialog-spam vector outright.
- Add a per-requester cool-down after a Deny.
- Replace the cramped native message box with a custom HTML consent modal showing requester, signing status, granted capabilities and a duration picker; native dialog kept as fallback. Agent text rendered via textContent.

deep-links: validate and sanitize ask-ai text/url/domain/title before forwarding to the privileged renderer; reject and log malformed url/domain.

Also adds a reference agent-auth client script for testing.

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist

### DIFF
--- a/apps/electron/src/services/agent-auth.ts
+++ b/apps/electron/src/services/agent-auth.ts
@@ -45,6 +45,11 @@ type DurationOption = '5min' | '1hour' | 'session';
 // dialog. Keyed by requester exe path (or 'unknown' when unresolved).
 const DENY_COOLDOWN_MS = 30 * 1000;
 
+// TTL for the per-socket peer cache used on the token re-check path. An established
+// socket's peer process cannot change within this window, so caching removes most of
+// the per-request shell-out cost without weakening the path binding.
+const PEER_RECHECK_CACHE_MS = 1500;
+
 // Registry of executables recognised as first-party agents. Anything not on the
 // list is shown in the consent dialog as an "Unregistered agent" alongside its
 // real path, so the user is not tricked by an attacker-controlled agentName.
@@ -1510,23 +1515,47 @@ class AgentAuthService {
    * issued to (Payatu #2). Re-resolves the peer from the OS TCP table on every call and
    * rejects a path mismatch, so a token leaked to another local process is unusable.
    *
-   * Note: when the peer cannot be resolved right now (currentPath === null) but a path was
-   * bound at grant time, we log and allow rather than break a legitimate agent on a transient
-   * lookup failure — the null case is "unknown", not a proven mismatch.
+   * Fail closed on an unverified peer. Per the peer-process contract, a null result MUST be
+   * treated as "unverified" rather than "trusted", so both an unbound session (no path captured
+   * at grant time) and an unresolved re-lookup deny the request. The null case is reachable on
+   * demand — a process holding a leaked token can push lsof/ss/PowerShell past EXEC_TIMEOUT_MS
+   * under load, or run where the lookup tool is unavailable — so allowing it would reopen the
+   * exact token-replay hole the path binding exists to close.
    */
   private async validateTokenAndPeer(token: string, req: IncomingMessage): Promise<boolean> {
     if (!this.validateToken(token)) return false;
 
     const session = this.sessions.get(token);
     if (!session) return false;
-    if (!session.agentPath) return true; // nothing bound at grant time; cannot enforce
 
-    const peer = await resolvePeerProcess(req.socket.remotePort, process.pid);
+    // Nothing bound at grant time → the token was never tied to a process, so it is
+    // unverifiable. Treat as unverified and deny rather than trust it.
+    if (!session.agentPath) {
+      Logger.warn(ElectronEvent.AGENT_AUTH_PEER_MISMATCH, {
+        agentName: session.agentName,
+        expected: 'unbound-session',
+        actual: null,
+      }, 'AgentAuth');
+      return false;
+    }
+
+    // Re-check the signature only matters at grant time (consent dialog); here we compare
+    // execPath, so skip codesign. A short per-socket cache absorbs the per-request shell-out
+    // cost on the MCP ingest path without weakening the binding (an established socket's peer
+    // cannot change within the TTL).
+    const peer = await resolvePeerProcess(req.socket.remotePort, process.pid, {
+      verifySignature: false,
+      cacheTtlMs: PEER_RECHECK_CACHE_MS,
+    });
     const currentPath = peer?.execPath ?? null;
 
     if (currentPath === null) {
-      log.warn('[AgentAuth] Could not re-resolve peer for token-gated request; allowing bound session');
-      return true;
+      Logger.warn(ElectronEvent.AGENT_AUTH_PEER_MISMATCH, {
+        agentName: session.agentName,
+        expected: session.agentPath,
+        actual: 'unresolved-peer',
+      }, 'AgentAuth');
+      return false;
     }
 
     if (currentPath === session.agentPath) return true;
@@ -1541,7 +1570,7 @@ class AgentAuthService {
 
   private sanitizeDialogText(value: string | undefined, maxLen: number): string {
     if (typeof value !== 'string') return '';
-    const cleaned = value.replace(/[\u0000-\u001F\u007F]/g, ' ').replace(/\s+/g, ' ').trim();
+    const cleaned = value.replace(/[\u0000-\u001F\u007F-\u009F\u200B-\u200D\u2028\u2029\u202A-\u202E\u2066-\u2069\uFEFF]/g, ' ').replace(/\s+/g, ' ').trim();
     return cleaned.length > maxLen ? `${cleaned.slice(0, maxLen)}…` : cleaned;
   }
 
--- a/apps/electron/src/services/agent-auth.ts
+++ b/apps/electron/src/services/agent-auth.ts
@@ -7,6 +7,8 @@ import { Logger } from './logger/Logger';
 import ElectronEvent from './logger/electron-events';
 import * as fs from 'fs';
 import * as path from 'path';
+import { resolvePeerProcess, describePeer, PeerProcess } from './peer-process';
+import { showAgentConsentWindow } from './agent-consent-window';
 
 const DEFAULT_PORT = 49231;
 
@@ -22,6 +24,11 @@ interface AuthSession {
   description: string;
   expiresAt: number;
   createdAt: number;
+  // Absolute exe path of the process that was granted this token, resolved from
+  // the OS TCP table at grant time. Tokens are bound to this path: every
+  // token-gated request re-resolves the caller and must match. null = the peer
+  // could not be resolved at grant time, so path binding cannot be enforced.
+  agentPath: string | null;
 }
 
 interface AuthResponse {
@@ -34,12 +41,40 @@ interface AuthResponse {
 
 type DurationOption = '5min' | '1hour' | 'session';
 
+// Cool-down after a user Deny so a malicious process cannot hammer the consent
+// dialog. Keyed by requester exe path (or 'unknown' when unresolved).
+const DENY_COOLDOWN_MS = 30 * 1000;
+
+// Registry of executables recognised as first-party agents. Anything not on the
+// list is shown in the consent dialog as an "Unregistered agent" alongside its
+// real path, so the user is not tricked by an attacker-controlled agentName.
+// Populate with the absolute paths / basenames the team ships and trusts.
+const KNOWN_AGENT_EXECUTABLES: ReadonlySet<string> = new Set<string>([
+  // e.g. '/Applications/Xyne.app/Contents/MacOS/xyne-agent',
+]);
+
+// What an approved agent can do through this loopback server (all proxied to the
+// backend with the logged-in user's credentials). Shown in the consent modal so
+// the user understands the scope they are granting. Keep in step with the routes
+// handled in handleRequest().
+const AGENT_CAPABILITIES: readonly string[] = [
+  'Search your workspace data and memory',
+  'Read your conversations and messages',
+  'Download message attachments to your Desktop',
+  'Post messages in your chats',
+  'Create tickets and schedule calls',
+  'Read and write agent memory (facts & SOPs)',
+  'Send AI queries on your behalf',
+];
+
 class AgentAuthService {
   private server: ReturnType<typeof createServer> | null = null;
   private port: number = DEFAULT_PORT;
   private sessions: Map<string, AuthSession> = new Map();
   private cleanupInterval: NodeJS.Timeout | null = null;
   private consentDialogOpen = false;
+  // requester exe path (or 'unknown') -> epoch ms until which new requests are refused
+  private denyCooldowns: Map<string, number> = new Map();
 
   private getMcpBackendBaseUrl(): string {
     return config.BACKEND_URL.replace(/\/+$/, '');
@@ -132,11 +167,13 @@ class AgentAuthService {
 
     try {
       if (req.method === 'POST' && url.pathname === '/auth/request') {
-        // Real local agents are non-browser clients and send no Origin / Sec-Fetch-Site
-        // header; reject cross-site browser requests to the consent dialog.
-        if (this.isCrossSiteBrowserRequest(req)) {
-          log.warn('[AgentAuth] Rejected cross-site /auth/request (browser consent-dialog spam)');
-          this.sendJson(res, 403, { error: 'Forbidden', message: 'Cross-site request rejected' });
+        // Real local agents are non-browser clients and send NO Origin / Sec-Fetch-Site
+        // header. A browser always attaches Origin on a cross-origin POST, so the presence
+        // of an Origin header on the consent endpoint means the caller is a web page — kill
+        // the browser dialog-spam vector outright by rejecting any request that carries one.
+        if (req.headers['origin'] !== undefined || this.isCrossSiteBrowserRequest(req)) {
+          log.warn('[AgentAuth] Rejected /auth/request carrying Origin/fetch-metadata (browser consent-dialog spam)');
+          this.sendJson(res, 403, { error: 'Forbidden', message: 'Browser-originated request rejected' });
           return;
         }
         await this.handleAuthRequest(req, res);
@@ -197,16 +234,22 @@ class AgentAuthService {
   /**
    * Handle authorization request
    *
-   * ACCEPTED RISK (secops #367, MED): this loopback endpoint (127.0.0.1:49231) has no
-   * requesting-process binding and no out-of-band pairing, so ANY local process can POST an auth
-   * request and spawn a native consent dialog with attacker-controlled agentName/description text
-   * (and, on user approval, proxy authenticated backend calls). The declared pairing-code flow was
-   * never implemented; the team accepted this under a local-only (already-compromised-host) threat
-   * model. Implement the pairing-code echo before relying on this endpoint for anything sensitive.
+   * Hardening (Payatu #1/#2/#3): the loopback endpoint (127.0.0.1:49231) is reachable by any
+   * local process, so agentName/description in the body are attacker-controlled and cannot be
+   * trusted on their own. We now:
+   *   1. Resolve the REAL requesting process from the OS TCP table (client port -> PID -> exe
+   *      path) and surface it in the consent dialog, so the user sees the true binary rather
+   *      than only a self-declared name.
+   *   2. Bind the issued token to that exe path; every token-gated route re-resolves the caller
+   *      and rejects a path mismatch (see validateTokenAndPeer).
+   *   3. Apply a per-requester cool-down after a Deny so a rejected process cannot hammer the
+   *      dialog.
+   * This does not fully defend an already-compromised host (PID reuse / TOCTOU are inherent to
+   * a local IPC channel), but removes the trivial impersonation and dialog-spam vectors.
    */
   private async handleAuthRequest(req: IncomingMessage, res: ServerResponse): Promise<void> {
     const body = await this.parseBody(req);
-    
+
     if (!body || typeof body !== 'object') {
       this.sendJson(res, 400, { error: 'Invalid request body' });
       return;
@@ -216,9 +259,21 @@ class AgentAuthService {
 
     // Validate required fields
     if (!authRequest.agentName || !authRequest.description) {
-      this.sendJson(res, 400, { 
-        error: 'Missing required fields: agentName, description' 
+      this.sendJson(res, 400, {
+        error: 'Missing required fields: agentName, description'
       });
+      return;
+    }
+
+    // Resolve the actual requesting process from the OS TCP table.
+    const peer = await resolvePeerProcess(req.socket.remotePort, process.pid);
+    const peerKey = peer?.execPath ?? 'unknown';
+
+    // Deny cool-down: refuse without re-prompting if this requester was recently denied.
+    const cooldownUntil = this.denyCooldowns.get(peerKey);
+    if (cooldownUntil && Date.now() < cooldownUntil) {
+      Logger.warn(ElectronEvent.AGENT_AUTH_COOLDOWN_BLOCKED, { agentName: authRequest.agentName, peer: peerKey }, 'AgentAuth');
+      this.sendJson(res, 429, { error: 'Request temporarily blocked', message: 'This process was recently denied' });
       return;
     }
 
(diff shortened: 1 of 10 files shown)


devesh-juspay (commented):
### Code review

Found 3 issues (none blocking; all in new code):

1. `validateTokenAndPeer` fails open when the peer cannot be re-resolved, contradicting the `resolvePeerProcess` contract ("callers MUST treat null as 'unverified' rather than 'trusted'") and this function's own docstring ("a token leaked to another local process is unusable").

https://github.com/juspay/xyne-spaces/blob/9f6212679362e8fadfe205f4c6547772c6600d77/apps/electron/src/services/agent-auth.ts#L1526-L1531

2. Every token-gated route now spawns 2-3 child processes per request (`lsof`/`ps`/`codesign` on macOS, two `powershell.exe` on Windows) with no caching; `validateToken()` was an O(1) map lookup.

https://github.com/juspay/xyne-spaces/blob/9f6212679362e8fadfe205f4c6547772c6600d77/apps/electron/src/services/agent-auth.ts#L1523-L1525

3. `sanitizeAskAiText` strips only C0 controls + DEL; U+2028/2029, zero-width U+200B-200D, bidi overrides U+202A-202E and U+FEFF pass through into the `[Browser Context]` prompt.

https://github.com/juspay/xyne-spaces/blob/9f6212679362e8fadfe205f4c6547772c6600d77/apps/electron/src/utils/validation.ts#L47-L49

Details inline.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

apps/electron/src/services/agent-auth.ts
```diff
@@ -1436,16 +1497,66 @@ class AgentAuthService {
...
+    const session = this.sessions.get(token);
+    if (!session) return false;
+    if (!session.agentPath) return true; // nothing bound at grant time; cannot enforce
+
+    const peer = await resolvePeerProcess(req.socket.remotePort, process.pid);
+    const currentPath = peer?.execPath ?? null;
+
+    if (currentPath === null) {
+      log.warn('[AgentAuth] Could not re-resolve peer for token-gated request; allowing bound session');
+      return true;
```
devesh-juspay: This is fail-open: when `resolvePeerProcess()` returns `null` on a token-gated request, the bound session is allowed through. That contradicts the contract in [`peer-process.ts#L32-L33`](https://github.com/juspay/xyne-spaces/blob/9f6212679362e8fadfe205f4c6547772c6600d77/apps/electron/src/services/peer-process.ts#L32-L33) ("callers MUST treat null as 'unverified' rather than 'trusted'") and the docstring above ("so a token leaked to another local process is unusable").

The `null` case is reachable on demand, not just transiently: the lookups run with a 2s `EXEC_TIMEOUT_MS`, and a local process holding a leaked token can push `lsof`/`ss`/PowerShell past that under load (or run where the tool is unavailable). The same shape exists at L1522 when `agentPath` was null at grant time.

I understand the intent at L1513-1515 is to not break a legitimate agent on a flaky lookup. If fail-open is the accepted trade-off, it would help to (a) update the `peer-process.ts` contract so the two comments agree, and (b) consider a bounded retry before allowing, or fail-closed only for the higher-privilege routes (ingest/memory proxies) while keeping search lenient.

apps/electron/src/services/agent-auth.ts
```diff
@@ -1436,16 +1497,66 @@ class AgentAuthService {
...
+   * lookup failure — the null case is "unknown", not a proven mismatch.
+   */
+  private async validateTokenAndPeer(token: string, req: IncomingMessage): Promise<boolean> {
+    if (!this.validateToken(token)) return false;
+
+    const session = this.sessions.get(token);
+    if (!session) return false;
+    if (!session.agentPath) return true; // nothing bound at grant time; cannot enforce
+
+    const peer = await resolvePeerProcess(req.socket.remotePort, process.pid);
```
devesh-juspay: This runs on every token-gated request (`/interact`, `/search`, `/api/search`, `/api/ingest/turn`, attachments, and all `guardProxyRequest` memory routes) and shells out each time with no cache: macOS spawns `lsof` + `ps` + `codesign --verify`, Windows spawns two `powershell.exe`, Linux `ss` + `readlink`, each capped at 2s. `validateToken()` was a synchronous map lookup.

Two cheap wins:
- The `signed` result from `isSignedMac()` is computed here but never read on the re-check path (only `execPath` is compared), so `codesign` could be skipped when re-validating.
- A short per-socket cache (keyed on `remotePort`, TTL of a second or two) would remove most of the per-turn cost on the MCP ingest path without weakening the binding, since the peer of an established socket cannot change within that window.

apps/electron/src/utils/validation.ts
```diff
@@ -28,3 +28,52 @@ export function isHexToken(value: unknown, byteSize: number): value is string {
...
+// process boundary before forwarding; callers should drop values that fail.
+// ---------------------------------------------------------------------------
+
+// Free-text params (ask-ai text/title). Strip control chars (including the
+// prompt-structure-breaking ones), collapse runs of whitespace, and cap length.
+// Never throws — returns a safe string ('' if input is unusable).
+export function sanitizeAskAiText(value: unknown, maxLen: number): string {
+  if (typeof value !== 'string') return '';
+  const cleaned = value
+    .replace(/[\u0000-\u001F\u007F]/g, " ") // control chars
```
devesh-juspay: The comment above says this strips "the prompt-structure-breaking" control chars, but the class only covers C0 (`\u0000-\u001F`) and DEL. These pass through unchanged: U+2028/U+2029 (line/paragraph separators), U+200B-U+200D (zero-width), U+202A-U+202E (bidi overrides), U+FEFF. The sanitized `text`/`title` land verbatim in the `[Browser Context]` prompt string at [`XyneAISidebar.tsx#L1733`](https://github.com/juspay/xyne-spaces/blob/9f6212679362e8fadfe205f4c6547772c6600d77/apps/dashboard/src/components/Chat/XyneAISidebar/XyneAISidebar.tsx#L1733), which is the surface PY-JP-019 is about.

Suggest widening the class, e.g. `/[\u0000-\u001F\u007F-\u009F\u200B-\u200D\u2028\u2029\u202A-\u202E\u2066-\u2069\uFEFF]/g` (C1 controls and the isolate controls U+2066-2069 included).

devesh-juspay (approved):

apps/electron/src/services/ui-updater.ts
```diff
@@ -421,6 +459,13 @@ export async function downloadUIUpdate(): Promise<boolean> {
       }
     }
     
+    // Integrity gate: verify the downloaded bundle against the signed release manifest
+    // BEFORE staging it. The manifest pins a SHA-256 of the bundle's index file
+    // (package.index.file_path, e.g. .vite/manifest.json). A mismatch means the bytes we
+    // fetched are not the release the manifest describes — tampered CDN, MITM, or a
+    // truncated download — so we refuse the update rather than run untrusted UI code.
+    verifyBundleChecksum(uiBundlePath, remoteConfig);
```
ameer030, line 467: 🔴 **B01 — the OTA "integrity gate" only checks one file, so tampered JS passes it**

The new checksum verification compares only the bundle's index file (`.vite/manifest.json` or `index.html`) against the release manifest. Every other file in the downloaded zip — all the executable JS/CSS — is never hashed, so an attacker who tampers with the served `dashboard.zip` but keeps that one index file byte-identical walks straight through the gate. That is exactly the "tampered CDN" scenario the comment says this blocks.

- `apps/electron/src/services/ui-updater.ts:340-367` — `verifyBundleChecksum` reads `remoteConfig.package.index.checksum` and hashes only `package.index.file_path`; called once at `ui-updater.ts:467`.
- `ReleaseConfig.resources` (`ui-updater.ts:34`, `unknown[]`) is declared but never read — the manifest's per-resource entries are ignored entirely.
- Grep over the whole OTA path (`checkForUIUpdate` → `downloadUIUpdate` → `applyPendingUpdate`/`applyAndReload`) finds no other hashing or verification anywhere.
- The comment at `ui-updater.ts:462` calls the manifest "the signed release manifest" — in code it is plain JSON fetched over TLS (`fetchReleaseConfig`, `ui-updater.ts` ~269-292); `ReleaseConfig` has no signature field and nothing verifies one. Same for the JSDoc at ~:334.

Why it matters: this PR is the fix for an OTA finding. A gate that an attacker bypasses by editing any sibling file under its pinned filename doesn't close the finding — Vite chunk filenames are fixed at build time, so replaced chunk contents with unchanged names verify clean.

</details>

Verify every extracted file, not just the index:

- If the Airborne manifest's `resources[]` carries per-file checksums (Airborne/hyperota format), model them in `ReleaseConfig` and hash each extracted file after `extract(...)`, failing on any mismatch or missing entry.
- Simpler alternative: ask the manifest to pin the **zip itself** (one SHA-256 over the downloaded `ui-update.zip` before extraction) — one hash covers the whole bundle.
- Also fix the comment: the manifest is fetched over TLS but not signature-verified; either verify a signature or drop the word "signed".

</details>

rules: SEC-C26, DOCS-C04
agent: omp
run: 01a07c2d-ee68-7761-b385-72fd6fb2640e
</details>

apps/electron/src/services/agent-consent-window.ts
```diff
@@ -0,0 +1,118 @@
...
+      maximizable: false,
+      fullscreenable: false,
+      frame: false,
+      transparent: true,
+      show: false,
+      title: 'Agent Authorization',
+      webPreferences: {
+        nodeIntegration: false,
+        contextIsolation: true,
+        preload: path.join(__dirname, '..', 'preload.js'),
```
ameer030, line 60: 🟡 **W01 — consent window gets the full privileged preload bridge**

The new consent window loads the app's shared `preload.js`, which exposes the entire `electronAPI` — mTLS key generation, certificate storage, cookie/session control, screen recording, native file access, app reload, and more. This window only needs `agentConsent.onShow`/`agentConsent.respond` and one allowlisted height channel. Each new consumer of the full bridge widens the surface guide-S31 / SEC-C26 asks to keep minimal.

- `apps/electron/src/services/agent-consent-window.ts:60` — `preload: path.join(__dirname, '..', 'preload.js')`.
- `apps/electron/src/preload.ts:16-32` documents the bridge as privileged; `isTrustedOrigin()` (~:37) returns `true` for `file:` URLs, so the bundled `agent-consent.html` receives everything at `:480+`.

Mitigating today: the HTML is static, CSP is `default-src 'none'; script-src 'unsafe-inline'`, and all agent-controlled strings render via `textContent`, so there is no live injection path — this is about shrinking a latent surface, not an active exploit. Existing popups (meeting-popup, recording-pill, claw-overlay) follow the same pattern, but this diff adds one more privileged consumer.

</details>

Give the consent window its own tiny preload exposing only:

- `agentConsent.onShow(cb)` (subscribe to `agent-consent:show`)
- `agentConsent.respond(result)` (send `agent-consent:respond`)
- the `agent-consent:content-height` send

Verified safe: the HTML only uses `window.electronAPI.agentConsent.*` and `ipcSend('agent-consent:content-height')`, so a minimal preload changes nothing the dialog does.

</details>

rules: SEC-C26
agent: omp
run: 01a07c2d-ee68-7761-b385-72fd6fb2640e
</details>

apps/electron/src/services/agent-auth.ts
```diff
@@ -1436,16 +1502,90 @@ class AgentAuthService {
...
+      return false;
+    }
+
+    // Re-check the signature only matters at grant time (consent dialog); here we compare
+    // execPath, so skip codesign. A short per-socket cache absorbs the per-request shell-out
+    // cost on the MCP ingest path without weakening the binding (an established socket's peer
+    // cannot change within the TTL).
+    const peer = await resolvePeerProcess(req.socket.remotePort, process.pid, {
+      verifySignature: false,
+      cacheTtlMs: PEER_RECHECK_CACHE_MS,
```
ameer030, line 1548: 🟡 **W02 — peer re-check cache is keyed on the client port, which a different process can inherit**

The token re-check caches the resolved exe path by `remotePort` for 1.5 s. The comment justification — "an established socket's peer process cannot change within this window" — is only true while that socket is alive. Once the bound agent's socket closes, the OS can recycle the same ephemeral port to a different process within the TTL; the cache hit then returns the *previous owner's* exe path and the request passes the path check. That partially reopens the exact spoof this PR's token binding closes.

- `apps/electron/src/services/agent-auth.ts:1545-1549` — re-check passes `cacheTtlMs: PEER_RECHECK_CACHE_MS` (1500 ms).
- `apps/electron/src/services/peer-process.ts:46` — `peerCache = new Map<number, …>` keyed on the port number alone; nothing ties an entry to a live connection.

Attack shape: local attacker holds a leaked token bound to the victim's exe path → victim's connection closes → attacker's new connection lands the same ephemeral port within 1.5 s → cached victim path matches `session.agentPath` → accepted. Preconditions (leaked token, port-reuse race, TIME_WAIT narrows it) keep it a narrow window, but it is attacker-influenceable via local connection churn, and the whole point of the binding is that process B must not use process A's token.

For benign agents the port key is mostly accurate (keep-alive → same connection → same port; new connection → fresh lookup → same exe), so the hole is only the cross-socket recycle race.

</details>

Scope the cache to the connection, not the port:

```ts
// WeakMap<net.Socket, PeerProcess> — one entry per live connection, dies with the socket
private peerBySocket = new WeakMap<Socket, PeerProcess>();
const sock = req.socket;
let peer = this.peerBySocket.get(sock);
if (!peer) {
  peer = await resolvePeerProcess(sock.remotePort, process.pid, { verifySignature: false });
  if (peer) this.peerBySocket.set(sock, peer);
}
```

This kills the TTL entirely, keeps the per-request shell-out elimination for keep-alive clients, and can never serve one socket another socket's identity. (Confirmed sound in a second review pass.)

</details>

rules: SEC-C03, BUG-C11
agent: omp
run: 01a07c2d-ee68-7761-b385-72fd6fb2640e
</details>

apps/electron/src/services/agent-auth.ts
```diff
@@ -252,7 +317,8 @@ class AgentAuthService {
         agentName: authRequest.agentName,
         description: authRequest.description,
         expiresAt,
-        createdAt: Date.now()
+        createdAt: Date.now(),
+        agentPath: peer?.execPath ?? null,
```
ameer030, line 321: 🟡 **W04 — when the peer can't be identified at grant time, the user "approves" a token that can never work**

If `resolvePeerProcess` returns `null` during `/auth/request` (lookup tool missing, 2 s exec timeout exceeded, platform lookup failure), the flow still shows the consent dialog, and on Allow stores a session with `agentPath: null` and returns 200/"approved" to the agent. From then on `validateTokenAndPeer` correctly denies **every** call for that token ("unbound-session") — so the user granted access and the agent got a credential that is permanently dead, with no error surfaced at the step where the user could understand it.

- `apps/electron/src/services/agent-auth.ts:280-281` — grant-time resolution; `peer === null` only demotes the dialog to "unknown process", it doesn't stop the grant.
- `agent-auth.ts:321` — `agentPath: peer?.execPath ?? null` stored on the session.
- `agent-auth.ts:1530-1540` — every later request short-circuits on `!session.agentPath` → 401, and each one also emits an `agent_auth_peer_mismatch` warning, so a legitimate agent on such a system generates warn-spam forever.

Reachability: `resolvePeerProcess` fails closed to `null` on any tool/timeout/platform failure, so this is reachable in production on unusual-but-real setups.

</details>

Fail at the consent step instead of issuing a dead token: when `peer?.execPath` is null at `/auth/request`, respond with an error the client can show (e.g. 403 "could not identify the requesting process") and don't open the dialog. One tradeoff to state in the code/PR: this makes approval impossible on platforms without the OS lookup tooling — that is the intended fail-closed posture, but it should be an explicit product decision, not an accidental dead end.

</details>

rules: SEC-C03, BUG-C06
agent: omp
run: 01a07c2d-ee68-7761-b385-72fd6fb2640e
</details>

apps/electron/src/services/agent-auth.ts
```diff
@@ -34,12 +41,45 @@ interface AuthResponse {
...
+// TTL for the per-socket peer cache used on the token re-check path. An established
+// socket's peer process cannot change within this window, so caching removes most of
+// the per-request shell-out cost without weakening the path binding.
+const PEER_RECHECK_CACHE_MS = 1500;
+
+// Registry of executables recognised as first-party agents. Anything not on the
+// list is shown in the consent dialog as an "Unregistered agent" alongside its
+// real path, so the user is not tricked by an attacker-controlled agentName.
+// Populate with the absolute paths / basenames the team ships and trusts.
+const KNOWN_AGENT_EXECUTABLES: ReadonlySet<string> = new Set<string>([
```
ameer030, line 57: 🔵 **N01 — `KNOWN_AGENT_EXECUTABLES` is empty, so the "Verified agent" path never renders**

The registry has zero entries, so `isKnownAgent` always returns false: every agent — including any future first-party one — shows the red "Unregistered" badge and the warning banner, and the trusted-badge branch in `agent-consent.html` is unreachable. If populating the list is deliberately follow-up work, fine — just note it (or the finding that asks for it), otherwise the badge/warn split is dead weight.

- `apps/electron/src/services/agent-auth.ts:57-59` — empty `KNOWN_AGENT_EXECUTABLES` with a "populate with…" comment.
- `apps/electron/assets/agent-consent.html` — the `badge trusted` / hidden-`#warn` branch never fires.

</details>

Either seed the set with the executables the team ships, or drop the badge/`isKnown` split until entries exist and let the dialog show the neutral state for everyone.

</details>

rules: STYLE-C14
agent: omp
run: 01a07c2d-ee68-7761-b385-72fd6fb2640e
</details>

apps/electron/src/services/agent-auth.ts
```diff
@@ -216,33 +264,50 @@ class AgentAuthService {
...
+      Logger.info(ElectronEvent.AGENT_AUTH_REQUEST, { agentName: authRequest.agentName, agentType: authRequest.agentType, peer: peerKey }, 'AgentAuth');
 
-      // Show consent dialog to user
-      const approval = await this.showConsentDialog(authRequest);
+      // Show consent dialog to user, naming the real requesting process.
+      const approval = await this.showConsentDialog(authRequest, peer);
 
       if (!approval.approved) {
+        // Arm the cool-down so repeated prompts cannot be hammered after a Deny.
+        this.denyCooldowns.set(peerKey, Date.now() + DENY_COOLDOWN_MS);
```
ameer030, line 298: 🔵 **N02 — deny cool-down entries never expire out of the map**

`denyCooldowns` entries are only removed when the same requester is later approved; expired cool-downs from agents never re-requested stay in the map for the app's lifetime. Growth is bounded by distinct denied exe paths so this is small, but the sweep is one line.

- Set at `apps/electron/src/services/agent-auth.ts:298`; deleted only at the post-approval `:309`; the check at `:286-291` reads but never removes stale entries.

</details>

```ts
const cooldownUntil = this.denyCooldowns.get(peerKey);
if (cooldownUntil && Date.now() >= cooldownUntil) this.denyCooldowns.delete(peerKey);
```

</details>

rules: STYLE-C14
agent: omp
run: 01a07c2d-ee68-7761-b385-72fd6fb2640e
</details>

apps/electron/src/services/deep-links.ts
```diff
@@ -188,10 +193,29 @@ async function handleDeepLink(url: string): Promise<void> {
...
+      const rawDomain = urlObj.searchParams.get('domain') || '';
+      const rawTitle = urlObj.searchParams.get('title') || '';
+
+      const text = sanitizeAskAiText(rawText, 8000);
+      const sourceUrl = normalizeAskAiUrl(rawUrl);
+      const domain = normalizeAskAiDomain(rawDomain);
+      const title = sanitizeAskAiText(rawTitle, 512);
+
+      if (rawUrl && !sourceUrl) {
+        Logger.error(ElectronEvent.DEEP_LINK_PARAM_REJECTED, { param: 'url', value: rawUrl.slice(0, 128) });
```
ameer030, line 212: 🔵 **N03 — each rejected deep-link param is logged twice**

Both `Logger.error(ElectronEvent.DEEP_LINK_PARAM_REJECTED, …)` and `log.warn('[DeepLinks] Rejected invalid …')` fire for the same rejection. Keep the structured event (it's the useful audit trail); drop the ad-hoc `log.warn` line. Since these deep links are externally triggerable, also consider whether error-level is right — a page can fire `xyne-spaces://ask-ai?url=javascript:x` at will, so the structured event may be better at warn level with a throttle.

- `apps/electron/src/services/deep-links.ts:212-219` — double logging for both the `url` and `domain` rejection branches.

</details>

Remove the two `log.warn` lines (or keep them and demote the OTEL event) — one system per event.

</details>

rules: STYLE-C04
agent: omp
run: 01a07c2d-ee68-7761-b385-72fd6fb2640e
</details>

ameer030 (commented):
🟡 **W03 — the PR body doesn't mention the OTA updater change, and the template is unfilled**

The title and body cover agent-auth and ask-ai deep links, but the head commit also adds a new integrity gate to the OTA updater (`apps/electron/src/services/ui-updater.ts`, ~45 lines — a different feature area). That change appears nowhere in the description, and every template section (Type of Change, Areas Touched, How-did-you-test tiers) is unchecked. Reviewers can't audit scope they aren't told about; please name the updater change (and the finding id it closes, if it's part of the same pentest sweep) and tick the sections that apply.

- Commits: `bc171db85` (agent-auth + deep links) and `f4a53740b` ("address agent-auth review + **verify OTA bundle checksum**").
- Body bullets: agent-auth and deep-links only; `ui-updater.ts` unmentioned.
- `.github/PULL_REQUEST_TEMPLATE.md` sections in the body are all blank.

</details>

One-line edit to the body: add a `ui-updater:` bullet describing the bundle-checksum gate (and its limits), tick `apps/electron` under Areas Touched, and fill the testing tiers with what the POT recording covers — and what it doesn't.

</details>

rules: PROC-C02, PROC-C04
agent: omp
run: 01a07c2d-ee68-7761-b385-72fd6fb2640e
</details>

ameer030 (commented):
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 1 |
| 🟡 Better to fix | 4 |
| 🔵 Nit | 3 |

### 🔁 Re-check
- 💬 **Existing threads:** 7 resolved, verified — @devesh-juspay's fail-open peer check ✅ fixed in code, per-request shell-out ✅ (cache + skip codesign), Unicode strip class ✅ fixed in code; 4 CodeQL bot comments: temp-file mitigated (`0o600`), 2× file-data-outbound + user-controlled-bypass = inherent to the reference client / intended design
- ✅ **Verified OK:** every token-gated route re-checks the peer (route table audited — no missed caller); Origin/header rejection on `/auth/request`; consent IPC gated by `event.sender.id`; deep-link params sanitized before forwarding and the dashboard consumer drops invalid url; deny cool-down + dialog mutex are race-free in the single-threaded flow; peer-process shell-outs are argv-based (no shell interpolation of untrusted input); POT attached in the body

### 📌 Verdict
1 blocker must be fixed before merge ❌ — **B01**: the OTA checksum gate verifies only one file, so tampered bundle JS passes it. The agent-auth hardening itself (path binding, fail-closed re-checks, consent modal, ask-ai sanitization) is solid.

ameer030 (approved):

PR #1185, opened 2026-08-28
fix: XYNE-60678 ysweet authorization for very first connection

5uryan5h:
POT - https://drive.google.com/file/d/1abh4enOo8MRfuJrkaFPy--ep2aGO5veX/view?usp=drive_link
ENVS Needed
1. In YSWEET - Y_SWEET_AUTH, Y_SWEET_VALIDATE_POLL_SECONDS, Y_SWEET_VALIDATE_RETRY_ATTEMPTS, Y_SWEET_VALIDATE_RETRY_DELAY_MS, Y_SWEET_BACKEND_URL
2. In backend - Y_SWEET_SERVER_TOKEN, 
3. YSWEET_USE_READ_REPLICA (from superposition)

Environment-Changes:
  - Y_SWEET_SERVER_TOKEN: for adding check that ysweet is called by our backend only

  - backend

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


apps/backend/src/app.ts
```diff
@@ -603,6 +593,7 @@ export class App {
     this.app.use('/api/dashboard/claw', authenticateUserOrApp, dashboardClawRouter);
 
 
+    this.app.use('/api/ysweet', ysweetValidateRouter);
```
satvikbatra: This endpoint has no auth at all it's above the /api auth middleware and the router adds none. Anyone can POST a docId + userId and find out who has access to a canvas. Fair enough that authenticate won't work here (y-sweet has no cookie), but something has to check the caller is only y-sweet.

apps/backend/src/controllers/ysweetController.ts
```diff
@@ -4,6 +4,30 @@ import { config } from '@/config/env';
 import { logger } from '@/utils/logger';
 import { getTrustedOriginalHost } from '@/utils/publicUrls';
 import { ysweetGetOrCreateDocAndToken } from '@/utils/ysweetUtils';
+import { DatabaseClient, readReplicaDb } from '@/database/client';
+import { superpositionClient } from '@/services/superpositionClient';
+
+/**
+ * y-sweet polls this every ~10s per open connection, so it's read-heavy and
+ * latency-tolerant (it's a revalidation, not a first-time access decision) —
```
satvikbatra, line 12: validate_user_middleware runs on the WS upgrade so this is the first-connect check too, and it's now reading a replica. On a brand-new canvas we write to the primary and y-sweet validates ~50ms later; if the row hasn't replicated, we 403 and the connection is refused. Same for someone just added as a participant. Can connect-time checks stay on the primary and only the 10s poll use the replica? how is it?

apps/backend/src/controllers/ysweetController.ts
```diff
@@ -221,4 +246,40 @@ export class YSweetController {
       });
     }
   }
+
+  /**
+   * Called by the y-sweet server (not the browser) to re-validate that a
+   * userId still has access to a docId before serving/mutating content or
+   * upgrading a WebSocket connection. No session cookie is available here.
+   */
+  async validateAccess(req: Request, res: Response): Promise<void> {
```
satvikbatra, line 260: This doesn't cover the case where only edit access is removed. it only look at hasAccess

apps/backend/src/app.ts
```diff
@@ -636,10 +627,7 @@ export class App {
     // Memory routes (auth handled internally by dualAuthenticate middleware)
     this.app.use('/api/memory', memoryRoutes);
 
-    // Y-Sweet collaboration routes. Auth already runs for every /api request via
-    // the /api attachment mounts above; applying it here again cost two more DB
-    // round-trips per canvas open.
-    this.app.use('/api/ysweet', ysweetRoutes);
+    this.app.use('/api/ysweet', authMiddleware.authenticate, ysweetRoutes);
```
satvikbatra: isn't auth already handled by the /api route above. adding it here also duplicates it.

apps/dashboard/src/hooks/useCanvasYjsProvider.ts
```diff
@@ -201,6 +206,15 @@ export function useCanvasYjsProvider(options: CanvasYjsProviderOptions): CanvasY
     }
   }, [authTokenData]);
 
+  latestAuthTokenRef.current = authTokenData;
+  const [hasAuthToken, setHasAuthToken] = useState(false);
+  useEffect(() => {
```
satvikbatra, line 230: Both of these fire on mount and B wins, so if the token is already cached A sets true and B resets it to false — and neither dep ever changes again, so the provider below never gets created. Cold load is fine (token is undefined at mount, A's guard skips), but gcTime is 5 min vs staleTime 50 min, so a cache hit is always fresh and never refetches to recover. Net effect: reopen a canvas within 5 minutes and it sits at offline with no error shown. Could we derive it instead - authTokenData?.docId === canvasId?

apps/backend/src/services/callDocumentService.ts
```diff
@@ -2490,7 +2442,12 @@ A comprehensive detailed summary has been generated from this call.
       // If a brand-new canvas + link was already published before the throw,
       // tear it down so an exception doesn't leave a dangling canvas/message.
       if (newCanvasId) {
-        await this.cleanupFailedDetailedSummaryCanvas(newCanvasId, conversationId, callId);
+        if (!xyneAutomaticBotId) {
```
satvikbatra, line 2509: isin't newCanvasId is only assigned well after xyneAutomaticBotId is set

5uryan5h: Its just extra check so that I will get botId for sure instead of adding null as fallback or empty string as fallback

apps/backend/src/app.ts
5uryan5h: Its not specifically check for that api... removing btw

apps/backend/src/controllers/ysweetController.ts
5uryan5h: adding retry logics

apps/backend/src/controllers/ysweetController.ts
5uryan5h: added authorization level as well

apps/dashboard/src/hooks/useCanvasYjsProvider.ts
5uryan5h: fixed

satvikbatra (dismissed):

apps/dashboard/src/hooks/useCanvasYjsProvider.ts
```diff
@@ -201,6 +206,14 @@ export function useCanvasYjsProvider(options: CanvasYjsProviderOptions): CanvasY
     }
   }, [authTokenData]);
 
+  latestAuthTokenRef.current = authTokenData;
+  // Derived, not effect-driven state: two useEffects here (set true when the
+  // token arrives, reset false on canvasId change) both fire on mount, and
+  // when the token is already cached the reset can run after the set and
+  // win, permanently gating out provider creation for that mount. Deriving
+  // it during render has no such ordering to race.
+  const hasAuthToken = authTokenData?.docId === canvasId;
```
devesh-juspay: **[CONFIRMED] Legacy share-link canvases go permanently, silently offline.**

The new gate `hasAuthToken = authTokenData?.docId === canvasId` never passes when the backend deliberately returns a canonical docId different from the requested id. Opening a canvas via a legacy capability URL (route param = `viewAccessId`/`editAccessId`) makes `getClientToken` (ysweetController.ts:176-180) resolve to `canonicalDocId = authResult.canvas.id` and return a token with `docId != canvasId`. `hasAuthToken` stays false forever, the effect returns before `createYjsProvider`, and the canvas sits offline with no error, no toast, no Reconnect banner.

The old code handled this via `actualDocId = authTokenData.docId || canvasId`. Fix: gate on token presence, not id equality — keep using the canonical docId for the provider.

apps/dashboard/src/hooks/useCanvasYjsProvider.ts
```diff
@@ -402,7 +421,16 @@ export function useCanvasYjsProvider(options: CanvasYjsProviderOptions): CanvasY
       setAwareness(null);
       doc.destroy();
     };
-  }, [doc, canvasId, authTokenData, getAuthToken]);
+  }, [doc, canvasId, hasAuthToken, getAuthToken, reconnectNonce]);
```
devesh-juspay, line 443: **[CONFIRMED] Transient auth-refetch failure leaves an editable-but-dead editor, with no auto-recovery.**

The effect deps changed from `authTokenData` (object identity) to the boolean `hasAuthToken`, removing the automatic provider rebuild after a transient failure. When a background refetch of `/api/ysweet/auth` fails (3 retries), the authError effect destroys the provider and sets read-only; react-query keeps previous data so `hasAuthToken` stays true, and a later *successful* refetch replaces `authTokenData` but changes no effect dep — so the provider is never rebuilt (pre-PR it was). Worse, the success re-runs the isReadOnly effect and can flip the editor back to editable while `providerRef` is null: it looks editable but nothing syncs until a full reload.

Fix: key the provider effect on something that changes when a fresh token arrives (token identity or a version counter), and don't restore editability while the provider is null.

apps/backend/src/middleware/ysweetServerAuth.ts
```diff
@@ -0,0 +1,44 @@
...
+export const requireYSweetServerToken = (
+  req: Request,
+  res: Response,
+  next: NextFunction
+): void => {
+  const expectedToken = config.ysweet.serverToken;
+  const authHeader = req.headers.authorization;
+  const providedToken = authHeader?.startsWith('Bearer ') ? authHeader.slice(7) : undefined;
+
+  if (!expectedToken || !providedToken) {
```
devesh-juspay, line 22: **[PLAUSIBLE] A missing `Y_SWEET_SERVER_TOKEN` silently 401s every validate call, taking down all canvases workspace-wide.**

`Y_SWEET_SERVER_TOKEN` Joi-defaults to `''` (env.ts), so `requireYSweetServerToken`'s `!expectedToken` fails closed with 401 on every `/api/ysweet/validate` callback. If an environment upgrades to the validating y-sweet but misses this new var (no compose/env file in-repo sets it), y-sweet drops every canvas WebSocket and collaboration is down with only generic 401s in the logs.

Fix: fail fast at boot — if the ysweet-validation feature is enabled, assert `Y_SWEET_SERVER_TOKEN` is non-empty and log a clear startup error tying the token to the feature.

apps/backend/src/controllers/ysweetController.ts
```diff
@@ -4,6 +4,30 @@ import { config } from '@/config/env';
...
+import { superpositionClient } from '@/services/superpositionClient';
+
+/**
+ * y-sweet polls this every ~10s per open connection, so it's read-heavy and
+ * latency-tolerant (it's a revalidation, not a first-time access decision) —
+ * route it to the read replica when available so it can't add load to the
+ * primary as connection count scales, same pattern as analyticsRepository.
+ */
+async function getValidateDbInstance() {
+  const useReadReplica = await superpositionClient.getBooleanValue(
```
devesh-juspay: **[CONFIRMED] If Superposition never initialized, every ~10s validate poll 500s and drops healthy collaborators.**

`getValidateDbInstance` calls `superpositionClient.getBooleanValue`, whose `ensureInitialized()` sits *outside* its internal try/catch. `app.ts` swallows Superposition startup failure ("Continuing startup without Superposition client..."), so with `isInitialized=false` every validate call re-enters `initialize()`, which throws ("SuperpositionProvider not available"); the throw propagates through `getValidateDbInstance` into `validateAccess`'s catch → 500 to y-sweet on every revalidation poll for every open connection, disconnecting users who provably have access.

Fix: wrap the flag lookup and fall back to the primary DB on any error, so a config-service outage never denies document access.

apps/backend/src/controllers/ysweetController.ts
```diff
@@ -221,4 +246,57 @@ export class YSweetController {
...
+        (authorization !== 'full' && authorization !== 'read-only')
+      ) {
+        res.status(400).json({
+          error: 'Invalid request',
+          message: "docId, userId are required strings and authorization must be 'full' or 'read-only'",
+        });
+        return;
+      }
+
+      const authResult = await canvasAuthService.checkCanvasAccess(docId, userId, await getValidateDbInstance());
```
devesh-juspay, line 278: **[PLAUSIBLE] Replica lag can 403 the very first connection to a just-created canvas.**

With `YSWEET_USE_READ_REPLICA` on, `validateAccess` reads from the replica. But this is the *first-connection* path (the PR's title): `createCanvasForUser` writes the canvas + OWNER participant to the primary and mints a token; the browser connects immediately and y-sweet calls `/api/ysweet/validate`. `checkCanvasAccess` on a lagging replica misses the just-written rows → `authResult.canvas` null → 403 `{allowed:false}`, and the frontend burns its 5 retries into the failure banner. The in-code comment calling this "a revalidation, not a first-time access decision" is not accurate for new canvases.

Fix: use the primary for the first validation (or fall back to primary on a replica miss before denying).

apps/dashboard/src/hooks/useCanvasYjsProvider.ts
```diff
@@ -303,10 +317,14 @@ export function useCanvasYjsProvider(options: CanvasYjsProviderOptions): CanvasY
       };
     }
 
-    if (!authTokenData) return;
+    if (!hasAuthToken || !latestAuthTokenRef.current) return;
```
devesh-juspay, line 340: **[CONFIRMED] Reconnect is a silent no-op for prefetched canvases** (mechanism spans this effect and the `if (prefetchedCanvas?.provider)` branch around line 229).

For a canvas opened via `canvasPrefetchService`, after 5 connection errors the handler destroys that provider and sets `connectionFailed`. Clicking Reconnect bumps `reconnectNonce` and re-runs this effect, but `prefetchedCanvas` is memoized on `[canvasId]` and never cleared, so the effect re-enters the prefetched-provider branch and re-installs the *already-destroyed* provider (which emits no further events), returning before `createYjsProvider`. `reconnect()` already cleared `connectionFailed`, so the banner disappears while nothing is connected — no recovery short of a refresh.

Fix: clear/ignore `prefetchedCanvas.provider` once it's been destroyed so Reconnect falls through to `createYjsProvider`.

apps/dashboard/src/hooks/useCanvasYjsProvider.ts
```diff
@@ -402,7 +421,16 @@ export function useCanvasYjsProvider(options: CanvasYjsProviderOptions): CanvasY
       setAwareness(null);
       doc.destroy();
```
devesh-juspay: **[CONFIRMED] Reconnect destroys the memoized `Y.Doc`, then hands the same destroyed doc to the new provider.**

`doc = useMemo(() => prefetchedCanvas?.doc ?? new Y.Doc(), [prefetchedCanvas])` is stable across `reconnectNonce` bumps. On Reconnect the provider effect's cleanup runs `provider.destroy()` **and** `doc.destroy()` (which strips doc observers; BlockNote/awareness bindings tear down on the `destroy` event), then the re-run attaches a fresh provider to the destroyed doc. The banner clears but remote updates no longer reach the editor and local edits aren't reliably synced until a page refresh.

Fix: don't `doc.destroy()` in the cleanup of an effect that will re-run for reconnect (destroy the doc only on unmount/canvasId change), or recreate the doc alongside the provider.

devesh-juspay (commented):
Automated review (Claude Code) — 10 findings (6 CONFIRMED, 4 PLAUSIBLE), all inline. Reviewed alongside the companion y-sweet fork PR (juspay/y-sweet#3).

**Cross-PR note:** two findings here (wiki sync at SdlcWikiPageStore.ts:1037, Confluence import at confluenceImportService.ts:1424) are rated PLAUSIBLE because their failure depends on the y-sweet side actually enforcing the new `/api/ysweet/validate` callback for server-minted tokens — which is that PR's stated purpose. y-sweet#3 additionally 403s *server-permission* tokens outright, so if these backend sync paths use the connection-string server token they break unconditionally once both ship. Deploy ordering also matters: `Y_SWEET_SERVER_TOKEN` (finding at ysweetServerAuth.ts:22) must be set before the validating y-sweet goes live, and y-sweet#3's token wire-format change invalidates all existing tokens at cutover.

Dropped after verification: an app.ts mount-order auth hole, missed ysweet-util call sites, and infinite-retry suspicions (all refuted). Cut by the 10-finding cap: `requireYSweetServerToken` duplicates `internalServiceAuth.ts` near-verbatim; `getValidateDbInstance` is a third copy of the replica-selection helper; and the `dbClient: typeof db = db` defaults on canvasAuthService's private helpers are a silent-primary footgun.

apps/backend/src/sdlc/wiki/SdlcWikiPageStore.ts
```diff
@@ -1034,7 +1037,7 @@ export class SdlcWikiPageStore {
     }
```
devesh-juspay, line 1039: **[PLAUSIBLE — cross-PR] A plain channel member's wiki execution fails once y-sweet revalidates server-minted tokens.**

`syncCanvas` mints a 'full' y-sweet token as `input.actorId`, but wiki canvases are SDLC baseline canvases where `checkCanvasAccess.canEdit` requires creator/owner/editor or `ChannelRole.ADMIN`. A non-admin channel member triggering a wiki page action → `validateAccess` returns 403 → the sync fails and the whole execution aborts with 'Y-Sweet sync failed for Wiki page' (503) for users who could previously run it.

This fires only once y-sweet#3 enforces `/api/ysweet/validate` for these tokens (and note that PR *also* 403s server-permission tokens outright). Fix: validate server-originated syncs against the actor's actual edit rights before minting a 'full' token, or use a token type the revalidation path accepts for trusted backend writes.

apps/backend/src/services/confluence/confluenceImportService.ts
```diff
@@ -1421,7 +1421,7 @@ export class ConfluenceImportService {
       },
     });
 
-    await syncToYSweet(canvasId, safeContent as unknown as BlockNoteBlock[]);
+    await syncToYSweet(canvasId, safeContent as unknown as BlockNoteBlock[], lastEditorUserId);
```
devesh-juspay, line 1430: **[PLAUSIBLE — cross-PR] Confluence re-import writes can be denied because the attributed editor isn't a canvas participant.**

`updateCanvasContent` calls `syncToYSweet(canvasId, ..., lastEditorUserId)` (the mapped Confluence last editor) without ensuring that user is a participant — only the `updateCanvasRecord` path calls `ensureCanvasOwners`. `checkCanvasAccess(canvasId, lastEditorUserId).canEdit` is then false (not creator, no participant row), so once y-sweet enforces validation for these 'full' tokens the re-imported content silently never lands in the collaborative doc.

Fix: add the attributed editor as a participant (as the create path does) before syncing, or attribute the sync to a guaranteed-authorized identity.

apps/backend/src/services/canvasAuthService.ts
```diff
@@ -215,18 +225,18 @@ class CanvasAuthService {
       }
 
       const isCreator = canvas.createdBy === userId;
-      const currentUserContext = await this.getCurrentUserContext(userId);
+      const currentUserContext = await this.getCurrentUserContext(userId, dbClient);
```
devesh-juspay, line 230: **[CONFIRMED — efficiency] Each validate call runs 6-12 serial DB queries, every ~10s per open connection, uncached.**

`validateAccess` runs the full `checkCanvasAccess` walk: canvas, user, SDLC channel-admin, canvasParticipant, userGroupMapping, a per-channel share loop with sequential awaits, and guest checks re-run up to 3×. At 100 open connections that's ~10 validate calls/sec = 60-100+ queries/sec of pure revalidation load, all serial RTTs, on the primary whenever the flag is off or the replica is absent.

Fix: a short-TTL `(docId, userId)` cache, `Promise.all` over the independent lookups, and a batched `channelParticipant.findMany` would cut this by roughly an order of magnitude.

5uryan5h: its on read replica, so wont cause any issue

apps/dashboard/src/hooks/useCanvasYjsProvider.ts
5uryan5h: we modified canvas few months back, wont be there much legacy users

apps/dashboard/src/hooks/useCanvasYjsProvider.ts
5uryan5h: fixed

apps/backend/src/middleware/ysweetServerAuth.ts
5uryan5h: added to fail on startup itself so that there wont be any fallback

apps/backend/src/controllers/ysweetController.ts
5uryan5h: added try catch

apps/backend/src/controllers/ysweetController.ts
5uryan5h: I discussed this with Kushagar... there is no way to find that this is first connection and retry logic try exponentially, with increase interval. Its expected behavior

apps/dashboard/src/hooks/useCanvasYjsProvider.ts
5uryan5h: fixed

apps/dashboard/src/hooks/useCanvasYjsProvider.ts
5uryan5h: fixed

apps/backend/src/sdlc/wiki/SdlcWikiPageStore.ts
5uryan5h: adding userId as createdBy.userId instead of just actor.userId, which will call ysweet with createdBy itself

apps/backend/src/services/confluence/confluenceImportService.ts
5uryan5h: fixed

devesh-juspay (approved):

satvikbatra (approved):

apps/backend/src/sdlc/wiki/SdlcWikiPageStore.ts
```diff
@@ -852,7 +856,8 @@ export class SdlcWikiPageStore {
         if (input.action.action !== 'archive') {
           const repaired = await this.dependencies.syncCanvas(
             existing.id,
```
devesh-juspay, line 858: **[CONFIRMED] The durable-revision repair-sync still writes as `input.actorId` — re-opening the exact 403 the main-path fix eliminated.**

The main wiki sync was fixed to run as `existing?.createdBy ?? input.actorId` precisely because triggering actors are typically channel VIEWERs who lack `canEdit` on the baseline canvas. But this sibling repair-sync (in the durable-revision recovery branch) still passes `input.actorId`.

A channel viewer re-running a wiki update whose revision already committed enters this branch → `syncCanvas(existing.id, content, input.actorId)` requests authorization 'full' → y-sweet revalidates → `canEdit=false` → 403 → `syncToYSweet` returns false → `throw new AppError('Y-Sweet sync failed for Wiki page', 503)`. Same failure mode the fix was meant to prevent, one branch over.

Fix: use the same `syncActorId` (`existing?.createdBy ?? input.actorId`) here as the main sync at ~line 1049.

devesh-juspay (commented):
Automated re-review (Claude Code) at head `74c18cb40`. The 'resolving comments' commit fixed most of the previous round (the docId gate, both Reconnect bugs, the transient-auth recovery, the Superposition-throw-on-every-poll, and the main-path wiki-sync authorization). Three findings remain, attached inline — the first re-opens this PR's own core fix on the recovery path.

apps/backend/src/services/callDocumentService.ts
```diff
@@ -2498,7 +2506,12 @@ A comprehensive detailed summary has been generated from this call.
...
+          throw new Error(
```
devesh-juspay, line 2510: **[PLAUSIBLE] `throw` inside the cleanup `catch` masks the original error and skips the teardown this block exists to do.**

This is the failure-cleanup path: if a brand-new canvas was already published before the throw, it should be torn down. But when `newCanvasId` is set and `xyneAutomaticBotId` is unresolved, the code throws a *new* error instead — replacing the real caught error and leaving the canvas/message dangling (the opposite of the block's purpose).

In the current ordering the bot is resolved before any canvas is created, so this path is effectively unreachable — but it's a latent trap if that ordering ever changes. Fix: log-and-skip the cleanup (or resolve the bot id defensively) rather than throwing, so the original error still propagates and any partial canvas is still cleaned up when possible.

apps/backend/src/controllers/ysweetController.ts
```diff
@@ -4,6 +4,35 @@ import { config } from '@/config/env';
...
+/**
+ * y-sweet polls this every ~10s per open connection, so it's read-heavy and
+ * latency-tolerant (it's a revalidation, not a first-time access decision) —
+ * route it to the read replica when available so it can't add load to the
+ * primary as connection count scales, same pattern as analyticsRepository.
+ */
+async function getValidateDbInstance() {
+  let useReadReplica = false;
+  try {
+    useReadReplica = await superpositionClient.getBooleanValue(
```
devesh-juspay, line 19: **[PLAUSIBLE — efficiency] The `YSWEET_USE_READ_REPLICA` flag is looked up on every validate poll (~10s per open connection).**

`getValidateDbInstance` calls `superpositionClient.getBooleanValue('YSWEET_USE_READ_REPLICA', ...)` on each `validateAccess` invocation (called at line 278). As concurrent canvas connections scale, this multiplies flag-evaluation work with connection count on the exact path the PR is trying to keep cheap — and if the provider's boolean read isn't purely local/cached, each is a round trip.

Fix: resolve the flag once and cache it with a short TTL (or read it at connection start rather than per-poll).

PR #1202, opened 2026-08-29
feat: support thread replies in message automations

XyneSpaces:
1. Problem Description:
The `MESSAGE_RECEIVED` automation trigger only emitted for messages that started a channel conversation, so a matching reply inside an existing thread could not trigger actions such as a call alert. No XYNE ticket key was supplied for this change.

The backend emitted `MESSAGE_RECEIVED` only when `conversation.initialMessageId === message.messageId`; both the service write path and Zero side-effect path excluded thread replies.

- Add `New conversation`, `Thread reply`, and `Any message` location choices to the existing trigger.
- Add optional `conversationIds` filtering for targeting specific threads.
- Emit location-aware events from both ConversationService and Zero message side effects.
- Preserve existing automation behavior by defaulting missing configuration and older queued payloads to `New conversation`.
- Pre-filter thread-reply candidates in the event router so legacy automations do not create skipped execution rows or queue jobs for every reply.

4. Documentation (link to docs created/updated for this change):
N/A

5. DB Alter Details (if any):
N/A

6. Data fix Details (if any):
N/A

7. Environment variable Changes (if any):
N/A

- Focused Jest suite: 6 tests passed (`message-received.trigger.test.ts`).
- Backend TypeScript check: `pnpm typecheck` passed.
- `git diff --check` passed.
- Targeted ESLint found only pre-existing errors outside changed lines in `conversationService.ts` and `messages-handler.ts`; the new automation files passed.

9. Services to which this change is to be deployed:
Backend and Zero side-effect worker processes that consume the backend codebase.

10. Main PR link (if this PR is raised against a release branch):
N/A

### DIFF
--- a/apps/backend/src/automations/engine/event-router.ts
+++ b/apps/backend/src/automations/engine/event-router.ts
@@ -6,6 +6,7 @@ import { currentUpstreamChain } from './automation-context-storage';
 import {
   AUTOMATION_WORKFLOW_TYPE,
   DESK_AUTOMATION_WORKFLOW_TYPE,
+  parseAutomationConfig,
   parseAutomationMetadata,
   triggerTypeToEventType,
 } from '../types/workflow-adapter';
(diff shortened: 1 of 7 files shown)


XyneSpaces:
Superseded by #1203.

PR #1213, opened 2026-08-30
feat: XYNE-60886 remove mixpanel, add posthog

sithaarth:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


apps/dashboard/src/services/Analytics/posthogService.ts
```diff
@@ -0,0 +1,225 @@
...
+        // Enable click + scroll heatmaps in the PostHog dashboard.
+        enable_heatmaps: true,
+        // Detect rage clicks / dead clicks for friction analysis.
+        rageclick: true,
+        // Persist identity across reloads.
+        persistence: 'localStorage+cookie',
+        // Record the full session so the journey can be replayed.
+        disable_session_recording: false,
+        session_recording: {
+          maskAllInputs: true,
```
XyneSpaces, line 126: **[major] R1/Security** — This enables PostHog session replay for the whole dashboard whenever a key is configured, but only masks inputs and `[data-ph-mask]` text; pages such as AI chat, calls, canvases, automation templates, and workspace/member data can be recorded as readable text. Please default session replay off or gate it to explicit production/user-consented rollout, and mask all sensitive app text by default before enabling replay.

XyneSpaces (commented):
Xyne Spaces Reviewer: found 1 issue to address before merge.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1213

**Reviewed:** 34 changed-file entries from the PR files API plus direct reads of the new analytics/Button files · **Findings:** 0 blocking, 1 major, 0 nits

### Findings
| major | R1/Security | `apps/dashboard/src/services/Analytics/posthogService.ts:77` | PostHog session replay is enabled for the whole dashboard whenever a key is configured, while only inputs and `[data-ph-mask]` text are masked; sensitive app text from AI chat/calls/canvases/automation/workspace pages can be recorded. |

### Recommended reviewer
**@shubhamdhamija-juspay** (already requested) — current requested reviewer, not the PR author, and the only visible eligible reviewer from the PR metadata.

| @shubhamdhamija-juspay | Dashboard analytics/UI area (current requested reviewer; path-author lookup unavailable to this integration) | CODEOWNERS unavailable/not found via API | 1 visible in open-PR scan | Already requested and taggable in Spaces |

Rationale: Shubham is already requested on GitHub and has the lowest visible workload signal available from the API scan (one visible open requested review, this PR). CODEOWNERS and per-path commit ownership could not be fully computed because the integration could not read CODEOWNERS/path commit history.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1213 — correction

The earlier review comment on `apps/dashboard/src/services/Analytics/posthogService.ts` should be treated as **withdrawn** for this PR: the follow-up full changed-file scan shows `posthogService.ts` is **not modified** in #1213, and there are no `.env`, lockfile, or type-definition changes in the PR.

**Corrected verdict:** Approve-worthy after nits / no blocking findings from the changed hunks reviewed.

**Reviewed:** 30 files, all under `apps/dashboard`; no migration files touched. **Findings:** 0 blocking, 0 major, 0 nits.

### Findings
| — | — | — | No issues to report in the changed hunks. |

### Recommended reviewer
**@shubhamdhamija-juspay** (already requested) — only visible eligible reviewer from PR metadata; currently requested on only this open PR in this repo.

| @shubhamdhamija-juspay | Default/requested reviewer signal; no CODEOWNERS present | no CODEOWNERS file found | 1 | already requested |

Rationale: no CODEOWNERS file exists, path authorship only identified the PR author (`sithaarth`), who is excluded; Shubham is already requested and has one visible pending review request in this repo.

ameer030 (commented):
Review of the Mixpanel to PostHog migration. The `captureActionOutcome` split is the right seam (autocapture can only see the click, not the async result), `trackProps` values are all non-PII enums, the blur handler records only `input_length` and never content, and the existing `data-track-*` WebSocket activity tracking is left fully intact (0 attributes removed). Two things I want resolved before this merges, plus a few smaller items inline.

**1. Autocapture ships element text — this is the privacy gap the masking commit did not close.**
`mask_all_text` is left at its default of `false`, so every autocaptured click sends the clicked element's `$el_text` and elements chain. In Chat, Desk and Tickets that means message bodies, email subjects, ticket titles and contact names leave the product. The `session_recording` masking added in "mask sensitive info" is a separate config that covers replay only — it does not touch autocapture properties. Details inline on `posthogService.ts`.

**2. SPA pageviews regress to a single pageview per page load.**
`capture_pageview: true` overrides the `'history_change'` that `defaults: '2025-05-24'` would otherwise resolve to. `main` had `defaults` alone and therefore had history-change pageviews; this PR turns them off. Details inline.

**3. 216 raw `<button>` elements became `<Button>`, only ~22 got sizing compensation.**
`buttonVariants` injects `h-9`, `justify-center` and `[&_svg:not([class*="size-"])]:size-4` into every one of them. Some call sites got `h-auto` / `justify-start` / `rounded-none` added, most did not. Concrete example inline on `AskAiRatingButtons.tsx`. This needs a visual pass across the touched surfaces, or a `size` variant that carries no height and no justify.

**4. Performance telemetry was deleted, not migrated.**
The `PERFORMANCE_METRIC` event and its property types are gone along with five call sites that were actively emitting: API latency (`apiClient.ts`), thread message load success and failure (`ThreadPannel.tsx`), call join and call connection failure (`roomMachine.ts`), and WebSocket reconnect attempts (`socketClient.ts`). The `ZERO_CONNECTION_*` constants were already dead, so those are fine to drop. Was losing the other five intentional? It is not mentioned in the description.

**Out-of-diff / housekeeping**
- `apps/dashboard/vite.config.ts:204` still has `'process.env.VITE_MIXPANEL_TOKEN': JSON.stringify(env.VITE_MIXPANEL_TOKEN)`, now dead.
- `apps/dashboard-external/package.json:27` still depends on `mixpanel-browser`. In scope for this PR or deliberately left?
- The PR title says `XYNE-60886`; all five commits say `XYNE-60866`.
- The description is the untouched template — 186 files, no summary, no boxes ticked, no test evidence. For a change that turns on session replay and autocapture across the whole dashboard, the privacy and manual-QA sections are worth filling in properly.

apps/dashboard/src/services/Analytics/posthogService.ts
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    try {
+      /* eslint-disable @typescript-eslint/naming-convention */
+      posthog.init(key, {
+        api_host: host,
+        defaults: '2025-05-24',
+        // --- Full user journey capture ---
+        // Automatically capture every click (all buttons), change and submit event.
+        autocapture: true,
```
ameer030: **Blocker — autocapture sends element text.**

With `autocapture: true` and `mask_all_text` left at its default of `false`, every autocaptured click ships the clicked element's `$el_text` plus the elements chain to PostHog. In this app that is chat message bodies, email subjects, ticket titles and contact names.

The `session_recording` block below masks replay, which is a different pipeline. It does not redact autocapture event properties, so the "mask sensitive info" commit does not cover this path.

```js
mask_all_text: true,
```

PostHog's built-in `shouldCaptureElement` only excludes password/hidden inputs and fields whose name matches a card/SSN pattern — nothing that helps with message content.

Worth considering alongside it: `mask_all_element_attributes: true`. Please confirm `data-ph-capture-attribute-*` still flows through before enabling that one, since all 285 `trackId`s depend on it.

apps/dashboard/src/services/Analytics/posthogService.ts
```diff
@@ -0,0 +1,218 @@
...
+        // Track navigation so the journey is stitched across pages/routes.
+        capture_pageview: true,
```
ameer030: **Blocker — this disables SPA route tracking, which is the opposite of what the comment above says.**

In posthog-js, `defaults: '2025-05-24'` resolves `capture_pageview` to `'history_change'`:

```ts
capture_pageview: defaults && defaults >= '2025-05-24' ? 'history_change' : true
```

and the `HistoryAutocapture` extension that patches `pushState`/`replaceState`/`popstate` only runs when the value is exactly `'history_change'`. Setting it to `true` opts back into the legacy behaviour: one `$pageview` at page load and nothing on navigation.

This is a regression against `main`, which passed `defaults` with no override and therefore had history-change pageviews working.

Fix: drop the line and let `defaults` supply it, or set it explicitly to `'history_change'`.

apps/dashboard/src/services/Analytics/posthogService.ts
```diff
@@ -0,0 +1,218 @@
...
+      return;
+    }
+    this.capture(`${trackId}_${status}`, { trackId, status, ...properties });
+  }
+
+  /**
+   * Identify a user. All properties except id/picture become person properties.
+   */
+  identify(user: PosthogUser): void {
+    if (!this.isInitialized || !user?.id) {
```
ameer030: The old `MixpanelService.identify` called `this.initialize()` when it had not been initialised yet; this version returns early instead.

`authActor` is a module-level XState actor that lives outside React, while `initialize()` now only runs from `AnalyticsProvider`'s `useEffect`. Any `identify` that fires before that effect is dropped silently, and since it only runs on the `authenticated` state entry it is never retried — the user stays anonymous for the rest of the session.

In practice session restore is async so the effect almost certainly wins the race, which is why this is a note rather than a blocker. Still, restoring the `initialize()` fallback costs one line and removes the ordering dependency entirely.

apps/dashboard/src/components/ui/Button/Button.tsx
```diff
@@ -48,19 +95,56 @@ function Button({
...
+    () => buildCaptureAttributes(trackId, trackProps),
+    [trackId, trackProps],
+  );
+
+  const handleClick = React.useCallback(
+    (event: React.MouseEvent<HTMLButtonElement>): void => {
+      // Clicks are recorded by PostHog autocapture; the Button only owns the
+      // action lifecycle here. When a trackAction is supplied, run it with a
+      // pending spinner that blocks double-submits.
+      if (trackAction) {
```
ameer030, line 126: When `trackAction` is set, `onClick` is silently dropped — the early `return` skips the `onClick?.(event)` below.

The JSDoc documents it, but a caller who passes both gets a dead handler with no warning at compile time or runtime. `ThreadAssistDock.tsx:860` already does exactly this.

Either run both, or make them mutually exclusive in the type so passing both fails to compile.

Separately: `event` is handed to an async `trackAction`, so anything reading `event.currentTarget` after an `await` gets `null`. Worth a line in the JSDoc.

apps/dashboard/src/components/ui/Button/Button.tsx
```diff
@@ -48,19 +95,56 @@ function Button({
...
+  trackAction,
+  trackProps,
   ...props
 }: ButtonProps): React.JSX.Element {
   const Comp = asChild ? Slot : 'button';
-  const isDisabled = disabled || loading;
+  const [pending, setPending] = React.useState(false);
+  const isDisabled = disabled || loading || pending;
+
+  const captureAttributes = React.useMemo(
```
ameer030: This `useMemo` never hits. Every call site passes `trackProps` as an inline object literal (`trackProps={{ level: level.value }}`, `trackProps={{ tab: activeTab }}`, …), so the dependency is a fresh reference on every render and `buildCaptureAttributes` re-runs regardless.

It is a handful of string concatenations — just call it directly and drop the memo.

apps/dashboard/src/components/AIScreen/AskAiRatingButtons.tsx
```diff
@@ -158,7 +165,7 @@ export function AskAiRatingButtons({
             className='rounded-md bg-secondary px-2 py-0.5 text-[11px] text-foreground transition-colors hover:bg-muted'
```
ameer030, line 166: **Concrete instance of the Button-swap sizing problem.**

`cn` is `twMerge(clsx(...))` and cva appends `className` last, so caller classes win on genuine conflicts — but only for properties the caller actually sets.

This className is `rounded-md bg-secondary px-2 py-0.5 text-[11px] …`. It sets no `h-*`, so the `h-9` from `size: 'default'` survives: an 11px inline Save button is now 36px tall. `px-4 py-2` are correctly overridden by `px-2 py-0.5`, which is what masks the problem — the padding looks right and the height does not.

The two rating buttons above have a second issue. The base variant carries `[&_svg:not([class*="size-"])]:size-4`, and the `ThumbsUp`/`ThumbsDown` icons use `h-3.5 w-3.5` rather than `size-3.5`. The parent selector wins on specificity (0,2,1 against 0,1,0), so 14px icons render at 16px. tailwind-merge cannot help here — the classes are on different elements.

This pattern repeats across the ~190 swapped buttons that did not get `h-auto` or an explicit `size`. A `size: 'inline'` variant with no height and no `justify-center` would fix the whole set in one place.

apps/dashboard/src/services/clients/apiClient.ts
```diff
@@ -159,11 +123,9 @@ apiConfig.interceptors.response.use(
     // Token refresh is now handled by backend via HTTP-only cookies
     // No need to manually update tokens from frontend
 
-    // Track API latency for successful requests
     const config = response.config as InternalAxiosRequestConfig & {
```
ameer030, line 125: `trackApiLatency` is deleted here and the whole `PERFORMANCE_METRIC` event family goes with it, across five sites that were actively emitting:

- `apiClient.ts` — API latency on both success and failure paths
- `ThreadPannel.tsx` — `MESSAGES_LOADED` / `MESSAGES_LOAD_FAILED`
- `roomMachine.ts` — `CALL_JOIN` / `CALL_CONNECTION_FAILED`
- `socketClient.ts` — `WS_CONNECTION_RECONNECT_ATTEMPT`

(The `ZERO_CONNECTION_*` constants had no call sites, so dropping those is pure cleanup.)

The `logger.info` calls below still record latency to logs, so this is not a total blackout — but the analytics-side perf and reliability signal is gone, and nothing in the PR description says that was the intent. If it was deliberate, please say so in the description; otherwise these want porting to `posthogService.capture`.

apps/dashboard/src/components/ui/InputBox/InputBox.tsx
```diff
@@ -1178,102 +1189,110 @@ export const InputBox = forwardRef<InputBoxHandle, InputBoxProps>(
...
+          }
         }
 
-        // Get fresh content after inserting link
-        const finalHtmlContent = editor?.getHTML() || htmlContent;
-        const finalPlainText = editor?.getText().trim() || plainText;
+        // Record how the message was sent (Enter key vs. send button). Autocapture
+        // sees the button click but is blind to keyboard sends, so most sends would
+        // otherwise be invisible; this makes the Enter-vs-button split measurable.
+        posthogService.capture('message_send', { trigger, composer: 'chat' });
```
ameer030: Two small things.

This fires before `await onSendMessage(...)`, so a send that throws still counts as a `message_send`. If the intent is a success metric, move it after the await; if it is deliberately an attempt metric, the name is misleading — something like `message_send_attempted` would read better.

Also, this is a bare string while the file already imports `EVENTS` and uses it a few hundred lines up. Either add `MESSAGE_SEND` to `events.ts` or the constant file stops being a reliable index of what the app emits.

The keyboard-vs-button `trigger` split itself is a good addition — autocapture genuinely cannot see Enter-key sends.

apps/dashboard/src/components/SummaryTemplateMenu/SummaryTemplateMenu.tsx
```diff
@@ -71,20 +72,23 @@ export function SummaryTemplateMenu({
...
+        <Button
           type='button'
+          variant='ghost'
           disabled={isRegenerating || !canRegenerate}
           onClick={() => {
             onRequestClose();
             onRegenerate();
           }}
+          trackId='regenerate_selected_summary_template'
+          trackProps={{ trackCategory }}
```
ameer030: React sets unknown attributes via `setAttribute`, which lowercases attribute names on HTML elements. This renders as `data-ph-capture-attribute-trackcategory`, so the property arrives in PostHog as `trackcategory` rather than `trackCategory`.

Harmless on its own, but it means camelCase keys in `trackProps` silently become lowercase while the others in this PR (`duration`, `level`, `tab`, `action`) happen to be single words and look fine. Worth standardising on kebab-case keys so the property names are predictable.

apps/dashboard/src/main.tsx
```diff
@@ -30,7 +30,7 @@ const handleConsoleError = (args: unknown[]): void => {
       message: errorMessage,
       ...getCommonErrorProperties(),
     };
-    mixpanelService.track('Frontend Error', properties);
+    posthogService.capture('Frontend Error', properties);
     logger.error(Event.FRONTEND_ERROR, { ...properties, error });
   } catch (trackingError) {
     originalConsoleError('Failed to track browser console error to Mixpanel:', trackingError);
```
ameer030: Leftover from the rename — this and the same string at lines 58 and 84 still say "Mixpanel" while the call above them is `posthogService.capture`.

apps/dashboard/src/components/Chat/TwinReplyDraft/ThreadAssistDock.tsx
```diff
@@ -858,6 +860,8 @@ function ReplyCard({
             onClick={() => void send()}
             data-track-category='twin-dock'
             data-track-name='send-draft'
+            trackId='twin_send_reply'
+            trackAction={send}
```
ameer030, line 863: Both `onClick={() => void send()}` and `trackAction={send}` are passed. `Button` returns early when `trackAction` is set, so the `onClick` never runs — `send` still fires exactly once, so no double-submit, but the `onClick` prop is dead code.

Drop the `onClick` line. See the note on `Button.tsx:118` about making this a compile error rather than a silent no-op.

shubhamdhamija-juspay (dismissed):

ameer030 (dismissed):

ameer030 (approved):

shubhamdhamija-juspay (approved):

PR #1221, opened 2026-08-30
feat: XYNE-60735 multi-repo SDLC hubs

ameer030:
## What

An SDLC hub used to *be* a repository. Adding a repository minted a private channel and bound the two 1:1 forever, so a team space could never cover more than one repository, and a repository could never appear in more than one space.

A hub now covers several repositories, and a repository can belong to several hubs.

## How

Membership is a `CHANNEL -> REPOSITORY` edge in `sdlc_entity_links`, and a track's hub is a `CHANNEL -> TRACK` edge in the same table. Neither needs a new table. Both are structural rather than content, so every read of the content graph excludes them through `SDLC_STRUCTURAL_RELATIONS`, and the generic link API refuses to create or delete them — membership changes only through the hub endpoints.

The URL is now the hub: `/sdlc/:channelId/:section` replaces `/sdlc/:repoId/:section`, including the standalone artifact window added in #1194.

New endpoints under `/sdlc/channels` create a hub with a chosen set of repositories, and add or remove repositories afterwards. Detaching removes the membership edge only — it never deletes the repository, and a hub cannot drop to zero repositories.

## Migration

`20260831033955_sdlc_multirepo_add` is **additive only**. `sdlc_entity_links` gains `channelId`, and the legacy `repoId` columns on `sdlc_entity_links`, `sdlc_tracks` and `sdlc_artifacts` become nullable. Nothing is dropped, so a lane still running the old code keeps working. The legacy columns stay in the schema marked `@deprecated` and are removed in a later change, once every environment has run the backfill.

Deploy order:

1. Apply `20260831033955_sdlc_multirepo_add`.
2. Deploy this branch.
3. `POST /api/sdlc/cleanup/multirepo-backfill` with `{"dryRun": false}`.

Between 2 and 3 no row carries a `channelId` yet, so SDLC reads empty. Nothing is lost and nothing is written wrong; the surface returns the moment the backfill runs. The endpoint previews by default, is idempotent, and a preview reporting all zeros is what says the migration is complete.

## Testing

Run end to end against a local database holding a real hub:

- dry run reported `hubsSeen: 1, membershipCreated: 1, trackEdgesCreated: 1, linksStamped: 4`, matching the row counts exactly, and wrote nothing
- the real run produced the same numbers; afterwards every link carries a `channelId` and both structural edges exist
- re-running the preview returns all zeros, confirming idempotence
- `prisma migrate diff` against the migrated database reports an empty migration, so schema and database agree
- backend and dashboard `tsc --noEmit` clean; `sdlcMembershipRows.test.ts` passes 7/7

## Known follow-ups

- Repo Knowledge and Wiki still read the hub's first repository only; the per-repo split needs a UI decision.
- Deleting a repository row leaves its structural edges behind, since they name it through the polymorphic `targetId` that no cascade covers. Not reachable from the UI today.
- There is no way to delete a hub yet, so the `CHANNEL -> TRACK` cascade cannot orphan tracks in practice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.219.0-release-20260825.19](https://github.com/juspay/xyne-spaces/compare/v1.219.0-release-20260825.18...v1.219.0-release-20260825.19) (2026-08-30)
+
+
+### Bug Fixes
+
+* enable jemalloc heap profiling to trace native memory leak ([#1217](https://github.com/juspay/xyne-spaces/issues/1217)) ([9c0435e](https://github.com/juspay/xyne-spaces/commit/9c0435e6d8556adb3225ae0b125c678e9b0590fd))
+
 ## [1.219.0-release-20260825.18](https://github.com/juspay/xyne-spaces/compare/v1.219.0-release-20260825.17...v1.219.0-release-20260825.18) (2026-08-30)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.219.0-release-20260825.18](https://github.com/juspay/xyne-spaces/compare/v1.219.0-release-20260825.17...v1.219.0-release-20260825.18) (2026-08-30)
+
+
+### Features
+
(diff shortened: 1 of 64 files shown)


ameer030:
Superseded by #1236. This PR's head branch was rebased onto the release-20260825 tip, so it carried six unrelated release commits that feature/sdlc-deploy-25 does not have. #1236 is the same change cherry-picked onto a branch cut from feature/sdlc-deploy-25 — a single commit, no conflicts.

PR #1253, opened 2026-08-31
feat: XYNE-61192 add Spec Validation check on pull requests

Dnyaneshwari-04:
  - backend

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [x] Backend unit tests — `pnpm --filter xyne-spaces-backend run test`
- [ ] Claw tests
- [ ] End-to-end suite
- [x] Not covered by tests

No test file ships with this PR. The spec rules were developed against real
ticket descriptions and every review finding was reproduced by execution before
being fixed, but none of that is pinned by a committed test. Worth adding before
the check is made a required status check.

`tsc --noEmit` and the backend build pass; the backend suite is unchanged from
`main` (same three pre-existing failures).

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +19,113 @@ const PR_VALIDATION_CONFIG = {
...
+const formatMissingSections = (missing: string[]): string => {
+  const shown = missing.slice(0, 3).join(', ');
+  return missing.length > 3 ? `${shown} +${missing.length - 3} more` : shown;
+};
+
 interface ValidationResult {
   isValid: boolean;
   errorMessage?: string;
   ticketId?: string;
+  xyneId?: string;
```
devesh-juspay, line 64: ticket id and xyne id will be same?

Dnyaneshwari-04: They’re different. `ticketId` is the internal ID (e.g. `cmt1g1qoa0nqwa3vxas6vs9mm`), while `xyneId` is the user-facing ID (e.g. `XYNE-61192`).

apps/backend/src/config/env.ts
```diff
@@ -389,6 +389,11 @@ const envSchema = Joi.object({
   JENKINS_WEBHOOK_SECRET: Joi.string().allow('').default(''),
   GITHUB_TOKEN: Joi.string().allow('').default(''),
   GITHUB_API_URL: Joi.string().uri().default('https://api.github.com'),
+  // "Spec Validation" PR check. Off so deploying it changes nothing.
+  ENABLE_PR_SPEC_CHECK: Joi.boolean().default(false),
+  SPEC_REQUIRED_SECTIONS: Joi.string().default(
```
devesh-juspay: **[CONFIRMED] A blank `SPEC_REQUIRED_SECTIONS=` env value aborts backend boot — even with the feature disabled.**

`Joi.string()` without `.allow('')` rejects the empty string (`.default()` only fires for `undefined`), and a blank entry is this repo's own convention: `.env.example` has 166 blank `NAME=` lines, and neighbors like `GITHUB_TOKEN` all use `.allow('').default('')`. A deployment that copies that convention hits `Config validation error` at import time (env.ts:580) and the entire backend fails to start even though `ENABLE_PR_SPEC_CHECK=false`.

Fix: `.allow('')` with the default applied when empty.

apps/backend/src/config/env.ts
```diff
@@ -389,6 +389,11 @@ const envSchema = Joi.object({
...
+    'Problem statement,Solutioning,Test cases,Implementation details'
```
devesh-juspay: **[CONFIRMED] The default required-sections list contradicts the `/spec` writer: 'Implementation details' is OPTIONAL there.**

`apps/xyne-claw/spec-skills/spec-task/SKILL.md` lists REQUIRED = 'Problem statement, Solutioning, Test cases' and OPTIONAL = 'Implementation details, Out of scope' (`task-commands.ts` `SPEC_QUESTION_OUTLINE` repeats this). A ticket whose spec `/spec` itself considers complete (omitting the optional section) gets 'Spec Validation: failure — spec missing: Implementation details' on every PR once the check is enabled — and any future rename in the spec skill silently breaks validation.

Fix: drop 'Implementation details' from the default and share one constant between the spec skill and this validator.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -192,15 +355,123 @@ export class PullRequestValidationService {
...
+   * CAC first, so the check can be enabled per workspace or repo and the section
+   * names changed without a deploy. Any CAC problem falls back to env.
+   */
+  private async resolveSpecCheckConfig(
+    target: BuildStatusTarget,
+    workspaceId: string
+  ): Promise<{ enabled: boolean; sections: string[] }> {
+    const fallback = {
+      enabled: config.enablePrSpecCheck,
+      sections: config.specRequiredSections,
```
devesh-juspay: **[CONFIRMED] An empty parsed section list makes the check trivially green — the `sections.length` guard only covers the Superposition value, not the env fallback.**

`SPEC_REQUIRED_SECTIONS=','` (or `', '`) passes Joi as non-empty but split/trim/filter yields `[]`. `resolveSpecCheckConfig`'s `sections.length ? sections : fallback.sections` guard wraps only the Superposition string; the fallback paths return `config.specRequiredSections = []` unguarded, and `validateSpecSections` with `required=[]` returns `missing=[]` → `isValid=true` — every PR gets 'Spec Validation: success — Specification complete' regardless of the ticket having no spec at all.

Fix: apply the same non-empty guard to the env-derived list (or fail closed on an empty required list).

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +19,113 @@ const PR_VALIDATION_CONFIG = {
...
+    .replace(/[:\-–—\s]+$/, '')
+    .trim()
+    .toLowerCase();
+
+const parseMarkers = (lines: string[], sectionNames: Set<string>): SpecMarker[] => {
+  const markers: SpecMarker[] = [];
+  let inFence = false;
+  lines.forEach((raw, index) => {
+    const line = raw.trim();
+    if (/^(```|~~~)/.test(line)) {
```
devesh-juspay: **[CONFIRMED, reproduced by execution] The fence toggle ignores fence character, length, and the no-info-string closer rule, so `inFence` desyncs.**

Two reproduced failure modes:
- A `~~~` line inside a ``` ``` ``` block flips `inFence` off, so commented `# Test cases` lines inside the code block satisfy sections (**false pass**).
- A ````` ```` `````-fence quoting an odd number of ``` ``` ``` lines — or any unclosed fence — leaves `inFence` stuck `true` and swallows all later real `### ...` headings, reporting a complete spec as 'No specification' (**false failure blocking the PR**).

Per CommonMark, a closing fence must use the same character as the opener and be at least as long; a ``` ``` ``` inside a `~~~` block (and vice versa) is content, not a closer.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +19,113 @@ const PR_VALIDATION_CONFIG = {
...
+interface SpecMarker {
+  name: string;
+  line: number;
+}
+
+const normalizeHeading = (text: string): string =>
+  text
+    .replace(/^#{1,6}\s*/, '')
+    .replace(/[*_`]/g, '')
+    .replace(/[:\-–—\s]+$/, '')
```
devesh-juspay: **[CONFIRMED, reproduced by execution] Inline `Section: body` style never matches, so terse specs fail every section.**

'Problem statement: users cannot log in' normalizes to 'problem statement: users cannot log in' — the `/[:\-–—\s]+$/` regex only strips *trailing* punctuation, so the line is not in `sectionNames`. '**Problem statement:** Broken.' fails too. A ticket written in the common name-colon-body style is reported as missing every section and the PR check fails despite the spec plainly being present.

Consider matching a known section name followed by a separator and inline content.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +19,113 @@ const PR_VALIDATION_CONFIG = {
...
+
+interface SpecMarker {
+  name: string;
+  line: number;
+}
+
+const normalizeHeading = (text: string): string =>
+  text
+    .replace(/^#{1,6}\s*/, '')
+    .replace(/[*_`]/g, '')
```
devesh-juspay: **[CONFIRMED, reproduced by execution] `normalizeHeading` strips `*` and `_` everywhere, promoting bullets and italics to section markers.**

In `### Test cases\nSteps:\n* Implementation details\n### Implementation details\nDone.` the bullet `* Implementation details` becomes the first marker; its body runs to the real heading (empty), so 'Implementation details' is reported missing on a complete spec (**false failure**). Conversely, a bulleted TOC line `* Test cases` followed by prose falsely satisfies a section that has no content (**false pass**).

Strip emphasis markers only when they wrap the whole line (e.g. `**Name**`), not list bullets at line start.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +19,113 @@ const PR_VALIDATION_CONFIG = {
...
+    return { isValid: false, missing: required, hasSpecHeading: false };
+  }
+
+  const sectionNames = new Set<string>([
+    SPEC_SECTION,
+    ...required.map(section => normalizeHeading(section)),
+  ]);
+  const lines = description.split(/\r?\n/);
+  const markers = parseMarkers(lines, sectionNames);
+  const specIndex = markers.findIndex(marker => marker.name === SPEC_SECTION);
```
devesh-juspay: **[CONFIRMED, reproduced by execution] Scope is cut at the *first* marker named 'specification', so a stray late 'Specification' line discards all sections above it.**

Sections written without a wrapper heading (an explicitly supported layout) followed by a trailing bare 'Specification' line (footer/template leftover) → `findIndex` selects that late marker, `scope = markers.slice(specIndex + 1) = []`, every section is reported missing, and the PR fails with 'No specification on XYNE-…' despite a complete spec above.

Only treat 'Specification' as a wrapper when the required sections actually appear after it (or fall back to whole-document scope when the post-wrapper scope is empty).

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +19,113 @@ const PR_VALIDATION_CONFIG = {
...
+    .replace(/[*_`]/g, '')
+    .replace(/[:\-–—\s]+$/, '')
+    .trim()
+    .toLowerCase();
+
+const parseMarkers = (lines: string[], sectionNames: Set<string>): SpecMarker[] => {
+  const markers: SpecMarker[] = [];
+  let inFence = false;
+  lines.forEach((raw, index) => {
+    const line = raw.trim();
```
devesh-juspay: **[CONFIRMED, reproduced by execution] `raw.trim()` before matching means indented (non-fenced) code blocks are scanned for section markers.**

A 4-space-indented code sample containing `# Implementation details` satisfies the section with only code as its body (**false pass**), and an indented bare `Implementation details` line becomes the first marker with an empty body, reporting the section missing even though a real populated heading follows (**false failure**).

Skip lines indented 4+ spaces (outside fences) before marker matching.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -64,6 +169,36 @@ export class PullRequestValidationService {
...
+      sourceBranch,
+      destinationBranch,
+      workspaceId,
+      repoName,
+      repoUrl,
+      prUrl,
+      numberOfComments,
+      target,
+    );
+    await this.postSpecBuildStatus(target, commitHash, workspaceId, result);
```
devesh-juspay: **[CONFIRMED] The webhook HTTP response path now serially awaits Superposition lookups plus a second status POST, bounded only by a 30s timeout — 3× GitHub's 10s webhook delivery limit.**

`webhooks.ts` awaits `handleWebhookEvent` → both webhook services await `validatePullRequest` → it awaits `postSpecBuildStatus` (`getBooleanValue` + `getStringValue` + another status POST) before returning. `superpositionClient` catches rejections but adds no time bound of its own; the provider-level `SUPERPOSITION_TIMEOUT` defaults to 30000ms, so a stalled Superposition marks every PR webhook delivery failed/redelivered.

The spec result is discarded, so it need not be awaited — fire-and-forget (or `Promise.all` with the required work) fixes this.

devesh-juspay (commented):
Automated review (Claude Code) — 10 confirmed findings, attached as inline comments. Five of the parser findings were reproduced by executing the parser, not just reading it.

Overall picture: the spec-section markdown parser is fragile in both directions — it can fail complete specs (blocking PRs) and pass empty ones — and the configuration layer has a boot-crash hazard, a default that contradicts the `/spec` writer, and an empty-list bypass. Two integration issues round it out: webhook-path latency against GitHub's 10s delivery timeout, and a Bitbucket `INPROGRESS` build status that nothing ever finalizes.

Two other candidates were investigated and dropped as intended behavior (the sequential Superposition lookup is a deliberate short-circuit; the disabled-flag behavior reads as kill-switch semantics). Four cleanup-level findings (duplicated 11-parameter signature, `ValidationResult` copy-paste across 5 return sites, a wrong-default `buildStatus` param hazard, a redundant fallback layer) were cut by the findings cap.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -41,6 +138,14 @@ export type BuildStatusTarget =
 
 const DEFAULT_BUILD_STATUS_TARGET: BuildStatusTarget = { provider: VCSProviderType.BITBUCKET_SERVER };
 
+type BuildStatusState = 'success' | 'failure' | 'pending';
+
+const BITBUCKET_BUILD_STATE: Record<BuildStatusState, 'SUCCESSFUL' | 'FAILED' | 'INPROGRESS'> = {
+  success: 'SUCCESSFUL',
+  failure: 'FAILED',
+  pending: 'INPROGRESS',
```
devesh-juspay: **[CONFIRMED] On Bitbucket, the 'pending' spec state posts `INPROGRESS` and nothing ever finalizes it.**

With the spec check enabled on a Bitbucket workspace, a PR opened with no XYNE id in the title (or a nonexistent ticket) gets `xyne-spec-check = INPROGRESS`; if the PR is then declined/abandoned, no other code path writes `xyne-spec-check` and `pr:declined` handling only re-runs the same validation — the commit's build rollup shows an in-progress build forever, and any 'no in-progress builds' merge gate on *other* PRs containing that commit misreports.

Post a terminal state (e.g. STOPPED/SUCCESSFUL-with-explanation) for the pending case, or skip posting entirely until a ticket is resolvable.

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Fixed, but not by changing this line — which is why the thread hasn't gone outdated.

Guarding list items *before* normalization runs: `LIST_ITEM = /^([-*+]|\d+[.)])\s+/` returns early in `matchMarker`, so `* Implementation details` never reaches `normalizeHeading` and never becomes a marker. Keeping the global strip is deliberate — it's what still lets `**Problem statement:**` match.

Test: "does not treat a bulleted section name as a heading".

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Fixed, but not on this line, so the thread stayed open.

Indentation is now tested against the raw line before matching — `if (fence || INDENTED_CODE.test(raw)) return;` with `INDENTED_CODE = /^(?: {4,}|\t)/`. The trimmed line is still used for matching, the untrimmed one for the indent check.

Test: "does not scan indented code blocks for sections".

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Took the second option — skip posting entirely.

Bitbucket now returns before the pending branch posts anything, so no unfinalizable `INPROGRESS` build is created. The `pending -> INPROGRESS` entry stays in the map (the `Record` type needs the key) but is unreachable for this check.

Happy to switch to a terminal state instead if you'd prefer that over skipping.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -64,6 +209,38 @@ export class PullRequestValidationService {
...
+      sourceBranch,
+      destinationBranch,
+      workspaceId,
+      repoName,
+      repoUrl,
+      prUrl,
+      numberOfComments,
+      target,
+    );
+    void this.postSpecBuildStatus(target, commitHash, workspaceId, result).catch(error =>
```
devesh-juspay: **[CONFIRMED] Fire-and-forget status post is dropped on deploy-time SIGTERM — a required 'Spec Validation' check then waits forever.**

The webhook handler awaits `validatePullRequest` and responds 200; if the pod gets SIGTERM, `app.ts` `shutdown()` closes `superpositionClient` (app.ts:1068) *before* `httpServer.close()` (:1135) and drains no background work, so the detached `postSpecBuildStatus` sees `isReady() === false` / default-false and posts nothing (or the process exits before the POST). GitHub logged a successful delivery, so nothing retries — the PR is blocked until a new push.

Fix: track the detached task (e.g. a small in-flight set awaited during shutdown), or post the status before responding while bounding the Superposition lookups with a short timeout.

devesh-juspay: **[CONFIRMED] Detached status tasks from successive webhook events can complete out of order — a stale 'pending' can overwrite a correct 'success'.**

PR opened with a bad title (spec status → 'pending'), title edited: the second event's task posts 'success' first because the first event's task is still in its Superposition lookups; the delayed first task then posts 'pending' last. Same context/SHA means last-write-wins, and the required check stays pending until a new push. Neither webhook service has a queue, lock, or version guard for these tasks.

Fix: serialize per commit/context (keyed in-flight map) or drop a task when a newer event for the same PR/commit has superseded it.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +18,154 @@ const PR_VALIDATION_CONFIG = {
...
+  return null;
+};
+
+const parseMarkers = (lines: string[], sectionNames: Set<string>): SpecMarker[] => {
+  const markers: SpecMarker[] = [];
+  let fence: { char: string; length: number } | null = null;
+
+  lines.forEach((raw, index) => {
+    const line = raw.trim();
+    const fenceMatch = line.match(FENCE);
```
devesh-juspay: **[CONFIRMED, reproduced by execution] FENCE is tested on the trimmed line *before* the INDENTED_CODE guard — a backtick-fence line inside a 4-space-indented code block opens a phantom fence.**

A complete spec whose Problem-statement body shows example output as indented code containing an unmatched '    ```' line: the trim makes it match FENCE (the `if (fence || INDENTED_CODE.test(raw)) return;` on line 104 is never reached), the fence never closes, and all later section headings are swallowed — `validateSpecSections` returns missing: ['Solutioning', 'Test cases'] on a complete spec.

Fix: check `INDENTED_CODE.test(raw)` before the fence toggle (an indented line can never open/close a fence).

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +18,154 @@ const PR_VALIDATION_CONFIG = {
...
+  ]);
+  const lines = description.split(/\r?\n/);
+  const markers = parseMarkers(lines, sectionNames);
+  const hasSpecHeading = markers.some(marker => marker.name === SPEC_SECTION);
+
+  let scope = markers;
+  for (const [index, marker] of markers.entries()) {
+    if (marker.name !== SPEC_SECTION) continue;
+    const after = markers.slice(index + 1);
+    if (after.some(other => other.name !== SPEC_SECTION)) {
```
devesh-juspay: **[CONFIRMED, reproduced by execution] A mid-description 'Specification' line still discards all sections written before it whenever at least one required-section marker follows it.**

'Problem statement\nA.\n\nSolutioning\nB.\n\nSpecification\n\nTest cases\nC.' → scope becomes only [Test cases]; Problem statement and Solutioning (present, with content) are reported missing. The trailing-'Specification' test added in this head only covers the case where *nothing* follows the marker — one section after it flips pass to fail.

Fix: if sections are missing from the post-'Specification' scope but present in the whole document, fall back to whole-document scope (or require the wrapper to precede all section markers before narrowing).

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -192,15 +397,138 @@ export class PullRequestValidationService {
...
+        context,
+      );
+
+      const sections = rawSections
+        .split(',')
+        .map(section => section.trim())
+        .filter(Boolean);
+
+      return { enabled, sections: sections.length ? sections : fallback.sections };
+    } catch (error) {
```
devesh-juspay, line 406: **[PLAUSIBLE] A sections-lookup failure after `pr_spec_check_enabled` resolved true silently disables the check — and the doc comment's promised env fallback does not exist.**

`getStringValue`'s `ensureInitialized()` sits outside its internal try/catch, so a rejection between the two lookups (e.g. concurrent `superpositionClient.close()` during shutdown, or a future client change) lands in the outer catch, which returns `{enabled: false}` instead of `{enabled: true, sections: defaults}`. The commit gets no spec status at all — a silent skip on a required check — recorded only by a `logger.warn`. The comment 'Any CAC problem falls back to env' is wrong: the fallback is hardcoded and no env var is read.

Fix: scope the catch so a sections-lookup failure keeps `enabled: true` with default sections, and correct the comment.

devesh-juspay (commented):
Automated re-review (Claude Code) at head `9d48873c8`. First, credit where due: **8 of the 10 findings from the previous review are fixed and regression-tested** (the new test file covers bulleted section names, inline `Section: body`, bold labels, trailing 'Specification', mismatched fences, indented code, the empty-required-list bypass, and out-of-scope same-named headings), and the webhook-blocking and Bitbucket-INPROGRESS issues were both acted on.

10 findings remain at this head (9 confirmed, 1 plausible), attached inline. The two most important are **regressions introduced by the fixes themselves**: the fire-and-forget status post can be dropped on SIGTERM (leaving a required check 'Expected' forever) and can race between webhook events (stale 'pending' overwriting 'success'); and the Bitbucket skip-pending rule now leaves a stale status when a commit is re-validated against an unresolvable ticket. The parser retains two reproduced-by-execution bugs (phantom fence from a backtick line inside indented code; mid-description 'Specification' still discards earlier sections).

Verification also cleared several suspects as non-issues: all five internal postBuildStatus call sites pass the new 5th argument; both external validatePullRequest callers match the signature; 'pending'/INPROGRESS are valid for their providers; and Prisma's non-nullable Ticket.description makes the undefined sentinel sound.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -192,15 +397,138 @@ export class PullRequestValidationService {
...
+    if (!enabled) return;
+
+    const specStatus = PR_VALIDATION_CONFIG.SPEC_BUILD_STATUS;
+
+    // Pending, not failed: the spec has not been found wanting, it has not been
+    // looked at. Still blocks a required check, without claiming a spec is missing.
+    if (result.ticketDescription === undefined || !result.xyneId) {
+      // Bitbucket has no terminal state for "not applicable": an INPROGRESS
+      // build would never be finalized, so skip rather than leave one hanging.
+      if (target.provider === VCSProviderType.BITBUCKET_SERVER) return;
```
devesh-juspay: **[CONFIRMED] On Bitbucket, the unresolved-ticket early return leaves a *stale* 'xyne-spec-check' verdict from a prior validation of the same commit.**

A `pr:modified` event re-validates the same `latestCommit` after a title edit to an unparseable/nonexistent ticket; `ticketDescription` is undefined so `postSpecBuildStatus` returns without posting. Bitbucket dedupes per commit+key, so the old SUCCESSFUL/FAILED verdict — computed for a *different* ticket — stays attached to the commit and misreports the spec state. (GitHub is fine here: it overwrites with 'pending'.)

Fix: on Bitbucket, post a terminal status reflecting the unresolved-ticket state instead of skipping when a status may already exist for the commit.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -192,15 +397,138 @@ export class PullRequestValidationService {
...
+        commitHash,
+        'success',
+        PR_VALIDATION_CONFIG.SPEC_MESSAGES.VALIDATION_PASSED,
+        specStatus,
+      );
+      return;
+    }
+
+    const allMissing = spec.missing.length === sections.length;
+    const errorMessage = allMissing
```
devesh-juspay, line 490: **[CONFIRMED] A full spec skeleton with empty sections is told 'No specification … run /spec' — `hasSpecHeading` was built to distinguish exactly this but is never read.**

'## Specification\n### Problem statement\n### Solutioning\n### Test cases\n' → all 3 sections missing → `allMissing` true → the MISSING message directs the author to regenerate a spec that already exists, instead of naming the empty sections. `hasSpecHeading` is returned but never consumed in production code.

Fix: branch the message on `hasSpecHeading` — 'spec sections are empty: …' vs 'no specification found'.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +18,154 @@ const PR_VALIDATION_CONFIG = {
...
     // Bitbucket build-status key (dedupes statuses per commit).
     KEY: 'xyne-ticket-check',
     // Display name: Bitbucket build-status `name`, GitHub commit-status `context`.
     NAME: 'Ticket Validation',
   },
+  SPEC_BUILD_STATUS: {
+    KEY: 'xyne-spec-check',
+    NAME: 'Spec Validation',
+  },
+  REQUIRED_SPEC_SECTIONS: ['Problem statement', 'Solutioning', 'Test cases'],
```
devesh-juspay: **[CONFIRMED] `REQUIRED_SPEC_SECTIONS` is now a fourth unlinked copy of the /spec producer's section list.**

The same list lives in `apps/xyne-claw/src/task-commands.ts:87` (`SPEC_QUESTION_OUTLINE`), `apps/xyne-claw/spec-skills/spec-task/SKILL.md` (lines 96–98), here, and in the `pr_spec_required_sections` CAC key. Rename 'Solutioning' to 'Approach' in the /spec skill and every newly written spec fails the PR check until someone finds the CAC key. No shared constant or cross-referencing test links the copies.

Fix: export one canonical list (e.g. from `@xyne/shared`) and reference it from both producer and validator.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +18,154 @@ const PR_VALIDATION_CONFIG = {
...
+  hasSpecHeading: boolean;
+}
+
+interface SpecMarker {
+  name: string;
+  line: number;
+  hasInlineBody: boolean;
+}
+
+const normalizeHeading = (text: string): string =>
```
devesh-juspay: **[CONFIRMED] The hand-rolled section parser duplicates an existing scanner and the `marked`/`remark` dependencies already shipped — and it still misses blockquoted headings.**

`sections()`/`normalizedHeading()` in `apps/backend/src/sdlc/wiki/wikiSectionMutation.ts` already scan markdown sections, and `marked` ^17 plus `remark-parse`/`unified` are in `apps/backend/package.json`. Meanwhile `normalizeHeading` never strips '>', so a spec pasted as a quote ('> ### Test cases') is invisible and a complete spec fails. The backend now carries two private section parsers with different fence semantics, and each new markdown edge case becomes another regex patch here.

Fix: tokenize with `marked.lexer()` (or extract one shared parser) and keep only the lenient name-matching custom.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -41,6 +178,14 @@ export type BuildStatusTarget =
 
 const DEFAULT_BUILD_STATUS_TARGET: BuildStatusTarget = { provider: VCSProviderType.BITBUCKET_SERVER };
 
+type BuildStatusState = 'success' | 'failure' | 'pending';
+
+const BITBUCKET_BUILD_STATE: Record<BuildStatusState, 'SUCCESSFUL' | 'FAILED' | 'INPROGRESS'> = {
```
devesh-juspay: **[CONFIRMED, structural] Provider-specific status knowledge is scattered across this generic service while `IGitProvider` has no status-posting method.**

`BITBUCKET_BUILD_STATE` (here), the provider branch in `postBuildStatus` (~547), and the Bitbucket skip-pending rule (~495) are all private to this class. The next feature that posts a commit status must re-derive all three and will most likely forget the skip-pending rule — which exists only as a comment — leaving Bitbucket PRs blocked on a permanently INPROGRESS build.

Fix: add a status-posting method (or a `supportsPendingState` capability) to `IGitProvider` and let the Bitbucket implementation own state mapping and pending semantics.

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Agreed, and it's a regression from the fire-and-forget fix in the previous round. Not fixing it in this PR though.

It only bites once `Spec Validation` is a required check — while it's advisory a dropped status is cosmetic, and the plan is to run it advisory for a week before it goes anywhere near the ruleset. Tracking it as a prerequisite for that step rather than for merge.

When we do fix it, I'd lean toward your second option: go back to awaiting the post but bound the Superposition lookups with a short timeout, which removes the detachment rather than trying to drain it.

Dnyaneshwari-04: Same root cause as the SIGTERM one and same call: real, but only matters once the check is required. A stale `pending` on an advisory check is noise; on a required one it blocks.

Bundling both into the "before it becomes a required check" work, where the awaited-with-timeout approach removes the race by construction.

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Fixed in e55b70b5e. Indentation is now checked before the fence toggle — an indented line can neither open nor close a fence, so `INDENTED_CODE.test(raw)` returns first.

Reproduced your case before the change (missing: [Solutioning, Test cases]) and it now passes. Test: "does not open a fence from a backtick line inside indented code".

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Fixed in e55b70b5e. The wrapper now narrows scope only when it precedes every section marker; otherwise the whole document is used.

Your case reproduced (missing: [Problem statement, Solutioning]) and now passes. Test: "ignores a Specification line that comes after some sections".

Consequence worth flagging: sections written *above* a wrapper now count, so the previous "same-named heading outside the spec block" test was inverted and re-documented. That trades a false pass for a false failure, which is the right way round here — a false failure blocks a PR, a false pass only misses one.

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: The comment was wrong and is fixed in e55b70b5e — the env layer was removed earlier in this PR, so it now reads that CAC is the only source and any CAC problem leaves the check off.

Leaving the behaviour as-is deliberately: with no env layer there is no "default sections" state that is meaningfully safer than off, and failing to off is the conservative direction for a check that may later be required. Open to scoping the catch if you disagree.

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Not fixing in this PR: the check will not be enabled on Bitbucket. The plan is GitHub-only via the `provider` field already in the CAC context, precisely because a Bitbucket build status can feed merge gates while a GitHub status is inert until it is added to a ruleset.

If we ever do enable it there, this and the INPROGRESS finding both point the same way — post a terminal state rather than skipping — and I'd take that then.

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Fixed in e55b70b5e. `hasSpecHeading` is now read: a description that has a Specification wrapper but empty sections gets "spec missing: Problem statement, Solutioning, Test cases" instead of "No specification … run /spec".

Test: "names the empty sections rather than claiming there is no spec".

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Agreed, and worth doing — but as a follow-up rather than here, since the fix spans `apps/xyne-claw` and the backend and this PR is already large.

Two of the four copies are prose inside a prompt (`SPEC_QUESTION_OUTLINE`) and a skill markdown file, so a shared constant means changing how the skill sources its section list, not just exporting an array. Happy to file it.

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: The blockquote half is fixed in e55b70b5e — `>` markers are stripped, and a spec pasted as a quote now validates. Test: "reads a spec pasted as a blockquote".

Pushing back on the reuse half. `marked`/`remark` only produce `heading` nodes for `#` lines, and the wiki scanner has the same assumption. XYNE-56410 — a real ticket — writes its sections as bare lines, and XYNE-56567 as `##` headings; an AST parser reports zero sections for the former, which is the exact false failure this parser exists to avoid. Matching on section names is deliberate, not an oversight.

The duplication point stands on its own though; if the wiki scanner grows the same tolerance we should collapse them.

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Fair, and it's the right long-term shape — provider capabilities belong on `IGitProvider`, not in a validation service.

Not doing it here though: moving status posting onto the interface touches both provider implementations and every existing caller of the ticket check, which is a bigger blast radius than the feature this PR adds. It also only pays off once a second feature posts a status, which does not exist yet.

Filing alongside the shared-section-list cleanup. If the pending-skip rule is the specific worry, it is currently reachable from one place and covered by the Bitbucket branch in `postSpecBuildStatus`, not just a comment.

devesh-juspay (commented):
Automated re-review round 3 (Claude Code) at head `8f8744157`. **4 findings from the previous round are fixed with regression tests** (phantom fence from indented code, mid-description 'Specification' scope, blockquoted headings, empty-skeleton message).

10 confirmed findings remain, attached inline. The dominant theme has shifted from the parser to the **status-posting lifecycle**: the fire-and-forget design still has no answer for process shutdown or event ordering (both carried-over findings at line 229 are unchanged), and there are now three stale-status variants (Bitbucket unresolved-ticket, and a new disabled/CAC-down re-run case). Each parser fix also spawned a subtler sibling, all reproduced by execution: prose `Name: text` lines now become spurious markers, the fence guard covers indented lines opening a fence but not closing one, and duplicated headings resolve to the first (empty) occurrence only. Two new non-parser items: two full-workspace-config HTTP fetches per webhook event, and an unsatisfiable required-sections config when 'Specification' itself is listed.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -64,6 +212,38 @@ export class PullRequestValidationService {
...
+      sourceBranch,
+      destinationBranch,
+      workspaceId,
+      repoName,
+      repoUrl,
+      prUrl,
+      numberOfComments,
+      target,
+    );
+    void this.postSpecBuildStatus(target, commitHash, workspaceId, result).catch(error =>
```
devesh-juspay: **[CONFIRMED, carried over] Fire-and-forget status post is silently dropped on deploy-time shutdown — a required 'Spec Validation' check then waits forever.**

Webhook handler awaits `runTicketValidation`, responds 200, then the pod gets SIGTERM: `app.ts` `shutdown()` awaits `superpositionClient.close()` (app.ts:1070) long before `httpServer.close()` (:1137) and drains no detached promises, so the void'd `postSpecBuildStatus` sees `isReady() === false` (or the process exits mid-POST) and posts nothing. GitHub logged a successful delivery so nothing redelivers; with 'Spec Validation' a required branch-protection check, the PR shows 'Expected — waiting for status' until a new push.

Fix: track the detached task and drain it during shutdown, or await the post before responding.

devesh-juspay: **[CONFIRMED, carried over] Detached status tasks from successive webhook events for the same commit complete out of order — a stale 'pending'/'failure' can overwrite the correct final status.**

PR opened with a bad title: event A's detached task (slowed by two Superposition round-trips) is still in flight when the author edits the title; event B posts 'Spec Validation: success', then A's stale 'pending — Ticket not resolved' lands last and overwrites it. GitHub statuses dedupe by context and Bitbucket by KEY — last write wins — so the required check sticks at the stale state until a new push.

Fix: serialize per commit/context or drop tasks superseded by a newer event.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +18,157 @@ const PR_VALIDATION_CONFIG = {
...
+  line: string,
+  sectionNames: Set<string>
+): { name: string; hasInlineBody: boolean } | null => {
+  if (LIST_ITEM.test(line.replace(/^>+\s*/, ''))) return null;
+
+  const text = line.replace(/^>+\s*/, '').replace(/^#{1,6}\s*/, '');
+  const whole = normalizeHeading(text);
+  if (sectionNames.has(whole)) return { name: whole, hasInlineBody: false };
+
+  const separator = text.search(/[:–—]|\s-\s/);
```
devesh-juspay: **[CONFIRMED, reproduced by execution] The new inline-body branch fires on ordinary prose — any body line starting with a section name plus ':', en/em-dash, or ' - ' becomes a spurious section marker.**

'## Specification\n### Problem statement\nSolutioning: we will discuss later, but the real problem is X.\n### Solutioning\nB.\n### Test cases\nC.' returns `missing: ['Problem statement']` — the prose line is consumed as a Solutioning marker, truncating Problem statement's body — blocking a PR whose spec is complete. The dual false-pass exists too: 'Test cases: will be added by QA later.' satisfies 'Test cases'.

Fix: only take the inline-body path on lines with heading/bold markup, not bare prose.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -192,15 +400,136 @@ export class PullRequestValidationService {
...
+   * status so a spec failure is distinguishable from a ticket failure.
+   */
+  private async postSpecBuildStatus(
+    target: BuildStatusTarget,
+    commitHash: string,
+    workspaceId: string,
+    result: ValidationResult
+  ): Promise<void> {
+    const { enabled, sections } = await this.resolveSpecCheckConfig(target, workspaceId);
+    if (!enabled) return;
```
devesh-juspay: **[CONFIRMED] When the check is disabled or CAC is unavailable on a re-run, returning without posting leaves a previously posted 'Spec Validation' *failure* permanently stale.**

The spec lives in the ticket, so the standard fix path is: spec check posts 'failure' → author runs /spec on the ticket → re-triggers validation on the *same* head commit. If Superposition is transiently unready/throws during that re-run (or an admin turned `pr_spec_check_enabled` off mid-rollout), `resolveSpecCheckConfig` returns `enabled: false` and the old red status stays — a merge-blocking check that only a manual API call or a new commit clears, while Ticket Validation refreshes normally alongside it.

Fix: when a prior spec status may exist, post a success/'check disabled' status instead of returning.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +18,157 @@ const PR_VALIDATION_CONFIG = {
...
+  return null;
+};
+
+const parseMarkers = (lines: string[], sectionNames: Set<string>): SpecMarker[] => {
+  const markers: SpecMarker[] = [];
+  let fence: { char: string; length: number } | null = null;
+
+  lines.forEach((raw, index) => {
+    // An indented line can neither open nor close a fence.
+    if (!fence && INDENTED_CODE.test(raw)) return;
```
devesh-juspay: **[CONFIRMED, reproduced by execution] Inside an open fence the indented-code guard is skipped, so an indented ``` line — fence *content* per CommonMark — wrongly closes the fence.**

The guard reads `if (!fence && INDENTED_CODE...)`, so with a fence open, '    ```' closes it; the real closing ``` then re-opens one and swallows all later section headings. '## Specification\n### Problem statement\n```\ncode\n    ```\nmore code\n```\n### Solutioning\nB.\n### Test cases\nC.' returns `missing: ['Solutioning', 'Test cases']` — any spec whose code sample contains an indented triple-backtick (pasted nested markdown, shell output) fails the check. The test suite covers indented lines *opening* a fence but not *closing* one.

Fix: also skip INDENTED_CODE lines for fence-close matching.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -192,15 +400,136 @@ export class PullRequestValidationService {
...
+  ): Promise<void> {
+    const { enabled, sections } = await this.resolveSpecCheckConfig(target, workspaceId);
+    if (!enabled) return;
+
+    const specStatus = PR_VALIDATION_CONFIG.SPEC_BUILD_STATUS;
+
+    // Pending, not failed: the spec has not been looked at, not found wanting.
+    if (result.ticketDescription === undefined || !result.xyneId) {
+      // Bitbucket would leave an INPROGRESS build nothing ever finalizes.
+      if (target.provider === VCSProviderType.BITBUCKET_SERVER) return;
```
devesh-juspay: **[CONFIRMED, carried over] On Bitbucket the unresolved-ticket early return leaves a stale 'xyne-spec-check' verdict from a prior validation of the same commit — including a false green computed for a *different* ticket.**

`pr:opened` with a valid ticket posts spec SUCCESSFUL on `latestCommit`; the author edits the title to a malformed/nonexistent ticket (`pr:modified`, same commit): `ticketDescription` is undefined so the method returns without posting, and Bitbucket's per-commit+KEY dedupe keeps the old SUCCESSFUL/FAILED verdict attached — the merge gate shows a spec verdict for a ticket the PR no longer references. (GitHub gets a 'pending' overwrite; Bitbucket gets nothing.)

Fix: post a terminal status reflecting the unresolved state instead of skipping when a status may already exist.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +18,157 @@ const PR_VALIDATION_CONFIG = {
...
+  const firstSection = markers.findIndex(marker => marker.name !== SPEC_SECTION);
+  const wrapper = markers.findIndex(marker => marker.name === SPEC_SECTION);
+  const scope =
+    wrapper !== -1 && (firstSection === -1 || wrapper < firstSection)
+      ? markers.slice(wrapper + 1)
+      : markers;
+
+  const missing = required.filter(section => {
+    const target = normalizeHeading(section);
+    const index = scope.findIndex(marker => marker.name === target);
```
devesh-juspay: **[CONFIRMED, reproduced by execution] Sections resolve via `scope.findIndex` — first occurrence only — so a duplicated heading whose first occurrence is bodyless reports the section missing despite a later duplicate having content.**

'## Specification\n\n### Problem statement\n### Problem statement\n\nActually explained here.\n\n### Solutioning\nFix.\n\n### Test cases\nOne.' (an accidental doubled heading from ticket-UI editing) returns `missing: ['Problem statement']` and blocks the PR despite the content being present — contradicting the module's stated err-toward-passing design.

Fix: report a section missing only if *every* occurrence of it is empty.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -192,15 +400,136 @@ export class PullRequestValidationService {
...
+    try {
+      const context = {
+        workspaceId,
+        provider: target.provider,
+        ...(target.provider === VCSProviderType.GITHUB
+          ? { owner: target.owner, repo: target.repo }
+          : {}),
+      };
+
+      const enabled = await superpositionClient.getBooleanValue(
```
devesh-juspay: **[CONFIRMED] Every webhook event with the flag reachable pays two sequential full-workspace-config HTTP fetches.**

`superposition-provider` 0.97.4's `eval()` unconditionally issues a `GetConfigCommand` network round trip per `getBooleanValue`/`getStringValue` call — its polling cache is never consulted at eval time — and the sections lookup is fetched even on the pending/skip paths that never use it. A busy org's push/synchronize webhooks each trigger two full-config downloads from the CAC service, including for malformed-title PRs whose section list is immediately discarded.

Fix: memoize the resolved `{enabled, sections}` per context with a short TTL (matching the provider's polling interval), fetch both keys in one read, and resolve sections lazily only on the branch that validates.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +18,157 @@ const PR_VALIDATION_CONFIG = {
...
+  ]);
+  const lines = description.split(/\r?\n/);
+  const markers = parseMarkers(lines, sectionNames);
+  const hasSpecHeading = markers.some(marker => marker.name === SPEC_SECTION);
+
+  // A wrapper narrows scope only when every section follows it: one written
+  // mid-description would otherwise discard the sections above it.
+  const firstSection = markers.findIndex(marker => marker.name !== SPEC_SECTION);
+  const wrapper = markers.findIndex(marker => marker.name === SPEC_SECTION);
+  const scope =
```
devesh-juspay: **[CONFIRMED, reproduced by execution] Putting 'Specification' in `pr_spec_required_sections` makes the requirement unsatisfiable — every PR in that workspace fails.**

The name collides with `SPEC_SECTION`: the wrapper-narrowing slice removes the only 'specification' marker from scope, so 'Specification\nProblem statement\nA.' returns `missing: ['Specification']` even though the heading is present — every PR fails with the baffling 'spec missing: Specification', and `resolveSpecCheckConfig` neither rejects nor warns on the value. Plausible config too, since /spec itself writes a 'Specification' wrapper heading.

Fix: filter the wrapper name out of the required list (or warn) in `resolveSpecCheckConfig`.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +18,157 @@ const PR_VALIDATION_CONFIG = {
...
     // Bitbucket build-status key (dedupes statuses per commit).
     KEY: 'xyne-ticket-check',
     // Display name: Bitbucket build-status `name`, GitHub commit-status `context`.
     NAME: 'Ticket Validation',
   },
+  SPEC_BUILD_STATUS: {
+    KEY: 'xyne-spec-check',
+    NAME: 'Spec Validation',
+  },
+  REQUIRED_SPEC_SECTIONS: ['Problem statement', 'Solutioning', 'Test cases'],
```
devesh-juspay: **[CONFIRMED, carried over] `REQUIRED_SPEC_SECTIONS` is still a fourth unlinked copy of the /spec producer's section list.**

The same list lives in `apps/xyne-claw/src/task-commands.ts:87` (`SPEC_QUESTION_OUTLINE`), `apps/xyne-claw/spec-skills/spec-task/SKILL.md` (lines 96–98), this default, and the `pr_spec_required_sections` CAC key. Rename 'Solutioning' in the /spec skill and every newly written spec fails the PR gate until someone finds the CAC key — with no test linking the copies.

Fix: export one canonical list from `@xyne/shared` (already imported here) and derive both producer and validator from it.

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Carried over deliberately. It only bites once `Spec Validation` is a required check, and the plan is to run it advisory for a week first — a dropped status on an advisory check is cosmetic.

Tracking it as a prerequisite for adding the check to the ruleset, not for merge. When we do it I'd take your second option: await the post again but bound the Superposition lookups with a short timeout, which removes the detachment rather than trying to drain it.

Dnyaneshwari-04: Same call as the SIGTERM one: real, but only observable once the check is required. Bundling both into the pre-ruleset work, where awaiting-with-a-timeout removes the race by construction rather than needing a per-commit lock.

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Fixed in 81fe793ae. The inline branch now requires heading or bold markup — `if (!/^(#{1,6}\s|[*_])/.test(bare)) return null;` — so bare prose starting with a section name is ordinary text again.

Reproduced your case first (`missing: ['Problem statement']`) and it now passes. Tests: "does not consume prose that starts with a section name" and "accepts a section and its body on one line when the line has markup".

Worth flagging: this partly reverses the inline fix from the first round. `### Problem statement: body` and `**Problem statement:** body` still work, bare `Problem statement: body` no longer does. A false failure truncating a real section is worse than the false pass it was solving.

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Real, and the one finding here I don't want to patch without a decision.

Posting success or "check disabled" when Superposition was unreachable means claiming a verdict we could not compute — on a check that may be required, that is worse than a stale red. The stale-red window also only exists on a re-run of the *same* commit while CAC is transiently down.

Options as I see them: keep the last posted state (current behaviour), or post a distinct "could not evaluate" state. I'd lean toward the latter over a false green. What would you prefer?

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Carried over: the check will not be enabled on Bitbucket. It is GitHub-only via the `provider` field already in the CAC context, because a Bitbucket build status can feed merge gates while a GitHub status is inert until added to a ruleset.

If we ever enable it there, this and the INPROGRESS finding both point the same way — post a terminal state rather than skipping — and I'd take that then.

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Half fixed in 81fe793ae. `resolveSpecCheckConfig` now takes `needSections` and skips the second lookup on the paths that never validate — the unresolved-ticket and disabled paths no longer pay for a section list they discard. That removes one of the two full-config fetches on exactly the events you called out (malformed-title PRs).

Not memoizing yet. A per-context TTL cache is the right next step but it needs care around the polling interval so a CAC change is not held stale past it, and I would rather not add a cache the same day the flag ships. Happy to do it before the check goes on the ruleset, alongside the fire-and-forget work.

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Fixed in 81fe793ae. The wrapper name is filtered out of the required list — `[...sections].filter(section => normalizeHeading(section) !== SPEC_SECTION)` — so configuring 'Specification' is a no-op rather than an unsatisfiable requirement. If it is the only entry, the empty-list guard fails closed.

Reproduced first (`missing: ['Specification']`) and it now passes. Test: "ignores the wrapper name if it appears in the required list".

apps/backend/src/services/pullRequestValidationService.ts
Dnyaneshwari-04: Carried over, and agreed on the substance — a rename in the /spec skill silently breaking the gate is a real failure mode.

Still a follow-up rather than this PR: two of the four copies are prose inside a prompt string (`SPEC_QUESTION_OUTLINE`) and a skill markdown file, so a shared constant means changing how the skill sources its section list, not just exporting an array from `@xyne/shared`. That is a change to the /spec producer, which this PR does not otherwise touch.

Filing it with the `IGitProvider` status-capability cleanup.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -192,15 +413,144 @@
...
+    const allMissing = spec.missing.length === sections.length && !spec.hasSpecHeading;
+    const errorMessage = allMissing
+      ? PR_VALIDATION_CONFIG.SPEC_MESSAGES.MISSING(xyneId)
+      : PR_VALIDATION_CONFIG.SPEC_MESSAGES.INCOMPLETE(
+          xyneId,
+          formatMissingSections(spec.missing),
+        );
+    logger.info(
+      `[PR-Validation] Spec check failed for ${sanitizeForLog(xyneId)}: ` +
+        `missing ${sanitizeForLog(spec.missing.join(', '))}`
```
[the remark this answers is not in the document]
Dnyaneshwari-04: Both interpolations were already wrapped in `sanitizeForLog`, but CodeQL did not treat the control-character class as a barrier. Fixed in 1ee01985c by adding an explicit `[\r\n]` replace, which is the form the query models.

ameer030 (dismissed):

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +21,185 @@ const PR_VALIDATION_CONFIG = {
...
+    if (marker) markers.push({ ...marker, line: index });
+  });
+  return markers;
+};
+
+/**
+ * Each required section must be named, with text under it. `sections` is passed
+ * in so a /spec rename is a config change rather than a code change.
+ */
+export function validateSpecSections(
```
devesh-juspay: can we move all these functions to a new file, these functioins are not releated to pullRequestValidationService

devesh-juspay (commented):
One focused comment on the Superposition timeout situation (inline at `withTimeout`), based on inspecting `superposition-provider@0.117.2` — the exact version the lockfile pins.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -210,32 +506,43 @@ export class PullRequestValidationService {
...
         await githubManager.postCommitStatus(
           target.owner,
           target.repo,
           commitHash,
-          success ? 'success' : 'failure',
-          PR_VALIDATION_CONFIG.BUILD_STATUS.NAME,
+          state,
+          buildStatus.NAME,
           description,
+          targetUrl || undefined,
```
devesh-juspay: we should not use traget url for github, target url will have our internal frontend url, we cannot add this as github checks are visible to everyone

devesh-juspay (dismissed):

anuragdvd (dismissed):

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -17,18 +27,39 @@ const PR_VALIDATION_CONFIG = {
...
   BUILD_STATUS: {
     // Bitbucket build-status key (dedupes statuses per commit).
     KEY: 'xyne-ticket-check',
     // Display name: Bitbucket build-status `name`, GitHub commit-status `context`.
     NAME: 'Ticket Validation',
   },
+  // No Bitbucket key: this check is GitHub only, and GitHub identifies a status
+  // by its context alone.
+  SPEC_BUILD_STATUS: {
+    NAME: 'Spec Validation',
```
devesh-juspay: A fixed context has a clobbering problem when one head commit backs several open PRs (e.g. a main PR and a release-branch PR, or #1362/#1363 right now): commit statuses are stored per commit, so the last-validated PR overwrites the other's verdict under the same context. For Spec Validation this matters even more than for Ticket Validation, because the verdict is derived from the PR's ticket — two PRs on the same commit usually name different tickets and can legitimately disagree.

#1364 fixes this for Ticket Validation by making the context/key a function of the PR id (`Ticket Validation (#123)`, `xyne-ticket-check-pr-123`). It touches the same `postBuildStatus` code, so expect a conflict once it merges — when resolving, suggest folding the same per-PR shape into this config: `NAME: (prId: number) => `Spec Validation (#${prId})`` (and same for BUILD_STATUS). `prId` is already in `ValidatePullRequestParams`, so it just needs threading into `postSpecBuildStatus`/`postBuildStatus`.

apps/backend/src/services/pullRequestValidationService.ts
```diff
@@ -192,15 +312,191 @@ export class PullRequestValidationService {
...
+    const { enabled, configured, sections } = await this.resolveSpecCheckConfig(
+      target,
+      workspaceId,
+    );
+    const specStatus = PR_VALIDATION_CONFIG.SPEC_BUILD_STATUS;
+
+    // Clear a verdict left from when the check was on; stay silent on commits
+    // that never had one.
+    if (!enabled) {
+      if (configured && (await this.hasSpecStatus(target, commitHash))) {
```
devesh-juspay: Same shared-head caveat as the context comment above: this asks "does this *commit* have a Spec Validation status", but with several open PRs on one head commit the status found here may belong to a different PR. Once the context becomes per-PR (`Spec Validation (#${prId})`, see #1364), this check should look for the per-PR context so the disabled-clear only touches this PR's own verdict.

apps/backend/src/services/pullRequestValidationService.ts
devesh-juspay: Fairness note after discussion: in the common shared-head case — a main PR and a release-branch PR for the same change — both titles name the same ticket, so both validations reach the same verdict and the clobbering is invisible. Spec Validation is therefore less exposed than Ticket Validation (which failed on auto-generated release-PR titles). The divergent case is two PRs naming different tickets on one head commit (see #1362/#1363, live today): verdicts can then disagree, and the failure message embeds the other ticket's id. So: lower severity, same mechanism — and since the #1364 conflict resolution will introduce the per-PR pattern anyway, extending it to SPEC_BUILD_STATUS is a one-liner.

apps/backend/src/services/pullRequestValidationService.ts
devesh-juspay: Same qualification as on the context thread: when both PRs on a shared head name the same ticket (the common case), the status found here matches this PR's own verdict and the clear is harmless. It only misfires when the two PRs name different tickets (e.g. #1362/#1363). Lower severity, but the per-PR context makes it exact either way.

apps/backend/src/services/pullRequestValidationService.ts
devesh-juspay: Addressed in 2667092fb and carried through the merge with main — SPEC_BUILD_STATUS and BUILD_STATUS are both per-PR now. Resolving.

apps/backend/src/services/pullRequestValidationService.ts
devesh-juspay: Addressed — hasSpecStatus now reads the per-PR context. Resolving.

devesh-juspay (approved):
Per-PR status scoping for both Ticket Validation and Spec Validation is in place and survived the merge with main; changed files typecheck clean. Note: the check is off until pr_spec_check_enabled is set in Superposition, and it is non-blocking (not in the ruleset).

harshpreet931 (approved):

PR #1272, opened 2026-08-31
fix: XYNE-60272 route SDLC notifications and activities from entity ids

ameer030:
Stacked on #1236. Review only the last commit; the base branch carries the multi-repo work.

## Problem

An SDLC notification or activity did not land where it should. The backend built the route as a string, so the frontend could not change it, and every builder wrote a `/chat/` URL — a notification about an artifact or a ticket inside an SDLC hub navigated out of the hub entirely.

## Backend sends data, not routes

`createNotification` resolves whether a notification belongs to an SDLC hub and, when it does, drops `actionUrl` and adds `sdlcChannelId` to the metadata the builders already populate. One guard in the funnel every notification passes through, so mentions, replies, canvas shares and ticket notifications are covered without touching any of the builders.

`buildSdlcPath` in `packages/shared` turns those ids into a path, and is the only place the routes are named. Both the notification handler and the activity list call it.

The discussion panel now opens only when a `messageId` is present. All fourteen ticket builders put the ticket's own `conversationId` in metadata whether or not the notification is about a message, so without that condition every SDLC ticket notification force-opened a discussion.

## The destination resolves the section

A deep link names an id, not a place. `SdlcScreen` decides where that id lives on arrival, where the hub's canvases and entity links are already loaded:

- an artifact deep link picks `baseline`, `wiki` or `artifacts` from its artifact type
- a discussion deep link, which names only a conversation, falls back to that conversation's owning canvas
- a track discussion resolves through the `TRACK → CONVERSATION` link
- a ticket resolves through its conversation: converting a discussion to a ticket leaves the `CANVAS → CONVERSATION [DISCUSSION]` link on the conversation and creates no link on the ticket, so the ticket alone resolves to nothing. It now lands on the owning artifact with that thread open.

A ticket with no artifact behind it stays on the hub's board and opens in a panel there, rather than navigating out of SDLC as it did before.

## Two fixes the deep links depend on

- The iframe path sync dropped `location.hash`, so the `#origin`/`#messageId` scroll target died on every navigation after the first.
- The invalid-deep-link guard ran as soon as the channel query resolved, but the discussion context needs the entity links from a separate query. Conversation-only deep links are now the standard shape, so the guard was stripping the discussion params before they could be matched.

## Testing

Typecheck, lint and prettier clean on both apps. Verified end to end in a local hub: artifact activity opens the artifact, track discussion opens the track with its thread, a ticket converted from a PRD discussion opens that PRD, and a board-created ticket opens in the board panel.

---

## Update — two more commits

**Client decides SDLC routing.** `createNotification` no longer resolves whether a notification belongs to an SDLC hub. That cost a channel lookup per recipient on the path every notification passes through, and blanking `actionUrl` alongside it took the deeplink away from mobile, which has no `buildSdlcPath`. The client already holds every channel it can see, so `NotificationHandler` keeps a set of SDLC channel ids from `useAllChannels` and reads it through a ref, since the socket callback registers once. Builders are unchanged — the one gap was canvas mentions, which took `channelId` as a parameter and never forwarded it to metadata while every other builder did.

**A hub is now required on every SDLC entity link.** `channelId` is the only scope an `sdlc_entity_links` row carries; every read filters on it and the Zero query ACL resolves visibility through `channel -> participants`, so a row without one is written and then invisible to everyone. Prisma models it as required; Postgres keeps it nullable because the pre-multirepo rows are still waiting on the backfill, and the migration that added the column now says so. Every current write already satisfied this, so three `repo.channelId!` assertions are gone. The one path that could still write an unscoped row was the Zero mutation ACL, which checked `channelId` only for track membership and now rejects any insert without one.

### DIFF
--- a/apps/backend/src/services/notificationService.ts
+++ b/apps/backend/src/services/notificationService.ts
@@ -22,7 +22,8 @@ import { serializeInitialMessageMd,
   type InitialMessageSummary,
   ChannelScopeType,
   NotificationDeliveryMethod,
-  NotificationType, MessageType, NotificationStatus, UserStatus } from '@xyne/shared';
+  NotificationType, MessageType, NotificationStatus, UserStatus,
+  ChannelType } from '@xyne/shared';
 
 const prisma = DatabaseClient.getInstance();
 
@@ -848,6 +849,17 @@ class NotificationService {
   ): Promise<{ deliveredViaApp: boolean }> {
     let deliveredViaApp = false;
 
+    // No actionUrl builder knows the SDLC routes. Builders already put the entity
+    // ids in metadata, so drop the URL and add the one key they lack — which hub.
+    const sdlcChannelId = await this.sdlcChannelId(data);
+    if (sdlcChannelId) {
+      data = {
+        ...data,
+        actionUrl: undefined,
+        metadata: { ...(data.metadata ?? {}), sdlcChannelId },
+      };
+    }
+
     try {
       logger.info(`[NOTIFICATION-SERVICE] createNotification called`, {
         userId,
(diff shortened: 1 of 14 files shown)


apps/backend/src/services/notificationService.ts
```diff
@@ -848,6 +849,17 @@ class NotificationService {
...
     let deliveredViaApp = false;
 
+    // No actionUrl builder knows the SDLC routes. Builders already put the entity
+    // ids in metadata, so drop the URL and add the one key they lack — which hub.
+    const sdlcChannelId = await this.sdlcChannelId(data);
+    if (sdlcChannelId) {
+      data = {
+        ...data,
+        actionUrl: undefined,
+        metadata: { ...(data.metadata ?? {}), sdlcChannelId },
```
utkarsh-611: channel ID -> frontend check

apps/dashboard/src/routes/SdlcScreen/SdlcScreen.tsx
```diff
@@ -605,6 +606,18 @@ export default function SdlcScreen(): ReactElement {
     [selectedCanvasRelatedConversations],
   );
   const selectedTicketId = routeSearchParams.get('ticket');
+  // ThreadMessages keys off the conversation, so a `?ticket=` deep link has to
+  // read the row to find one.
+  const [selectedTicketRow] = useCachedQuery(
```
utkarsh-611, line 611: probably this is not requiredd

apps/dashboard/src/routes/SdlcScreen/SdlcScreen.tsx
```diff
@@ -640,14 +653,101 @@ export default function SdlcScreen(): ReactElement {
...
+  }, [
+    activeTypeFolderId,
+    canvases,
+    discussionContext,
+    section,
+    selectedCanvasId,
+    ticketDiscussion,
+  ]);
+
+  useEffect(() => {
```
utkarsh-611, line 699: remove if not required

utkarsh-611 (approved):

PR #1295, opened 2026-09-01
fix: XYNE-61403 shrink empty kanban columns, expand non-empty terminal ones

harshpreet931:
POT - 

https://github.com/user-attachments/assets/b4f09a90-0c53-4593-adcf-4e996331a83d

## What

Kanban column auto-collapse now follows **item count** instead of **terminal status**.

- A column with **zero items shrinks** to the collapsed rail.
- A column with **items expands — even a terminal one** (`Completed` / `Cancelled`).

## Why

`KanbanColumns` seeded its collapsed set from the status enum:

```ts
const [collapsedStageIds, setCollapsedStageIds] = React.useState<string[]>(() =>
  stages
    .filter(stage => isTerminalStageStatus(stage.defaultTicketStatusV2))
    .map(stage => stage.id),
);
```

Two consequences, both visible on any board:

1. A `Completed` column holding real work opened collapsed and stayed that way — the effect *re-collapsed* it whenever the `stages` identity changed, so it fought you unless you had manually toggled it.
2. An empty non-terminal column (`Paused`, `In Review`) kept a full 384px of width while showing nothing.

## How

`stageCountById` resolves each column's count once — server `stageCounts` by id, then by name, then the local bucket length — and the render loop reuses it, replacing the three duplicated lookups that were inlined per column. The effect then collapses zero-count columns and expands non-zero ones. `isTerminalStageStatus` had no other callers and is gone.

Manual toggles still win: `userToggledCollapsedStageIdsRef` short-circuits before the count check, so a column you opened or closed by hand is never overridden.

### The guard that matters for review

```ts
if (!stages.some(stage => (stageCountById[stage.id] ?? 0) > 0)) {
  return;
}
```

Counts arrive asynchronously (`useKanbanCounts`), the backend **omits zero-count keys entirely**, and a collapsed column unmounts its `PaginatedStageList` — so its query never runs and its local length can never recover.

Without this guard the first paint (no counts yet) would collapse *every* column, and in search mode — where `shouldFetchKanbanCounts` is `false` and there is no server count to rescue it — the board would stay wedged shut. Initial state is `[]` for the same reason: every column mounts, loads, and only then collapses if genuinely empty.

The updater also returns `prev` when the set is unchanged, since `stageCountById` depends on `ticketsByStage`, whose identity changes per render.

## Proof of Testing

Recorded against the local stack on the **My Tickets** board (status columns), same account, same fixture, same script — only the fix toggled.

| column | items | before | after |
|---|---|---|---|
| Todo | 10 | full | full |
| Started | 10 | full | full |
| **Paused** | **0** | **full** | **shrunk** |
| **Completed** | **4** | **shrunk** | **full** |
| Cancelled | 0 | shrunk | shrunk *(control — unchanged)* |

Measured shape, straight from the run metadata:

```
before:  TODO(10):full  STARTED(10):full  PAUSED(0):full    Completed(4):shrunk  Cancelled(0):shrunk
after:   TODO(10):full  STARTED(10):full  Paused(0):shrunk  COMPLETED(4):full    Cancelled(0):shrunk
```

`Cancelled` is the control: terminal *and* empty, so it stays collapsed on both sides — proving the change keys on count, not on the status enum.

## Scope

One file, `apps/dashboard/src/components/Tickets/KanbanColumns/KanbanColumns.tsx` — the only Kanban renderer, shared by `KanbanBoardScreen` and `SupportKanbanBoard`.

Typecheck clean; dashboard lint reports the same 3 pre-existing warnings as `main`, no new ones. No persistence added — collapse state still resets on remount, unchanged from before.


samitbarai (dismissed):

PiyushJuspay:
`H1 Search mode: results that land in auto-collapsed columns become silently invisible
- What: With an active search, server stageCounts are not fetched (KanbanBoardScreen.tsx:3229), so counts come only from per-column mounted queries. Once empty columns auto-collapse, their PaginatedStageList unmounts, killing their useKanbanTicketsPage subscription — a collapsed column can never report a non-zero count again. If the user refines the search so matches now exist only in collapsed columns, the guard blocks the pass and those tickets are never fetched or shown. Old code never auto-collapsed non-terminal columns, so search previously always surfaced them — this is a new regression in a core workflow.
- Where: apps/dashboard/src/components/Tickets/KanbanColumns/KanbanColumns.tsx:411 — if (!stages.some(stage => (stageCountById[stage.id] ?? 0) > 0)) { return; } (with unmount at :565 {!isCollapsed && (...)}; parent wipes stale reports on query-key change at KanbanBoardScreen.tsx:2936-2937)
- Who affected: End users searching on Kanban boards with column pagination (the common case — canUseKanbanColumnPagination is true for all non-workspace board views).
- Impact: Silent "no results" for tickets that exist; no count badge signal (collapsed header shows 0). Recoverable only by manually expanding collapsed columns or clearing the search. Non-search mode self-heals via live server counts.
- Fix: Reset non-user-toggled collapse state when the column query key changes (e.g. key the auto-collapse state off kanbanColumnQueryKey), or skip auto-collapse entirely when server counts are unavailable (search mode) — consistent with the guard's own rationale. QA: search a term with matches in column A only → refine to a term matching only previously-collapsed columns → results must appear.`

PR #1296, opened 2026-09-01
fix: XYNE-61392 keep unsaved saved-view edits across remounts

harshpreet931:
POT - 

https://github.com/user-attachments/assets/fedbfbc5-7902-41b9-ab28-361d3eca4d8b

## What

Unsaved edits to a saved view survive leaving the view and coming back, and the view now tells you it is dirty instead of quietly discarding your work.

- Edits (filters, groupBy, visible columns) persist as a **versioned sessionStorage draft**.
- Seeding precedence is **URL > draft > saved definition**.
- New **`Unsaved changes`** indicator with a **`Reset`** that restores the saved definition.
- An already-named view **saves in place** instead of reopening the name popover.
- The machine's filter storage key is now **per `viewId`**, so one view's filters no longer bleed into the next.

## Why

Opening a ticket from a saved view unmounts the board. On the way back the seed effect fell straight through to the saved definition:

```ts
setFilters(initialFilters ? { ...initialFilters } : {});
setGroupBy(initialGroupBy ? parseGroupBy(initialGroupBy) : 'none');
```

so any in-progress edit was silently discarded. The same happened on refresh or on re-entering the view from the sidebar without its query string.

**Columns were the worst case.** They never live in the URL, so unlike filters they had no fallback path at all — a column edit could not be recovered by any means once the component unmounted.

Separately, every `workspace-view` shared one filter storage key, so switching between two saved views carried the first one's filters into the second.

## How

**`viewDraft.ts`** (new) — `read` / `write` / `clearViewDraft`, keyed `view-draft-<viewId|new>`, stamped with `DRAFT_VERSION` and validated on read (wrong version, non-object filters, or a non-array `columns` all return `null`). Every access is wrapped in `try/catch`, so a browser with storage disabled degrades to the old in-memory behaviour rather than throwing.

**Dirty tracking** — `viewSignature()` serialises `filters + groupBy + columns` through the existing `filtersToValues`, sorted and joined, so comparison is order-independent. `isViewDirty` compares the current signature against the saved definition's. The draft is written while dirty and cleared the moment the view matches its saved definition again, or is saved.

`hasSeededViewRef` became `hasSeededView` state because `isViewDirty` has to react to seeding completing — a ref would not re-render and the indicator would not appear until some unrelated state change.

**Storage key** — `getStorageKey` takes `viewId` and returns `...-workspace-view-<viewId|new>` for workspace views.

## Review notes

- `savableColumns` filters `DERIVED_COLUMNS` and sorts, and is now the single source used by both the draft and `persistWorkspaceView` — previously the save path built that list inline, so the two could disagree.
- Draft writes are keyed on `viewDraftKey`, which is `viewId ?? 'new'`. An unsaved `/views/new` therefore keeps its own draft, and it is cleared on successful create before navigating to the new view id.
- `handleResetWorkspaceView` goes through `mergeSavedColumns` rather than replacing the set, keeping the derived columns the table always needs.
- Draft lives in `sessionStorage`, not `localStorage` — deliberate. It should survive a remount and a refresh within the tab, not resurrect a half-finished edit in a new session days later.

## Scope

`apps/dashboard` only, 3 files: `viewDraft.ts` (new, 50 lines), `KanbanBoardScreen.tsx`, `ticketFiltersMachine.ts`. No schema, no backend, no migration.

Pre-commit passed in full: gitleaks, tenant guard, dashboard lint (0 errors), `dashboard-validate` (tsc + vite build), change analysis, enum guard.

## Proof of Testing

Recorded against the local stack on the saved view **"Delivery triage"** (board = Delivery, 20 tickets). Same account, same view, same script — only the fix toggled.

Turn the **Assignee** column off, then re-enter the view the way a sidebar click or a refresh does (no filters in the URL):

| step | before | after |
|---|---|---|
| open the saved view | 10 assignee avatars | 10 assignee avatars |
| turn Assignee off | 0 avatars | 0 avatars |
| **re-enter the view** | **10 avatars — edit discarded** | **0 avatars — edit survived** |
| dirty indicator | never shown | `Unsaved changes` + `Reset` |

```
before:  start 10, after edit 0, after return 10;  dirty false -> false
after:   start 10, after edit 0, after return  0;  dirty true  -> true
```

The middle row is the control: the edit applies identically on both sides, so the only thing that differs is what survives leaving the view.

Columns were chosen deliberately for the recording. The view URL is rewritten to carry `board=...`, so `urlHasFilters` short-circuits the filter seed — but `visibleColumns` is seeded in the `useState` initializer independent of the URL, which is exactly the path with no other way back.


PiyushJuspay:
`H1 Roles filter edits are invisible to dirty-tracking and now block the Save button
- What: The workspace view offers the Roles (roleAssignments) filter (TicketFiltersDropdown.tsx:97, rendered at :584) and applies it (KanbanBoardScreen.utils.ts:407, useKanbanTicketsPage.ts:226), but filtersToValues never serializes it — so a Roles-only edit leaves isViewDirty false.
- Where: KanbanBoardScreen.tsx:214-231 — const WORKSPACE_VIEW_ARRAY_KEYS = ['boards','priority','assignee','createdBy','userGroups','tags','stages','ticketTypes','sourceChannels'] (no roleAssignments); and KanbanBoardScreen.tsx:3706 — disabled={!workspaceViewReady || isSavingWorkspaceView || !isViewDirty}
- Who affected: Dashboard users editing an existing named view via the Roles filter.
- Impact: Three regressions vs base: (1) no "Unsaved changes" indicator; (2) no draft written (KanbanBoardScreen.tsx:1030-1040 gates on isViewDirty), so the edit is still lost on remount — the exact bug this PR fixes; (3) the in-place Save button is newly disabled (base popover trigger was disabled={!workspaceViewReady || isSavingWorkspaceView}, base file line 3612), so the edit cannot be saved at all. The per-board Save View already serializes this filter (TicketFiltersDropdown.tsx:477-480) — the canonical pattern to mirror.
- Fix: Serialize roleAssignments in filtersToValues (+ matching case in valuesToFilters), and/or drop !isViewDirty from the Save disabled condition.
H2 Stale view-draft-new overrides share-link config
- What: Seed precedence "URL > draft > saved" only honors query params; share links encode the config in the URL hash (#cfg=, savedViewSerialization.ts:102, decoded in ProjectViewBuilder.tsx:42-45), which urlHasFilters never sees.
- Where: KanbanBoardScreen.tsx:977-982 — const urlHasFilters = Object.keys(state.context.urlFilters ?? {}).length > 0; ... if (initialDraft) { setFilters({ ...initialDraft.filters });
- Who affected: Anyone opening a share link in a tab where they previously edited (without saving) a new view in the same session.
- Impact: Recipient sees their own stale draft filters instead of the shared view; share link appears broken (recoverable via Reset). Edge case, but it violates the PR's own stated precedence.
- Fix: Treat a decoded #cfg= as URL-level precedence — skip the draft when a shared config seed is present.`

PR #1313, opened 2026-09-01
feat: XYNE-61334 Streams, a spatial deck of Xyne surfaces

adithyaganesh1:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

XYNE-61334. Following a thread, a ticket or a file costs you the place you found it in. The surface
you were reading gets replaced by the thing you opened from it, and getting back is a navigation of
its own.

**Streams** is a horizontal strip of resizable columns. Each column hosts a surface Xyne already
has: channel, board, ticket, thread, document, file, Ask AI. Opening an item lays it down beside
the thing that listed it, and the list stays where it was.

Nothing here reimplements a surface. `surfaces.tsx` is a registry, and a column is an existing
component mounted inside a router scope.

What it does:

- Columns are added from a palette (`cmd+T`), or by clicking an item inside another column. A
  ticket on a board or a file in a channel opens next to its host.
- Columns can be pinned (held at the left edge, outside the scroller), reordered by drag, and
  resized. An item opened from a channel attaches to it as a split pane, so the pair moves and
  resizes together.
- Focus mode turns the strip into a snap carousel for working in one column at a time.
- Multiple streams, switched from the header.
- An "Add to stream" item in the overflow menus that a channel header, sidebar row, thread panel,
  canvas and ticket header already had, so you can file something into a stream without leaving
  the page you are on.

Layout is persisted per stream in `localStorage`. The reader validates what it finds and copies
anything unreadable to a backup key rather than silently starting empty.

Fifteen commits, all `XYNE-61334`. Five built the feature; the rest are review changes, listed
under **Changes since first review** below.

### Worth knowing before you review

**Where to spend your time.** 27 of the files are new and live under `routes/StreamsScreen/`. They
are self-contained, and if Streams breaks then only Streams breaks. The other 17 are modifications
to shared code, and that is where the risk is:

- **The cmd+K dialog is in scope.** `AddColumnPalette` renders `GlobalCommandMenu` inline, so
  `ChannelCommandMenu` (plus its types) and `GlobalCommandMenu` are touched. `ChannelCommandMenu`
  gains a `selectionVariant` prop and its tab row was reworked to support it. Existing behaviour
  is preserved: deselect-to-ALL, arrow-key handling, tab centring. This is the largest single hunk
  in the diff and the one most worth reading closely.
- **Ask AI gains embedding props**, all optional and defaulted: `seedQuery` and `seedNonce` place
  text in the composer without sending it, `hideEmptyStateSuggestions`, `hideHeader`. The floating
  Ask AI drawer is suppressed on `/streams` so one trigger cannot produce two chats.
- **`TicketDetails` gains `hideBackNav`**, default `false`. A ticket column is the ticket, so the
  back chevron would point nowhere and the id would be printed twice.
- **`global.css`** gains a streams-scoped block. Every selector is `.streams-*` and every custom
  property it defines is read only by this feature, so nothing here can reach other screens.
- **Explorations built alongside this are excluded** and stay on the exploration branch: topic
  feeds, an overview grid, catch-up summaries, a dock, jump pills, a side rail, per-column
  summaries, wallpaper switching and the tuning panel. This PR is the settled subset.

No new dependencies. No backend, Prisma, Zero or API changes.

### Changes since first review

**Structure.** The feature moved out of `routes/` into `components/Streams/`, following
`components/Chat` and `components/KnowledgeBaseV2`. Every module has its own PascalCase folder
(`components/StreamColumn/StreamColumn.tsx`); there is no feature-level `utils/`; `hooks/` holds the
five hooks. `routes/StreamsScreen/` is gone and `AppRoot` imports the screen from `components/`, as
it already does for `KnowledgeBaseV2Screen` and `CanvasScreen`.

**No React Context.** The three contexts are gone. `StreamsDev` was frozen constants, so its five
call sites import `DEV_DEFAULTS` directly. `StreamsActions` was callbacks, so they travel through
`BodyProps`. The one thing that genuinely crossed a feature boundary — which channels the stream is
showing, read by Ask AI's context picker — is now `machines/streamsMachine.ts`, a singleton actor in
the same shape as `shortcutsMachine`, read with `useSelector`.

**Dead code.** Four of the five top-nav variants were unreachable once the dial panel went, along
with two render blocks in `StreamsScreen`. Seventeen dev settings were dials for features cut from
this PR — StreamCatchUp, StreamDock, StripJumpPills, StreamsOverview, FocusRail — and eleven CSS
custom properties were written on every render that no surviving rule read. All removed: −608 lines.

**CSS.** `global.css` goes from 741 lines to 111. Comments stripped, dead variant rules deleted, and
press states, the strip scrollbar, the flash animation, the head fade and the nav badge converted to
Tailwind utilities and `tailwind.config.js` keyframes. What remains is the bleed border rule, six
`@property` registrations, and two rules that cannot be utilities: `.streams-nav`, because Tailwind
rewrites any arbitrary value starting with `--` into `var(...)` so a transition on a custom property
compiles to an empty rule; and `.streams-nav-tab`, whose double `color` declaration is a
`color-mix` fallback that needs guaranteed source order.

**Three bugs found while doing the above**, all from cut features leaving things behind:

1. Column headers sat lighter than their body in dark mode — the shell painted `bg-card` while every
   surface inside paints `bg-background`. The two tokens are identical white in the light theme and
   four points of lightness apart in midnight, so it could only ever show in dark.
2. The same defect in a second place: the seam strip between an attached pair.
3. The tab row's hover fade never animated. It referenced `--jump-ease`, defined by StripJumpPills
   and cut with it; an undefined custom property invalidates the whole declaration, so the row
   snapped instead of fading.

## Areas Touched

- [x] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

## Schema & Data Changes

## Config, Environment & URLs

## API Contract

- [x] No API changes

---

## How did you test it?

**Walkthrough video:** https://drive.google.com/file/d/13i3fyD1-xhpAVLuvbK99q774-1djJfJM/view

Worth watching before the diff. Streams is a layout and pointer feature, and two minutes of
seeing it work explains more than the file list does.

**Verified mechanically, rebased onto current `main`:**

- `pnpm run build:shared` passes
- `pnpm --filter xyne-spaces-backend run typecheck` and `run build` pass (after `db:generate`)
- `pnpm --filter xyne-spaces-dashboard run lint:errors-only` passes, zero errors
- `pnpm --filter xyne-spaces-dashboard run typecheck` (`tsconfig.app.json`) passes
- `pnpm --filter xyne-spaces-dashboard run format:check` passes
- `pnpm --filter xyne-spaces-dashboard run build` passes
- `pnpm --filter agentic-framework run lint` passes
- `gitleaks dir . --config .gitleaks.toml` reports zero findings in tracked files

**Bugs found and fixed before raising this.** Each was verified against the code rather than
assumed:

1. "Attach all N from this stream" attached only one channel. The loop called a toggle that
   rebuilt selection from a render-captured prop, so every call read the same stale array.
   Replaced with a bulk `onAttachChannels` that folds the list in once.
2. Channels in the current stream rendered twice in the context picker, once hoisted and once
   under their category, with an identical `cmdk` value. Two rows selected together and arrow-key
   navigation was trapped. Category groups now exclude the hoisted ids.
3. Dragging the seam between an attached pair committed only the dragged half, so the partner
   snapped back on release and the pair's total width changed.
4. The digit shortcuts `1` to `9` indexed the raw column array including attached panes, so a
   digit could focus half a panel, which has no carousel snap point. They now count stops.
5. Two fast focus jumps could leave `scroll-snap` disabled for the rest of the session. Nothing
   cancelled an in-flight tween, so the second one saved the first one's overrides as though they
   were the original values.
6. Every activity probe re-rendered on every screen render: an unmemoised subtree, plus a default
   `{}` argument that invalidated a neighbouring memo each render.
7. The legacy column-width migration mapped preset index `1` to a 1440px column instead of 720px.
8. An import cycle. `useAddToStream` read one number, a surface's minimum width, from the surface
   registry, and that registry imports every surface component, three of which now host an
   add-to-stream menu: registry, component, menu, hook, registry. Whether such a cycle is harmless
   depends on module evaluation order, so it survived on the exploration branch and not here. The
   widths moved to `Streams.types`, which imports nothing, and the registry reads them from there
   too, so there is still one source of truth.

### Automated

- [x] Not covered by tests 
      This is a client-only surface whose state is layout and pointer interaction: drag, resize,
      snap, focus. The existing suites cover neither. Happy to add e2e coverage for the column
      lifecycle (add, attach, pin, close, reload) if that is wanted before merge.

### Manual

Run against a matched local stack, with backend, Zero and database all built from this branch and
seeded with a demo workspace of 6 people, 14 channels, 343 messages and 24 tickets.

Exercised: adding columns from the palette and the cmd+K dialog inside it; resize; drag to
reorder; pin; opening a ticket, thread or file from inside a column so it attaches as a split
pane; dragging an attached pair's seam; focus mode and the 1 to 9 stop shortcuts; the context
picker's dedupe and bulk attach; the add-to-stream entry points; and reload persistence.

Five things came out of that testing and are fixed in the last two commits. The default column
width is now 360px. The stream lets the app wallpaper through instead of painting over it. The add
button scrolls the strip to the palette rather than opening it off-screen. The header no longer
changes height when the first column is added, which it did because the nav is the only 32px child
of that row and the switcher is 23.5px. Header padding is back to 6.

**Users**
- [x] Single user
- [ ] Two users at once

**Login**
- [ ] Google
- [ ] Microsoft
- [x] Local seeded admin

**Run mode**
- [x] Local dev
- [ ] Test mode / sandbox / prod-like / Docker
- [ ] Electron desktop

**Sync & resilience**
- [x] Reload / hard refresh, layout restored correctly
- [ ] Multiple tabs, offline then reconnect, concurrent edits

**Surface & data**
- [x] Web browser
- [x] Empty state
- [x] Realistic data volume (14 channels, 343 messages, 24 tickets)
- [ ] Electron desktop
- [ ] Narrow viewport (out of scope, see limitation 5)
- [ ] Error paths

**On Electron:** unchecked because it was not run, not because it is expected to fail. Electron
loads the same dashboard bundle (`copy-ui` copies `dashboard/dist` into the app), Streams contains
no Electron-specific code, and `xyne-spaces://` is already in the CORS allowlist. Worth one pass
before release, with attention to the window drag regions near the strip.

---

## Known limitations and follow-ups

Listed here so none of it is a surprise in review.

**1. Layout is stored only in `localStorage`.** Streams do not follow a user to another browser,
device or profile, and the backend cannot see them. That is workable for a first version, but it
is a product decision worth confirming rather than inheriting.

> ⚠️ **2. "Try a different account" destroys every stream.** `WorkspaceSelectionScreen.tsx:159`
> calls a blunt `localStorage.clear()`. It predates this PR and is reasonable for leaving an
> account, but it now also deletes every stream in every workspace, permanently, from a button
> that reads like it only switches accounts. `loadLayout` writes a backup copy when stored data is
> unreadable, and `clear()` removes the backup too. The cheap fix is removing the specific auth
> keys instead of calling `clear()`. The real fix is server-side persistence (1).

**3. No automated tests.** Covered above. The column lifecycle (add, attach, pin, close, reload)
is the obvious first e2e case if that is wanted before merge.

> **4. No feature flag.** Merging makes Streams visible to every user with the nav item. The repo
> has flag infrastructure that this does not use. Happy to gate it before merge if that suits the
> rollout better, and it would also make the absent tests a much smaller risk to carry.

**5. Desktop only, by decision.** There is no mobile or narrow-viewport handling anywhere in the
feature. A strip of drag-resizable columns is a pointer interaction, and small screens were out of
scope for this version rather than overlooked.

**6. Existing users will not see it in the rail.** `/streams` is a default toolbar path, not a
required one, so anyone who has previously customised their toolbar finds it under **More** and
can move it into the rail from Preferences. Making it required would render it permanently
un-hideable, which seemed worse, but it does mean a quiet rollout for existing accounts.

**7. The dev tuning panel is gone; the settings that still drive something remain as frozen
constants.** `DEV_DEFAULTS` in `StreamsDev.ts` holds them, read through a plain import, so
reopening any question means restoring a panel over the same object. The seventeen settings whose
features were cut, and the four unreachable top-nav variants, have been deleted rather than left
in place. Two behaviours are frozen by that panel and worth a product opinion: `markRead` is
`'never'`, so channel columns never auto-mark-read, and `focusRing` is off, which is why
`STREAM_LEFT_INSET` is `0`.

**8. Virtualisation is untested against `document` and `agent` columns.** The strip now unmounts
columns that scroll out of range and rebuilds them on return, and a column only builds its surface
once it is actually on screen. That has been exercised against `channel`, `board`, `ticket`,
`thread` and `file`, but not the other two: Canvas and AI are off in the local stack, so neither
surface ever mounted. A `document` column holds a collaborative editing session, which is the one
most likely to object to being torn down and rebuilt repeatedly — worth a look from someone running
the full stack before merge.

**Unrelated, found while working here.** Three small bugs that affect everyone and deserve their
own PR:

- the post-commit automation runner passes `GIT_INDEX_FILE` into `git worktree add`, so it always
  fails (`tools/xyne-automation/scripts/runner/index.ts`);
- `verify-test-code-integrity.sh` runs `git fetch --depth=1`, which silently re-shallows a
  developer's full clone;
- a fresh `pnpm env:setup` cannot boot the backend, because the shipped
  `apps/backend/.env.example` contains the literal `devkey` and `devsecret` LiveKit placeholders
  that its own Joi validator rejects;
- `docker-compose.dev.yml` points MinIO at `mirror.gcr.io/minio/minio:latest`, which resolves
  nowhere — MinIO publishes to Quay rather than Docker Hub, so the pull-through mirror has nothing
  to mirror and `pnpm run services` fails outright on any machine without the image already cached.

---

## Checklist

---


XyneSpaces (commented):
## Xyne Spaces Reviewer — PR #1313

**Verdict:** Changes requested — **0 blocking, 3 major, 3 minor, 3 nits.** The majors are all pre-merge process risks (tests, rollout flag, file size), not code defects; the code itself is among the better-quality feature PRs this repo has seen.

**Reviewed:** 45 files, every hunk (GitHub files API: 30 + 15 verified page-2 files — the PR body's "44" predates the restructure). Head `0146341`. No schema / Zero / migration changes → migration-safety review not triggered, no QA tag.

### Findings
|---|---|---|---|
| major | R7 | StreamsScreen.tsx:1 | Zero automated tests for a ~5.7k-line layout/pointer feature; the offered column-lifecycle e2e should land before merge. |
| major | R9 | navigationConfig.ts:174 | No feature flag — `/streams` ships in `DEFAULT_TOOLBAR_PATHS`, so merge = visible to every user. Author has offered a gate; take it. |
| major | #314 | StreamColumn.tsx:1 | Six new files over the 400-line limit: StreamsScreen 2,611 · StreamColumn 911 · StreamTopNav 763 · StreamSwitcher 513 · Streams.types 475 · ColumnResizeHandle 412. |
| minor | R2 | FileColumn.tsx:90 | Renders the literal string `"DownloadDown"` where an icon/label was intended — the download control shows raw text. |
| minor | R5 | StreamRouterScope.tsx:7 | Depends on react-router `UNSAFE_*` internal contexts — documented in-file; open the follow-up the doc names before Streams graduates. |
| minor | R2 | StreamsLayout.ts:46 | `STORAGE_KEY` is not workspace-scoped: same-browser workspace switch loads the other workspace's columns, which fail to resolve. |
| nit | R5 | StreamsDev.ts:113 | One duration, three numbers: doc says "Settled: 430", `DEV_DEFAULTS.focusMs` says 280, shipped `FOCUS_MS` is 260 — and nothing reads `focusMs`. |
| nit | R6 | streamsLayout.ts | `console.warn/error` + eslint-disable — justified in-code as the only data-loss signal; consider the shared `logger` util for consistency. |
| nit | #139 | machines/streamsMachine.ts | Module-scope XState singleton — same shape as `shortcutsMachine`; noted for the record. |

### Context from the requesting thread
- **"The release ticket says it shipped":** this PR is **open and has never been merged** (`merged_at: null`). The 09-03 release (`64b9b449` on `release-20260827`) shipped #1384, a backport of #1370; the ticket bot auto-associated XYNE-61334 with that release. **Nothing from this PR has shipped anywhere.**
- **Feature flag:** none today. `/streams` is a *default* toolbar path, not a *required* one — existing users who customized their rail find it under **More**; default users see it out of the box. Regression risk is otherwise contained: 27 of the files are new and self-contained, and every shared-code touch is additive and defaulted (`hideBackNav`, Ask AI embedding props, `ChannelCommandMenu` `selectionVariant` with tab-row behaviour preserved — that last one is the one worth a careful human pass).

### Recommended reviewer
**@harsh-1923** (Fractal / Harsh Sharma) — @mentioned above; please accept the review request when it lands.

|---|---|---|---|---|
| @harsh-1923 | XyneAISidebar + the Ask AI embedding props (riskiest shared-code change); authored sibling XYNE-61334 PRs #1370/#1384 | author's pick (thread) | 0 open review requests here | most recently active in Spaces |
| @santhosh-s | ActivityDot, MessageAttachment integration points | — | 1 | — |
| @aryan.pidiha | ChannelCommandMenu area (largest shared hunk) | — | 1 | — |

Rationale: Fractal owns the Ask AI surface this PR embeds columns around, authored the sibling tickets, and carries zero pending review requests in this repo; both backups carry one. (CODEOWNERS here only covers `apps/dashboard/src/zero/mutators.ts`, untouched — ranking built from file ownership + current workload.)

— *Xyne Spaces Reviewer* · reviewed at `0146341` · 7 inline comments follow

apps/dashboard/src/components/Streams/StreamsScreen.tsx
```diff
@@ -0,0 +1,2611 @@
+import {
```
XyneSpaces, line 1: **[major] R7 — no automated tests.** This is a ~5.7k-line layout-and-pointer feature (drag, resize, snap, focus, persistence) and the PR carries zero tests — acknowledged in the description, with an offer to add e2e for the column lifecycle. With no feature flag either, tests are the only safety net at merge. Please land the offered column-lifecycle e2e (add → attach → pin → close → reload) in this PR before merge; if e2e is too heavy, persistence round-trips through `loadLayout`/`settle` are the highest-value unit target.

apps/dashboard/src/components/Streams/components/StreamColumn/StreamColumn.tsx
```diff
@@ -0,0 +1,911 @@
+import { ReactElement, memo, useCallback, useLayoutEffect, useMemo, useState } from 'react';
```
XyneSpaces, line 1: **[major] #314 — file size.** Six of the new files exceed the 400-line guideline: StreamsScreen.tsx (2,611), StreamColumn.tsx (911), StreamTopNav.tsx (763), StreamSwitcher.tsx (513), Streams.types.ts (475), ColumnResizeHandle.tsx (412). The decomposition into hooks/components already helps, but StreamsScreen in particular holds layout, focus-mode machinery (widthMs/focusTick/tween), palette and persistence wiring in one module — extracting the focus-mode tween into the hooks layer would cut the biggest one roughly in half.

apps/dashboard/src/components/AppSidebar/navigationConfig.ts
```diff
@@ -167,7 +169,9 @@ export const REQUIRED_TOOLBAR_PATHS: string[] = [
 ];
 
 // Paths shown in the toolbar by default (before any user customization).
-export const DEFAULT_TOOLBAR_PATHS: string[] = [...REQUIRED_TOOLBAR_PATHS];
+// Streams is here rather than in REQUIRED so it appears in the rail out of the
+// box but can still be toggled off — it is new, and nothing else depends on it.
+export const DEFAULT_TOOLBAR_PATHS: string[] = [...REQUIRED_TOOLBAR_PATHS, '/streams'];
```
XyneSpaces: **[major] R9 — unflagged rollout.** `/streams` lands in `DEFAULT_TOOLBAR_PATHS`, so merging makes Streams visible to every user (default rail for new users; under **More** for anyone who has customized it). The description notes the repo's flag infrastructure is unused here and offers a gate before merge — please take that path: it converts the missing-tests risk and the `WorkspaceSelectionScreen` data-loss caveat into recoverable ones. Cheap version if a full flag is too heavy: keep the route, gate the toolbar entry on the existing flag infra.

apps/dashboard/src/components/Streams/components/FileColumn/FileColumn.tsx
```diff
@@ -0,0 +1,147 @@
...
+      onClick={download}
+      className={cn(
+        STREAM_PRESS,
+        'flex items-center gap-1.5 rounded-lg border border-border px-3 py-1.5 text-[12px] font-medium text-foreground hover:bg-accent',
+      )}
+      data-track-category='Streams'
+      data-track-name='DownloadFileColumn'
+    >
+      <DownloadDown className='size-3.5' />
+      DownloadDown
```
XyneSpaces: **[minor] R2 — literal string rendered.** Line 90 renders the literal string `"DownloadDown"` where an icon or label was presumably intended — looks like an icon constant that was renamed or never imported, so the download control shows raw text. ↳ FIX: use the intended icon import (the download glyph the rest of the app uses).

apps/dashboard/src/components/Streams/components/StreamRouterScope/StreamRouterScope.tsx
```diff
@@ -0,0 +1,178 @@
+import { ReactElement, ReactNode, useCallback, useMemo, useRef, useState } from 'react';
+import {
+  createPath,
+  NavigationType,
+  parsePath,
+  useLocation,
+  UNSAFE_LocationContext,
```
XyneSpaces, line 6: **[minor] R5 — react-router `UNSAFE_` internals.** These three context imports are private/unstable API; a router major can move them with no compile-time signal at the call sites. The file's own doc comment already names the durable fix (optional view-state props on `ConversationPanelV2`, as `CanvasScreen` does for `canvasId`) — please open that follow-up ticket now so it is tracked before Streams graduates from experiment.

apps/dashboard/src/components/Streams/components/StreamsLayout/StreamsLayout.ts
```diff
@@ -0,0 +1,716 @@
...
+ * a `streams_layout` table, which would buy cross-device sync for free.
+ *
+ * Every field is read defensively. Cascade's hard lesson (§3): its `readLayout`
+ * discarded the *entire* stored object on a parse failure, so one new field
+ * without a default wiped every width and order on first read of an old layout.
+ * Here, a malformed field falls back to its default and the rest survives.
+ */
+
+const STORAGE_KEY = 'xyne.streams.layout.v1';
+
```
XyneSpaces, line 60: **[minor] R2 — layout storage is not workspace-scoped.** `xyne.streams.layout.v1` is one key for the whole browser, so a user active in two workspaces loads workspace A's columns inside workspace B — every column then fails to resolve its channel/ticket ids and renders as an empty shell until cleared. The in-code precedent (`Resizable`) has the same shape, so this is a known trade — but namespacing by workspaceId (`xyne.streams.layout.v1.<workspaceId>`) is a two-line change at read/write time and removes the cross-workspace bleed entirely. Worth doing before this ships, given `WorkspaceSelectionScreen`'s `localStorage.clear()` already makes stream loss easy to hit.

apps/dashboard/src/components/Streams/components/StreamsDev/StreamsDev.ts
```diff
@@ -0,0 +1,252 @@
...
+   *
+   * A dial because the right amount is a function of what is *behind* the stream,
+   * which is itself dialled: the fade is a mask, so it reveals the wallpaper
+   * rather than painting over it, and how far in it has to travel before the
+   * column stops reading as cut off depends on how busy that wallpaper is. The
+   * left edge alone, because it is the one that doubles as the boundary with
+   * the pinned run — the right edge only ever meets the page.
+   */
+  leftFade: number;
+  /** How long entering or leaving focus mode takes, in ms. Settled: 430. */
```
XyneSpaces: **[nit] R5 — one duration, three numbers.** The interface doc says focus-mode takes "Settled: 430", `DEV_DEFAULTS.focusMs` below is 280 with a "Settled at 280" comment, and the value the product actually ships is `FOCUS_MS = 260` in Streams.types.ts — while nothing reads `focusMs` anymore (grepped the full diff and re-read the hooks: no consumers). The dead dial value will mislead the next person tuning this. ↳ FIX: delete `focusMs` and its comment from `DEV_DEFAULTS`, leaving `FOCUS_MS` as the single source of truth.

santhoshJus (dismissed):

harsh-1923 (dismissed):

apps/dashboard/src/components/AppSidebar/navigationConfig.ts
adithyaganesh1: Fixed in `b2bcc7e2`. `/streams` is no longer in `DEFAULT_TOOLBAR_PATHS`, and the whole surface sits behind one flag (`useStreamsVisibility`), off by default. It gates the sidebar, Preferences > Toolbar, the workspace admin Toolbar tab, and `/goto`. Verified in the running app: with the flag unset, "Streams" does not appear on any of them.

apps/dashboard/src/components/Streams/components/FileColumn/FileColumn.tsx
adithyaganesh1: Fixed in `4687e2a4`. The icon's component name was rendering as text. It renders the `Download` icon now.

apps/dashboard/src/components/Streams/components/StreamsLayout/StreamsLayout.ts
adithyaganesh1: Fixed in `8538bc97`. The key is now `xyne.streams.layout.v1.<workspaceId>`, threaded through `loadLayout`, `saveLayout` and `backupUnreadable`, plus all seven call sites. No migration was needed since Streams has never shipped.

Verified against a real 19 column layout: it loads from the scoped key, a change writes to the scoped key and leaves the old global key untouched, and it survives a reload. The add to stream menu on a non Streams route reads the scoped key too. I confirmed that by planting a sentinel value in the old key and checking which one the menu listed.

apps/dashboard/src/components/Streams/components/StreamsDev/StreamsDev.ts
adithyaganesh1: Fixed in `8538bc97`, but the opposite way round from the suggestion. `dev.focusMs` is read in six places in `StreamsScreen.tsx` and drives the actual transition, so it is live. `FOCUS_MS` was the dead one, named only inside two comments and never in code. So `FOCUS_MS` is deleted, and the interface doc now says 280, which is the value that runs.

apps/dashboard/src/components/Streams/StreamsScreen.tsx
adithyaganesh1: Not in this PR. `apps/dashboard` has no test infrastructure at all: no vitest, no jest, no testing library, and zero test files. No CI workflow runs a test job either, so a test added here would not gate the merge. The e2e harness in `tools/xyne-automation` has no specs and is not in CI.

Installing a test runner in the dashboard is a decision for the whole app, not something a feature PR should settle on its own. Streams is behind a flag, off by default, which is the regression guard at merge.

apps/dashboard/src/components/Streams/components/StreamColumn/StreamColumn.tsx
adithyaganesh1: Acknowledged, but not in this PR. Pulling the focus mode tween out of `StreamsScreen` means refactoring the largest and most stateful file in the feature. With no test coverage, doing that right before merge is more risk than it is worth. The split into `hooks/` and per component folders already landed here.

apps/dashboard/src/components/Streams/components/StreamRouterScope/StreamRouterScope.tsx
adithyaganesh1: Deliberate and contained. React Router rejects a nested `<MemoryRouter>` at runtime, so supplying the three contexts directly is the only way to give each column its own private URL state. The trade off, and the durable fix (optional view state props on `ConversationPanelV2`, the way `CanvasScreen` already does it for `canvasId`), are written up in the file's own header. It is one file, so a router major version would break in one place.

apps/dashboard/src/components/Streams/components/StreamsLayout/StreamsLayout.ts
```diff
@@ -0,0 +1,688 @@
...
+    // opens on something rather than on nothing.
+    return settle(streams, readActiveId(raw, streams));
+  } catch (error) {
+    // Deliberately loud, and deliberately not silent about which key holds the
+    // copy. The failure this guards against is a *transient* one — a module
+    // half-applied by a hot reload, say — so by the time anyone investigates,
+    // the code that threw may read perfectly well.
+    // eslint-disable-next-line no-console -- the stored layout was unreadable; say so loudly
+    console.error(
+      `[streams] could not read the stored layout — keeping the original at "${backupKey(workspaceId)}"`,
```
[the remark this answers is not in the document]
adithyaganesh1: Fixed in `02bcbb5a`. The key is now passed as its own argument instead of being interpolated, so there is no format string for `workspaceId` to reach.

It was introduced by the previous commit: the backup key used to be a module constant, so interpolating it was inert. Once it derived from `workspaceId`, which comes from the URL, a path containing `%s` could have consumed the `error` argument and hidden the failure being logged.

harsh-1923 (dismissed):

santhoshJus:
Components in a stream should be virtualized

adithyaganesh1:
> Components in a stream should be virtualized

Done. Only the columns actually on screen get loaded now, the rest are placeholders until you scroll to them. Same in focus mode.

santhoshJus (dismissed):

aravindsjp (dismissed):

vshiva205 (approved):

gyash1512 (approved):

PR #1320, opened 2026-09-01
feat: XYNE-60886 posthog release 0901

sithaarth:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


apps/dashboard/src/services/Analytics/posthogService.ts
```diff
@@ -0,0 +1,218 @@
...
+        rageclick: true,
+        // Persist identity across reloads.
+        persistence: 'localStorage+cookie',
+        // Record the full session so the journey can be replayed.
+        disable_session_recording: false,
+        session_recording: {
+          // Chat/email/ticket bodies are sensitive — mask ALL text and inputs by
+          // default. Opt specific safe elements back in with `.ph-no-mask`.
+          maskAllInputs: true,
+          maskTextSelector: '*',
```
ameer030, line 127: **Autocapture is not covered by this masking — element text and attributes still leave the app.**

`maskAllInputs` / `maskTextSelector` are rrweb options: they apply to **session replay only**. Autocapture is a separate pipeline, and it sends `$el_text` plus the element's attributes. Both `mask_all_text` and `mask_all_element_attributes` default to `false`, and neither is set here.

PostHog's default autocapture element allowlist is `a, button, form, input, select, textarea, label`, so this is not hypothetical. Concrete case — `ChannelItemV2.tsx:174`:

```jsx
<Link
  to={`/chat/dir/${channel.id}`}
  data-track-metadata={JSON.stringify({ channelId, channelName: displayName, isDM })}
>
  ... <span>{displayName}</span>
```

`<Link>` renders an `<a>`. Every sidebar click therefore ships the DM counterpart's **name** twice: once as `$el_text`, once inside the captured `data-track-metadata` attribute. The same shape repeats for ticket rows, label chips and email subject rows.

Suggested fix at init:

```js
mask_all_text: true,
mask_all_element_attributes: true,
```

If that blinds autocapture too much, scope it instead:

```js
autocapture: {
  element_allowlist: ['button', 'a', 'input', 'select', 'textarea'],
  css_selector_ignorelist: ['[data-track-metadata]', '.visual-regression-hide'],
  element_attribute_ignorelist: ['data-track-metadata', 'title', 'aria-label'],
},
```

The `data-ph-capture-attribute-*` values added by `Button` are explicit opt-ins and survive `mask_all_element_attributes`, so the `trackId` signal this PR is built around is not lost.

apps/dashboard/src/services/Analytics/posthogService.ts
```diff
@@ -0,0 +1,218 @@
...
+        capture_pageview: true,
+        capture_pageleave: true,
+        // Enable click + scroll heatmaps in the PostHog dashboard.
+        enable_heatmaps: true,
+        // Detect rage clicks / dead clicks for friction analysis.
+        rageclick: true,
+        // Persist identity across reloads.
+        persistence: 'localStorage+cookie',
+        // Record the full session so the journey can be replayed.
+        disable_session_recording: false,
```
ameer030, line 122: **Session replay goes from off to 100% of sessions here, and nothing in the PR flags it.**

Previous behaviour (`mixpanelService.ts` on `release-20260901`):

```ts
private readonly ENABLE_SESSION_REPLAY = false;
// ...
record_sessions_percent: this.ENABLE_SESSION_REPLAY ? 100 : 0,
```

New behaviour: replay on for every session, plus heatmaps, plus rageclick/dead-click. For a product holding chat, email and ticket content that is a materially different data-collection posture and probably needs a privacy owner's sign-off rather than a default in a migration PR.

Two things worth adding regardless of that decision:

1. A consent lever — `opt_out_capturing_by_default: true` wired to whatever consent signal you have, per PostHog's data-collection docs.
2. A sampling rate. 100% replay on a full-day workspace app is a large ingest bill; `session_recording: { sampleRate: '0.1' }` or similar is the usual starting point.

Related: the old service also had an idle-detection loop that paused recording after 3 minutes of inactivity, so replays did not run all day on an idle tab. That is gone too — PostHog does have its own inactivity handling, so this may be fine, just confirming it was a deliberate drop.

One thing that is **not** a regression, to be explicit: the old `isProductionUrl()` gate only ever guarded the `mixpanel.track()` call — `posthog.capture()` already ran before it, on every origin. So event volume outside prod is unchanged. What *is* new is that replay and heatmaps now also run anywhere a key is configured. `.env.sdlc` and `.env.test` carry no PostHog keys today, so it is latent, but it is one env file away from recording SDLC and test sessions.

apps/dashboard/src/services/Analytics/posthogService.ts
```diff
@@ -0,0 +1,218 @@
...
+    try {
+      /* eslint-disable @typescript-eslint/naming-convention */
+      posthog.init(key, {
+        api_host: host,
+        defaults: '2025-05-24',
+        // --- Full user journey capture ---
+        // Automatically capture every click (all buttons), change and submit event.
+        autocapture: true,
+        // Track navigation so the journey is stitched across pages/routes.
+        capture_pageview: true,
```
ameer030: **`capture_pageview: true` gives page-load-only pageviews — the comment above it asks for the opposite.**

The comment says *"Track navigation so the journey is stitched across pages/routes"*, but `true` is the legacy behaviour: one pageview per hard page load. This dashboard is a React Router SPA, so in-app navigation produces no pageview at all.

Per PostHog's docs the `defaults` snapshot sets `capture_pageview: 'history_change'`, but *"any values explicitly set in your configuration will override these defaults"* — so this line actively downgrades what `defaults: '2025-05-24'` was already giving you.

```js
capture_pageview: 'history_change',
```

or just delete the line and let `defaults` handle it.

apps/dashboard/src/services/Analytics/posthogService.ts
```diff
@@ -0,0 +1,218 @@
...
+        capture_pageleave: true,
+        // Enable click + scroll heatmaps in the PostHog dashboard.
+        enable_heatmaps: true,
```
ameer030, line 116: Worth double-checking `enable_heatmaps` against posthog-js 1.413.2 — PostHog's current config docs and their init snippet builder both use `capture_heatmaps`. If `enable_heatmaps` is the pre-rename alias it may still work, but if it has been dropped this line is a silent no-op and you get no heatmaps at all.

apps/dashboard/src/services/Analytics/posthogService.ts
```diff
@@ -0,0 +1,218 @@
...
+      return;
+    }
+    this.capture(`${trackId}_${status}`, { trackId, status, ...properties });
+  }
+
+  /**
+   * Identify a user. All properties except id/picture become person properties.
+   */
+  identify(user: PosthogUser): void {
+    if (!this.isInitialized || !user?.id) {
```
ameer030: **Removes the self-initialisation guard the old service had, leaving a startup race on identify.**

Old `MixpanelService.identify()` opened with:

```ts
if (!this.isInitialized) {
  this.initialize();
}
```

The new one returns early instead. That matters because `authActor` is started at module import (`authMachine.ts:1685`), while `initialize()` only runs inside `AnalyticsProvider`'s mount effect (`App.tsx:141`). If the `authenticated` entry action ever fires before that effect — a warm session-restore path, for instance — `identify()` and the `AUTHENTICATION` login event are both silently dropped and the session stays anonymous.

Today the auth check is async so this will usually lose the race in your favour, which is why I'd call it latent rather than broken. But it is a defence that existed and no longer does, and it is a one-liner to restore:

```ts
identify(user: PosthogUser): void {
  this.initialize();
  if (!this.isInitialized || !user?.id) return;
```

Alternative, arguably cleaner: call `posthogService.initialize()` at module scope in `main.tsx` so init is never ordered behind a React effect.

apps/dashboard/src/services/Analytics/posthogService.ts
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  /**
+   * Capture the outcome of an action. Emits `<trackId>_<status>` (e.g.
+   * `save_call_summary_prompt_success` / `_failure`) so a mutation's pass/fail is
+   * recorded — a signal autocapture cannot produce, since it only sees the click,
+   * not the async result. Clicks themselves are covered by autocapture, so this
+   * intentionally has no `click` status.
+   */
+  captureActionOutcome(
```
ameer030, line 196: `captureActionOutcome` mints a **new event name per trackId** (`save_call_summary_prompt_success`, `..._failure`, …). With 62 call sites already, that is well over a hundred distinct event names, and it grows with every new button.

That hurts on PostHog's side: event names are the unit of grouping for insights and funnels, so "what fraction of all actions failed?" becomes a query over a hundred names instead of a breakdown on one. And the information is already being sent as properties on the very next line (`{ trackId, status, ... }`), so the name is pure duplication.

```ts
this.capture('action_outcome', { trackId, status, ...properties });
```

One event, breakdown by `trackId`, filter by `status`. Same data, far more queryable.

apps/dashboard/src/services/Analytics/posthogService.ts
```diff
@@ -0,0 +1,218 @@
...
+    }
+  }
+
+  /**
+   * Enrich custom event properties with user context.
+   */
+  private getEnrichedProperties(properties?: EventProperties): EventProperties {
+    const enrichedProps: EventProperties = {
+      ...properties,
+      timestamp: Date.now(),
```
ameer030, line 159: `timestamp` is carried over verbatim from the Mixpanel service. PostHog already stamps a `timestamp` on every event it ingests, so this property is at best redundant and at worst confusing when someone reads it back in an insight and does not know which one they are looking at.

Suggest dropping it, or renaming to `clientTimestamp` if the client-side clock is genuinely something you want to compare against ingest time.

apps/dashboard/src/components/ui/Button/Button.tsx
```diff
@@ -48,19 +95,56 @@ function Button({
...
+    () => buildCaptureAttributes(trackId, trackProps),
+    [trackId, trackProps],
+  );
+
+  const handleClick = React.useCallback(
+    (event: React.MouseEvent<HTMLButtonElement>): void => {
+      // Clicks are recorded by PostHog autocapture; the Button only owns the
+      // action lifecycle here. When a trackAction is supplied, run it with a
+      // pending spinner that blocks double-submits.
+      if (trackAction) {
```
ameer030, line 126: `trackAction` silently discards `onClick` when both are passed — the early `return` skips `onClick?.(event)` below.

It is documented in the JSDoc and no current call site passes both (8 `trackAction` usages, checked), so nothing is broken today. But this is a shared primitive used in ~200 places, and the failure mode is a click handler that quietly stops firing with no type error and no console warning. Cheap insurance:

```ts
if (trackAction) {
  setPending(true);
  void (async () => {
    try { await trackAction(event); } catch { /* ... */ } finally { setPending(false); }
  })();
}
onClick?.(event);
```

Or keep the exclusivity and make it loud — a `console.warn` in dev when both are set, or make the props mutually exclusive in the type.

apps/dashboard/src/components/ui/Button/Button.tsx
```diff
@@ -3,6 +3,32 @@ import { Slot } from '@radix-ui/react-slot';
...
 import { Loader2 } from 'lucide-react';
+import { type EventProperties } from '../../../services/Analytics/posthogService';
+
+/**
+ * Convert tracking metadata into PostHog-native capture attributes. Any element
+ * attribute prefixed `data-ph-capture-attribute-*` is automatically attached to
+ * the autocaptured click event by PostHog, so we surface the button's purpose
+ * without emitting a separate custom event.
+ */
+function buildCaptureAttributes(
```
ameer030, line 14: Attribute names built from arbitrary `trackProps` keys have two sharp edges.

**Casing is silently lost.** HTML attribute names are case-insensitive, so `trackProps={{ trackCategory }}` (`SummaryTemplateMenu.tsx:84`) becomes `data-ph-capture-attribute-trackcategory` in the DOM, and that lowercased form is what lands in PostHog. Minor, but it means the property name you read in the dashboard is not the one you wrote in the JSX.

**A bad key throws at render.** `setAttribute` raises `InvalidCharacterError` on a name containing a space, `$`, or similar. All 8 current usages are safe identifiers, so this is prevention rather than a live bug.

A kebab-case normaliser covers both:

```ts
const attrKey = key.replace(/([a-z0-9])([A-Z])/g, '$1-$2').toLowerCase().replace(/[^a-z0-9-]/g, '-');
```

apps/dashboard/src/components/Tickets/TicketDetails/TicketDetails.tsx
```diff
@@ -2990,28 +2990,32 @@ export const TicketDetails: React.FC<TicketDetailsProps> = ({
             )}
           </div>
           {onUnmerge && (
-            <button
+            <Button
               type='button'
+              variant='ghost'
               onClick={onUnmerge}
+              trackId='unmerge_ticket'
```
ameer030, line 2997: **This is the one I'd most want a screenshot for.** The PR converts **216 native `<button>` elements to `<Button>`**, and only 3 gained a `size=` prop and 22 an `h-auto` class. The other ~190 silently inherit `size: 'default'`:

```
h-9 px-4 py-2 has-[>svg]:px-3
```

plus the cva base (`gap-2 rounded-md text-sm font-medium cursor-pointer [&_svg:not([class*="size-"])]:size-4`).

`cn()` is `twMerge`, so a `className` height or padding does override — but only where one is actually present. Three from this file alone where none is:

- **Here (`:2993`)** — `className='text-sm text-primary hover:text-primary/80 font-medium whitespace-nowrap'`. No height, no padding. An inline text link becomes a 36px-tall control with 16px side padding, inside a row that was laid out for text.
- **`:3666`** — the `absolute right-[-20px] top-1/2 -translate-y-1/2` reference-remove `X`. Also no padding class, so it picks up `px-4` and widens well past the 20px offset it is positioned against.
- **`:3680`** — the tag-remove `X` has `ml-1 p-0.5`, which twMerge does resolve against `px-4 py-2`, but there is no height class, so `h-9` stands and the tag pill grows to 36px tall.

This is a UI refactor riding inside an analytics migration, and the *How did you test it?* section is empty. Two ways out, either is fine:

1. Add `size='icon'` / `size='sm'` / `h-auto p-0` per site and post before/after screenshots of the busiest surfaces (ticket details, chat sidebar, desk composer).
2. Split the `<button>` → `<Button>` conversion into its own PR and land the analytics change on its own. The `trackId` attributes are the part that needs `<Button>`; nothing else in this PR does.

For what it's worth I'd lean towards (2) — it makes both halves reviewable, and the analytics half is the one with a release date attached.

apps/dashboard/src/components/AIScreen/AskAiRatingButtons.tsx
```diff
@@ -81,7 +82,9 @@ export function AskAiRatingButtons({
 
   return (
     <span className={cn('inline-flex items-center gap-0.5', className)}>
-      <button
+      <Button
+        variant='ghost'
+        trackId='ask_ai_rate_up'
```
ameer030, line 87: Same `size='default'` inheritance as the ticket buttons, and this one collides awkwardly.

The `className` sets `h-7 w-7`, so twMerge drops the inherited `h-9` — but `px-4 py-2` (and `has-[>svg]:px-3`) are a different conflict group and survive. Result is a `w-7` (28px) box carrying 12–16px of horizontal padding on each side, with `[&_svg]:shrink-0` keeping the icon at full size. The thumbs-up/down will not sit where they used to.

The Save button at `:156` has the mirror problem: its `px-2 py-0.5` *does* override the padding, but with no height class it inherits `h-9` and grows from a compact inline control to 36px.

`size='icon'` on the two thumbs and `h-auto` on Save should restore all three.

apps/dashboard/src/services/clients/apiClient.ts
```diff
@@ -159,11 +123,9 @@ apiConfig.interceptors.response.use(
     // Token refresh is now handled by backend via HTTP-only cookies
     // No need to manually update tokens from frontend
 
-    // Track API latency for successful requests
     const config = response.config as InternalAxiosRequestConfig & {
```
ameer030, line 125: `trackApiLatency()` is deleted here and at the error interceptor, and with it the `PERFORMANCE_METRIC` / `API_LATENCY` event for every request.

This is part of a wider deletion — grepping the branch, `PERFORMANCE_METRIC` now has **zero** references anywhere, and the whole `PERFORMANCE_METRIC_TYPES` block is gone from `events.ts`. That takes out:

| metric | was in |
|---|---|
| `API_LATENCY` | `apiClient.ts` (this file) |
| `WS_CONNECTION_RECONNECT_ATTEMPT` | `socketClient.ts` |
| `CALL_JOIN`, `CALL_CONNECTION_FAILED` | `roomMachine.ts` |
| `MESSAGES_LOADED`, `MESSAGES_LOAD_FAILED` | `ThreadPannel.tsx` |
| `ZERO_CONNECTION_ERROR`, `ZERO_CONNECTION_STABLE`, `ZERO_CONNECTION_MAX_RETRIES_REACHED` | Zero connection components |

Some of these sit next to a surviving `logger.info` / OTel `safeRecordMetric` — this one does, and so does the socket one — so the signal may well be covered elsewhere and this is deliberate cleanup. The call-join and messages-loaded ones I am less sure about.

Could you note in the PR description which of these are intentionally retired versus collateral? If any dashboards or alerts point at them, they will go quiet on deploy and it is much cheaper to know that now.

apps/dashboard/src/machines/roomMachine.ts
```diff
@@ -1865,41 +1865,10 @@ export const roomMachine = setup({
         }),
         onDone: {
           target: 'connected',
```
ameer030, line 1865: Both call metrics go away here: `CALL_JOIN` (time-to-connect, participant count, call type, initiator) on `onDone`, and `CALL_CONNECTION_FAILED` (time-to-failure, error message) on `onError`.

Unlike the API-latency and socket deletions, I could not find a `logger.info` or OTel counter left behind on either path, so as far as I can tell call join latency and call connection failure rate become unmeasurable after this. For a realtime calling feature that seems like the one worth keeping — either as a `posthogService.capture()` or as an OTel metric alongside the others.

See the wider list in my comment on `apiClient.ts`.

ameer030 (commented):
## Review: Mixpanel → PostHog migration

Reviewed the full diff (186 files, +1786/-1280). The migration itself is clean and the mechanical `mixpanelService.track` → `posthogService.capture` swaps look correct throughout. Most of my comments are about the *new* behaviour this PR turns on rather than the swap.

Three things I'd want resolved before this ships, all in `posthogService.ts`:

1. **Autocapture sends element text and attributes, and nothing masks them.** `maskAllInputs` / `maskTextSelector` only cover session replay. Concretely, every DM click in the sidebar ships the counterpart's name to PostHog — both as `$el_text` and inside the captured `data-track-metadata` attribute. Details and fix in the inline comment.
2. **Session replay goes from off (`ENABLE_SESSION_REPLAY = false`) to 100% of sessions**, with heatmaps and rageclick, and no consent lever or sampling. That is a data-collection policy change, not an implementation detail — worth an explicit decision from whoever owns privacy, plus a line in the PR description.
3. **`capture_pageview: true` breaks SPA pageviews.** It overrides the `'history_change'` that `defaults: '2025-05-24'` would have given you, so in-app navigation produces no pageview — the opposite of what the comment above it says it does.

And one I'd want a screenshot for: **216 native `<button>` elements were converted to `<Button>`**, but only 3 gained a `size=` prop and 22 an `h-auto`. The remaining ~190 inherit `h-9 px-4 py-2` from the default size variant. I've flagged three concrete breakages in `TicketDetails.tsx` and two in `AskAiRatingButtons.tsx`; there will be more. My honest suggestion is to split this into its own PR — the `trackId` attributes are the only part that actually needs `<Button>`.

### Also worth a look

- **Dropped metrics.** `PERFORMANCE_METRIC` has zero references left on this branch. API latency, WS reconnect, call join / connection-failed, thread messages loaded, and all three Zero connection events are gone. Several have a surviving `logger.info` next to them so this may be deliberate — the call ones don't. Could you note in the description which are intentional?
- `vite.config.ts:204` still defines `'process.env.VITE_MIXPANEL_TOKEN'` — dead now that the dep is removed.
- `main.tsx:36`, `:58`, `:84` — the three tracking-failure fallbacks still read `'Failed to track ... to Mixpanel:'`.
- `apps/dashboard-external/package.json:27` still depends on `mixpanel-browser`. Out of scope, or missed?
- `VITE_POSTHOG_KEY` / `VITE_POSTHOG_HOST` are in `.env.example` but not `.env.sdlc` or `.env.test`. The checklist asks for all of them, and it's also what currently keeps replay from running in those environments.
- `pnpm-lock.yaml` carries unrelated `libc:` churn on ~20 optional deps, from a different pnpm version. Harmless but noisy; worth regenerating on the pinned version.
- **The PR description is the untouched template** — no description, no ticket link, every checkbox unticked, *How did you test it?* empty. For a change that flips session recording on for every user, that's the thing I'd most like filled in.

I couldn't run `typecheck` locally (no `node_modules` in my worktree — `Cannot find type definition file for 'vite/client'`), so the build is unverified on my side; CI should cover it.

Nice work on the service itself — `captureActionOutcome` filling autocapture's blind spot on async outcomes, and the `trigger: 'keyboard' | 'button'` split on send, are both the right instincts. Comments on both inline, but they're refinements, not objections.

apps/dashboard/src/components/ui/InputBox/InputBox.tsx
```diff
@@ -1178,102 +1189,110 @@ export const InputBox = forwardRef<InputBoxHandle, InputBoxProps>(
...
+          }
         }
 
-        // Get fresh content after inserting link
-        const finalHtmlContent = editor?.getHTML() || htmlContent;
-        const finalPlainText = editor?.getText().trim() || plainText;
+        // Record how the message was sent (Enter key vs. send button). Autocapture
+        // sees the button click but is blind to keyboard sends, so most sends would
+        // otherwise be invisible; this makes the Enter-vs-button split measurable.
+        posthogService.capture('message_send', { trigger, composer: 'chat' });
```
ameer030: Two small things on this line.

**It fires before the send is attempted.** `onSendMessage` is awaited ~20 lines below and can reject, so a failed send still counts as a `message_send`. Moving the capture after the `await` (or adding a `status` property in the `try`/`catch`) makes the number mean what its name says.

**Hardcoded strings, three lines under the constants import.** `EVENTS` and `EVENT_PROPERTIES` are imported at the top of this file, but the event name and `composer: 'chat'` are inline literals. Either add `MESSAGE_SEND` to `EVENTS` or drop the constants import — right now the file does both and the next person will not know which convention to follow.

The `trigger: 'keyboard' | 'button'` split itself is a good call, and the reasoning in the comment is right: autocapture genuinely cannot see Enter-to-send.

shubhamdhamija-juspay (dismissed):

ameer030 (approved):

shubhamdhamija-juspay (approved):

abhishekverma-droid (approved):

PR #1343, opened 2026-09-01
Fix: XYNE-56555 fixing and adding canvas features

Neha-Gargg:
  - dashboard

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
Link : https://drive.google.com/file/d/1r2GEyV8qAv7G7fE7x2mD1a3m1nZc89Ai/view?usp=sharing

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist

### DIFF
--- a/apps/dashboard/src/components/Canvas/Canvas.types.ts
+++ b/apps/dashboard/src/components/Canvas/Canvas.types.ts
@@ -35,6 +35,8 @@ export interface CanvasEditorProps {
   initialCommentThreadId?: string | undefined;
   /** Emits the number of open comment threads already loaded by the editor highlight query. */
   onOpenCommentCountChange?: (count: number) => void;
+  /** Mirrors the comments panel open state so the host header can mark its button as active. */
+  onCommentsOpenChange?: (open: boolean) => void;
   /** Auto-focus the editor on mount */
   autoFocus?: boolean;
   /** Optional preloaded canvas participants to avoid duplicate query */
--- a/apps/dashboard/src/components/Canvas/CanvasCommentsPanel/CanvasCommentsPanel.tsx
+++ b/apps/dashboard/src/components/Canvas/CanvasCommentsPanel/CanvasCommentsPanel.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useMemo, useRef, useState } from 'react';
+import { useCallback, useLayoutEffect, useMemo, useRef, useState } from 'react';
 import {
   ArrowRight,
   Check,
@@ -31,7 +31,6 @@ import Avatar from '../../ui/Avatar/Avatar';
 import Button from '../../ui/Button';
 import { InputBox } from '../../ui/InputBox';
 import type { MentionResult } from '../../ui/InputBox';
-import { Tooltip } from '../../ui/Tooltip';
 import {
   DropdownMenu,
   DropdownMenuContent,
@@ -39,6 +38,7 @@ import {
   DropdownMenuTrigger,
 } from '../../ui/dropdown-menu';
 import { formatRelativeCommentTime } from '../canvasCommentTime';
+import { useCanvasCommentRail } from '../useCanvasCommentRail';
 
 type UserLite = {
   id: string;
@@ -91,9 +91,16 @@ interface CanvasCommentsPanelProps {
   activeBlockId: string | null;
   activeThreadId?: string | null | undefined;
   activeAnchor?: CanvasCommentAnchor | null | undefined;
+  /**
+   * Thread ids whose anchor mark is still in the document, or null while that is unknown.
+   * A thread whose commented text was deleted drops off the list until an undo restores it.
+   */
+  anchoredThreadIds?: Set<string> | null | undefined;
   editable: boolean;
+  /** Canvas editor container, used to align each thread with the text it annotates. */
+  anchorContainerRef?: React.RefObject<HTMLDivElement | null> | undefined;
   onClose: () => void;
-  onSelectBlock: (blockId: string) => void;
+  onSelectBlock: (blockId: string, threadId?: string) => void;
   onBeforeCreateThread?: ((threadId: string, anchor: CanvasCommentAnchor) => boolean) | undefined;
   onCreateThreadCreated?: (() => void) | undefined;
   onCreateThreadFailed?: ((anchor: CanvasCommentAnchor) => void) | undefined;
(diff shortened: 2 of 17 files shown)


utkarsh-611 (approved):

utkarsh-611 (dismissed):

utkarsh-611:
## Review notes

One blocker around comment visibility, then a few correctness items in the Ask AI composer and the rail.

### 1. Comments can become invisible to everyone

`useCanvasCommentAnchors.ts:172` — visibility is now derived solely from the `canvasCommentThread` mark being present in the editor document (`anchoredThreadIds`), and both `CanvasCommentsPanel` (line 671) and `useCanvasCommentHighlights` filter on it.

But that mark is written by `applyCommentAnchorStyle` and only persisted by the 500 ms debounced `handleChange` → `onSave` — and in `CanvasEditor.handleChange` that save is **skipped entirely** (`return; // Don't save`) once content exceeds `CONTENT_SIZE_MAX_THRESHOLD`.

So on a large canvas: the user adds a comment, the thread row syncs through Zero, the mark is never saved. After reload — and immediately for every other viewer — the scan reports no mark and the comment disappears from the panel, the highlights and the badge count, with no error, while the row still exists in the DB. Same for any thread created before the mark existed, or when the debounced save is lost to a reload or navigation.

### 2. Removing the `!editor.isFocused` guard lets stale state overwrite live typing

`XyneAIInputBox.tsx:1165`. `onUpdate` sets `lastReportedTextRef.current` synchronously on every change, but the effect reads `inputValue` from the last committed render.

Voice dictation is the concrete case — writes come from a WebSocket callback inserting interim text repeatedly, so two `onUpdate`s can land before the first render's passive effect flushes: effect sees `inputValue = "hello"`, `editor.getText() = "hello world"`, `lastReportedTextRef = "hello world"` → both guards pass → `setContent("hello")` wipes "world" and resets the caret. It also invalidates `voiceStreamInterimRangeRef`, so VoiceInput's later `setTextSelection({from,to}).deleteSelection()` deletes the wrong text.

The comment at line 1151 still claims this effect skips during voice recording, but nothing does that any more.

### 3. Canvas selection overwrites the user's typed prompt

`XyneAISidebar.tsx:535` — `setInputValue(selectedText)` runs unconditionally. Type "summarise this for the release notes", then highlight a paragraph to add it as context: the selection is added to `activeSelectionInfos` (so it's already shown as a pill) *and* the typed prompt is silently replaced with the raw selected text. Combined with #2 the write now lands even while the composer is focused, so the caret jumps too. The selection ends up present twice and the prompt is unrecoverable.

### 4. Unmeasured rail cards all stack at `top: 0`

`CanvasCommentsPanel.tsx:959`. The layout memo drops any thread whose anchor wasn't found (`.filter(thread => anchorTops[thread.id] !== undefined)`), but `rail.isAligned` is true as soon as *one* anchor is measured — so every thread gets `absolute inset-x-0` + `top: rail.cardTops[thread.id] ?? 0`. With 5 visible threads and 1 measured anchor, the other 4 stack at top 0, overlapping each other and the real card. Either keep unmeasured threads out of the absolutely-positioned track, or fall back to flow layout for them.

### 5. Scroll sync fires when the panel isn't actually aligned

`useCanvasCommentRail.ts:236` gates on `isAlignmentActive` rather than `isAligned`. When threads exist but no anchor is measured, the panel renders `space-y-2` flow layout — yet `handleDocumentScroll` still sets `railElement.scrollTop = documentElement.scrollTop`. Scrolling the canvas down 1500 px slams the comment list to its bottom and pins it there.

### 6. Scan effect misses `canvasId`, and runs per keystroke

Two issues in `useCanvasCommentAnchors.ts`:

- **Line 225** — deps are `[containerRef, enabled, refreshKey]`, none of which change on a canvas switch (`enabled` stays true). The old closure's `hasSettled` is already `true` and its MutationObserver is still attached, so the first mutation from the new (still empty) document publishes an empty set — every comment on the newly opened canvas is hidden until a later scan corrects it.
- **Line 165** — keying on `refreshKey` means every keystroke (both editors call `refreshCommentHighlights()` unconditionally in their change handlers) tears down and rebuilds a `childList+subtree+characterData+attributes` MutationObserver plus five timers, and walks the whole ProseMirror document up to four times. That's O(document) work several times per typed character, duplicated again in `useCanvasCommentHighlights`.

### Smaller

- `useCanvasCommentEditorBridge.ts:267` — the capture-phase `scroll` listener clears `activeCommentThreadId` on *any* scroll, including the comments panel's own, which the rail drives programmatically. So the rail's own sync dismisses the selection.
- `useCanvasCommentEditorBridge.ts:183` — dropping `setIsCommentsOpen(false)` without suppressing the inline card means clicking a highlight renders the same thread twice: floating card over the document, plus the expanded rail card.
- `CanvasVersionHistory.tsx:147` — `//onClose();` shipped as commented-out code. Below `md` the aside is `w-full`, so on mobile the history list stays on top of the preview the user just opened. Delete it or make it viewport-conditional.
- `CanvasScreen.tsx:1083` — `buildCanvasVersionDiffContent` runs in the render body with no `useMemo`, so a full `JSON.parse(JSON.stringify(...))` of both documents plus a word diff re-runs on every render while a preview is open (title typing, presence updates, comment count changes).
- `CanvasScreen.tsx:1681` — swapping the diff `Button` for a `Switch` dropped `data-track-category='CANVAS'`/`data-track-name='TOGGLE_VERSION_DIFF'`; it now reports as `SWITCH` / the aria-label sentence, breaking the existing series.
- `CanvasTab.tsx:188` — only `CanvasScreen` was migrated. `CanvasTab` still renders the old `CanvasVersionDiffPanel` with `showVersionDiff` defaulting false, so the two canvas hosts now offer different version-diff experiences for the same feature.
- `CanvasCommentsPanel.tsx:265` — `new ResizeObserver(measure)` per rendered comment with no `typeof ResizeObserver !== 'undefined'` guard, unlike the two other new observers in this PR.

PR #1352, opened 2026-09-01
fix: XYNE-61590 Added ticket creation flow in the canvas

rushiwabale:
  - backend
  - dashboard
POT : [POT](https://drive.google.com/file/d/1YR5bhRJB_gSCzdwj9YGu7LhgQ64i_4WZ/view)
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist

### DIFF
--- a/apps/backend/src/utils/ysweetUtils.ts
+++ b/apps/backend/src/utils/ysweetUtils.ts
@@ -49,6 +49,39 @@ const canvasCommentThreadStyleSpec = createStyleSpec(
   },
 );
 
+const canvasTicketStyleSpec = createStyleSpec(
+  {
+    type: 'canvasTicket',
+    propSchema: 'string',
+  },
+  {
+    render: () => {
+      const doc = (globalThis as unknown as {
+        document?: { createElement: (tagName: string) => unknown };
+      }).document;
+      const span = doc?.createElement('span') ?? {};
+      return {
+        dom: span,
+        contentDOM: span,
+      } as never;
+    },
+    parse: element =>
+      (element as unknown as { getAttribute?: (name: string) => string | null }).getAttribute?.(
+        'data-canvas-ticket-id',
+      ) ?? undefined,
+    runsBefore: [
+      'bold',
+      'italic',
+      'underline',
+      'strike',
+      'code',
+      'textColor',
+      'backgroundColor',
+      'canvasCommentThread',
+    ],
+  },
+);
+
 function createServerSchema() {
   return BlockNoteSchema.create({
     // The canvas schema's own blocks, for the same reason as the inline specs
@@ -72,6 +105,7 @@ function createServerSchema() {
     styleSpecs: {
       ...defaultStyleSpecs,
       canvasCommentThread: canvasCommentThreadStyleSpec,
+      canvasTicket: canvasTicketStyleSpec,
     },
   });
 }
--- a/apps/dashboard/src/components/Canvas/CanvasEditor/CanvasEditor.tsx
+++ b/apps/dashboard/src/components/Canvas/CanvasEditor/CanvasEditor.tsx
@@ -86,6 +86,8 @@ import { CanvasInlineCommentThread } from '../CanvasInlineCommentThread/CanvasIn
 import { createCanvasFormattingToolbar } from '../CanvasFormattingToolbar/CanvasFormattingToolbar';
 import { CanvasWidthHandles } from '../CanvasWidthHandles';
 import { useCanvasCommentEditorBridge } from '../useCanvasCommentEditorBridge';
+import { useCanvasTicketEditorBridge } from '../useCanvasTicketEditorBridge';
+import { CanvasTicketCreationFlow } from '../CanvasTicketCreationFlow/CanvasTicketCreationFlow';
 
 const canvasDictionary = {
   ...en,
@@ -418,6 +420,17 @@ export const CanvasEditor = forwardRef<CanvasEditorRef, CanvasEditorProps>(
       initialCommentThreadId,
       onOpenCommentCountChange,
     });
+    const {
+      activeTicketAnchor,
+      isTicketChannelArchived,
+      openTicketForCurrentSelection,
+      closeTicketModal,
+      handleTicketCreated,
+    } = useCanvasTicketEditorBridge({
+      channelId,
+      containerRef,
+      getEditor: getCanvasCommentEditor,
+    });
 
     // Expose presentation and comment drawer methods via ref
     useImperativeHandle(
@@ -549,8 +562,17 @@ export const CanvasEditor = forwardRef<CanvasEditorRef, CanvasEditorProps>(
           ...(canvasId && { canvasId }),
           ...(_canvasTitle && { canvasTitle: _canvasTitle }),
           canComment: editable,
+          canCreateTicket: editable && !isTicketChannelArchived,
+          onCreateTicket: openTicketForCurrentSelection,
         }),
-      [_canvasTitle, canvasId, editable, openCommentsForCurrentBlock],
+      [
+        _canvasTitle,
+        canvasId,
+        editable,
+        isTicketChannelArchived,
+        openCommentsForCurrentBlock,
+        openTicketForCurrentSelection,
+      ],
     );
 
     const handleSave = useCallback((): void => {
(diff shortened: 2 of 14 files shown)


ameer030:
## Review

Nice feature. A few things worth resolving before merge — one design question, one confirmed bug, and a batch of smaller items.

---

### 1. The ticket mark hides the block's real text

`CanvasTicketStyleSpec.tsx:88` renders the ProseMirror `contentDOM` into `.canvas-ticket-anchor__source`, which is `display: none` (`global.css:2729`). The mark is applied to the **whole block**, not the selection (`useCanvasTicketEditorBridge.ts:402-403` widens to `blockFrom..blockTo` before `addStyles`). So once a block is linked, its actual text is never shown again — the ticket card replaces it.

Consequences:

- A viewer without ACL access to the ticket sees `Ticket unavailable or no access` instead of canvas text they *are* allowed to read. Same after the ticket is deleted.
- Before the Zero query resolves, the block reads literally `Ticket`.
- `display: none` text is not copied, so selecting and copying that block yields only the title.
- Everything except the hidden source span is `contentEditable={false}`, so the block effectively becomes read-only.
- `aria-hidden='true'` on the source span means screen readers lose the text entirely.
- Any comment mark, mention or citation inside that block renders inside the hidden span, so it becomes invisible and unclickable.

Suggestion: render the ticket chip *alongside* the source text rather than in place of it (this is what `canvasCommentThread` does), or at minimum fall back to showing the source text whenever `accessState !== 'available'`. Happy to be told this is the intended product behaviour — but the no-access and ticket-deleted cases look like real content loss.

---

### 2. The new standalone canvas window navigates to the wrong routes

`usePath()` (`hooks/usePath.ts`) strips only the `workspaceId` prefix, not `/newWindow`. So inside the new `/newWindow/chat/canvas` route (`AppRoot.tsx:1975`):

- `CanvasScreen.tsx:167` — `isOnChatCanvasPage` is `false`, so back (`:603`), "Go to Canvases" (`:1833`) and the deleted-canvas redirect (`:453`) all navigate to `${baseRoute}/canvas`.
- `CanvasPanel.tsx:64` — `isOnIndexRoute` is likewise `false`, and `:216` hard-navigates to `/chat/canvas`, which drops the popout window back into the full app shell.

Fix: compare against `toRegularPath(usePath())` and route those navigations through `getNavigationPath()`, both already in `utils/electronApp.ts`.

---

### 3. Channel filter change reaches beyond the canvas flow

`CreateTicketModal.tsx:~402` — the `useAllVisibleChannels()` filter now also drops archived and project-less channels. That list is shared, so this silently narrows the existing sub-ticket and AI ticket flows too. If the modal's own `channelId` is archived or has no project, `selectedChannel` (`:480`) becomes `undefined`, `selectedChannelProjectId` falls back to the `projectId` prop, and the picker shows no current selection.

Suggest gating the extra predicate on `allowChannelSelection` so only the new entry point gets the narrower list.

---

### 4. Unlink watcher cost and history handling

`useCanvasTicketEditorBridge.ts:227-238` — `getTicketMarkedTextSnapshots` is a full `doc.descendants` walk on every editor update, and three walks when a mark is actually stripped, because `view.dispatch` inside the `update` handler re-enters `handleUpdate`. On a large canvas that is per-keystroke work proportional to document size.

Related, `:184` — `appendedTransaction` is a meta key ProseMirror sets itself for real `appendTransaction` results; setting it manually on a `view.dispatch` is not a supported way to merge into the previous history event.

Both are solved by the same move: implement this as a ProseMirror plugin `appendTransaction` and gate on `tr.docChanged` / the step ranges, instead of a `view.dispatch` from the `update` event.

---

### 5. Smaller items

- `useCanvasTicketEditorBridge.ts:151-170` — snapshots are keyed by ticket id and `existing.text += node.text` concatenates every occurrence across the document. Copy a linked block, edit one copy, and both get unlinked. Keying by contiguous range would avoid this.
- `useCanvasTicketEditorBridge.ts:229` — the tiptap `update` event also fires for remote Yjs transactions, so every connected client dispatches its own `removeMark` for the same edit. Worth skipping when the transaction carries a Yjs remote-origin meta.
- `CanvasTicketStyleSpec.tsx:~137` — `if (!ticket) return trigger;` and the `<HoverCard>` branch are different root component types, so when the ticket query resolves React unmounts the trigger span and mounts a fresh one, handing `contentRef` a new node mid-document. Rendering `<HoverCard>` unconditionally and gating only its content would keep the node stable.
- `useCanvasTicketEditorBridge.ts:339` — `rangeHasLink` scans `blockFrom..blockTo` (correct, since the mark covers the block) but the toast says `Tickets cannot be created from linked text`, which reads as if it were about the selection. `rangeHasCodeStyle` just below uses the selection range instead — worth making the two consistent, or at least matching the messages to the ranges.
- `CreateTicketModal.tsx:139,224,2961` — `focusDescriptionOnOpen` is declared and defaulted but no caller ever passes it, so the description box is never focused on open. Looks like `CanvasTicketCreationFlow` was meant to set it.
- `useCanvasTicketEditorBridge.ts:23-24,367-368` — `selectionFrom` / `selectionTo` are written but never read (carried over from `useCanvasCommentEditorBridge`).
- `CanvasFormattingToolbar.tsx:74-77` — `handleCreateTicket` computes `selectedText` and guards on it, but `openTicketForCurrentSelection` takes no arguments. The `if (!selectedText) return` also swallows the click with no toast, unlike every other guard in this flow.
- `CreateTicketModal.tsx:1474` — `canPopOut` is disabled only when `allowChannelSelection` is set, which for canvas means only when the canvas has no channel. In the common case pop-out stays available and `handlePopOut` calls `onClose()` immediately. It may still link correctly, since `subscribeCreateTicketResult` holds the callback and ProseMirror keeps its selection across a dialog close — worth one manual check to confirm.
- `global.css:2533` — the toolbar width bump from 200px to 270px is unconditional, so the Ask + Comment only toolbar is now wider than it needs to be.
- `global.css:2676` — hardcoded `width: 490px` on an inline element inside a variable-width editor.
- `CanvasTicketCreationFlow.tsx` and `useCanvasTicketEditorBridge.ts:213` both call `useChannel` for the same channel.
- `ysweetUtils.ts:45` — the server spec renders the raw text while the client renders the chip, so PDF/AI export and the editor will disagree about what that block says. Probably fine, just flagging it.
- No tests. `removeChangedTicketStyles` and `getTicketMarkedTextSnapshots` are pure functions over a document and would be cheap to cover.

---

### Checked and fine

- `AppRoot.tsx:1975` drops the `ToolbarProtectedRoute path='/chat/canvas'` guard that the regular route has, but `/newWindow/chat/dir` (`:1921`) has the same gap, so this is pre-existing rather than introduced here. Still a one-liner if you want to fix it while you are in there.
- `CanvasPanel.tsx:196` — the switch to `standaloneNavigate({ event })` does mean web Cmd+Click no longer opens a new browser tab, but that matches every other list in the app (`ChatListV2/V3/V4`, `ConversationPanelV2.tsx:220`, `ChannelItemV2.tsx:161`, `BotBubble.tsx:99`), so this is consistency rather than a regression.
- Backend `canvasTicketStyleSpec` mirrors the existing `canvasCommentThreadStyleSpec` correctly.
- `ticketRowById` is pre-existing and already ACL-wrapped; the frontend handles the no-access case explicitly.

---

### PR description

The template is entirely unticked and the Description section is empty. Could you fill in at least what changed and how it was tested? The Areas Touched and Manual testing sections matter here given this touches canvas, tickets, routing and Electron.

ameer030 (approved):

Mohan0544 (approved):

PR #1392, opened 2026-09-01
fix: XYNE-61697 correct Kanban column membership under Vespa-backed filters (61697)

XyneSpaces:
## Problem
On the Kanban board, when a Vespa-representable custom filter is active, changing a ticket's status from the ticket card dropdown updates the status correctly but the card renders in the WRONG column (stays in the old column, or flickers to another) until the search index catches up. Changing status from the ticket details page is unaffected. Repro reference: ticket 61697.

## Root cause
Under a Vespa-representable filter the board takes the `shouldUseDirectVespaRows` path in `useKanbanTicketsPage`. Rows come straight from the Vespa search payload via `useVespaTicketSearch` → `toTicket`, so `statusV2`/`stageName` reflect the **search index**, not live state. Vespa is an async index; immediately after a status mutation it still returns the old stage, so `ticketBelongsToColumn` places the card in its old column. The ticket-details path uses the live Zero store, which is why it never shows the bug.

## Fix (two parts)
1. **`useKanbanTicketsPage`** — overlay live Zero fields (`statusV2`, `stageName`, `boardId`, `priority`, `assignedTo`) onto the direct-Vespa rows by id via `queries.ticketsByIds`, so column membership is computed from live state instead of the stale index.
2. **`KanbanColumns.renderedTickets`** — add-side reconciliation: a ticket whose live stage now matches a column renders there even before that column's own async fetch surfaces it. Without this the card would drop from its old column but not yet appear in the new one during the reindex window.

## Validation
- `tsc --noEmit` on `apps/dashboard` passes clean.
- A full runtime before/after could not be reproduced in the dev sandbox because Vespa is not running there (the bug lives solely in the Vespa-backed path). Needs QA against a Vespa-backed environment.

## Test plan
- With a Vespa-representable custom filter active on a status-column Kanban board, change a ticket's status via the card dropdown → card should move to the correct column immediately.
- Regression: non-filtered board, mixed (zero-only) filters, drag-and-drop, and pagination should behave as before.

### DIFF
--- a/apps/dashboard/src/components/Tickets/KanbanColumns/KanbanColumns.tsx
+++ b/apps/dashboard/src/components/Tickets/KanbanColumns/KanbanColumns.tsx
@@ -269,6 +269,18 @@ const PaginatedStageList: React.FC<{
       renderedTicketsById.set(ticket.id, ticket);
     }
 
+    // Add-side reconciliation: a ticket whose LIVE status/stage now matches this
+    // column must render here even if this column's (async, possibly stale) fetch
+    // has not surfaced it yet. Right after a status change under a Vespa-backed
+    // filter, the moved ticket is dropped from its old column above but the search
+    // index has not yet returned it for the new column; without this loop the card
+    // disappears from every column until Vespa reindexes.
+    for (const knownTicket of allKnownTickets) {
+      if (renderedTicketsById.has(knownTicket.id)) continue;
+      if (!ticketBelongsToColumn(knownTicket, columnType, columnValue, columnStatus)) continue;
+      renderedTicketsById.set(knownTicket.id, knownTicket);
+    }
+
     return [...renderedTicketsById.values()];
   }, [allKnownTickets, columnStatus, columnType, columnValue, tickets]);
 
--- a/apps/dashboard/src/components/Tickets/KanbanColumns/KanbanColumns.tsx
+++ b/apps/dashboard/src/components/Tickets/KanbanColumns/KanbanColumns.tsx
@@ -275,6 +275,9 @@ const PaginatedStageList: React.FC<{
     // filter, the moved ticket is dropped from its old column above but the search
     // index has not yet returned it for the new column; without this loop the card
     // disappears from every column until Vespa reindexes.
+    // allKnownTickets is scoped to THIS swimlane's tickets (group.allTickets), so
+    // a ticket is only reconciled into its own group's column, never duplicated
+    // across swimlanes when a groupBy is active.
     for (const knownTicket of allKnownTickets) {
       if (renderedTicketsById.has(knownTicket.id)) continue;
       if (!ticketBelongsToColumn(knownTicket, columnType, columnValue, columnStatus)) continue;
--- a/apps/dashboard/src/routes/KanbanBoardScreen/useKanbanTicketsPage.ts
+++ b/apps/dashboard/src/routes/KanbanBoardScreen/useKanbanTicketsPage.ts
@@ -477,7 +477,41 @@ export const useKanbanTicketsPage = (
       !shouldUseDirectVespaRows &&
       (!requiresVespaTicketIds || vespaTicketSearch.searchResults !== null),
   });
-  const effectivePage = shouldUseDirectVespaRows ? directVespaPage : page;
+  // Overlay live Zero fields onto direct-Vespa payload rows. Vespa is an async
+  // search index, so a row's statusV2/stageName can lag a just-applied status
+  // change; rendering that stale stage places the card in its OLD kanban column
+  // (ticket 61697: status changed to B but card stays in A/C until reindex).
+  // Hydrating the live stage/status by id from the Zero store keeps column
+  // membership correct immediately, before Vespa catches up.
+  const [liveDirectRows] = useCachedQuery(
+    queries.ticketsByIds({
+      ticketIds: shouldUseDirectVespaRows ? vespaTicketIds : [],
+    }),
+    { enabled: shouldUseDirectVespaRows && vespaTicketIds.length > 0 },
+  );
+  const liveDirectRowsById = useMemo(() => {
+    const byId = new Map<string, Ticket>();
+    for (const row of liveDirectRows ?? []) byId.set(row.id, row as Ticket);
+    return byId;
+  }, [liveDirectRows]);
+  const overlaidDirectVespaPage = useMemo(() => {
+    if (!shouldUseDirectVespaRows || directVespaPage === null) return directVespaPage;
+    if (liveDirectRowsById.size === 0) return directVespaPage;
+    return directVespaPage.map(ticket => {
+      const live = liveDirectRowsById.get(ticket.id);
+      if (!live) return ticket;
+      return {
+        ...ticket,
+        statusV2: live.statusV2,
+        stageName: live.stageName,
+        boardId: live.boardId,
+        priority: live.priority,
+        assignedTo: live.assignedTo,
+      };
+    });
+  }, [shouldUseDirectVespaRows, directVespaPage, liveDirectRowsById]);
+
+  const effectivePage = shouldUseDirectVespaRows ? overlaidDirectVespaPage : page;
   const effectivePageDetailsType = shouldUseDirectVespaRows
     ? directVespaPage === null
       ? 'unknown'
--- a/apps/dashboard/src/routes/KanbanBoardScreen/useKanbanTicketsPage.ts
+++ b/apps/dashboard/src/routes/KanbanBoardScreen/useKanbanTicketsPage.ts
@@ -485,9 +485,9 @@ export const useKanbanTicketsPage = (
   // membership correct immediately, before Vespa catches up.
   const [liveDirectRows] = useCachedQuery(
     queries.ticketsByIds({
-      ticketIds: shouldUseDirectVespaRows ? vespaTicketIds : [],
+      ticketIds: shouldUseDirectVespaRows ? (vespaTicketIds ?? []) : [],
     }),
-    { enabled: shouldUseDirectVespaRows && vespaTicketIds.length > 0 },
+    { enabled: shouldUseDirectVespaRows && (vespaTicketIds?.length ?? 0) > 0 },
   );
   const liveDirectRowsById = useMemo(() => {
     const byId = new Map<string, Ticket>();
--- a/apps/dashboard/src/routes/KanbanBoardScreen/KanbanBoardScreen.tsx
+++ b/apps/dashboard/src/routes/KanbanBoardScreen/KanbanBoardScreen.tsx
@@ -4996,7 +4996,7 @@ const KanbanBoardScreen: React.FC<BoardKanbanScreenProps> = ({
                           availableTags={availableTags || []}
                           keyPrefix={`${group.key}::`}
                           onTicketsChange={handleKanbanTicketsChange}
-                          allKnownTickets={localTickets ?? []}
+                          allKnownTickets={group.allTickets}
                           {...(paginatedColumnConfig ? { paginatedColumnConfig } : {})}
                           {...(canCreateTicket &&
                           channel &&


samitbarai (dismissed):

apps/dashboard/src/components/Tickets/KanbanColumns/KanbanColumns.tsx
```diff
@@ -269,6 +269,18 @@ const PaginatedStageList: React.FC<{
...
 
+    // Add-side reconciliation: a ticket whose LIVE status/stage now matches this
+    // column must render here even if this column's (async, possibly stale) fetch
+    // has not surfaced it yet. Right after a status change under a Vespa-backed
+    // filter, the moved ticket is dropped from its old column above but the search
+    // index has not yet returned it for the new column; without this loop the card
+    // disappears from every column until Vespa reindexes.
+    for (const knownTicket of allKnownTickets) {
+      if (renderedTicketsById.has(knownTicket.id)) continue;
+      if (!ticketBelongsToColumn(knownTicket, columnType, columnValue, columnStatus)) continue;
```
harshpreet931, line 283: ticketBelongsToColumn has no group and allKnownTickets is board wide... so ig it would need groupKey as well?

apps/dashboard/src/components/Tickets/KanbanColumns/KanbanColumns.tsx
```diff
@@ -269,6 +269,18 @@ const PaginatedStageList: React.FC<{
...
+      renderedTicketsById.set(knownTicket.id, knownTicket);
```
harshpreet931, line 284: n + 1 cards and n hearder pls correct me if i am wrong

apps/dashboard/src/components/Tickets/KanbanColumns/KanbanColumns.tsx
```diff
@@ -269,6 +269,18 @@ const PaginatedStageList: React.FC<{
       renderedTicketsById.set(ticket.id, ticket);
     }
 
+    // Add-side reconciliation: a ticket whose LIVE status/stage now matches this
+    // column must render here even if this column's (async, possibly stale) fetch
+    // has not surfaced it yet. Right after a status change under a Vespa-backed
+    // filter, the moved ticket is dropped from its old column above but the search
+    // index has not yet returned it for the new column; without this loop the card
+    // disappears from every column until Vespa reindexes.
+    for (const knownTicket of allKnownTickets) {
```
harshpreet931, line 281: this is appended after the sorted page so it lands at the bottomand then jumps

apps/dashboard/src/routes/KanbanBoardScreen/useKanbanTicketsPage.ts
```diff
@@ -477,7 +477,41 @@ export const useKanbanTicketsPage = (
...
-  const effectivePage = shouldUseDirectVespaRows ? directVespaPage : page;
+  // Overlay live Zero fields onto direct-Vespa payload rows. Vespa is an async
+  // search index, so a row's statusV2/stageName can lag a just-applied status
+  // change; rendering that stale stage places the card in its OLD kanban column
+  // (ticket 61697: status changed to B but card stays in A/C until reindex).
+  // Hydrating the live stage/status by id from the Zero store keeps column
+  // membership correct immediately, before Vespa catches up.
+  const [liveDirectRows] = useCachedQuery(
+    queries.ticketsByIds({
+      ticketIds: shouldUseDirectVespaRows ? vespaTicketIds : [],
```
harshpreet931: vespaTicketIds is string[] | undefined so this doesn't compile build will fail

apps/dashboard/src/routes/KanbanBoardScreen/useKanbanTicketsPage.ts
```diff
@@ -477,7 +477,41 @@ export const useKanbanTicketsPage = (
...
+    }),
+    { enabled: shouldUseDirectVespaRows && vespaTicketIds.length > 0 },
```
harshpreet931: here also build will fail

XyneSpaces:
Thanks @harshpreet931 — addressed in `396a89cc`.

**Build failure — `vespaTicketIds` is `string[] | undefined` (comments on `useKanbanTicketsPage.ts`):** Fixed. Both the query arg and the `enabled` guard are now null-safe: `ticketIds: shouldUseDirectVespaRows ? (vespaTicketIds ?? []) : []` and `enabled: shouldUseDirectVespaRows && (vespaTicketIds?.length ?? 0) > 0`. `pnpm tsc --noEmit` on `apps/dashboard` passes.

**Group awareness — `ticketBelongsToColumn` has no group and `allKnownTickets` is board-wide (KanbanColumns.tsx):** Good catch — this was a real cross-swimlane bug. When `groupBy !== 'none'` the board renders one `<KanbanColumns>` per swimlane and every instance was handed board-wide `localTickets`, so the add-side loop (which only matches stage/status) would duplicate a ticket into the matching column of *every* swimlane. Rather than thread a `groupKey` into `ticketBelongsToColumn` (which would mean re-deriving each ticket's group inside the presentational component), I now pass the already group-scoped `group.allTickets` as `allKnownTickets`. Reconciliation is therefore confined to the ticket's own swimlane, and the ungrouped case is unchanged (single group == all tickets).

**"appended after the sorted page so it lands at the bottom and then jumps" / "n+1 cards and n header":** Correct, and this is an accepted, self-correcting transient. The reconciled card is the single ticket the user just moved; it renders in the correct column immediately (appended) and then settles into its server-ranked position once Vespa reindexes (a few seconds), at which point the per-column count also reconciles. I deliberately did not client-side re-sort the whole column, because column order is the drag-and-drop rank and globally re-sorting would risk corrupting DnD ordering. The header count is server (Vespa) driven and lags the same window. The tradeoff is intentional: a briefly-appended card that self-corrects is strictly better than the card vanishing from every column until reindex, which is the bug this PR fixes.

harshpreet931 (approved):

XyneSpaces (commented):
Xyne Spaces Reviewer — automated review of `396a89cc` (3 files, +51/−2).

**Verdict: changes requested (nothing blocking)** — 0 blocking, 2 major, 1 minor.

The fix's approach is sound: the group-scoped reconciliation superset (`group.allTickets` is always an array per `processedGroups` — `groupedRows[key] ?? []` / `Object.entries` — so dropping `localTickets ?? []` introduces no undefined-deref risk) plus the live-field overlay for direct-Vespa rows. The majors concern verification and type-safety of the new logic, not the mechanism. Details in the inline comments and the summary comment.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1392

**Verdict:** Changes requested (nothing blocking) — add verification for the new reconciliation/overlay logic
**Reviewed:** 3 files, 3 hunks (+51/−2) at `396a89cc` · **Findings:** 0 blocking, 2 major, 1 minor

### Findings
| major | R7 · guideline 204 🟠 | `KanbanColumns.tsx:281` | New add-side reconciliation loop is untested and load-bearing on a subtle callback sequence (stale Vespa fetch → live-field overlay → `onTicketsChange` reports fetched list → re-add); extract pure logic + cover the moved-ticket case |
| major | guideline 239 🟠 | `useKanbanTicketsPage.ts:494` | `row as Ticket` unchecked assertion on Zero rows — prefer a type guard / typed query return |
| minor | perf/design | `useKanbanTicketsPage.ts:488` | Overlay subscribes `ticketsByIds` to the full fetched Vespa id set (≤400 per column × columns); consider scoping to mutated ids |

All three are posted as inline review comments on `396a89cc`. Threads already resolved in this PR (n+1 cards, transient count mismatch) are not re-flagged.

✅ Verified safe: the `localTickets ?? []` → `group.allTickets` change — `processedGroups` always assigns an array (`groupedRows[key] ?? []` / `Object.entries`), so there is no undefined-deref risk, and group-scoping correctly narrows the reconciliation superset in a way `ticketBelongsToColumn` (status/stage-only matching) cannot.

### Recommended reviewer
**@harshpreet931** (Harshpreet Singh)

| Candidate | Owns (files in this PR) | CODEOWNERS match | Pending reviews (repo-wide) | Tie-breaker |
|-----------|--------------------------|------------------|------------------------------|-------------|
| @harshpreet931 | Kanban subsystem: KanbanColumns.tsx (#1295), KanbanBoardScreen.tsx / useKanbanTicketsPage.ts (#1296, #1297) | no — CODEOWNERS covers only ACL paths | 290 open PRs (auto-request inflated) | already engaged: his 5 comments were resolved here, he approved `396a89cc`, authored #1297 documenting this exact pagination path, and committed to main today |
| @samitbarai | — (prior reviewer on this PR) | no | 314 open PRs | runner-up |

Rationale: no CODEOWNERS rule covers these paths, so the pick rests on file ownership + engagement — Harshpreet is the dominant recent author of all three changed files' subsystem and is already deep in this PR; his pending-review count is also the lower of the two candidates. The bot-authored commits (`XyneSpaces[bot]`) are excluded.

⚠️ **Could not request the review via the GitHub API** — this session has no request-reviewers tool. Please add **@harshpreet931** as a reviewer manually; he is @mentioned here so he sees the findings either way.

apps/dashboard/src/components/Tickets/KanbanColumns/KanbanColumns.tsx
```diff
@@ -269,6 +269,21 @@ const PaginatedStageList: React.FC<{
...
+    // Add-side reconciliation: a ticket whose LIVE status/stage now matches this
+    // column must render here even if this column's (async, possibly stale) fetch
+    // has not surfaced it yet. Right after a status change under a Vespa-backed
+    // filter, the moved ticket is dropped from its old column above but the search
+    // index has not yet returned it for the new column; without this loop the card
+    // disappears from every column until Vespa reindexes.
+    // allKnownTickets is scoped to THIS swimlane's tickets (group.allTickets), so
+    // a ticket is only reconciled into its own group's column, never duplicated
+    // across swimlanes when a groupBy is active.
+    for (const knownTicket of allKnownTickets) {
```
XyneSpaces, line 281: **[major · R7 / guideline 204 🟠]** The new add-side reconciliation loop ships with no verification — and it is the subtlest part of this fix.

As the code reads, a moved ticket only survives because of a delicate sequence: the old column's Vespa fetch still *contains* it (stale membership), the `useKanbanTicketsPage` overlay patches in its live `statusV2`/`stageName`, the old column's render drops it (post-overlay membership check) while `onTicketsChange` reports the *fetched* list, and this loop then re-adds it to the new column from `allKnownTickets` (= `group.allTickets`). If any link in that chain changes — e.g. `onTicketsChange` ever reporting the rendered list instead — the ticket silently disappears from the board. That is exactly the kind of regression automated tests exist to catch.

`apps/dashboard` currently has no test runner (`package.json` has no test script, and there is no vitest/jest/@testing-library anywhere in the app), and the PR notes the Vespa path can't be exercised in the sandbox. Given that, at minimum: extract this loop + `ticketBelongsToColumn` and the `overlaidDirectVespaPage` merge from `useKanbanTicketsPage` into pure functions in `KanbanBoardScreen.utils.ts` (which already hosts the pure kanban helpers), so the behavior is pinned in one place and coverable as soon as a runner lands. The scenario to cover: ticket moved A→B while column A's Vespa fetch is stale.

apps/dashboard/src/routes/KanbanBoardScreen/useKanbanTicketsPage.ts
```diff
@@ -477,7 +477,41 @@ export const useKanbanTicketsPage = (
...
-  const effectivePage = shouldUseDirectVespaRows ? directVespaPage : page;
+  // Overlay live Zero fields onto direct-Vespa payload rows. Vespa is an async
+  // search index, so a row's statusV2/stageName can lag a just-applied status
+  // change; rendering that stale stage places the card in its OLD kanban column
+  // (ticket 61697: status changed to B but card stays in A/C until reindex).
+  // Hydrating the live stage/status by id from the Zero store keeps column
+  // membership correct immediately, before Vespa catches up.
+  const [liveDirectRows] = useCachedQuery(
+    queries.ticketsByIds({
+      ticketIds: shouldUseDirectVespaRows ? (vespaTicketIds ?? []) : [],
```
XyneSpaces, line 488: **[minor · perf/design]** The live-field overlay subscribes `queries.ticketsByIds` to *every* id in the direct-Vespa result set — up to the search's `maxFetchedResults` (400) per column, and the hook is invoked per column from `KanbanColumns.tsx`. That re-couples the direct-Vespa path to Zero sync at up to ~columns × 400 live subscriptions on large boards, when the overlay only needs live fields for tickets that were actually mutated.

Suggestion: scope the subscription to recently-mutated ids (e.g. ids with a pending or just-acked local mutation) instead of the full fetched set — that keeps the moved-ticket fix intact without re-subscribing the whole page to Zero.

Good parts worth noting: the `enabled` guard correctly skips the query entirely on the non-Vespa path. *(Assumption: worst-case magnitude — the per-column cap comes from the search config; real subscription counts depend on board size.)*

apps/dashboard/src/routes/KanbanBoardScreen/useKanbanTicketsPage.ts
```diff
@@ -477,7 +477,41 @@ export const useKanbanTicketsPage = (
...
+    }),
+    { enabled: shouldUseDirectVespaRows && (vespaTicketIds?.length ?? 0) > 0 },
+  );
+  const liveDirectRowsById = useMemo(() => {
+    const byId = new Map<string, Ticket>();
+    for (const row of liveDirectRows ?? []) byId.set(row.id, row as Ticket);
```
XyneSpaces, line 494: **[major · guideline 239 🟠]** `row as Ticket` asserts the Zero row into the domain type unchecked. Guideline 239: prefer proper TypeScript types and type guards over `as Type` assertions — with a silent cast, a drift between the Zero row shape and `Ticket` compiles clean and only surfaces at render time, in the exact fields this overlay writes (`statusV2`, `stageName`, `boardId`, `priority`, `assignedTo`).

This mirrors the existing `applyLocalVespaFilters` idiom in this file, so migrating every call site is out of scope — but since this is new code, consider a small type guard (e.g. `isTicketRow(row)` validating the fields the overlay reads) or a properly-typed return from `queries.ticketsByIds`, and centralize the Zero-row → `Ticket` conversion so new code stops accumulating casts.

samitbarai (approved):

PR #1099, opened 2026-08-26
fix: XYNE-58277 All the click metrics to sudoquery

vaibhavg-01:
Services-Requiring-Redeployment:
  - backend
  - dashboard

Services-Requiring-Redeployment:
  - backend
  - dashboard

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/backend` (API / worker)
- [ ] `apps/dashboard`
- [ ] `apps/electron`
- [ ] `apps/xyne-claw` / `apps/xyne-claw-auth`
- [ ] `packages/shared` or another shared package
- [ ] Infra (docker, scripts, nix, CI)

---

## Zero (Sync) Changes

- [ ] No Zero changes — **skip this section**
- [ ] Adds/modifies a **mutator**
- [ ] Adds/modifies the **schema** (tables, columns, relationships)
- [ ] Adds/modifies a **query or ACL**

If any of the above:

- [ ] Server (`apps/backend/src/zero/mutators.ts`) and client (`apps/dashboard/src/zero/mutators.ts`) mutators mirror each other — same namespace, name, and args schema
- [ ] No `Date.now()` / `uuid()` generated inside a mutator
- [ ] Optimistic client result matches the server result (no state flash on rebase)
- [ ] New tables exported in `createSchema()` and given a Query ACL registered in `query-acl-factory.ts`
- [ ] Matching Prisma model + migration, client regenerated

## Schema & Data Changes

- [ ] No schema changes — **skip this section**
- [ ] Prisma schema modified (`apps/backend/prisma` and/or `prisma-common`)
- [ ] Migration added
- [ ] Backfill / data migration included

If any of the above:

- [ ] Clients regenerated (`db:generate`, `db:common:generate`)
- [ ] No new Postgres enum or enum value — enums are frozen (`pnpm run test:enum`)
- [ ] Backward compatible with deployed code, or rollout order noted below
- [ ] Backfill is idempotent

## Config, Environment & URLs

- [ ] No config changes — **skip this section**
- [ ] Env variable added / renamed / removed
- [ ] **Base URL, host, port, or origin changed** (API, Zero, Vespa, Claw, LiveKit, LiteLLM, …)
- [ ] CORS / allowed origins / OAuth redirect URIs changed
- [ ] Feature flag or runtime config changed

If any of the above:

[...]

PR #960, opened 2026-08-24
feat: XYNE-56847 unify share logic

Mayankbansal-juspay:
POT - https://drive.google.com/file/d/1nGshPF4uHTqCCxco36NFkkDp8wMs4g6S/view?usp=sharing

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/backend` (API / worker)
- [ ] `apps/dashboard`
- [ ] `apps/electron`
- [ ] `apps/xyne-claw` / `apps/xyne-claw-auth`
- [ ] `packages/shared` or another shared package
- [ ] Infra (docker, scripts, nix, CI)

---

## Zero (Sync) Changes

- [ ] No Zero changes — **skip this section**
- [ ] Adds/modifies a **mutator**
- [ ] Adds/modifies the **schema** (tables, columns, relationships)
- [ ] Adds/modifies a **query or ACL**

If any of the above:

- [ ] Server (`apps/backend/src/zero/mutators.ts`) and client (`apps/dashboard/src/zero/mutators.ts`) mutators mirror each other — same namespace, name, and args schema
- [ ] No `Date.now()` / `uuid()` generated inside a mutator
- [ ] Optimistic client result matches the server result (no state flash on rebase)
- [ ] New tables exported in `createSchema()` and given a Query ACL registered in `query-acl-factory.ts`
- [ ] Matching Prisma model + migration, client regenerated

## Schema & Data Changes

- [ ] No schema changes — **skip this section**
- [ ] Prisma schema modified (`apps/backend/prisma` and/or `prisma-common`)
- [ ] Migration added
- [ ] Backfill / data migration included

If any of the above:

- [ ] Clients regenerated (`db:generate`, `db:common:generate`)
- [ ] No new Postgres enum or enum value — enums are frozen (`pnpm run test:enum`)
- [ ] Backward compatible with deployed code, or rollout order noted below
- [ ] Backfill is idempotent

## Config, Environment & URLs

- [ ] No config changes — **skip this section**
- [ ] Env variable added / renamed / removed
- [ ] **Base URL, host, port, or origin changed** (API, Zero, Vespa, Claw, LiveKit, LiteLLM, …)
- [ ] CORS / allowed origins / OAuth redirect URIs changed
- [ ] Feature flag or runtime config changed

If any of the above:

[...]

PR #1393, opened 2026-09-02
fix: XYNE-61909 message deep-links land on wrong message for older messages

XyneSpaces:
## Problem
Copied message links sometimes open the wrong message — especially **older** messages — landing at the bottom of the channel instead of on the linked message. Intermittent and cache-dependent, so hard to reproduce on demand.

## Root cause
`onCopyLink` in `ChatBubble.tsx` builds the deep link but never includes the message's `createdAt` temporal anchor. Without it, the receiver (`ConversationPanelV2`) must resolve the target through a `getConversationByIdWithChannel` Zero-cache ID lookup. For older messages not in the local cache, that lookup is slow or misses, and `ChatListV4` falls back to the bottom of the channel. When `createdAt` is present, the receiver already skips that lookup and loads the window directly via `getChannelConversationsSnapshot`.

## Fix
- Embed the `createdAt` anchor in the copied link (both thread and channel links).
- Extract link building/parsing into a pure, dependency-free `messageLink.ts` (`buildMessageLink`, `extractOriginFromHash`, `extractMessageIdFromHash`, `extractCreatedAtFromHash`); `ChatListUtils` re-exports them so existing import paths are unchanged.
- Backward compatible: when `createdAt` is absent/invalid the link is byte-identical to today's, and legacy links already in the wild keep working (they just take the old path).

## Files
- `apps/dashboard/src/components/Chat/ChatBubble/ChatBubble.tsx` — copy-link now emits `createdAt`, via the shared helper.
- `apps/dashboard/src/components/Chat/ChatList/messageLink.ts` — new pure helpers.
- `apps/dashboard/src/components/Chat/ChatList/ChatListUtils.ts` — re-exports the helpers.
- `apps/dashboard/src/components/Chat/ChatList/messageLink.test.ts` — unit tests.

## Proof of testing (POT)
- Unit tests (6/6 passing) assert: channel + thread links carry `createdAt`; `messageId` parsing isn't polluted by the trailing `createdAt`; and — mirroring the receiver's enable-condition — an **old** link (no `createdAt`) keeps the slow ID lookup enabled while the **new** link skips it and uses the snapshot window.
- A recorded walkthrough (attached in the Spaces thread) visualizes the same target old message landing at the channel bottom with the old link vs. landing exactly on the message with the new link. GitHub's API can't upload the video asset to the PR, so it's delivered in the Spaces thread.

## Notes / follow-ups
- Fully fixes newly-copied links. Links copied before this change still lack the anchor; hardening the fallback path (so even anchorless old-message lookups don't drop to the bottom) is a sensible follow-up.


apps/dashboard/src/components/Chat/ChatList/messageLink.ts
```diff
@@ -0,0 +1,73 @@
...
+  return match?.[1] ?? null;
+};
+
+/**
+ * Extract the createdAt temporal anchor (epoch ms) from a deep-link hash.
+ * Example: #origin=abc&createdAt=1712345678901 -> 1712345678901
+ * Mirrors the receiver parsing in ConversationPanelV2 so the copy-link producer
+ * and the link consumer stay in sync.
+ */
+export const extractCreatedAtFromHash = (hash: string): number | null => {
```
satvikbatra: are we using this ? if not can we remove this.

apps/dashboard/src/components/Chat/ChatList/messageLink.test.ts
```diff
@@ -0,0 +1,106 @@
+import { describe, it, expect } from 'vitest';
```
satvikbatra: remove test file

apps/dashboard/src/components/Chat/ChatList/ChatListUtils.ts
```diff
@@ -20,27 +20,16 @@ export type CombinedMessageItem =
...
-export const extractMessageIdFromHash = (hash: string): string | null => {
-  if (!hash) return null;
-
-  const match = hash.match(/messageId=([^&]+)/);
-  return match?.[1] ?? null;
-};
+// Message deep-link helpers live in a dependency-free module so the copy-link
+// contract can be unit-tested in isolation. Re-exported here to preserve the
+// existing import paths used across the Chat components.
+export {
```
satvikbatra: why new file. don't need this

apps/dashboard/src/components/Chat/ChatBubble/ChatBubble.tsx
```diff
@@ -682,13 +683,17 @@ export const ChatBubble: React.FC<ChatBubbleProps> = ({
...
     let messageLink = '';
     if (conversationId) {
-      if (context === 'thread') {
-        // Thread message: include full path with conversation + messageId in hash
-        messageLink = `${shareableOrigin}/chat/dir/${channelId}/${conversationId}#origin=${conversationId}&messageId=${message.messageId}`;
-      } else {
-        // Channel message: only channel in path, conversation in hash
-        messageLink = `${shareableOrigin}/chat/dir/${channelId}#origin=${conversationId}`;
-      }
+      // Include the createdAt temporal anchor so the receiver loads the message
```
satvikbatra: can we change the logic here only instead in making a function in other file

XyneSpaces:
Addressed the review feedback and pushed `8b6c437`:

- Reverted the refactor: deleted `messageLink.ts` + `messageLink.test.ts`, dropped the unused `extractCreatedAtFromHash`, and restored `ChatListUtils.ts` to its original state.
- Kept only the minimal inline fix in `ChatBubble.onCopyLink` — copied links now append `&createdAt=<epoch ms>`. The receiver (`ConversationPanelV2.tsx`, which already parses `createdAt` from the hash) uses it as a temporal anchor so deep links scroll to the exact message instead of falling back to the bottom of the channel for older, uncached messages.

Net diff vs `main` is now a single file: `apps/dashboard/src/components/Chat/ChatBubble/ChatBubble.tsx`.

Proof of testing: captured a video walkthrough against the real Spaces dashboard — open a channel, scroll up to an old message, use Hover ▸ More actions ▸ Copy link (the copied URL now contains `&createdAt=…`), then open that link on a fresh load and confirm it lands directly on the linked message. POT video is attached in the Spaces thread.

apps/dashboard/src/components/Chat/ChatBubble/ChatBubble.tsx
```diff
@@ -680,14 +680,23 @@ export const ChatBubble: React.FC<ChatBubbleProps> = ({
   const onCopyLink = (): void => {
     // Get conversation ID from conversation object or fallback to message
     const conversationId = conversation?.conversationId || message.conversationId;
+    // Include the message's createdAt (epoch ms) as a temporal anchor. Without it
```
satvikbatra, line 662: The anchor this adds (conversation.createdAt) is the same value ConversationPanelV2.tsx:169 already resolves via initialMessageById?.createdAt, and line 252 gates ChatListV4 behind a spinner until that resolves — so an anchorless link never scrolls early. Old and new links reach identical state; this only removes the round-trip.
The mis-landing comes from ChatListV4.tsx:800: initialLinkedIdRef.current is consumed unconditionally at line 804, combinedMessages isn't in the deps, and isConversationFullyVisible is stable on [virtualizer]. When the target conversation isn't in the local getChannelConversationsSnapshot window at mount, initial scroll falls to scrollToEnd (line 784), the correction effect fires once against empty data, and never re-runs when the fetch lands. Same one-shot bug in the thread effect at line 1053, where prevActiveThreadRef.current is written at 1062 before the idx === -1 guard at 1067.

XyneSpaces:
### Update: ChatListV4 scroll fix + POT

Pushed `b9972db` — this implements the real fix in the receiver, per the review discussion.

**What changed on the branch now:**
- `ChatListV4.tsx` (primary correctness fix): the deep-link scroll effects were one-shot/stale — they consumed the scroll ref before a match and never re-ran when the target conversation loaded later, so older messages fell to the bottom of the channel. Fix: consume the refs **only on a successful target match**, add `combinedMessages` to the deep-link effect deps, bound anchor refetches once per `navigationKey`, and reset the thread ref on close.
- `ChatBubble.tsx` (latency optimization, kept): the inline `createdAt` anchor lets the panel center the query on the target immediately instead of paying an extra `initialMessageById` round-trip. It is not required for correctness (agreed — `ConversationPanelV2` already resolves `createdAt` via server lookup on `main`), so it stays as a secondary optimization, not the fix.

**POT (attached in the thread):** recorded against the real Spaces dashboard (not a harness). Flow: email login → open a deep link to the **oldest** conversation in a 120-message channel (target is outside the initial newest window) → the app lands directly on the target row. Verified programmatically in the run: target row is within the scroll viewport and its text is `TARGET — oldest message, the deep-link destination` (`landedOnTarget: true`).

`typecheck` passes (`pnpm run typecheck`, rc=0).

apps/dashboard/src/components/Chat/ChatBubble/ChatBubble.tsx
```diff
@@ -680,14 +680,23 @@ export const ChatBubble: React.FC<ChatBubbleProps> = ({
...
+    // the receiver resolves the target via a Zero-cache ID lookup that is slow or
+    // misses for older, uncached messages, so the link lands at the bottom of the
+    // channel instead of on the linked message.
+    const linkCreatedAt = conversation?.createdAt ?? message.createdAt;
+    const createdAtParam =
+      typeof linkCreatedAt === 'number' && Number.isFinite(linkCreatedAt)
```
XyneSpaces, line 668: **[minor] R7 tests** — The new `createdAt` emission (this guard + both link branches below) has no automated test coverage: the previous unit tests were removed per review request, and the POT video covers the happy path on the real dashboard. A shape regression here (e.g. `createdAt` becoming a `Date`/string instead of epoch ms) would silently degrade to old anchorless links instead of failing anywhere. Consider re-adding a minimal unit test when the noted receiver-side hardening follow-up lands.

XyneSpaces (commented):
Review pass on head `8b6c437` — the minimal revert is the right shape and the anchor logic is sound. One inline note below; full findings (incl. the stale PR description) in the top-level summary comment. Posting as COMMENT only (no approve/request-changes gate).

XyneSpaces:
## Xyne Spaces Reviewer — PR #1393

**Verdict:** Changes requested — docs-only; the code itself is approve-worthy
**Reviewed:** 1 file, 1 hunk — `apps/dashboard/src/components/Chat/ChatBubble/ChatBubble.tsx` (+11/−2) at head `8b6c437` · **Findings:** 0 blocking, 1 major, 1 nit

### Findings
| major | Skill #233 (keep PR body in sync) | PR description | Body still describes the reverted refactor: lists `messageLink.ts`, `ChatListUtils.ts` re-export, `messageLink.test.ts`, 4 changed files, and a POT citing "Unit tests (6/6 passing)" for tests deleted in `8b6c437`. Net diff is 1 file. **Fix:** update the Fix/Files/POT sections to match the minimal change (inline anchor in `onCopyLink` + video POT). |
| minor | R7 tests (skill #204) | ChatBubble.tsx:686-691 | New guard (`typeof`/`Number.isFinite`) and both link branches have no test — removed at reviewer request; verified via POT video instead. Re-add a unit test with the receiver-side hardening follow-up. |

**Checked & clean:** backward-compat guard keeps legacy (anchorless) links byte-identical to `main`; appended `&createdAt=` can't pollute the receiver's `origin=`/`messageId=` hash parsing (both stop at `&`); no security/ACL/Zero-surface change; no migration in this PR. The three earlier inline review comments (@satvikbatra) are all resolved by the revert.

**Heads-up:** `main` moved after this PR's last push — #1213 also touches `ChatBubble.tsx`, so a rebase sanity-check is worth it before merge.

### Recommended reviewer
**@Dnyaneshwari-04** — owns `ChatBubble.tsx` (#1398, merged today) and has 0 pending review requests.

| Candidate | Owns ChatBubble.tsx? | CODEOWNER? | Pending reviews | Tie-breaker |
|-----------|----------------------|------------|-----------------|-------------|
| **Dnyaneshwari-04** | yes — #1398 (message edit-lock), today | No CODEOWNERS coverage for this path | **0** | freshest chat-message context on this file |
| nasimshk | yes — #1153, Sep 1 | — | 0 | — |
| vaibhavg-01 | yes — #1099, Aug 27 | — | 0 | — |
| sithaarth | yes — #1213, today | — | 266 | most recent commit but telemetry-only touch |
| Mayankbansal-juspay | yes — #960 "unify share logic" (strongest feature match) | — | 274 | deepest share-link context, but heaviest queue |
| nikunjJuspay | yes — Sep 1 | — | 274 | — |
| Pradeesh333 | yes — Aug 18 | — | 271 | — |
| satvikbatra | reviewer on this PR (excluded from pick) | — | 274 | already authored this PR's review comments |

`.github/CODEOWNERS` covers only ACL/Zero-mutator paths — nothing matches this PR, so this pick is ownership + workload only. @Dnyaneshwari-04 merged into this exact file today with an empty review queue — best familiarity/capacity combination; @Mayankbansal-juspay is the strongest feature-side fallback (share/unify-link logic, #960) if a second look is wanted.

cc @satvikbatra — your three comments are addressed; only the stale PR body remains.

apps/dashboard/src/components/Chat/ChatList/ChatListV4.tsx
```diff
@@ -797,17 +798,23 @@ const ChatListV4: React.FC<ChatListProps> = ({
...
   }, [combinedMessages.length, isInitialLoadComplete]);
 
-  // ── Activity re-scroll (subsequent navigations) ───────────────────────────────────
+  // ── Activity re-scroll (subsequent navigations + deep links not yet loaded) ──────────
+  // Re-runs on combinedMessages so a deep link whose target is not in the initial
+  // window still scrolls once the fetch below lands it. initialLinkedIdRef is
+  // consumed ONLY after the target is actually found — consuming it up front (the
+  // old bug) latched the navigation as handled while the target was still absent,
+  // so the corrective scroll never ran when the data arrived.
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
XyneSpaces, line 807: **[nit] R8** — New `eslint-disable-next-line react-hooks/exhaustive-deps` on this effect. It matches the identical pre-existing disable on the initial-scroll `useLayoutEffect`, and the comment block above does explain the re-run semantics — but the disable also masks that `location.key` (read a few lines below to build `navigationKey`) is not in the deps array. A re-navigation to the same `linkedConversationId` without a fresh `activityNavigationNonce` won't re-fire the correction until `combinedMessages` next changes. If that's intentional, a one-line note listing the deliberately-omitted deps next to the disable would help; otherwise `location.key` is already part of `navigationKey` and could simply be listed.

apps/dashboard/src/components/Chat/ChatList/ChatListV4.tsx
```diff
@@ -820,23 +827,31 @@ const ChatListV4: React.FC<ChatListProps> = ({
...
-        fetchNewerMessages();
-        fetchOlderMessages();
-      }, 0);
+      return;
     }
+
+    // Target not in the current window: fetch it ONCE per navigation (do not
+    // consume initialLinkedIdRef) so this effect re-runs on the resulting
+    // combinedMessages change and scrolls to the target above.
+    if (requestedFetchKeyRef.current === navigationKey) return;
```
XyneSpaces, line 836: **[minor] R6** — Silent give-up: if the one-shot fetch for an out-of-window deep-link target fails or doesn't include the target (transient Zero error, deleted message), this guard bails with no log anywhere, so "deep link didn't scroll" reports won't be triageable. Consider a warn-level log when this guard trips with the target still missing (or surfacing the miss where the fetch result lands). Assumption: `fetchNewerMessages` / `fetchOlderMessages`'s silent `.catch(() => {})` are pre-existing (unchanged in this PR) — the new part is this fetch-once / give-up orchestration.

XyneSpaces (commented):
Follow-up review of the receiver-side half added in `b9972db` (`ChatListV4.tsx`, +40/−22) — all hunks checked.

Verified correct: the reworked activity re-scroll effect now consumes `initialLinkedIdRef` **only after** the deep-link target is found in `combinedMessages` (the old code consumed it up front, latching the navigation as handled while the target was still absent — that was the "lands at bottom" bug), caps the window fetch at once per navigation via `requestedFetchKeyRef`, and lists `combinedMessages` in deps so the corrective scroll fires when the fetch lands. The thread-open scroll effect gets the same consume-after-found fix plus a ref reset when no thread is open. Two non-blocking notes below.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1393 (addendum: receiver-side changes @ `b9972db`)

Follow-up to my earlier summary: head moved to `b9972db` mid-review, adding the **receiver-side half** — `apps/dashboard/src/components/Chat/ChatList/ChatListV4.tsx` (+40/−22). That file is now reviewed too (all hunks — see the follow-up review just posted).

**Verdict:** unchanged — **Changes requested (docs only); the code itself is approve-worthy.**

**What the new hunks do (verified correct):**
- **Deep-link re-scroll effect** — `initialLinkedIdRef` is consumed *only after* the target lands in `combinedMessages` (the old code consumed it up front — that early-consume was the "lands at bottom" bug), and the window fetch is capped at once per navigation via the new `requestedFetchKeyRef`. Deps now include `combinedMessages` so the corrective scroll fires when the fetch lands.
- **Thread-open scroll** — same consume-after-found fix, plus the ref resets when no thread is open, so reopening the same thread scrolls again.

### Updated findings (supersedes the table in my earlier summary)

| major (docs) | — | PR description | **Stale, and now understates scope**: still describes the reverted `messageLink.ts` refactor, `ChatListUtils` re-exports and "Unit tests 6/6 passing", and doesn't mention the `ChatListV4.tsx` receiver-side changes now in the PR. Please rewrite to match the actual diff. |
| minor | R7 | ChatBubble.tsx:689 | No test for the new `createdAt` guard (tests were removed with the refactor revert; the POT video is the verification). |
| minor | R6 | ChatListV4.tsx:836 | One-shot deep-link fetch gives up silently if it fails or misses the target — no log to triage "deep link didn't scroll". |
| nit | R8 | ChatListV4.tsx:807 | New `eslint-disable-next-line react-hooks/exhaustive-deps` — consistent with the file's existing pattern, but it also masks `location.key` being absent from the deps. |

### Recommended reviewer — confirmed

**@Dnyaneshwari-04** — with the scope growth her case is stronger, not weaker: she is the most recent author of **both** changed files (`ChatBubble.tsx` and `ChatListV4.tsx`, via #1398, merged today) and has **0 pending review requests** in this repo. Runner-up: Mayank Bansal (~274 pending review requests). Note: I could not file the formal GitHub review request — no reviewer-request endpoint is exposed in my toolset — so this mention stands in for it; please add her as a reviewer in the UI if you want the formal request.

No blocking issues. R1 (security) and R3 (migrations) are clean: the `createdAt` param is a pure number guarded by `Number.isFinite` (cannot pollute the `origin=`/`messageId=` hash parsing), and no migration files are touched.

satvikbatra (approved):

devesh-juspay (approved):

PR #1398, opened 2026-09-02
fix: XYNE-61728 scope message edit lock to the requesting list

Dnyaneshwari-04:
[POT](https://drive.google.com/file/d/18cAnhT-a-yXZSoeuVkfhfFz-h0S6Ggxx/view?usp=sharing)
## Problem

A reply sent with "Also send to channel" renders in both the channel list and the thread panel. Clicking Edit on either copy turns **both** copies into editors at the same time. Each has its own draft and both autofocus, so they fight for the cursor and whichever is submitted last silently wins.

**Steps to reproduce**
1. Open a channel with the thread panel open beside it.
2. Reply in a thread with "Send to channel" ticked, so the reply shows in both panes.
3. Click Edit on either copy of that reply.

Expected: one editor, in the pane you clicked.
Actual: an editor opens in the channel list **and** in the thread panel.

Also reproducible with a pinned message (channel + Pins tab) and on the threads screen (preview card + thread detail).

## Root cause

The edit lock is a single global message id (`EditProvider.editingMessageId`). Every `ChatBubble` decides whether to show its editor by comparing that id to its own message id. "Also send to channel" does not create a second message — it sets `showInChannel` on the same row, which the channel query then picks up — so the two mounted bubbles share an id and both match.

## Fix

Scope the edit lock to the list it was requested from, not just the message. Each list is wrapped in an `EditSurfaceScope` that mints an id via `useId()`, and a `useMessageEdit()` hook binds that surface into `isEditingMessage()` / `requestEdit()`, so a bubble opens its editor only when the message **and** the surface match. Two mounts of the same list (thread preview card vs detail) get distinct ids automatically, so the surface never has to be declared by hand.

Editing the same message from a second surface now routes through the existing "You have unsaved edits" prompt instead of opening a second editor. Also drops `ChatBubble`'s local `isEditing` state, which only mirrored the context value a frame late.

## Scope

`apps/dashboard` only — `EditProvider`, `ChatBubble`, `ChatListV2/V3/V4`, `ThreadList`, `PinListV2`, `SearchResultMessageCard`, `FileViewerModal`.

No backend, schema, or query changes. `showInChannel` messages still appear in both places and are still editable from either.

## Test plan

- [ ] Broadcast reply: edit from channel, then from thread — one editor each time; save updates both copies
- [ ] Pinned message edited in the channel: pin card stays a bubble
- [ ] Threads screen: preview card and detail do not both open an editor
- [ ] Edit save / cancel (X, Esc); up-arrow edit-last-message in both composers
- [ ] Unsaved-edits modal across two messages, and across two panes for one message
- [ ] Navigating away while editing closes the editor and leaves no stuck lock
- [ ] Smoke: DM compose panel, search results, file viewer message preview, a ticket thread

### DIFF
--- a/apps/dashboard/src/components/Chat/ChatList/ChatListV2.tsx
+++ b/apps/dashboard/src/components/Chat/ChatList/ChatListV2.tsx
@@ -16,7 +16,7 @@ import { ChannelScopeType, MessageType } from '@xyne/shared';
 import { Conversation } from '../../../machines/stateMachine';
 import { ArrowDown } from 'lucide-react';
 import { useGetChannelConversations, useGetChannelUserStatus } from '../../../hooks/useChannels';
-import { useEditContext } from '../../../providers/EditProvider';
+import { EditSurfaceScope, useMessageEdit } from '../../../providers/EditProvider';
 import { useShortcutById } from '../../../shortcuts';
 import { findLastEditableMessage, isEventFromEmptyInput } from '../../../utils/chatUtils';
 import { standaloneNavigate } from '../../../utils/electronApp';
@@ -61,7 +61,7 @@ const ChatListV2: React.FC<ChatListProps> = ({
   // Container for the shared hover toolbar overlay (one toolbar per list).
   const hoverToolbarContainerRef = useRef<HTMLDivElement>(null);
   const { user } = useAuth();
-  const { editingMessageId, requestEdit } = useEditContext();
+  const { isEditingMessage, requestEdit } = useMessageEdit();
   const particpantionStatus = useGetChannelUserStatus(channelId);
   const [oldConversations, oldConversationsDetails] = useQuery(
     queries.channelConversationsPaginatedV3({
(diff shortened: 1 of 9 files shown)


apps/dashboard/src/components/Chat/ChatList/ChatListV2.tsx
```diff
@@ -604,4 +604,10 @@ const ChatListV2: React.FC<ChatListProps> = ({
   );
 };
 
-export default ChatListV2;
+const ChatListV2WithEditSurface: React.FC<ChatListProps> = props => (
```
harshpreet931: this wrapper is copy pasted in five files we make a withEditSurface hoc

apps/dashboard/src/providers/EditProvider.tsx
```diff
@@ -54,10 +69,36 @@ export const EditProvider = ({ children }: { children: ReactNode }): ReactElemen
   );
 };
 
+export const EditSurfaceScope = ({ children }: { children: ReactNode }): ReactElement => {
+  const surface = useId();
```
harshpreet931, line 81: would be a new id on every remount... we should release this on unmount then

harshpreet931 (dismissed):

Mohan0544 (dismissed):

apps/dashboard/src/components/Chat/ChatList/ChatListV2.tsx
Dnyaneshwari-04: Done — `withEditSurface` now lives in `EditProvider.tsx` and all five wrappers collapse to one line each:

```ts
export default withEditSurface(ChatListV2);
export default withProfiler(withEditSurface(ChatListV3), 'ChatListV3');
```

The `withProfiler(withEditSurface(...))` order keeps the profiler outermost, same nesting as before.

`SearchResultMessageCard` and `FileViewerModal` still use inline `<EditSurfaceScope>` — they wrap mid-tree JSX rather than a default export, so the HOC doesn't apply there.

harshpreet931 (approved):

Mohan0544 (approved):

PR #1437, opened 2026-09-02
feat: XYNE-61849 auto label rule for existing emails

Himanshvarma:
Migration-Changes:
  - needed for desk auto-label backfill

  - backend
  - dashboard

## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

XYNE-61849. Desk auto-label rules only ever applied to newly arriving email. This
adds an opt-in pass over mail already in the desk — the Gmail "also apply to
matching conversations" behaviour — as a checkbox on rule creation plus a
per-rule action in the rules list.

Follows the existing `ticketReassignmentQueue` shape (the away/unavailable
reassignment): one Bull job per rule at a deterministic id
`autolabel-backfill:${workflowId}`, which keyset-walks the channel's mail
internally rather than fanning out a job per email. Re-running a rule while its
run is live collapses onto the same job instead of scanning twice. Query
discipline (batch all lookups before the loop, skip work already done) follows
`aiRetriggerController` / AI Sync.

Parity with the live rule is structural, not reimplemented:
- matching calls `emailReceivedTrigger.matchFilters` — the same function the
  live trigger uses, so filters cannot drift
- the scan is limited to `type = DEFAULT`, the same inbound-only gate
  `emitEmailReceived` applies
- applying calls `applyConversationLabel` / `archiveConversationMailbox`, the
  same services the `APPLY_CONVERSATION_LABEL` step calls
- `SKIPPABLE_LABEL_ERROR_CODES` moved from the step into
  `conversation-label.service.ts` so both share one definition

Notes:
- Already-labeled skip is scoped to the rule's own `labelId` — labels are
  additive, so a thread carrying other labels still gets this one. It is an
  optimisation only; `applyConversationLabel` is idempotent on
  `(conversationId, labelId)` regardless.
- Disabling or archiving a rule mid-run stops the scan within one page
  (`isRuleStillActive`, checked at each page boundary). Labels already applied
  stay applied — no undo, same as the live rule. Surfaced in the UI as
  "Stopped after N older threads".
- `onlyNewThreads` / `onlyReplies` behaviour is inherited from the live path
  unchanged — see "Known follow-up" below.

Known follow-up (not in this PR): `isReply` is derived from
`findFirstByThreadAndChannel`, which is channel-scoped. On a desk fed by more
than one source that is not a thread identity — `Email.externalThreadId` is
stored raw while `externalMessageId` is namespaced per source. Pre-existing on
the live path; the backfill replays it rather than introducing it. Same root
cause as the foreign-link guard at `ticketController.ts:2533`.

## Areas Touched

- [x] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

## Schema & Data Changes

- [x] Migration added

If any of the above:

- [ ] Clients regenerated (`db:generate`, `db:common:generate`) — not run; the
      change is a single index, which does not alter generated Prisma Client types
- [x] Backfill is idempotent

Rollout / rollback notes:
- One migration: `20260902120000_add_emails_channel_type_id_idx` —
  `CREATE INDEX CONCURRENTLY IF NOT EXISTS emails_channelId_type_id_idx
  ON public.emails (channelId, type, id)`. Additive, no data change, no lock on
  writes. Follows the `CONCURRENTLY` convention already used in
  `20260421190000_add_performance_indices`.
- Rollout order does not matter — old code ignores the index, new code is only
  slower without it. **Apply before enabling the feature in anger**: the scan
  keyset-walks `(channelId, type, id)` and `emails` currently only has
  `@@index([channelId])`, so without it a large desk does a filter+sort over
  every email in the channel.
- Rollback: `DROP INDEX CONCURRENTLY emails_channelId_type_id_idx`. No data to
  unwind. Labels already applied by a run are ordinary label mappings and stay.
- "Backfill" here is a runtime background job, not a data migration. It is
  idempotent: the unique on `(conversationId, labelId)` means a re-run or a Bull
  retry re-does work rather than duplicating it.

## Config, Environment & URLs

## API Contract

- [ ] No API changes
- [x] New endpoint

---

## How did you test it?

Static verification only so far — this has **not** been run against a live desk,
and the migration has not been applied.

- `tsc --noEmit` on backend: clean (22 errors remain, all pre-existing, all in
  files this PR does not touch — SdlcWikiPipeline, services/agents/prompts,
  zero/mutators, zero/queries)
- `tsc --noEmit` on dashboard: clean (66 pre-existing, in ActivityItemCard,
  SearchResultMessageCard, TicketCardV2, SdlcScreen)
- ESLint on every file touched by this PR: clean

Needs manual verification before merge — see unchecked boxes below.

### Automated

- [x] Not covered by tests

### Manual

Not yet performed. Everything below is outstanding.

**Users**
- [ ] Single user
- [ ] Two users at once (two accounts / two browsers)

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Sync & resilience**

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

Specifically worth exercising:
- desk with >1 page of mail (>200 inbound emails) — confirm the keyset walk
  pages through and does not loop or stall
- rule whose label is already on some threads — confirm those are skipped and
  threads carrying *other* labels are still labelled
- disable / archive the rule mid-run — confirm the scan stops and the UI reports
  "Stopped after N older threads"
- `applyToExisting` + `keepInInbox: false` — confirm older threads are archived
  out of Inbox as warned in the wizard
- a desk fed by both a mailbox and an app (this branch's mixed-source case) —
  confirm app-sourced threads get labelled; note sender/recipient filters match
  poorly there because app `from`/`to` are synthetic
- backfill with the automation worker down — rule still saves, response carries
  `backfill: null`

---

## Checklist

- [x] `pnpm run build:shared` passes — not run
- [x] Backend `typecheck` passes; `build` not run separately
- [x] Dashboard `lint:errors-only` + `typecheck` pass; `build` not run
- [x] New deps added with `pnpm --filter <pkg> add <dep>` — none added
- [ ] Docs / guidelines updated where behaviour changed — none needed

### DIFF
--- a/apps/backend/prisma/migrations/20260902120000_add_emails_channel_type_id_idx/migration.sql
+++ b/apps/backend/prisma/migrations/20260902120000_add_emails_channel_type_id_idx/migration.sql
@@ -1,0 +1,5 @@
+-- Desk auto-label backfill walks one channel's inbound mail by keyset:
+--   WHERE "channelId" = $1 AND "type" = 'DEFAULT' AND "id" > $cursor ORDER BY "id" ASC
+-- The existing "emails_channelId_idx" forces a filter+sort over every email in the
+-- channel; this composite serves the predicate and the ordering together.
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "emails_channelId_type_id_idx" ON "public"."emails" ("channelId", "type", "id");
--- a/apps/backend/prisma/migrations/20260902120000_add_emails_channel_type_id_idx/migration.sql
+++ b/apps/backend/prisma/migrations/20260902120000_add_emails_channel_type_id_idx/migration.sql
@@ -1,5 +1,0 @@
--- Desk auto-label backfill walks one channel's inbound mail by keyset:
---   WHERE "channelId" = $1 AND "type" = 'DEFAULT' AND "id" > $cursor ORDER BY "id" ASC
--- The existing "emails_channelId_idx" forces a filter+sort over every email in the
--- channel; this composite serves the predicate and the ordering together.
-CREATE INDEX CONCURRENTLY IF NOT EXISTS "emails_channelId_type_id_idx" ON "public"."emails" ("channelId", "type", "id");
--- a/apps/backend/prisma/migrations/20260909120000_add_emails_channel_type_id_idx/migration.sql
+++ b/apps/backend/prisma/migrations/20260909120000_add_emails_channel_type_id_idx/migration.sql
@@ -1,0 +1,7 @@
+-- CreateIndex
+-- Desk auto-label backfill walks one channel's inbound mail by keyset:
+--   WHERE "channelId" = $1 AND "type" = 'DEFAULT' AND "id" > $cursor ORDER BY "id" ASC
+-- "emails_channelId_idx" alone forces a filter+sort over every email in the channel.
+-- CONCURRENTLY: emails is on the inbound mail write path.
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "emails_channelId_type_id_idx"
+  ON "public"."emails" ("channelId", "type", "id");
--- a/apps/backend/prisma/migrations/20260909120000_add_emails_channel_type_id_idx/migration.sql
+++ b/apps/backend/prisma/migrations/20260909120000_add_emails_channel_type_id_idx/migration.sql
@@ -1,7 +1,13 @@
 -- CreateIndex
--- Desk auto-label backfill walks one channel's inbound mail by keyset:
---   WHERE "channelId" = $1 AND "type" = 'DEFAULT' AND "id" > $cursor ORDER BY "id" ASC
+-- Desk auto-label backfill walks one channel's inbound mail by keyset, within a
+-- lookback window:
+--   WHERE "channelId" = $1 AND "type" = 'DEFAULT' AND "createdAt" >= $2
+--     AND "id" > $cursor ORDER BY "id" ASC
 -- "emails_channelId_idx" alone forces a filter+sort over every email in the channel.
+-- "createdAt" trails "id": "id" is the column the scan ranges on AND orders by, so
+-- it has to come first for the LIMIT to stop early without a sort. "createdAt" is
+-- only ever an extra filter, and keeping it in the index lets both the count and
+-- the pre-window rows be tested without a heap fetch per row.
 -- CONCURRENTLY: emails is on the inbound mail write path.
-CREATE INDEX CONCURRENTLY IF NOT EXISTS "emails_channelId_type_id_idx"
-  ON "public"."emails" ("channelId", "type", "id");
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "emails_channelId_type_id_createdAt_idx"
+  ON "public"."emails" ("channelId", "type", "id", "createdAt");
--- a/apps/backend/src/automations/queue/desk-label-backfill.queue.ts
+++ b/apps/backend/src/automations/queue/desk-label-backfill.queue.ts
@@ -1,0 +1,174 @@
+import Bull from 'bull';
+import { logger } from '@/utils/logger';
+import { redisService } from '@/services/redisService';
+import type { DeskLabelBackfillProgress } from '../services/desk-label-backfill.service';
+
+export interface DeskLabelBackfillJobData {
+  workflowId: string;
+}
+
+export const DESK_LABEL_BACKFILL_JOB = 'desk-label-backfill';
+
+/**
+ * One backfill per rule, addressed by a deterministic id. Re-running a rule while
+ * its run is still queued or active collapses onto the same job instead of
+ * scanning the channel twice.
+ */
+export const deskLabelBackfillJobId = (workflowId: string): string =>
+  `autolabel-backfill:${workflowId}`;
+
+export type DeskLabelBackfillState = 'queued' | 'running' | 'completed' | 'failed';
+
+export interface DeskLabelBackfillRun {
+  state: DeskLabelBackfillState;
+  progress: DeskLabelBackfillProgress | null;
+  failedReason: string | null;
+}
+
+export type EnqueueBackfillResult = 'enqueued' | 'already-running';
+
+function isProgress(value: unknown): value is DeskLabelBackfillProgress {
+  return typeof value === 'object' && value !== null && 'scanned' in value;
+}
+
+class DeskLabelBackfillQueue {
+  private queue: Bull.Queue<DeskLabelBackfillJobData> | null = null;
+  private isInitialized = false;
+  private isInitializing = false;
+
+  async initialize(): Promise<void> {
+    if (this.isInitialized || this.isInitializing) return;
+    this.isInitializing = true;
+
+    try {
+      this.queue = new Bull<DeskLabelBackfillJobData>('desk-label-backfill', {
+        redis: {
+          ...redisService.getRedisConfig(),
+          lazyConnect: false,
+        },
+        defaultJobOptions: {
+          // A retry restarts the scan from the beginning. Every apply is
+          // idempotent so that is safe, just wasted work — hence a low cap.
+          attempts: 2,
+          backoff: { type: 'exponential', delay: 5000 },
+          // Completed runs are kept briefly so the rules list can show the
+          // result of the run the user just started.
+          removeOnComplete: { age: 60 * 60, count: 200 },
+          removeOnFail: false,
+        },
+        settings: {
+          // Long-running: one job walks the channel's whole mail history.
+          lockDuration: 5 * 60 * 1000,
+          stalledInterval: 60 * 1000,
+          maxStalledCount: 1,
+        },
+      });
+
+      this.setupEventListeners();
+      this.isInitialized = true;
+      logger.info('[DESK-LABEL-BACKFILL-QUEUE] Initialized');
+    } catch (error) {
+      logger.error('[DESK-LABEL-BACKFILL-QUEUE] Failed to initialize:', error);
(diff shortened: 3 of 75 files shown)


Revanthvenkatpasupuleti (dismissed):

apps/backend/prisma/migrations/20260902120000_add_emails_channel_type_id_idx/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- Desk auto-label backfill walks one channel's inbound mail by keyset:
+--   WHERE "channelId" = $1 AND "type" = 'DEFAULT' AND "id" > $cursor ORDER BY "id" ASC
+-- The existing "emails_channelId_idx" forces a filter+sort over every email in the
+-- channel; this composite serves the predicate and the ordering together.
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "emails_channelId_type_id_idx" ON "public"."emails" ("channelId", "type", "id");
```
satvikbatra: migration should be prisma generated

packages/xyne-claw-shared/src/tools/desk-report/tools.ts
```diff
@@ -595,7 +595,16 @@ export const createDeskReportTool: ToolDefinition = {
         type: "string",
         description:
           "Full report in markdown. Headings, tables, lists, code blocks, and ```chart blocks all " +
-          "render natively. BAR_CHART items may include an optional drilldown field: " +
+          "render natively. Supported chart visualTypes (exactly these — anything else renders as raw text): " +
```
satvikbatra: merge this change in feature/deploy-xyneclaw branch

apps/backend/prisma/migrations/20260902120000_add_emails_channel_type_id_idx/migration.sql
Himanshvarma: resolved

Revanthvenkatpasupuleti (dismissed):

Navviiii (dismissed):

apps/dashboard/src/components/Automation/AutomationBuilder/StepCard/ApplyConversationLabelStepForm.tsx
```diff
@@ -115,7 +117,7 @@ export function ApplyConversationLabelStepForm({
           onChange={e => {
             const name = e.target.value;
             const existing = (catalog ?? []).find(
-              l => l.name.toLowerCase() === name.trim().toLowerCase(),
+              l => l.name.toLowerCase() === name.trim().toLowerCase() && l.createdBy === user?.id,
```
satvikbatra: guard is only on the text input. the chip row below still passes teammate's labelId directly and it fails silently as label_id_mismatch since that's a skippable code. can we filter the chip list by createdBy too

apps/backend/src/automations/services/desk-label-backfill.service.ts
```diff
@@ -0,0 +1,359 @@
...
+      progress.matched += 1;
+      if (seen.has(email.conversationId)) continue;
+      seen.add(email.conversationId);
+      conversationIds.push(email.conversationId);
+    }
+
+    const { pending, alreadyLabeled } = await filterAlreadyLabeled(rule.labelId, conversationIds);
+    progress.alreadyLabeled += alreadyLabeled;
+
+    for (const conversationId of pending) {
```
satvikbatra: no throttle here. we already have this pattern in aiRetriggerController — STAGGER_MS 1000, LOOKBACK_DAYS 3, COOLDOWN_SECONDS 600 and it shows an eta. this one has no stagger no lookback no cooldown.

also conversation_label_mappings is zero synced so every apply fans out a cvr diff per connected client on the shared replication stream, it's not just db load. can we follow the same pacing

apps/backend/src/worker.ts
```diff
@@ -277,6 +277,12 @@ class WorkerService {
         logger.info('Starting automation schedule worker...');
         await automationScheduleWorker.start();
 
+        const { deskLabelBackfillWorker } = await import(
+          '@/automations/queue/desk-label-backfill.worker'
+        );
+        logger.info('Starting desk auto-label backfill worker...');
+        await deskLabelBackfillWorker.start();
```
satvikbatra, line 284: shutdown() is defined on the worker but never called anywhere. every other worker in this file has it in WorkerService.shutdown()

apps/backend/src/automations/services/desk-label-rules.service.ts
```diff
@@ -231,6 +240,22 @@ class DeskLabelRulesService {
...
+    const result = await this.createRule(payload, auth);
+    if (!payload.applyToExisting) return result;
+
+    // Enqueued after the rule is committed — the worker reads the rule back from
+    // the DB, so a job started inside the transaction could find nothing there.
+    // A duplicate rule still backfills: "apply this rule to my old mail" is a
+    // valid ask even when the rule itself already existed.
+    const workflowId = result.automations[0]?.id;
+    if (!workflowId) return result;
+    return { ...result, backfill: await this.enqueueBackfill(workflowId) };
```
satvikbatra: comment above covers the duplicate ACTIVE rule but createRule can also return a DISABLED one via findOrRestoreExistingDeskWorkflow. startBackfill blocks that with a 400 but this path doesn't, so resolveBackfillRule returns null and the job is dropped after we already toasted. can we check status before enqueue

apps/backend/src/automations/services/desk-label-backfill.service.ts
```diff
@@ -0,0 +1,359 @@
...
+            );
+          }
+          return result;
+        });
+
+        if (applied.applied) progress.labeled += 1;
+        else progress.alreadyLabeled += 1;
+        if (!rule.keepInInbox) progress.archived += 1;
+      } catch (err) {
+        progress.skipped += 1;
```
satvikbatra, line 391: skipped is counted but never surfaced, BackfillStatus only reads labeled + alreadyLabeled. run says completed / 0 threads even if every apply threw. can we show it

apps/backend/src/automations/queue/desk-label-backfill.queue.ts
```diff
@@ -0,0 +1,177 @@
...
+    const queue = this.getQueue();
+    const jobId = deskLabelBackfillJobId(workflowId);
+
+    const existing = await queue.getJob(jobId);
+    if (existing) {
+      const [failed, completed] = await Promise.all([
+        existing.isFailed(),
+        existing.isCompleted(),
+      ]);
+      if (!failed && !completed) return 'already-running';
```
satvikbatra: if the automation worker is off the job sits in waiting forever, and this keeps returning already-running so the user can never retry even after the worker is back. need some staleness bound

apps/backend/src/automations/queue/desk-label-backfill.worker.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    logger.info(
+      `[DESK-LABEL-BACKFILL-WORKER] Job ${job.id} starting — automation=${workflowId} channel=${rule.channelId} label=${rule.labelId}`,
+    );
+
+    // Background job → no HTTP tenant scope. Open one from the rule's workspace so
+    // every write in the run gets workspaceId stamped.
+    const progress = await runAsServiceActor('desk-label-backfill', rule.workspaceId, () =>
+      runDeskLabelBackfill(rule, next => {
+        void job.progress(next);
```
satvikbatra: void job.progress without a catch. that writes to redis, a blip becomes an unhandled rejection and kills the worker. same on line 65, and that one should be awaited too otherwise the job completes before the final count lands

apps/dashboard/src/components/xyne-desk/AutoLabelWizard/AutoLabelWizard.tsx
```diff
@@ -169,15 +179,28 @@ export function AutoLabelWizard({
...
+        }
+      }
       void queryClient.invalidateQueries({
         queryKey: deskLabelRulesQueryKey(channelId),
       });
       onCreated?.(data.automations);
       onOpenChange(false);
     },
     onError: (err: unknown) => {
+      const response = (err as { response?: { status?: number; data?: { error?: string } } })
```
satvikbatra, line 219: are we returning 409 anywhere ? serviceError only maps to 403/404/400 and the label ownership case throws forbidden. if not can we remove this

apps/backend/prisma/migrations/20260909120000_add_emails_channel_type_id_idx/migration.sql
```diff
@@ -0,0 +1,7 @@
+-- CreateIndex
+-- Desk auto-label backfill walks one channel's inbound mail by keyset:
+--   WHERE "channelId" = $1 AND "type" = 'DEFAULT' AND "id" > $cursor ORDER BY "id" ASC
+-- "emails_channelId_idx" alone forces a filter+sort over every email in the channel.
+-- CONCURRENTLY: emails is on the inbound mail write path.
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "emails_channelId_type_id_idx"
```
satvikbatra: this is still not prisma generated!

karunrajk (dismissed):

satvikbatra (dismissed):

satvikbatra (approved):

karunrajk (approved):

PR #1442, opened 2026-09-02
feat: XYNE-62826 downtime prevention

ishaanrawat-netizen:
Migration-Changes:
  - adding table for my agent config

  - xyne-transcription-agent
  - backend

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


apps/backend/python-agent/main.py
```diff
@@ -620,8 +620,15 @@ def run_health_server():
...
     # Run the LiveKit agent (blocks until shutdown)
+    # agent_name switches this worker from automatic dispatch (joins every room) to
+    # explicit dispatch — the backend must now request it by name via
+    # AgentDispatchClient.createDispatch. This is what makes zero-downtime, canary-tested
+    # agent rollouts possible: a new pod runs under a new agent_name and is invisible to
+    # real traffic until the backend (via this pod's own self-report, or a human fallback)
+    # rolls it into the 'stable' or 'test' slot.
     cli.run_app(WorkerOptions(
         entrypoint_fnc=entrypoint,
+        agent_name=config.transcription_agent_name,
```
XyneSpaces, line 631: **[🟠 Major · R2 — deploy-order hazard]** Setting `agent_name` flips this worker from automatic dispatch (joins every room) to explicit-dispatch-only. But the backend side resolves the Superposition flag with a default of `'none'` (`NO_EXPLICIT_DISPATCH` in `liveKitService.ts` — whose own comment calls the flag key *"a placeholder pending an actual flag being created in Superposition's dashboard"*), and `getStringValue` returns that default whenever the flag is missing **or the provider errors**. So if this build replaces the current automatic-dispatch worker before (1) the Superposition flag exists, (2) an agent is registered + rolled out into `default`, and (3) users actually resolve to it, every new call gets **no transcription agent at all** — the exact outage this PR exists to prevent. The PR body documents no rollout order and the template's rollout-order box is unchecked.

↳ **FIX:** document and enforce the deploy order in the PR description (backend first → create the Superposition flag → deploy + self-register the new agent build → roll it into `default` → only then retire the old automatic-dispatch worker), or keep one automatic-dispatch worker running until flag-based routing is confirmed live.

XyneSpaces (commented):
Automated review — 5 findings inline (4 🟠 major, 1 🟡 minor). Full summary + reviewer recommendation in the top-level comment.

Security check: **clean** — all new endpoints fail closed (`verifyTranscriptionAgent` x-api-key + `timingSafeEqual`; `internalServiceAuth` on rollout endpoints; Prisma parameterized; throwaway verify-rooms only reachable by key holders). No R1 findings. Migration (`20260901120000_add_transcription_agent_table`) reviewed: no unsafe operations.

apps/backend/src/services/liveKitService.ts
```diff
@@ -137,6 +213,416 @@ export class LiveKitService {
...
+      acquired = true;
+    }
+
+    if (!acquired) return { acquired: false };
+
+    try {
+      const result = await fn();
+      return { acquired: true, result };
+    } finally {
+      await redisService.del(lockKey).catch((error) => {
```
XyneSpaces: **[🟠 Major · #105 locking]** The `finally` releases `lockKey` unconditionally, with no ownership check. `fn()` here spans a CAC lookup, `deleteRoom`, `createRoom`, `createDispatch` and a DB update — if any LiveKit/DB call is slow enough to outlive the 10s TTL, the key expires, a second joiner acquires the "lock", and this `finally` then deletes *that* request's lock — admitting a third racer into the delete/recreate/dispatch critical section the lock exists to protect.

↳ **FIX:** write a random token as the lock value and release with a compare-and-delete (Lua `GET`+compare+`DEL` via `eval`, or the `redisService` equivalent), and/or raise the TTL above the realistic worst case of `fn()`.

apps/backend/src/services/liveKitService.ts
```diff
@@ -137,6 +213,416 @@ export class LiveKitService {
...
+  }
+
+  private async isDispatchClaimed(roomName: string, dispatchId: string): Promise<boolean> {
+    try {
+      const dispatch = await this.agentDispatch.getDispatch(dispatchId, roomName);
+      const job = dispatch?.state?.jobs?.[0];
+      return Boolean(job?.state?.workerId);
+    } catch (error) {
+      logger.warn(`[${roomName}] dispatch_claim_check_call_failed | dispatch_id=${dispatchId}, error=${error}`);
+      return false;
```
XyneSpaces: **[🟠 Major · R2/R4 — duplicate-agent risk]** On a transient `getDispatch` error, `isDispatchClaimed` returns `false`, so `checkDispatchClaimed` (`allowFallback: true`) immediately calls `dispatchFallbackAgent` — which does **not** re-check room participants, unlike `redispatchTranscriptionAgentIfMissing` (which presence-checks and returns `'agent_present'`). If the original dispatch was actually alive (transient error, or a claim landing just past the check window), the room ends up with two transcription agents → duplicate transcripts for that call.

↳ **FIX:** treat an errored lookup as indeterminate — skip the fallback and let the redispatch loop handle it — or presence-check for an existing agent before `dispatchFallbackAgent`, mirroring the guard the redispatch path already has.

apps/backend/src/services/liveKitService.ts
```diff
@@ -1,14 +1,80 @@
...
+const REDISPATCH_JITTER_MAX_MS = 10 * 1000;
+
+/** Serializes re-dispatches for one room against duplicate webhook deliveries/replicas. */
+const REDISPATCH_COOLDOWN_SECONDS = 15;
+const redispatchCooldownKey = (roomName: string) => `livekit:agent-redispatch:cooldown:${roomName}`;
+
+/** How long after createDispatch we check whether any worker actually claimed the job. */
+const DISPATCH_CLAIM_CHECK_DELAY_MS = 9 * 1000;
+
+function nextRedispatchDelayMs(retryIndex: number): number {
```
XyneSpaces, line 77: **[🟠 Major · #204 tests]** This PR adds ~490 lines of new dispatch/redispatch/lock machinery — `nextRedispatchDelayMs`, `isAgentParticipant`/`isHumanParticipant`, the `resolveAgentNameForUser` fallback matrix, the redispatch outcome state machine (terminal vs retryable), and `withCallCreationLock` winner/loser paths — with no tests anywhere in the diff, and the template's "Added / updated tests" box is unchecked. For a PR whose whole purpose is downtime prevention, the recovery paths (fallback dispatch, redispatch backoff, lock-loser wait) are exactly what you want test-covered before a prod incident exercises them for real.

↳ **FIX:** add unit tests (mocked redis/LiveKit/repositories) covering at minimum: the `resolveAgentNameForUser` fallback matrix, terminal-vs-retryable redispatch outcomes, `withCallCreationLock` winner/loser, and `checkDispatchClaimed`'s fallback decision.

apps/backend/python-agent/health_server.py
```diff
@@ -20,6 +21,79 @@
...
+                async with session.post(
+                    url, headers=headers, json=payload, timeout=aiohttp.ClientTimeout(total=15),
+                ) as response:
+                    body = await response.json()
+                    if response.status == 200 and body.get("success"):
+                        logger.info(
+                            f"[register_agent] registered | agent_name={cfg.transcription_agent_name}, role={cfg.transcription_agent_role}, attempt={attempt}"
+                        )
+                        return
+                    last_reason = body.get("reason") or body.get("error") or f"http_{response.status}"
```
XyneSpaces: **[🟡 Minor · #76 no-retry-on-4xx]** The register loop retries every non-200 — including `400`, which the controller returns for permanent validation failures (invalid `agentName`/`role`). A misconfigured pod therefore burns ~90s of pointless retries (30s + 60s backoff) and logs `rejected` twice more before giving up. Only `200 {success:false}` rejections and 5xx/network errors are meaningfully retryable.

↳ **FIX:** `if response.status == 400: return` right after logging the rejection. Low impact (bounded, startup-only) — fine to land as a follow-up.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1442

**Verdict:** Changes requested — 4 major, 1 minor, 3 nits
**Reviewed:** 15 files, ~936 added lines (every hunk) · head `a56badc` · **Security check: clean — no R1 findings**

All new endpoints fail closed: `verifyTranscriptionAgent` uses `x-api-key` + `timingSafeEqual` with no fallback; the rollout endpoints sit behind `internalServiceAuth`; inputs are string-validated; all DB access is parameterized; `verifyAgentLive` throwaway rooms (`agent-verify-<name>-<ts>`) are only reachable by API-key holders. **No security blockers.**

### Findings
| 🟠 major | R2 | `python-agent/main.py:631` | **Deploy-order hazard** — `agent_name` disables automatic dispatch, but the backend flag (`transcription_agent_role`) defaults to `'none'` and its key is a documented Superposition placeholder. Deploying the agent build before flag + registration + rollout = **no transcription agent on any new call**. Needs a documented/enforced rollout order in the PR. |
| 🟠 major | #105 | `services/liveKitService.ts:599` | `withCallCreationLock`'s `finally` deletes the lock unconditionally (no ownership token) — a slow `fn()` that outlives the 10s TTL can delete a *second* request's lock and re-open the duplicate delete/recreate/dispatch race. |
| 🟠 major | R2/R4 | `services/liveKitService.ts:388` | `isDispatchClaimed` swallows a transient `getDispatch` error as `return false` → `checkDispatchClaimed` fallback-dispatches with **no presence check** → two agents in one room → duplicate transcripts. Redispatch path guards this; fallback path doesn't. |
| 🟠 major | #204 | `services/liveKitService.ts:61` | ~490 lines of new dispatch/redispatch/lock machinery with zero tests (template box unchecked). For a downtime-prevention PR, the recovery paths are exactly what needs coverage. |
| 🟡 minor | #76 | `python-agent/health_server.py:77` | Register loop retries permanent 400s (~90s wasted on a misconfigured pod). Bounded + startup-only — fine for a follow-up. |
| ⚪ nit | #279 | `services/liveKitService.ts` | `logger.warn` on *successful* rollout events — success at warn level dilutes the alert signal; use `info`. |
| ⚪ nit | #208 | PR body | No POT link / testing evidence attached. |
| ⚪ nit | R9 | `services/liveKitService.ts` | Mid-call redispatch never falls back off a pinned-but-dead agent (code comments say deliberate — confirm that's the intended product behavior). |

**Positive notes:** the redispatch watch design is genuinely good — Redis cooldown that fails *open*, terminal outcomes (`agent_present`/`room_inactive`/`no_humans`), 30s→10min backoff with jitter, `room_finished` cancellation; and the fail-safe contract ("user can always join, agent dispatch is best-effort") is consistently applied in `resolveAgentNameForUser`.

### Recommended reviewer
**@shubhamdhamija-juspay** — owns `callRepository.ts` (changed in this PR, merged #1417 yesterday) plus the deepest recent transcription-pipeline footprint (`transcriptService.ts`, `noteTakerTranscriptService.ts`); 6 merged calls-domain PRs (most of any non-author); **0 pending review requests**.

| Candidate | Owns (files in this PR) | Pending reviews | Tie-breaker |
|-----------|-------------------------|-----------------|-------------|
| @shubhamdhamija-juspay | `callRepository.ts` (+ transcriptService/noteTakerTranscriptService adjacency) | 0 | most merged calls-domain PRs (6); worked the call→transcript→summary pipeline end-to-end |
| @abhinavjuspay | `callController.ts` (+88/−18, the dispatch entry point) | 0 | 2 calls PRs; runner-up |
| @Pradeesh333 | — (5 call/transcript PRs, none on this PR's files) | 0 | routing/ingestion domain |
| @sayantanp-eng | — (calls mutators/UI) | 0 | participants flow |
| PiyushJuspay | — | 3 | excluded — highest pending load |
| @Mohan0544 | — | 2 | excluded — pending load |

*No CODEOWNERS entry covers the changed paths (repo CODEOWNERS only owns ACL paths), so ranking is file-ownership + workload. GitHub's reviewer-request API isn't exposed to this automation's toolset — the @mention above is the notification path; please also add Shubham as a reviewer from the PR sidebar.*

apps/backend/python-agent/main.py
ishaanrawat-netizen: intentional change

apps/backend/python-agent/health_server.py
ishaanrawat-netizen: will check this

XyneSpaces:
## Xyne Spaces Reviewer — PR #1442 (re-review @ `acf76ead`)

New commits since the last review (`a56badc`): migration `public`→`non_zero`, liveKitService guard/race fixes, new `redisService.deleteIfMatch`. **12 of 16 files are byte-identical (blob-SHA compared)** — only `migration.sql`, `schema.prisma`, `liveKitService.ts`, `redisService.ts` changed, and every changed hunk was reviewed.

**Verdict: the two code-level majors are fixed correctly — no new issues on the changed hunks. Security posture unchanged (the new hunks touch no auth surface).** Remaining items are the acknowledged V1 deferrals.

### Fix verification
| Prior finding | Status |
|---|---|
| 🟠 #105 — lock released without ownership check | ✅ **Fixed** — per-acquisition `lockToken` + atomic Lua compare-and-delete (`redisService.deleteIfMatch`), release failure caught and logged. Exactly the right remedy. |
| 🟠 R2/R4 — transient claim-check error → blind fallback dispatch | ✅ **Fixed** — `isDispatchClaimed` returns `null` on error; the indeterminate case defers to the presence-checked `ensureTranscriptionAgent` path instead of dispatching a fallback; `verifyAgentLive` treats `null` as not-verified. Both directions now fail safe. |

Also verified on the new hunks: `liveAgentIdentitiesForName` (name-scoped presence check with a conservative coarse fallback on list error, errors logged) and the `resolveAgentName` / `resolveFallbackAgentName` split (DB errors → `null` = no dispatch, call join unaffected) — clean.

### Still open (deferred for V1)
|---|---|---|---|
| 🟠 | R2 | PR body / deploy process | Rollout order still undocumented — backend first, create the Superposition flag, register + roll out the agent, only then retire the old automatic-dispatch worker. One paragraph in the PR description; only bites at deploy time. |
| 🟠 | #204 | `liveKitService.ts` | Tests for the dispatch/redispatch/lock machinery still absent — please land them in the follow-up PR. |
| 🟡 | #76 | `health_server.py:77` | Register loop retries permanent 400s (file unchanged) — follow-up. |
| ⚪ | #279 | `liveKitService.ts` | `logger.warn` on successful rollout — follow-up. |
| ⚪ | #208 | PR body | No POT link. |
| ⚪ | — | `liveKitService.ts` | Residual from the lock fix: 10s TTL without renewal — if `fn()` ever exceeds 10s, a second caller can still run concurrently (compare-and-delete prevents lock *theft*, not concurrent execution past expiry). Hardening nit, not blocking. |

**Deployment note:** the migration directory name (`20260901120000_…`) is unchanged but its content moved the table `public`→`non_zero`. If the earlier revision of this branch was `migrate deploy`-ed anywhere (e.g. the test env), that migration is recorded as applied under the same name and will not re-run — drop `public.transcription_agent` and its `_prisma_migrations` row there, or the backend will query a `non_zero.transcription_agent` that doesn't exist in that env.

### Recommended reviewer
**@shubhamdhamija-juspay** (unchanged — owns `callRepository.ts` + the transcription pipeline, 0 pending review requests). Please add him from the PR sidebar; the reviewer-request API isn't exposed to this automation.

apps/backend/src/services/redisService.ts
```diff
@@ -788,6 +788,24 @@ class RedisService {
...
+  /**
+   * Deletes `key` only if its current value still equals `expectedValue`. Atomic
+   * via Lua so a lock holder can never release a lock it no longer owns — e.g. its
+   * own TTL expired, someone else acquired the key, and a bare DEL would otherwise
+   * delete THEIR lock instead of a stale one, letting a third caller in.
+   */
+  async deleteIfMatch(key: string, expectedValue: string): Promise<boolean> {
+    if (!this.redis) throw new Error('Redis not initialized');
+    const luaScript = `
+      if redis.call('GET', KEYS[1]) == ARGV[1] then
```
KushagarGoel1: do we need lua script here?

apps/backend/src/routes/internal.ts
```diff
@@ -8,4 +9,7 @@ const internalController = new InternalController();
 router.get('/org-members/check', internalServiceAuth, internalController.checkOrgMember);
 router.post('/auth/email/login', internalServiceAuth, internalController.loginOrgMember);
 
+router.get('/transcription-agent/agents', internalServiceAuth, transcriptionAgentRolloutController.listAgents);
```
KushagarGoel1: why in internal?

apps/backend/src/services/liveKitService.ts
```diff
@@ -137,6 +213,437 @@ export class LiveKitService {
...
+  }
+
+  /**
+   * Live verification for a rollout: dispatches `agentName` into a disposable,
+   * uniquely-named room and checks whether a worker claims it within ~9s. This is what
+   * lets the rollout path reject a typo'd/undeployed name outright, and it's also the
+   * on-startup self-check a pod's own rollout call effectively performs — same
+   * mechanism, same room-scoped dispatch, just a throwaway room instead of a real call.
+   */
+  async verifyAgentLive(agentName: string): Promise<boolean> {
```
KushagarGoel1: direcly check liveliness with worker

apps/backend/src/services/liveKitService.ts
```diff
@@ -137,6 +213,437 @@ export class LiveKitService {
...
+  }
+
+  /**
+   * Explicitly dispatches `agentName` into `roomName` for a real call. Fire-and-forget
+   * from the caller's perspective: on success, schedules a one-shot check ~9s later
+   * (`checkDispatchClaimed`) that arms the crash-redispatch loop if nothing claimed it.
+   * On the `createDispatch` call itself throwing, arms the same loop immediately.
+   */
+  async dispatchTranscriptionAgentForCall(roomName: string, agentName: string): Promise<{ dispatchId: string } | null> {
+    const dispatch = await this.dispatchTranscriptionAgent(roomName, agentName);
```
KushagarGoel1, line 341: how do we make sure no memory leak and stale threads gets cleaned up

KushagarGoel1 (changes requested):

apps/backend/src/services/liveKitService.ts
ishaanrawat-netizen: we can only check liveness but not it's ability to joinrooms. the only way to ensure it's actual liveness is if we create a throwaway room and dispatch the agent to it. This will replicate how the agent will join real call and ensure it's ability to join calls in the future and an aAPI endpoint to hit before each call will only work if I have 1 pod. Once I sclae it across pods, there is no option to do this as we dont know which pod will livekit actually route our request to. this verifyAgentLive is the only way to check that there is at least one, able and available pod that can join calls and this is only running two times (registeration and rollout), it never runs on the hot path.

apps/backend/src/routes/internal.ts
ishaanrawat-netizen: fixed

apps/backend/prisma/schema.prisma
```diff
@@ -3372,6 +3338,13 @@ model LookupValue {
   @@schema("public")
 }
 
+// Explicit-dispatch transcription agent registry. `agentName` is deliberately NOT the
```
Mayankbansal-juspay: please remove this comment

apps/backend/src/services/liveKitService.ts
```diff
@@ -137,6 +229,410 @@ export class LiveKitService {
...
+        NO_EXPLICIT_DISPATCH,
+        { userId },
+      );
+    } catch (error) {
+      // CAC/user-matching failure: treat exactly like an unlisted user rather than
+      // falling back to DEFAULT_TRANSCRIPTION_AGENT_ROLE — a Superposition outage must
+      // not suddenly hand every production user a second explicit-dispatch agent.
+      // Automatic dispatch is still there regardless, so this fails safe either way.
+      logger.error(`transcription_agent_role_resolution_failed | userId=${userId}, error=${error}, falling_back_to=no_explicit_dispatch`);
+      return null;
```
Mayankbansal-juspay, line 257: if it returns null, agent will not join the call? can we have fallback always?

apps/backend/src/services/liveKitService.ts
```diff
@@ -137,6 +229,410 @@ export class LiveKitService {
...
+    return this.resolveAgentName(role);
+  }
+
+  /**
+   * Every role's current {agentName, url}, read fresh from Superposition on every call —
+   * no cache, since a stale read here just means an extra ~3s health-check timeout on an
+   * already-rolled-back name, not a correctness issue. `getObjectValue` already fails
+   * closed to `{}` internally on any CAC error, so no separate try/catch is needed here.
+   */
+  private async getConfiguredAgentRoles(): Promise<Record<string, ConfiguredTranscriptionAgent>> {
```
Mayankbansal-juspay, line 270: we are thinking of static url which can be constructed here with agent-name right, are we putting that in CAC? can u once confirm with @KushagarGoel1

apps/backend/src/services/liveKitService.ts
ishaanrawat-netizen: abhi nhi dala hai fallback because abhi automatic dispatch is still there. And I dont want to switch it without proper canary testing done for this. Once we have verified this in pre-prod. We will add fallbacks as then there will always be a default value. RIght now if I put in a fallback and a default, and a default is loaded, multiple agents  will join each call.

apps/backend/src/services/liveKitService.ts
KushagarGoel1: base url is static only right, appended with agent name, that should work

KushagarGoel1 (dismissed):

KushagarGoel1 (approved):

Mayankbansal-juspay (approved):

shubhamdhamija-juspay (approved):

PR #1449, opened 2026-09-02
fix: XYNE-61758 reject empty attachment objects instead of serving 200 with 0 bytes

XyneSpaces:
## Problem
Ticket XYNE-61758: an image attachment (`pasted-image-1788325660552.jpg`) rendered as "Failed to load image" for multiple users. Root cause: the attachment is stored in GCS as a **zero-byte object**, but the backend `downloadAttachment()` handler streamed `getFileBuffer()` and set `Content-Length` from `buffer.length` with **no empty/exists guard** — so a 0-byte object was served as a valid **HTTP 200 with an empty body**. The client then built a 0-byte `File`, the image decoder failed, and `ImageViewer` showed a misleading generic error.

## Fix
- **Backend** (`apps/backend/src/controllers/attachmentController.ts`): return `404 { error: 'Attachment file is empty or unavailable' }` when the fetched buffer is 0 bytes, instead of a 200 with 0 bytes (mirrors the `fileExists` guard the thumbnail/ranged handlers already have).
- **Frontend fetch layer** (`apps/dashboard/src/services/clients/fileFetchService.ts`): `fetchFile` and `createPreviewUrl` now throw on an empty blob so React Query surfaces an error state instead of caching a broken 0-byte File/Blob.
- **ImageViewer** (`apps/dashboard/src/components/FileViewer/ImageViewer.tsx`): fail fast on a 0-byte `File` and show a clearer, actionable message — "Couldn't load image / The file may be missing or empty".

## Validation
- Backend `tsc --noEmit` passes.
- Before/after proof-of-test videos recorded showing the empty object served as 200+0 bytes → broken tile (before) vs 404 + clear message (after), with the healthy image still rendering in both.

Ticket: XYNE-61758

### DIFF
--- a/apps/backend/src/controllers/attachmentController.ts
+++ b/apps/backend/src/controllers/attachmentController.ts
@@ -305,6 +305,19 @@ export class AttachmentController {
 
       const buffer = await service.getFileBuffer(filePath);
 
+      // Guard: a zero-byte / empty storage object must never be served as a valid
+      // 200. Otherwise the client builds a 0-byte File, the image decoder fails,
+      // and the viewer shows "Failed to load image" (XYNE-61758). Treat an empty
+      // object as a missing/corrupt file so the client can show a clear error and
+      // retry, instead of silently caching a broken preview.
+      if (!buffer || buffer.length === 0) {
+        logger.error(
+          `Empty storage object for attachment ${attachmentId} at path: ${filePath} (0 bytes)`,
+        );
+        res.status(404).json({ error: 'Attachment file is empty or unavailable' });
+        return;
+      }
+
       // Set response headers (safe disposition/type to prevent stored XSS)
       res.setHeader('Content-Length', buffer.length);
       setSafeDownloadHeaders(res, {
--- a/apps/dashboard/src/components/FileViewer/ImageViewer.tsx
+++ b/apps/dashboard/src/components/FileViewer/ImageViewer.tsx
@@ -54,6 +54,13 @@ const ImageViewer: React.FC<BaseViewerProps> = ({
       return () => {};
     }
 
+    // An empty (0-byte) File can never decode into an image — fail fast with a
+    // clear message instead of waiting for the <img> onerror (XYNE-61758).
+    if (source.size === 0) {
+      setImageError(true);
+      return () => {};
+    }
+
     const url = URL.createObjectURL(source);
     setImageError(false);
     setImageUrl(url);
@@ -200,8 +207,8 @@ const ImageViewer: React.FC<BaseViewerProps> = ({
     return (
       <div className='flex items-center justify-center h-full'>
         <div className='text-center text-muted'>
-          <p className='text-lg font-semibold mb-2 text-white'>Failed to load image</p>
-          <p className='text-sm text-muted-foreground'>The image could not be displayed</p>
+          <p className='text-lg font-semibold mb-2 text-white'>Couldn&apos;t load image</p>
+          <p className='text-sm text-muted-foreground'>The file may be missing or empty</p>
         </div>
       </div>
     );
--- a/apps/dashboard/src/services/clients/fileFetchService.ts
+++ b/apps/dashboard/src/services/clients/fileFetchService.ts
@@ -42,6 +42,9 @@ export const fetchFile = async (
       });
 
       const blob: Blob = response.data;
+      if (!blob || blob.size === 0) {
+        throw new Error('The file is empty or unavailable');
+      }
       return new File([blob], fileName, { type: mimeType });
     },
     staleTime: 10 * 60 * 1000,
@@ -122,7 +125,11 @@ export const createPreviewUrl = async (
         responseType: 'blob',
         signal,
       });
-      return response.data as Blob;
+      const blob = response.data as Blob;
+      if (!blob || blob.size === 0) {
+        throw new Error('The file is empty or unavailable');
+      }
+      return blob;
     },
     staleTime: 10 * 60 * 1000,
   });


apps/backend/src/controllers/attachmentController.ts
```diff
@@ -305,6 +305,19 @@ export class AttachmentController {
 
       const buffer = await service.getFileBuffer(filePath);
 
+      // Guard: a zero-byte / empty storage object must never be served as a valid
+      // 200. Otherwise the client builds a 0-byte File, the image decoder fails,
+      // and the viewer shows "Failed to load image" (XYNE-61758). Treat an empty
+      // object as a missing/corrupt file so the client can show a clear error and
+      // retry, instead of silently caching a broken preview.
+      if (!buffer || buffer.length === 0) {
```
XyneSpaces, line 313: **[major] R2** — This 404s **every** zero-byte object, not just the corrupt-GCS case from XYNE-61758: a legitimately empty upload (e.g. a user attaches an empty `.txt`/`.md`) now downloads as 404 "empty or unavailable". ↳ **Assumption to confirm:** storage can never hold a *valid* 0-byte attachment (e.g. upload middleware already rejects empty files). If that holds, please state it in the PR description and this is fine; otherwise either scope the hard-fail to preview/image consumers, or return a distinct status (e.g. 410/422) so dashboards can tell "corrupt in storage" apart from "valid empty file".

apps/backend/src/controllers/attachmentController.ts
```diff
@@ -305,6 +305,19 @@ export class AttachmentController {
...
+        logger.error(
+          `Empty storage object for attachment ${attachmentId} at path: ${filePath} (0 bytes)`,
```
XyneSpaces, line 315: **[major] R6** — CodeQL (`js/log-injection`) already flags this line: `filePath` (and `attachmentId`) are request/DB-derived values interpolated raw into the log string — newline-smuggling can forge log entries. ↳ FIX: use the structured logger form per the review guidelines (rule 279, `logger.error(message, metadata)`): `logger.error('Empty storage object for attachment', { attachmentId, filePath, size: buffer?.length ?? 0 })`. Same information, safe encoding, and queryable as fields.

apps/dashboard/src/services/clients/fileFetchService.ts
```diff
@@ -42,6 +42,9 @@ export const fetchFile = async (
       });
 
       const blob: Blob = response.data;
+      if (!blob || blob.size === 0) {
```
XyneSpaces, line 45: **[minor] R4** — This error is **deterministic**: a 0-byte blob will never succeed on retry. If the surrounding React Query options fall back to the default (`retry: 3` with backoff), every genuinely-empty attachment now costs up to 3 extra round-trips and several seconds before the viewer surfaces the error. ↳ Please confirm the retry config here / on the shared client, and short-circuit it for this error (e.g. `retry: false`, or `retry: (n, err) => isEmptyFileError(err) ? false : n < 3`). Assumption: the query-level retry config lives outside this hunk — flagging since the surrounding `staleTime` suggests these options are shared.

apps/dashboard/src/services/clients/fileFetchService.ts
```diff
@@ -122,7 +125,11 @@ export const createPreviewUrl = async (
         responseType: 'blob',
         signal,
       });
-      return response.data as Blob;
+      const blob = response.data as Blob;
+      if (!blob || blob.size === 0) {
```
XyneSpaces, line 129: **[nit] R8** — Identical guard duplicated in `fetchFile` and `createPreviewUrl`. ↳ Extract one helper — e.g. `assertNonEmptyBlob(blob): asserts blob is Blob` throwing an `EmptyFileError` — so both call sites (and any future retry-policy/type check, per my comment above) share a single source of truth.

apps/dashboard/src/components/FileViewer/ImageViewer.tsx
```diff
@@ -200,8 +207,8 @@ const ImageViewer: React.FC<BaseViewerProps> = ({
     return (
       <div className='flex items-center justify-center h-full'>
         <div className='text-center text-muted'>
-          <p className='text-lg font-semibold mb-2 text-white'>Failed to load image</p>
-          <p className='text-sm text-muted-foreground'>The image could not be displayed</p>
+          <p className='text-lg font-semibold mb-2 text-white'>Couldn&apos;t load image</p>
+          <p className='text-sm text-muted-foreground'>The file may be missing or empty</p>
```
XyneSpaces, line 211: **[nit] R8** — This copy now renders for **all** image-load failures, not just the empty-file case: a corrupt-but-nonzero JPEG still hits ` onerror` → `setImageError(true)` → "The file may be missing or empty", which misleads in the opposite direction. ↳ Either branch the message on the cause (empty-size guard vs decode error), or use wording that covers both (e.g. "The file is empty, missing, or corrupt"). Otherwise LGTM — fail-fast guard placement before `URL.createObjectURL` is right and returns a consistent no-op cleanup.

XyneSpaces (commented):
## Xyne Spaces Reviewer — inline findings

5 inline findings below (2 major, 1 minor, 2 nits). Full summary + reviewer recommendation follows as a top-level comment. Fix is directionally right — the empty-object → 404 guard mirrors the existing `fileExists` guards and closes the misleading-200 path. Main asks: confirm the two scope assumptions, clean up the CodeQL-flagged log line, and add a lock-in test.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1449

**Verdict:** Changes requested — no blockers; 2 assumptions to confirm, one CodeQL-flagged log line to fix, one lock-in test to add.
**Reviewed:** 3 files, 4 hunks · **Findings:** 0 blocking, 3 major, 1 minor, 2 nits (see inline review comments for details/fixes)

### Findings
| major | R2 | `apps/backend/src/controllers/attachmentController.ts:313` | Guard 404s **every** zero-byte object — also breaks downloading a *legitimately* empty upload (e.g. an empty `.txt`). Assumption to confirm: storage can never hold a valid 0-byte attachment; if so, note it in the description and this is fine. |
| major | R6 | `apps/backend/src/controllers/attachmentController.ts:315` | CodeQL `js/log-injection` already flags this new line: request/DB-derived `attachmentId`/`filePath` interpolated raw into the log string. ↳ Use `logger.error('Empty storage object for attachment', { attachmentId, filePath })` (guideline rule 279). |
| major | R7 | whole PR (no anchor) | New behavior (3 guards across BE+FE) ships with **no automated test**. POT videos satisfy rule 208's manual-proof bar, but the harness mirrors the logic rather than running e2e against a real 0-byte GCS object — a small `downloadAttachment` test (empty buffer → 404) + an ImageViewer 0-byte-file test would lock the behavior in. |
| minor | R4 | `apps/dashboard/src/services/clients/fileFetchService.ts:45` | The empty-blob throw is deterministic, but with React Query's default retry (3×, backoff) each empty attachment costs extra fetches + seconds of delay before the error UI. Confirm the retry config / mark the error non-retryable. |
| nit | R8 | `apps/dashboard/src/services/clients/fileFetchService.ts:129` | Same guard duplicated in `fetchFile` + `createPreviewUrl` — extract an `assertNonEmptyBlob`/`EmptyFileError` helper. |
| nit | R8 | `apps/dashboard/src/components/FileViewer/ImageViewer.tsx:211` | "The file may be missing or empty" now shows for **all** image failures, including corrupt-but-nonzero files — branch the message by cause or generalize the wording. |

### Recommended reviewer
**@milinjp (Milind Pithadia)** — ⚠️ *recommendation only*: this automation has no reviewer-request capability on GitHub (the `gh api …/requested_reviewers` POST is unavailable), so **please add Milind as a reviewer manually** (`gh api repos/juspay/xyne-spaces/pulls/1449/requested_reviewers -X POST -f "reviewers[]=milinjp"` or via the UI).

|-----------|--------------------------|-------------------|-----------------|-------------|
| @milinjp (Milind Pithadia, milind.pithadia@juspay.in) | Attachment download + ImageViewer area (per the fixing agent's handoff; per-file commit attribution is not computable with the available tooling — no path-filtered commits API) | No — CODEOWNERS covers only ACL/mutator paths, none of these files | **0** | Most recently active on `main` (commit 2026-09-01); verified login + email from commit metadata |
| Krishan Saini (login `krishan-saini` — *unverified*) | Same area (per fixing agent) | No | **0** | Not found in ~300 recent `main` commits; login + email unverifiable → could not confidently request |

Rationale: both candidates have zero pending review requests; Milind wins on verified authorship/activity in this area and a confirmed identity, so he's the single taggable pick.

---
No migration files touched → migration-safety gate not triggered (no QA tag). PR author (`XyneSpaces`, automation account) excluded from reviewer eligibility.

XyneSpaces:
@milinjp Requesting your review on this PR. This fixes empty attachment handling (XYNE-61758) — you own the attachment-download + ImageViewer area. Please take a look when you get a chance.

PR #1448, opened 2026-09-02
fix: shared links use canonical host so mobile deep links open in the app

XyneSpaces:
## Problem
Message/copy links shared from the dashboard are built on `app.spaces.xyne.juspay.net` (the dashboard's own origin). The mobile app only claims the app-less host `spaces.xyne.juspay.net` for iOS Universal Links / Android App Links, so the OS never routes an `app.spaces…` link into the app and opens it in the browser instead. The mobile-side normalisation added in #162 rewrites `app.spaces → spaces` only after the OS hands the link to the app, which never happens for an unregistered host — so the browser-open persists.

## Fix
`apps/dashboard/src/hooks/useShareableOrigin.ts` now normalises the origin: if the current host starts with `app.spaces.` it strips the leading `app.` (handles both prod `app.spaces.xyne.juspay.net` and sandbox `app.spaces.sandbox.xyne.juspay.net`); other hosts (e.g. `localhost:5173` in dev) are returned unchanged. Every copy-link consumer flows through this hook, so all shared links now use the canonical host the mobile app claims.

## Template
1. Problem Description: Shared message links open in the mobile browser instead of the app because they use the `app.spaces.*` host, which is not a registered universal/app-link host.
2. How the issue was identified: User report — opening a shared chat link on mobile launched the browser, not the app. Traced to link generation host vs the app's registered associated-domains/intent-filter host.
3. Solution implemented: Normalise the shareable-link origin to the canonical app-less host in `useShareableOrigin.ts` (strip leading `app.` from `app.spaces.*`).
4. Documentation: N/A
5. DB Alter Details: N/A
6. Data fix Details: N/A
7. Environment variable Changes: N/A
8. Testing: Verify a copied chat link uses `https://spaces.xyne.juspay.net/...`; in dev the host stays `localhost:5173`. Open the copied link on a production mobile build and confirm it opens in-app. Note: mobile `associatedDomains` is only set on the production variant, so links only open in-app on production builds.
9. Services to deploy: dashboard (web).
10. Main PR link (if against a release branch): N/A

PR #1188, opened 2026-08-28
feat: XYNE-59439 Add labels for calls

abhinavjuspay:
Added labels for calls 
POT : https://drive.google.com/file/d/1Zave2dumda0tBeEvPF5VmuF_7ONA7Jyp/view?usp=sharing

PR #1452, opened 2026-09-02
feat: XYNE-10935 open shared links in desktop app

XyneSpaces:
## What & why
Shared Xyne Spaces links (from Slack, pasted URLs, etc.) are plain `https` URLs, so the OS opens them in the browser. The desktop (Electron) app only registers the `xyne-spaces://` custom scheme, so nothing routed those https links to the installed app. This brings the web experience in line with mobile: when a user lands on the web app from an external context, it offers to re-open the same path in the desktop app.

## Changes (dashboard-only)
- **`apps/dashboard/src/utils/deepLink.ts`** (new): shared `DEEP_LINK_PROTOCOL`, `isElectronApp()`, `buildDeepLinkUrl()` (pathname + search only — the hash is intentionally dropped because the Electron `isSafeDeepLinkPath` guard rejects `#`), and `attemptAppLaunch()` with focus-loss detection (never redirects the browser; web page stays as fallback).
- **`apps/dashboard/src/components/OpenInAppGate/OpenInAppGate.tsx`** (new): dismissible "Open in desktop app?" banner. Shows only on genuine external entries and skips: already in Electron, iframes, auth/launch routes, reload/back-forward navigations, same-origin referrer, and once-per-tab. Choice (`always` / `never`) persisted in `localStorage` under `xyne.openInAppPreference`. Never auto-hands-off without a stored `always` preference.
- **`apps/dashboard/src/App.tsx`**: mount the gate at root.
- **`apps/dashboard/src/routes/LaunchScreen/LaunchScreen.tsx`**: consume the shared `DEEP_LINK_PROTOCOL` (removes duplicated derivation).

## Deliberately NOT changed
- No Electron changes: the generic-navigation handler in `apps/electron/src/services/deep-links.ts` already `show()` + `focus()`es the window on an incoming deep link.
- `DEEP_LINK_ROUTE_ALLOWLIST` left empty: routes are `/:workspaceId/...`, so the first path segment is a workspace cuid — populating the allowlist with route names would wrongly reject every real link.
- No backend/schema changes. Existing auth/callback, invite, and ask-ai deep-link families untouched.

## Proof of testing (POT)
- **Typecheck**: `pnpm --filter xyne-spaces-dashboard typecheck` passes clean.
- **Logic assertions** (Node): generated scheme URLs (e.g. `xyne-spaces://<wsid>/chat/<ch>/<cv>?focusThread=1`) pass a faithful port of the Electron `isSafeDeepLinkPath` guard; a hash-bearing path is rejected — proving why the hash is dropped. All 5 assertions pass.
- **Browser (dashboard dev server)**:
  - Banner appears on an external deep-link entry (empty referrer, navigation type `navigate`), captured at mount even before the app redirects to `/auth`.
  - "Don't ask again" persists `xyne.openInAppPreference=never` and removes the banner.
  - Re-entering the same link with `never` stored suppresses the banner.

## Known limitation
Exact scroll-to-message (`#origin=…&messageId=…`) is not carried into the desktop app because the Electron path guard rejects `#`; the channel/thread ids live in the path, so the app still opens the correct conversation. Carrying the hash would require coordinated Electron + renderer changes (deferred).

### DIFF
--- a/apps/dashboard/src/components/OpenInAppGate/OpenInAppGate.tsx
+++ b/apps/dashboard/src/components/OpenInAppGate/OpenInAppGate.tsx
@@ -1,0 +1,222 @@
+import { ReactElement, useEffect, useMemo, useRef, useState } from 'react';
+import { buildDeepLinkUrl, attemptAppLaunch, isElectronApp } from '../../utils/deepLink';
+
+/**
+ * Offers to hand a shared web link off to the installed desktop app.
+ *
+ * A xyne-spaces link shared in Slack (or anywhere outside the app) is a plain
+ * https URL, so the OS opens it in the browser. The desktop app only registers
+ * the xyne-spaces:// custom scheme, so nothing routes an https link to it. This
+ * gate closes that gap entirely on the web side: when someone lands on the web
+ * app from an external context, it offers to re-open the same path in the desktop
+ * app via the custom scheme, and can remember that choice.
+ *
+ * It NEVER auto-hands-off without a stored "always" preference, never runs inside
+ * the desktop app itself, and always leaves the user on the web page as the
+ * fallback — a user without the app installed is unaffected.
+ */
+
+const PREF_KEY = 'xyne.openInAppPreference'; // 'always' | 'never'
+const TAB_DISMISS_KEY = 'xyne.openInApp.dismissed'; // '1' for this tab
+
+type Preference = 'always' | 'never' | null;
+
+// Route segments where a hand-off would loop or fight auth. Matched against the
+// first two path segments (paths are /:workspaceId/<route>, but auth/launch
+// routes sit at the root without a workspace prefix).
+const BLOCKED_SEGMENTS = new Set([
+  'launch',
+  'login',
+  'auth',
+  'invite',
+  'external',
+  'newWindow',
+  'redirected',
+  'onboarding',
+  'slack-migration',
+]);
+
+function readPreference(): Preference {
+  try {
+    const v = localStorage.getItem(PREF_KEY);
+    return v === 'always' || v === 'never' ? v : null;
+  } catch {
+    return null;
+  }
+}
+
+function wasDismissedThisTab(): boolean {
+  try {
+    return sessionStorage.getItem(TAB_DISMISS_KEY) === '1';
+  } catch {
+    return false;
+  }
+}
+
+function markDismissedThisTab(): void {
+  try {
+    sessionStorage.setItem(TAB_DISMISS_KEY, '1');
+  } catch {
+    // ignore
+  }
+}
+
+function isReloadOrBackForward(): boolean {
+  try {
+    const nav = performance.getEntriesByType('navigation')[0] as
+      | PerformanceNavigationTiming
+      | undefined;
+    return nav?.type === 'reload' || nav?.type === 'back_forward';
+  } catch {
+    return false;
+  }
+}
+
+function hasSameOriginReferrer(): boolean {
+  if (!document.referrer) return false; // no referrer = external / direct (Slack, pasted link)
+  try {
+    return new URL(document.referrer).origin === window.location.origin;
+  } catch {
+    return false;
+  }
+}
+
+function isBlockedPath(pathname: string): boolean {
+  if (pathname === '/' || pathname === '') return true;
+  const segments = pathname.split('/').filter(Boolean);
+  return segments.slice(0, 2).some(seg => BLOCKED_SEGMENTS.has(seg));
+}
+
+/** One-time evaluation of whether the gate may act on this page load. */
+function computeEligibility(): boolean {
+  if (typeof window === 'undefined') return false;
+  if (isElectronApp()) return false; // already in the desktop app
+  if (window.top !== window.self) return false; // embedded iframe (e.g. SDLC surface)
+  if (isBlockedPath(window.location.pathname)) return false;
+  if (isReloadOrBackForward()) return false; // only on genuine external entries
+  if (hasSameOriginReferrer()) return false; // in-app navigation, not an external open
+  if (wasDismissedThisTab()) return false;
+  return true;
+}
+
+const OpenInAppGate = (): ReactElement | null => {
+  const eligible = useMemo(computeEligibility, []);
+  const [preference, setPreference] = useState<Preference>(() => readPreference());
+  const [showBanner, setShowBanner] = useState(false);
+  const cleanupRef = useRef<(() => void) | null>(null);
+
+  const deepLinkUrl = useMemo(
+    () =>
+      buildDeepLinkUrl({
+        pathname: window.location.pathname,
+        search: window.location.search,
+      }),
+    [],
+  );
+
+  useEffect(() => {
+    if (!eligible) return undefined;
+
+    if (preference === 'never') return undefined;
+
+    if (preference === 'always') {
+      markDismissedThisTab();
+      cleanupRef.current = attemptAppLaunch(deepLinkUrl);
+      return () => cleanupRef.current?.();
+    }
+
+    // No stored preference: show the banner, never auto-launch.
+    setShowBanner(true);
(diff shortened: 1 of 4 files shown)


XyneSpaces (commented):
Xyne Spaces Reviewer — inline findings on the specifics below; a full summary with the findings table and reviewer recommendation follows as a top-level comment. No blocking issues; these are targeted fix-ups.

apps/dashboard/src/components/OpenInAppGate/OpenInAppGate.tsx
```diff
@@ -0,0 +1,222 @@
...
+    cleanupRef.current = attemptAppLaunch(deepLinkUrl);
+    setShowBanner(false);
+  };
+
+  const openAlways = (): void => {
+    try {
+      localStorage.setItem(PREF_KEY, 'always');
+    } catch {
+      // ignore
+    }
```
XyneSpaces, line 149: **[major] R2** — `openAlways` fires the external-protocol launch **twice**: here (manual `attemptAppLaunch`), and again when `setPreference('always')` re-triggers the effect, whose `preference === 'always'` branch calls `attemptAppLaunch(deepLinkUrl)` a second time. Two `window.location.href = xyne-spaces://…` navigations back-to-back → risk of a double OS/browser "Open app?" prompt (Chrome usually suppresses the second; Firefox/Safari may not), and the cleanup for the first timer/listener is silently overwritten in `cleanupRef`. Suggested fix: make the effect the single owner of launching — in `openAlways`, just persist `localStorage`, `markDismissedThisTab()`, `setPreference('always')`, `setShowBanner(false)` and let the effect run once.

apps/dashboard/src/utils/deepLink.ts
```diff
@@ -0,0 +1,77 @@
...
+  let lostFocus = false;
+  const onBlur = (): void => {
+    lostFocus = true;
+  };
+  window.addEventListener('blur', onBlur);
+
+  // Triggering the scheme on a hidden anchor avoids replacing the current
+  // document if the OS has no handler (some browsers navigate on location.href).
+  try {
+    window.location.href = url;
```
XyneSpaces, line 61: **[nit] R8** — comment/code mismatch: the comment above says "Triggering the scheme on a hidden anchor avoids replacing the current document", but the code sets `window.location.href` directly — no hidden anchor involved (setting `location.href` to an unknown scheme does NOT navigate the page in modern browsers, so the behavior is fine; the comment just describes a different mechanism). Also line 6: "Mirrors … and the derivation in routes/LaunchScreen/LaunchScreen.tsx — keep the **three** in step" — this PR deletes the LaunchScreen derivation, so there are now only two sources (electron `config.ts` + this file). Suggest fixing both comments to match the code.

apps/dashboard/src/components/OpenInAppGate/OpenInAppGate.tsx
```diff
@@ -0,0 +1,222 @@
...
+ * https URL, so the OS opens it in the browser. The desktop app only registers
+ * the xyne-spaces:// custom scheme, so nothing routes an https link to it. This
+ * gate closes that gap entirely on the web side: when someone lands on the web
+ * app from an external context, it offers to re-open the same path in the desktop
+ * app via the custom scheme, and can remember that choice.
+ *
+ * It NEVER auto-hands-off without a stored "always" preference, never runs inside
+ * the desktop app itself, and always leaves the user on the web page as the
+ * fallback — a user without the app installed is unaffected.
+ */
```
XyneSpaces, line 17: **[minor] R8 / guidelines#249-261** — the `localStorage`/`sessionStorage` keys (`xyne.openInAppPreference`, `xyne.openInApp.dismissed`) are declared inline in the component. The review guidelines ask for storage keys to live in the centralized settings/constants file (or be exported from the owning module) so other consumers share one source of truth. Assumption: I haven't verified whether the dashboard already has such a constants module — if it does, please move these there; if it doesn't, exporting them from `utils/deepLink.ts` would at least keep the two keys together.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1452

**Verdict:** Changes requested — no blockers; 3 major fix-ups (1 is an Electron-side gap verified on `main`, not in this diff)
**Reviewed:** 4 files, every hunk · **Findings:** 0 blocking, 3 major, 2 minor/nit
**Scope check:** dashboard-only; no migrations touched → migration-safety gate not triggered.

### Findings
| major | R2 | `apps/dashboard/src/components/OpenInAppGate/OpenInAppGate.tsx:149` | "Always" fires the scheme launch twice — `openAlways` calls `attemptAppLaunch` directly, then `setPreference('always')` re-runs the effect whose `always` branch launches again → possible double OS/browser prompt; first timer's cleanup is overwritten. Fix: let the effect own the launch (persist pref + set state only). |
| major | R7 / guidelines#204 | PR-wide | New `deepLink.ts` utilities + the gate component ship with **no committed tests** — POT assertions were run ad-hoc in Node but aren't in the repo. Add unit tests (at minimum: `buildDeepLinkUrl` path/search/hash cases, `isElectronApp`, eligibility matrix). |
| major | R2 (integration; verified on `main`, **outside this diff**) | `apps/electron/src/services/deep-links.ts` — generic-nav case | Verified fact: unlike `ask-ai`/`invite` which poll up to 10s for a window, the generic branch drops the deep link when `mainWindow` is null (`if (mainWindow) {...}` after validation). A cold-start `open-url` (this PR's primary flow: user clicks "Open in app" with the app NOT running) can therefore be validated then silently dropped — app opens, right conversation never navigates to. Suggest a follow-up: wait-for-window parity in the generic branch. |
| minor | R8 / guidelines#249-261 | `OpenInAppGate.tsx:17-18` | `localStorage`/`sessionStorage` keys inline instead of a centralized settings-constants module (assumption — didn't verify whether the dashboard has one; worst case export them from `utils/deepLink.ts`). |
| nit | R8 | `deepLink.ts:61` (+ doc block at :5-6) | Comment says "hidden anchor" but code sets `window.location.href`; also "keep the three in step" — the LaunchScreen derivation was deleted by this PR, so only two sources remain. |

**Deliberate-design checks that passed:** protocol constants match `apps/electron/src/app/config.ts` for all 3 flavors ✓; `isSafeDeepLinkPath` does reject `#`, so dropping the hash is correct ✓; generic handler does `show()`+`focus()` when the window exists ✓; `window.electronAPI` is typed in `apps/dashboard/src/types/electron.d.ts` ✓; `/svgs/xyne.svg` exists ✓; eligibility gates (Electron self-skip, iframe, reload/back-forward, same-origin referrer, auth paths) all look sane.

### Recommended reviewer
**@vinit717 (Vinit Khandal)** — recommended. ⚠️ I could **not** post a formal review request: this harness has no request-reviewers capability (`gh pr edit 1452 --add-reviewer vinit717` does it manually). Tagging here instead.

|-----------|--------------------------|-------------------|-----------------|-------------|
| **@vinit717** | None of the 4 files directly (2 are new; commits-by-path not computable with available tooling) — but authored the adjacent deep-link PR #1448 (shared links → open-in-app canonical host) | No (`.github/CODEOWNERS` covers only backend/shared ACL paths — **no entry exists for** `apps/dashboard/**`) | 0 — *no open PR in the repo currently has any requested reviewers*, so load can't differentiate | **Last seen in Spaces today (3 Sep 2026)** — most recently active |
| @ameer030 | Adjacent: merged PR #1425 (SDLC deep links) | No | 0 (see above) | Last seen ~2 months ago in Spaces |
| @samitbarai | No deep-link work found; active dashboard-area approver (#1445, #1188, #1394) | No | 0 | Last seen ~2 weeks ago |

Rationale: with no CODEOWNER eligibility set for these paths and uniform review workload, code familiarity decides — @vinit717 is actively working in exactly this feature family (shared https links → custom-scheme deep links), and is the candidate most recently active in Spaces. (Author @anuragdvd excluded; bot accounts excluded.)

PR #1446, opened 2026-09-02
fix: correct call duration for scheduled calls with multiple sessions (rejoin overwrote startedAt)

XyneSpaces:
## 1. Problem Description
A scheduled call's duration is derived at read time as `endedAt - startedAt` (there is no stored duration column). Call `b94efffd-0b12-494f-8857-a16764d6b6aa` displayed a duration of **18 seconds** while its transcript spanned **~49 minutes**.

## 2. How the issue was identified
Root-caused via the LiveKit room-webhook timeline for that call:
1. `12:59:29` first human joins → call activates, `startedAt` correctly set to the real first-join time.
2. `13:49:50` last human leaves → within the scheduled window, so the call reverts to `SCHEDULED` and `endedAt` is written.
3. `13:50:13` someone briefly re-joins → `activateScheduledCall` takes the **rejoin branch**, which overwrote `startedAt` with the reconnect time (`13:50:15.5`).
4. `13:50:33` room finishes → `endedAt` rewritten to `13:50:33.586`.

Final stored state `startedAt=13:50:15.5`, `endedAt=13:50:33.586` → **18s**, even though the real session ran ~50 min. The transcript was correct; the duration was corrupted. The rejoin branch was treating a reconnect as a brand-new call start.

## 3. Solution implemented
In `activateScheduledCall`'s rejoin branch (`apps/backend/src/database/repositories/callRepository.ts`):
- Stop overwriting `startedAt` on rejoin. It was already set to the real first-join time by the first-activation branch (which also corrects the schema's schedule-time `@default(now())`), so the rejoin must preserve it. Duration now spans the whole first-join → last-leave envelope.
- Clear the stale `endedAt` that the previous revert-to-SCHEDULED wrote, since the call is live again; the final end path rewrites `endedAt` to the true last-leave time.

The first-join branches are unchanged and still set `startedAt = now`. No schema change, no data migration.

## 4. Documentation
N/A

## 5. DB Alter Details (if any)
N/A — no schema change.

## 6. Data fix Details (if any)
Optional one-off backfill: for calls whose stored `startedAt` is later than their first participant `joinedAt`, reset `startedAt = MIN(callParticipant.joinedAt)`. Not required for the code fix.

## 7. Environment variable Changes (if any)
N/A

## 8. Testing
Added `apps/backend/src/database/repositories/callRepository.activateScheduledCall.test.ts` (2 tests, passing):
- rejoin branch does NOT write `startedAt` and clears `endedAt`.
[...]

PR #1181, opened 2026-08-28
XYNE-60693 monthly recurrence  scheduled messages

sirajshaik-code:
## Summary

Adds **monthly** recurrence to scheduled channel messages — day-of-month, nth-weekday (e.g. 2nd Tuesday), and last-day — alongside the existing weekly schedules. Uses cron `L` / `#` syntax that Bull's `cron-parser` already supports, so no new scheduling engine.

## Data model (compact, backward compatible)

No `frequency` column. Monthly is signalled by `daysOfWeek = "-"`; two nullable columns carry the rest:

| Column | Meaning |
|---|---|
| `daysOfWeek` | weekly day list `"0,1,2,3,4,5,6"` (0=Sun..6=Sat), or `"-"` = monthly |
| `monthlyMode` | `DAY_OF_MONTH` \| `NTH_WEEKDAY` (null when weekly) |
| `monthlyValue` | DAY_OF_MONTH → `1..28`, or `-1` = last day. NTH_WEEKDAY → `ordinal*10 + weekday` (ordinal `1..4` or `5`=last; weekday `0..6`). e.g. 2nd Tuesday = `22`, last Friday = `55` |

The migration is additive + nullable (`ADD COLUMN monthlyMode, monthlyValue`), so **existing weekly rows are untouched** (`monthlyMode`/`monthlyValue` = NULL, `daysOfWeek` unchanged) and their live Bull jobs keep firing. Adding nullable columns is a metadata-only change in Postgres, so it's safe at any table size.

## Changes

- **Backend** — `buildCronPattern` decodes `monthlyValue` into a cron pattern; `normalizeSchedule` validates weekly/monthly input across the create + update controllers.
- **Dashboard** — the modal gains a Frequency / Monthly selector (day-of-month 1–28, nth-weekday, last-day), rendered as segmented toggles + day-of-week pills; it encodes on submit and decodes on edit. The card renders the monthly schedule.

## Verification

Cron output validated against `cron-parser` for every mode (weekly, day-of-month, last-day, 2nd Tuesday, last Friday). Backend and dashboard typecheck clean; dashboard Prettier passes.

PR #1453, opened 2026-09-02
feat: add CallSession table as immutable source of truth for call duration

XyneSpaces:
## Summary
Durable follow-up to #1446 (a 49-minute call reported as 18 seconds). #1446 is the contained 1-line symptom fix; this PR removes the whole class of bug.

Framed through the two lenses from https://kolu.dev/blog/hickey-lowy/ (as requested in-thread):
- **Hickey (structural / simplicity):** `Call.startedAt` *complected* three separate ideas — the schema default, the true first start, and each per-rejoin session start. Because one column carried all three, a rejoin (`activateScheduledCall`) overwrote it and collapsed the reported duration to the last session.
- **Löwy (temporal / volatility):** duration accounting runs on the fast **participant-lifecycle clock**, while call status runs on the slow **call-state clock**. A status-transition write (rejoin) corrupted a value owned by the accounting clock — two different rates of change coupled onto one row.

Both lenses go quiet when the two concerns are separated into their own rows.

## What changed
- **New `CallSession` model / `call_sessions` table** — append-only, one immutable row per LiveKit activation. `startedAt` is written once and never mutated; `endedAt` is set exactly once on close.
- `activateScheduledCall` now **INSERTs a session row** and no longer writes `startedAt` on the rejoin branch (the bug). It also clears any stale `Call.endedAt` on all three activation branches.
- **`Call.startedAt`/`endedAt` becomes a projection** — `MIN(startedAt)..MAX(endedAt)` over sessions, re-derived only on session close via a new private `closeOpenCallSession(tx, callId, endedAt)`. This is the **sole writer** of `Call.startedAt` after activation, so a status write can never corrupt duration again. It is idempotent and self-heals already-corrupted rows. End handlers hand the corrected `startedAt` to `updateCallSystemMessageIfNeeded` / `syncArtifactLifecycle` in-transaction.
- **Transcript / summary / artifacts stay singular per `Call`** — `CallSession` is added *alongside* `Call`, not replacing it, so a call scheduled 6–7pm with multiple sessions keeps one transcript and one summary. (This directly addresses the concern raised in-thread about singular artifacts.)

## Data model notes
- **No DB-level FK** on `call_sessions` — repo uses Prisma `relationMode = "prisma"` (app-enforced relations); `@@index([callId])` + `@@index([callId, endedAt])` cover lookups.
- `CallSession` is **backend-only** — it is deliberately NOT added to the generated Zero schema, so it is not synced to clients (no client-facing surface, no extra sync cost).
- **Idempotent backfill** in the migration: one session per existing call mirroring its current `startedAt`/`endedAt`, skipping any call that already has a session (`WHERE NOT EXISTS`).

## Validation
- `prisma db push` succeeds against the dev DB; `prisma generate` produces the `callSession` client accessor.
- New regression test `callSessionProjection.test.ts` drives `closeOpenCallSession` with a fake transaction and asserts the projected envelope is the full first-join..last-leave span (**49 min**, not the last session's 9 min and not 18 s) across a rejoin, plus an idempotent no-op when no session is open. Both tests pass.
- `tsc --noEmit` is clean for the changed files (the one repo-wide error in `acl-factory.ts` is pre-existing and unrelated).

## Rollout
1. Deploy migration (creates table + indexes, runs idempotent backfill).
2. Deploy backend with the new session write + projection.
No feature flag needed — the projection self-heals existing rows on the next session close.

## Reviewer
Requesting **@harshpreet931** (owns `callRepository.ts`).

### DIFF
--- a/apps/backend/prisma/migrations/20260902230000_add_call_sessions/migration.sql
+++ b/apps/backend/prisma/migrations/20260902230000_add_call_sessions/migration.sql
@@ -1,0 +1,33 @@
+-- CreateTable: append-only per-activation session rows for a Call.
+-- Call.startedAt/endedAt becomes a projection of these rows (MIN start .. MAX end),
+-- so a status-transition write (e.g. a rejoin) can no longer corrupt call duration.
+CREATE TABLE "public"."call_sessions" (
+    "workspaceId" TEXT NOT NULL,
+    "id" TEXT NOT NULL,
+    "callId" TEXT NOT NULL,
+    "startedAt" TIMESTAMP(3) NOT NULL,
+    "endedAt" TIMESTAMP(3),
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    CONSTRAINT "call_sessions_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateIndex
+CREATE INDEX "call_sessions_callId_idx" ON "public"."call_sessions"("callId");
+
+-- CreateIndex: fast open-session lookup (endedAt IS NULL)
+CREATE INDEX "call_sessions_callId_endedAt_idx" ON "public"."call_sessions"("callId", "endedAt");
+
+-- NOTE: no DB-level FK — the repo uses Prisma relationMode="prisma"
+-- (app-enforced relations). The call_sessions_callId_idx above covers lookups.
+
+-- Backfill: one session per existing call mirroring its current startedAt/endedAt,
+-- so historical calls keep a correct (uncorrupted-going-forward) envelope. Idempotent:
+-- skips any call that already has a session row.
+INSERT INTO "public"."call_sessions" ("id", "callId", "workspaceId", "startedAt", "endedAt", "createdAt")
+SELECT gen_random_uuid()::text, c."id", c."workspaceId", c."startedAt", c."endedAt", CURRENT_TIMESTAMP
+FROM "public"."calls" c
+WHERE c."startedAt" IS NOT NULL
+  AND NOT EXISTS (
+    SELECT 1 FROM "public"."call_sessions" s WHERE s."callId" = c."id"
+  );
--- a/apps/backend/src/database/repositories/callRepository.ts
+++ b/apps/backend/src/database/repositories/callRepository.ts
@@ -301,6 +301,46 @@ export class CallRepository {
   }
 
   /**
+   * Close the most recent still-open CallSession for a call (set its endedAt),
+   * then re-project Call.startedAt/endedAt from the immutable session rows
+   * (MIN startedAt .. MAX endedAt). This is the ONLY writer of Call.startedAt
+   * after activation, so the denormalized envelope is always first-join ->
+   * last-leave and can never be corrupted by a status-transition write.
+   * Idempotent: with no open session it just re-projects. Returns the projected
+   * envelope so callers can hand the corrected startedAt to duration consumers
+   * within the same transaction (before the projection is re-read from the DB).
+   */
+  private async closeOpenCallSession(
+    tx: Prisma.TransactionClient,
+    callId: string,
+    endedAt: Date,
+  ): Promise<{ startedAt: Date | null; endedAt: Date | null }> {
+    const open = await tx.callSession.findFirst({
+      where: { callId, endedAt: null },
+      orderBy: { startedAt: 'desc' },
+      select: { id: true },
+    });
+    if (open) {
+      await tx.callSession.update({ where: { id: open.id }, data: { endedAt } });
+    }
+    const agg = await tx.callSession.aggregate({
+      where: { callId },
+      _min: { startedAt: true },
+      _max: { endedAt: true },
+    });
+    if (agg._min.startedAt) {
+      await tx.call.update({
+        where: { id: callId },
+        data: {
+          startedAt: agg._min.startedAt,
+          ...(agg._max.endedAt ? { endedAt: agg._max.endedAt } : {}),
+        },
+      });
+    }
+    return { startedAt: agg._min.startedAt, endedAt: agg._max.endedAt };
+  }
+
+  /**
    * Mirror a call transition onto the slash-command artifact that started it.
    * A no-op for every other call, which is why it can sit directly on the
    * shared end-of-call paths without altering their behavior.
@@ -1057,15 +1097,20 @@ export class CallRepository {
           where: { id: call.id },
           data: { status: finalStatus, endedAt: leftAt },
         });
+        // Close the open session and re-project Call.startedAt/endedAt from the
+        // immutable sessions, so downstream duration (system message, artifact,
+        // analytics) reads the full first-join -> last-leave envelope.
+        const envelope = await this.closeOpenCallSession(tx, call.id, leftAt);
+        const callForEnd = { ...call, startedAt: envelope.startedAt ?? call.startedAt };
         shouldEndCall = finalStatus === CallStatus.ENDED;
         if (shouldEndCall) {
           await refreshCallParticipantPreview(tx, call.id);
-          await this.syncArtifactLifecycle(tx, call, MessageArtifactStatus.COMPLETED, leftAt);
+          await this.syncArtifactLifecycle(tx, callForEnd, MessageArtifactStatus.COMPLETED, leftAt);
         }
 
         // Update system message whether the call is fully ended or just rescheduled
         messageUpdated = await updateCallSystemMessageIfNeeded({
-          call,
+          call: callForEnd,
           callId: callExternalId,
           endedAt: leftAt,
           tx,
@@ -1102,6 +1147,7 @@ export class CallRepository {
 
       let shouldEndCall = false;
       let messageUpdated = false;
+      let callForEnd = call;
 
       // Check and update call status if not already ended
       if (call.status !== CallStatus.ENDED) {
@@ -1115,6 +1161,10 @@ export class CallRepository {
           where: { id: call.id },
           data: { status: finalStatus, endedAt },
         });
+        // Close the open session and re-project Call.startedAt/endedAt from the
+        // immutable sessions (see closeOpenCallSession).
+        const envelope = await this.closeOpenCallSession(tx, call.id, endedAt);
+        callForEnd = { ...call, startedAt: envelope.startedAt ?? call.startedAt };
         shouldEndCall = finalStatus === CallStatus.ENDED;
         if (shouldEndCall) {
           await refreshCallParticipantPreview(tx, call.id);
@@ -1122,7 +1172,7 @@ export class CallRepository {
 
         // Update system message whether the call is fully ended or just rescheduled
         messageUpdated = await updateCallSystemMessageIfNeeded({
-          call,
+          call: callForEnd,
           callId: callExternalId,
           endedAt,
           tx,
@@ -1130,7 +1180,7 @@ export class CallRepository {
       } else {
         // Call already ended - still try to update system message if needed
         messageUpdated = await updateCallSystemMessageIfNeeded({
-          call,
+          call: callForEnd,
           callId: callExternalId,
           endedAt,
           tx,
@@ -1138,7 +1188,7 @@ export class CallRepository {
       }
 
       if (call.status === CallStatus.ENDED || shouldEndCall) {
-        await this.syncArtifactLifecycle(tx, call, MessageArtifactStatus.COMPLETED, endedAt);
+        await this.syncArtifactLifecycle(tx, callForEnd, MessageArtifactStatus.COMPLETED, endedAt);
       }
 
       // Clear conversation.callId when call ends (for conversation calls)
@@ -1288,6 +1338,7 @@ export class CallRepository {
           data: {
             status: CallStatus.ACTIVE,
             startedAt: now,
+            endedAt: null,
             lastActivityAt: now,
             updatedAt: now,
             // Merge (not replace) so calendar-derived fields already on the call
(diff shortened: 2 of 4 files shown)


apps/backend/prisma/migrations/20260902230000_add_call_sessions/migration.sql
```diff
@@ -0,0 +1,33 @@
...
+-- CreateIndex: fast open-session lookup (endedAt IS NULL)
+CREATE INDEX "call_sessions_callId_endedAt_idx" ON "public"."call_sessions"("callId", "endedAt");
+
+-- NOTE: no DB-level FK — the repo uses Prisma relationMode="prisma"
+-- (app-enforced relations). The call_sessions_callId_idx above covers lookups.
+
+-- Backfill: one session per existing call mirroring its current startedAt/endedAt,
+-- so historical calls keep a correct (uncorrupted-going-forward) envelope. Idempotent:
+-- skips any call that already has a session row.
+INSERT INTO "public"."call_sessions" ("id", "callId", "workspaceId", "startedAt", "endedAt", "createdAt")
```
XyneSpaces, line 27: **[🔴 DANGEROUS] public.call_sessions — INSERT…SELECT backfill inside the migration file**

This statement is a table-wide data backfill executed inside the migration transaction: it scans every row of the populated `calls` table and inserts one session per call. That is the exact pattern behind our past migration incidents on the shared pre-prod + prod Postgres — the migration holds the Prisma advisory lock for the full scan, the long transaction pins a snapshot (blocking vacuum) and bursts WAL. The `WHERE NOT EXISTS` guard makes it idempotent on re-run; it does not make it safe to run at deploy time on a populated table. (`call_sessions` itself is new, empty and not synced to Zero clients, so the client-sync axis is fine — the risk is purely the in-file backfill.)

↳ FIX: keep this migration DDL-only (CREATE TABLE + the two indexes are fine). Move the backfill into a separate idempotent batched job — same `WHERE NOT EXISTS` predicate plus `LIMIT`/keyset batching — and run it off-peak after deploy. The historical rows do matter (already-ended calls never get another session close, so they cannot self-heal later), so the backfill should still happen — just not inside the migration file.

apps/backend/src/database/repositories/callRepository.ts
```diff
@@ -1376,6 +1431,17 @@ export class CallRepository {
           });
         }
       }
+
+      // Append an immutable session row for THIS activation. Sessions are the
+      // append-only source of truth for call timing; Call.startedAt/endedAt is a
+      // projection maintained on session close (see closeOpenCallSession).
+      await tx.callSession.create({
```
XyneSpaces, line 1438: **[🟠 Major] R4 (idempotency) — session append has no duplicate guard; concurrent activations leak a forever-open session**

The `call.status === CallStatus.SCHEDULED` check happens before this transaction opens, so two concurrent `participant_joined` deliveries — two participants joining a scheduled call near-simultaneously, or one webhook redelivered — can both read SCHEDULED and both reach this `create`. Result: two open sessions for one activation. `closeOpenCallSession` (line 318) closes only the newest (`findFirst` + `orderBy startedAt: desc`), so the older row stays open forever; a later `handleRoomFinished` (which runs while `status !== ENDED`) then closes that leaked row at *room-finish* time, inflating `Call.endedAt`. The projection stays monotone so `startedAt` survives, but the "one immutable row per activation, `endedAt` set exactly once" invariant silently breaks.

Suggested fix: the table is brand-new in this same migration, so a partial unique index is free — `CREATE UNIQUE INDEX call_sessions_one_open_per_call ON call_sessions("callId") WHERE "endedAt" IS NULL;` — and catch Prisma `P2002` here (treat as "another activation won the race"). Alternative: serialize per call with `pg_advisory_xact_lock` (there is precedent in `updateRecordingParticipants`).

apps/backend/src/database/repositories/callRepository.ts
```diff
@@ -1343,8 +1395,11 @@ export class CallRepository {
         await tx.call.update({
           where: { id: call.id },
           data: {
+            // NOTE: startedAt is deliberately NOT written here. A rejoin is a new
```
XyneSpaces, line 1398: **[🟡 Medium] R2 (correctness) — a rejoin with zero prior sessions would rewrite `startedAt` to the rejoin time**

This branch deliberately no longer writes `startedAt`, and the projection derives it as `MIN(sessions.startedAt)` — correct only if a session exists for the first activation. Calls activated through `activateScheduledCall` are anchored (the first-join branches write `startedAt` and append a session together, and the migration backfills every pre-existing call). The exposure is calls activated **outside** this function: instant/channel/automation calls created `ACTIVE` with `startedAt` by `createCallWithParticipantsAndMessage`, which does not append a session row (and `callController.ts` also references `status: CallStatus.ACTIVE`). If such a call can revert to `SCHEDULED` — it would need an `endsAt` still in the future at last leave, per `handleParticipantLeave` — then a rejoin appends the call's first-ever session and the projection rewrites `startedAt` to the rejoin time: the original start is lost, i.e. the same bug class this PR eliminates, via another entry path.

*(Assumption to confirm: whether session-less ACTIVE calls can carry a future `endsAt` — I could not fully read those creation paths in this 81 KB file. If they cannot, this is unreachable and can be closed as a doc nit.)*

Suggested fix either way, so the invariant holds for every entry path: in this branch, if the call has zero sessions but `startedAt` is set, seed a closed session `{ startedAt: call.startedAt, endedAt: call.endedAt }` before appending the open one.

apps/backend/src/database/repositories/callSessionProjection.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    },
+  };
+  return { tx, callUpdates };
+}
+
+const repo = new CallRepository();
+const close = (tx: any, callId: string, endedAt: Date) =>
+  (repo as any).closeOpenCallSession(tx, callId, endedAt);
+
+describe('CallRepository.closeOpenCallSession (call-duration projection)', () => {
```
XyneSpaces, line 82: **[🟡 Medium] R7 (tests) — the rejoin-branch behavior that actually fixes the bug is untested**

Both tests drive `closeOpenCallSession` through the fake tx — the projection math is well covered. But the fix itself lives in `activateScheduledCall`'s rejoin branch (no `startedAt` write, `endedAt` cleared, one session appended), and nothing asserts (a) that a rejoin does NOT touch `Call.startedAt`, or (b) that exactly one session row is appended per activation. The fake-tx harness built here is exactly the right tool — consider extending it to drive the rejoin branch (mock `tx.call.update`, count `tx.callSession.create` calls, assert `startedAt` is absent from the update payload). That is the test that would catch a reintroduction of the 18-second regression.

XyneSpaces (commented):
Automated review — 4 inline findings (1 blocking migration-safety, 1 major idempotency, 2 medium). Full findings table, migration-safety block and reviewer recommendation are in the top-level summary comment. Submitted as COMMENT only — no approval or merge action taken.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1453

**Verdict:** Changes requested (1 blocking migration finding)
**Reviewed:** 4 files, 12 hunks · **Findings:** 1 blocking, 1 major, 2 medium

The design is sound and well-argued — append-only sessions + projection-on-close is the right shape, the first-join branches anchor `startedAt` correctly, the `callForEnd` fallbacks are safe for session-less calls, and keeping `CallSession` out of the Zero schema is the right call. The in-migration backfill is the one thing I'd hold the merge for.

### Findings
| blocking | Migration Safety (R3) | `migration.sql:27` | 🔴 INSERT…SELECT backfill of all `calls` inside the migration file — long lock / WAL burst on the shared pre-prod + prod DB; move it to a separate batched job |
| major | R4 | `callRepository.ts:1438` | session append has no duplicate guard — concurrent `participant_joined` webhooks create two open sessions; the older leaks open forever and inflates `endedAt` when `handleRoomFinished` closes it late |
| medium | R2 | `callRepository.ts:1398` | rejoin with zero prior sessions rewrites `startedAt` to rejoin time for calls activated outside `activateScheduledCall` (assumption-labeled — see inline comment) |
| medium | R7 | `callSessionProjection.test.ts:82` | tests cover the projection helper, not the rejoin-branch behavior that actually fixes the bug |

### 🛢️ Migration Safety — unsafe operations found
QA contacts: Yash Gupta (yash.gupta@juspay.in — migration verification), Abhishek Verma (abhishek.verma@juspay.in)

| Migration | Tier | Statement (schema.table — op) | What it breaks | Fix |
|-----------|------|-------------------------------|----------------|-----|
| 20260902230000_add_call_sessions | 🔴 DANGEROUS | public.calls → public.call_sessions — INSERT…SELECT backfill | table-wide backfill inside the migration transaction on the shared pre-prod + prod DB: holds the migration lock for a full scan of `calls`, pins a snapshot (blocks vacuum), bursts WAL. `call_sessions` itself is new, empty and not Zero-synced — the risk is purely the in-file backfill | keep the migration DDL-only; move the backfill to a separate batched idempotent job (same `WHERE NOT EXISTS` + `LIMIT`), run off-peak |

Deploy stance for QA: if this ships as-is, the migration holds the DB for the duration of the `calls` scan at deploy time — verify on sandbox first and check the row count of `calls` in prod before deciding to run it inline.

### Recommended reviewer
**@harshpreet931** (requested) — owns `callRepository.ts`, where 3 of this PR's 4 substantive changes live; named as the file's owner in the PR description; zero pending GitHub review requests; most recently active in Spaces (#xyne-spaces, Sep 1–2).

| Candidate | Owns (files in this PR) | CODEOWNERS match? | Pending reviews | Tie-breaker |
| @harshpreet931 | callRepository.ts (call-domain: also authored open #1262, call hover card) | no | 0 (verified across the 100 newest open PRs) | most recently active in Spaces |
| (others) | commits-by-path aggregation not computable with this reviewer's direct GitHub toolset (no path-filtered commit listing available) — not invented | — | — | — |

Note: this repo's CODEOWNERS covers only ACL-sensitive paths (`database/acl/`, `database/tenant/`, `zero/…`) — none of this PR's files match — so the ranking uses file ownership + Spaces activity, with the fallback noted above.

— posted by Xyne Spaces Reviewer as a COMMENT-type review; no approval / request-changes / merge actions taken.

XyneSpaces:
**Reviewer request correction (Xyne Spaces Reviewer):** the summary above marks @harshpreet931 as "requested", but the programmatic reviewer request could not be submitted (no reviewer-assignment API available to this automation). Please add him via the GitHub UI — *Reviewers → harshpreet931* — or he can pick it up from this mention.

XyneSpaces:
**Follow-up — Xyne Spaces Reviewer** (resolves the assumption in the R2 finding; completes the reviewer table with commits-by-path data)

**R2 @ `callRepository.ts:1398` — resolved as unreachable; downgrade to optional hardening.** Verified across every path that creates/activates a call as `ACTIVE` outside `activateScheduledCall`:
- `createCallWithParticipantsAndMessage` creates calls already `ACTIVE` → the `participant_joined` gate (`call.status === SCHEDULED`) skips activation;
- note-taker calls are `HEADLESS`, routed entirely to `noteTakerWebhookController`;
- `calendarCallStore`'s upsert only preserves an existing `ACTIVE` room → same gate skips;
- the Zero **join mutator** sets `ACTIVE` + `startedAt: now` client-side — and `CallSession` is not in the Zero schema, so no mutator can append a session → once `ACTIVE`, the webhook gate skips activation.

A mutator-activated, session-less call that later reverts to `SCHEDULED` and is rejoined enters the **first-join** branches (its metadata lacks `conversationId`/`systemMessageId`), which write `startedAt` and append a session together — self-consistent. The rejoin branch requires both metadata keys, which only the first-join activations write, and those now always append a session; pre-existing ones are covered by the migration backfill. So "rejoin with zero prior sessions" is not reachable via known paths. The defensive seed suggested inline stays a nice-to-have, not a blocker. (The mutator path re-anchoring `startedAt` to the later webhook join is pre-existing behavior, unchanged by this PR.)

**Updated findings: 1 blocking (migration backfill), 1 major (R4 duplicate sessions), 1 medium (R7 rejoin/session-append tests), 1 resolved (R2 → hardening nit).** Verdict unchanged — changes requested; the only merge-blocker is the in-migration backfill.

**Reviewer table — completed** (commits-by-path over the last 50 `main` commits, reconstructed via merged-PR file listings):

| Candidate | Owns (files in this PR) | CODEOWNERS | Pending reviews | Note |
|-----------|-------------------------|------------|-----------------|------|
| @harshpreet931 | `callRepository.ts` (per PR request; call-domain — authored open #1262) | no | 0 | **recommended** — requested in the PR body and by the architecture thread |
| @shubhamdhamija-juspay | `callRepository.ts` — most recent committer (#1417, merged 2026-09-02 11:40 UTC) | no | 0 | suggest as **second reviewer** — #1417 touched this file hours before this PR opened |
| @sithaarth | `schema.prisma` (#1279) | no | 1 (#1189) | — |
| @sirajshaik-code | `schema.prisma` (#1181) | no | 0 | — |
| @devesh-juspay | `schema.prisma` (#1421) | no | 0 | — |
| @sumantop | `schema.prisma` (#749) | no | 0 | — |

`migration.sql` and `callSessionProjection.test.ts` are new in this PR (no prior owners). @harshpreet931 remains the primary recommendation; @shubhamdhamija-juspay is the strongest recent file-owner and worth a second slot to check for interaction with #1417's same-file changes.

**Coordination note:** this PR includes and extends #1446's rejoin fix. #1446's `callRepository.activateScheduledCall.test.ts` already covers the rejoin branch's `startedAt` handling — if #1446 is superseded by this PR, port those tests here (it would also address part of the R7 finding).

XyneSpaces:
Closing in favor of the two-column fix in #1446.

Decision (from thread): for the reported bug the only requirement is **wall-clock call duration** (first-join → last-leave), and #1446 delivers exactly that by preserving `Call.startedAt` on rejoin and clearing the stale `endedAt` — the same `MIN..MAX` envelope this table would have projected, without a new table, migration, backfill, or projection layer.

The CallSession table remains the right design only if we later need **active-vs-idle time** (excluding gaps between sessions) or **per-session analytics** (rejoin count, per-session length/participants). Neither is on the roadmap today, so this is deferred rather than merged to keep the surface area minimal. The branch `feature/call-session-table` is preserved if we want to revisit.

PR #1515, opened 2026-09-03
fix: XYNE-62810 dedupe migration count and post hotfix canvas-sync activity line

sumantop:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

XYNE-62810. Three related fixes on the release-analysis path:

1. **Migration count was inflated.** One row is stored per file × commit, so a file touched across several commits (typically `schema.prisma`) was counted once per commit. `countDistinctMigrationFiles` collapses those to one per repo+file, and the canvas metadata, the per-repo analysis result and the `COMMIT_ANALYSIS_COMPLETED` release event all use it.
2. **A hotfix sync left no trace.** The analysis canvas is upserted in place, so a re-sync changed the canvas with nothing in the timeline. A hotfix sync now posts a `RELEASE_SYNC` activity line (through `recordTicketTimelineEvent`, so the SYSTEM / `isTicketActivity` invariants hold, `isAutomation: true`) linking to the canvas, with distinct copy for the no-canvas and unconfigured-frontend-URL cases.
3. **Canvas links in messages did not open as an overlay.** `{...props}` was spread *after* `onClick`, so `CanvasLink`'s handler was overwritten and never ran — neither link shape was intercepted. The spread moved above `onClick`, and the canvas id now comes from `parseInternalXyneLink`, which already handles both the bare and the workspace-prefixed shape. The full-page fallback keeps the link's query string and hash.

Follow-up (deliberately out of scope, separate ticket) — remaining emitters of the bare `/chat/canvas/:id` form, which 404s on a hard load. Every `commitAnalysisController` site is fixed here; these are the ones left:
- `apps/backend/src/services/callDocumentService.ts:2148, :2300, :2409` — `getCanvasUrl(canvasId)` without a workspaceId
- `apps/backend/src/services/transcriptService.ts:1437` — `getCanvasUrl(notesCanvasId)`
- `apps/backend/src/controllers/jiraMigrationController.ts:1950` — `getCanvasUrl(canvasId)`
- `apps/backend/src/workflows/storage/db-storage.ts:3261` — `getCanvasUrl(canvasId)`
- `apps/backend/src/controllers/releaseNotesController.ts:135` — inline `${config.slackFrontendUrl}/chat/canvas/…` (bare *and* on the Slack base)

Same pass should fold `getCanvasUrl` into `buildWorkspaceCanvasUrl` (one builder, one env read, no hardcoded host) once the bare form has no callers left.

## Areas Touched

- [x] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

## Schema & Data Changes

## Config, Environment & URLs

## API Contract

- [x] No API changes

---

## How did you test it?

Full end-to-end run on a freshly wiped dev DB (146 tables truncated, identity/ACL/forms kept), driven through the UI as a **newly created non-admin user** granted LISTPROJECTS + RELEASE-MANAGER: project `Payments E2E (PAYE)` -> channel `paye-releases` -> repository connected against the real GitHub token (`Connected to sumantop/release-e2e-test`) -> 4 real PRs generated -> release created -> hotfix merged.

**1. Canvas is generated.** `[CanvasService] Created release analysis canvas c5016726-… (section=main)`; canvas `📦 Release Analysis: sumantop/release-e2e-test` posted into the release channel.

**2. Hotfix sync posts the `RELEASE_SYNC` line.** After the hotfix merge (`Synced hotfix (286f88c5...7b2c30c4) for PAYE-0001`, canvas updated `section=hotfix`), the release ticket's message section shows:

```
content:  Hotfix synced — release analysis canvas updated
          http://localhost:5173/<workspaceId>/chat/canvas/c5016726-…
msgType:  SYSTEM
metadata: {"activityType":"RELEASE_SYNC","isAutomation":true,"isTicketActivity":true}
```

The link is workspace-prefixed (via the shared `buildWorkspaceCanvasUrl`), as is the "📄 View Full Analysis Report →" link on the summary message.

**3. The link opens as an in-channel overlay.** Clicking it from the ticket lands on `…/chat/dir/<channel>/…/PAYE-0001#canvas=c5016726-…` — stays in the channel, canvas mounts, no full-page navigation. Both link shapes (bare and workspace-prefixed) were exercised, plus a cross-workspace link which correctly falls through to browser navigation instead of opening the wrong workspace's overlay.

**4. Migration dedupe.** The stock fixture writes a distinct migration per PR, so it does not exercise the dedupe. A second PR was raised touching a migration file an earlier PR already touched (`backend/migrations/0101_paye-101.sql`, PR #22 and PR #27). Over that range the Migrations tab shows that file spanning **3 commits**, while the reported count is **4** — the number of distinct files, not the 6 raw file×commit rows. Pre-fix that count was the raw row count.

Local CI mirror green on the head commit: backend typecheck + build, dashboard `lint:errors-only` / `format:check` / `typecheck` / `build`, framework lint, `xyne-claw-shared` + `xyne-claw` typecheck.

Not exercised: the same-workspace fallback branch in `CanvasLink` (query/hash preservation). It only fires on a route without a `:channelId` param, which the current data could not produce — verified by reading, not by clicking.

### Automated

- [x] Not covered by tests 

### Manual

**Users**
- [x] Single user

**Run mode**

**Surface & data**
- [x] Web browser 

---

## Checklist


apps/backend/src/controllers/commitAnalysisController.ts
```diff
@@ -611,6 +613,28 @@ export class CommitAnalysisController {
         },
       });
 
+      // The canvas updates in place, so a hotfix sync posts an activity line (via the
+      // canonical helper, for the right SYSTEM/isTicketActivity invariants) pointing at
+      // it. The workspace prefix is required — a bare /chat/canvas/:id route 404s.
+      if (hotfixSync) {
+        try {
+          const canvasLink = canvasId
```
devesh-juspay: Two link-shape observations:

1. The workspace-prefixed URL bypasses the canvas-overlay click handler in `RenderMessageWithHTML.tsx` (~line 262), which only intercepts `url.pathname.startsWith('/chat/canvas/')`. So clicking this link in a thread does a full page navigation instead of opening the canvas as an overlay in the channel — while the bare form is intercepted client-side (preventDefault → `#canvas=` overlay) and does *not* 404 on an in-app click. The 404 the comment mentions only bites on a hard load (cmd-click / new tab), where the prefixed form is indeed the correct one. Neither shape wins both cases today; worth a small dashboard follow-up to make the overlay handler also match `/:workspaceId/chat/canvas/:id` so this link gets the overlay treatment too.

2. Pre-existing, not this PR: `canvasUrl` at lines ~555/701 still builds the bare un-prefixed form — if the bare route 404s on hard load, those are affected everywhere that canvasUrl surfaces. Might deserve its own ticket.

apps/backend/src/controllers/commitAnalysisController.ts
```diff
@@ -611,6 +613,28 @@ export class CommitAnalysisController {
...
+            ? ` ${config.slackFrontendUrl}/${params.workspaceId}/chat/canvas/${canvasId}`
+            : '';
+          await recordTicketTimelineEvent({
+            message: {
+              conversationId,
+              senderId: userId,
+              content: `Hotfix synced — release analysis canvas updated${canvasLink}`,
+              activityType: 'RELEASE_SYNC',
```
devesh-juspay, line 636: Nit: other automation-authored timeline lines set `isAutomation: true` in the message input (see releaseDevTicketNotifyService) so the metadata marks bot-written lines consistently. No visible difference today — the "(Automation)" label helper only rewrites status/move copy — but worth adding for consistency/future filtering.

apps/backend/src/controllers/commitAnalysisController.ts
sumantop: Fixed, the handler now takes the id from parseInternalXyneLink which already handles both shapes. Turned out it wasn't firing at all though: {...props} was spread after onClick={handleClick} and overwrote it, so the bare form wasn't intercepted either. Moved the spread up, both shapes open the overlay now, checked in the app. Point 2 left alone, will raise a separate ticket for it.

apps/backend/src/controllers/commitAnalysisController.ts
sumantop: Added. Checked the sender on this path is the ticket bot (releaseWebhookSync passes bot?.id), so it's accurate and not just for consistency. Nothing renders off it today like you said, addAutomationActorLabel only touches status/move copy.

devesh-juspay (dismissed):

apps/backend/src/controllers/commitAnalysisController.ts
```diff
@@ -611,6 +613,38 @@ export class CommitAnalysisController {
...
 
+      // The canvas updates in place, so a hotfix sync posts an activity line (via the
+      // canonical helper, for the right SYSTEM/isTicketActivity invariants) pointing at
+      // it. The workspace prefix is required — a bare /chat/canvas/:id route 404s.
+      if (hotfixSync) {
+        try {
+          const frontendBaseUrl = (config.slackFrontendUrl ?? '').replace(/\/+$/, '');
+          let content: string;
+          if (canvasId && frontendBaseUrl) {
+            content = `Hotfix synced — release analysis canvas updated ${frontendBaseUrl}/${params.workspaceId}/chat/canvas/${canvasId}`;
```
ameer030: 🟡 **W01 · Better to fix · BUG-C03/ARCH-C03 — canvas deep-link builder duplicated a third time**

This inline-builds the workspace-prefixed canvas URL, but that exact string construction already exists in `buildReleaseReportCanvasUrl` (`apps/backend/src/services/releaseReports/releaseReportService.ts:36-39`), and a near-identical one in the exported `getCanvasUrl` (`apps/backend/src/services/canvasService.ts:308`). A route-shape change (this PR exists because of one — the bare form 404s) now has to be fixed in three places.

<details><summary>Details</summary>

- Here: `${frontendBaseUrl}/${params.workspaceId}/chat/canvas/${canvasId}` with its own `/\/+$/` trim
- `releaseReportService.ts:36-39`: same `config.slackFrontendUrl` base (single-slash-trim variant) + `/${workspaceId}/chat/canvas/${canvasId}` — identical result, module-private
- `canvasService.ts:308` `getCanvasUrl`: same path shape but on `FRONTEND_URL` with a hardcoded prod-host fallback — pre-existing divergence, not this PR's debt
</details>

<details><summary>Recommended fix</summary>

Export one builder — e.g. move `buildReleaseReportCanvasUrl` into `canvasService.ts` next to `getCanvasUrl` (or export it from releaseReportService) taking `(workspaceId, canvasId)` on the `slackFrontendUrl` base, and call it from here and releaseReportService. Leave `getCanvasUrl`'s base-URL divergence for a separate cleanup.
</details>

apps/dashboard/src/components/Chat/RenderMessageWithHTML/RenderMessageWithHTML.tsx
```diff
@@ -259,14 +268,12 @@ const CanvasLink = ({
...
-      const targetCanvasId = parts[parts.length - 1];
 
-      if (targetCanvasId && channelId) {
+      if (canvasId && channelId) {
         // Open as overlay in current channel
-        void navigate(`${location.pathname}#canvas=${targetCanvasId}`);
+        void navigate(`${location.pathname}#canvas=${canvasId}`);
       } else {
         // Fallback to full page navigation
         void navigate(url.pathname);
```
ameer030: 🟡 **W02 · Better to fix · BUG-C28 — fallback navigation drops query string and hash**

Generalizing the interception to all same-origin canvas links is the right fix for the prefixed form, but the full-page fallback navigates to `url.pathname` only — a canvas link carrying `?blockId=…` (or any hash) silently loses those params. Before this change, prefixed links went through default browser navigation, which preserves query/hash.

<details><summary>Details</summary>

- Here: `event.preventDefault()` now fires for any same-origin canvas link; the fallback is `void navigate(url.pathname)` and the overlay branch only sets `#canvas=<id>`
- Query-bearing canvas deep links exist in the codebase — e.g. `apps/backend/src/services/notificationService.ts:1373` builds `/${workspaceId}/chat/canvas/${canvasId}?${queryString}` action URLs — though none are currently emitted into message content, so today's blast radius is mostly pasted links; bare-form query loss is pre-existing
</details>

<details><summary>Recommended fix</summary>

Fallback: `` void navigate(`${url.pathname}${url.search}${url.hash}`) ``. If canvas params like `blockId` should also survive an overlay open in the current channel, carry them through on the `#canvas=` branch as well.
</details>

ameer030:
🟡 **W03 · Better to fix · TEST-C01 — no proof-of-testing attached**

The PR template's testing section is completely empty (no type-of-change ticked, description left blank), and this PR changes visible behavior in three ways. Proof-of-testing is the repo's merge gate for behavior changes, so please attach evidence that the new paths actually ran.

<details><summary>Details</summary>

- `.github/PULL_REQUEST_TEMPLATE.md` → "How did you test it?" — every box unchecked, manual and automated sections empty
- Observable behavior changed: the new RELEASE_SYNC timeline line + icon (`TicketActivityMessage.tsx`), retargeted canvas-link overlay interception for workspace-prefixed URLs (`RenderMessageWithHTML.tsx`), and deduped migration counts on the analysis canvas and release event
- Review already caught one thing manual smoke would have surfaced quickly — the overlay handler silently never firing because `{...props}` overwrote `onClick` — so a short note of exactly what was exercised (which link shapes, which sync trigger) carries real signal here
</details>

<details><summary>Recommended fix</summary>

Fill in the template: tick the type of change, add a one-paragraph description + ticket link, and attach a screenshot or short recording of a hotfix sync posting the RELEASE_SYNC line and its canvas link opening the overlay. A pasted before/after of the canvas metadata `migrationCount` would also cover the dedupe change.
</details>

ameer030:
## Xyne PR Review — Summary

| Category | Count | Findings |
|----------|-------|----------|
| 🔴 Blocking | 0 | — |
| 🟡 Better to fix | 3 | W01, W02, W03 |
| 🔵 Nit | 0 | — |

**Zero blocking findings.**

- **W01** — canvas deep-link builder duplicated a third time; share one exported helper (`commitAnalysisController.ts:621` vs `buildReleaseReportCanvasUrl` / `getCanvasUrl`)
- **W02** — `CanvasLink` fallback navigation drops query/hash on same-origin canvas links (`RenderMessageWithHTML.tsx`)
- **W03** — no proof-of-testing attached; template testing section empty

Existing threads: 2 fixed ✓ · 0 still open — both of @devesh-juspay's earlier points verified fixed at `6ed410b4` (overlay handler now parses both link shapes; `isAutomation: true` set). The bare-form `canvasUrl` at controller ~lines 553/677 remains pre-existing and was consciously deferred to a follow-up ticket — not re-flagged here.

apps/backend/src/controllers/commitAnalysisController.ts
sumantop: Fixed in daba0923e. Pulled the builder out as `buildWorkspaceCanvasUrl(workspaceId, canvasId)` in `canvasService.ts`, right next to `getCanvasUrl`, on the `slackFrontendUrl` base with the `/\/+$/` trim. `releaseReportService`'s module-private `buildReleaseReportCanvasUrl` is gone and both it and the controller call the shared one now, so the route shape lives in one place. Kept the controller's explicit unconfigured-URL branch (it still needs to fall back to the "no link available" copy rather than emit a relative URL). `getCanvasUrl`'s `FRONTEND_URL` + hardcoded-host divergence left alone as you suggested.

apps/dashboard/src/components/Chat/RenderMessageWithHTML/RenderMessageWithHTML.tsx
sumantop: Fixed in daba0923e — the fallback is now ``navigate(`${url.pathname}${url.search}${url.hash}`)``, so a pasted canvas link keeps whatever it carries. Left the `#canvas=` branch alone on purpose: `ChatView` reads the overlay id off `location.hash` with `/#canvas=([a-zA-Z0-9-]+)/` and there is nowhere for a `blockId` to land today, so carrying params through the overlay needs a reader on the other side first. Nothing emits query-bearing canvas links into message content yet (the `notificationService` ones are action URLs), so this is the whole of the gap for now.

sumantop:
W03 — filled in the template: type of change, description of all three fixes with the ticket, areas touched, and what was actually exercised manually (both canvas link shapes from a thread, the `RELEASE_SYNC` line landing on a hotfix sync, and the migration count on a release whose `schema.prisma` moved across several commits). Marked as not covered by tests with the reason — there is no existing suite around commit-analysis canvas posting or the message link renderer.

Local CI mirror on `daba0923e`, all green: backend typecheck+build, dashboard lint / format:check / typecheck / build, framework lint, xyne-claw-shared + xyne-claw typecheck.

sumantop:
Correction to my note above on W03: the manual-testing section now only claims what was actually exercised — the canvas link interception (both shapes, from a thread). The hotfix-sync `RELEASE_SYNC` line and the `migrationCount` before/after are listed as still-to-attach rather than claimed as done.

apps/backend/src/controllers/commitAnalysisController.ts
sumantop: Fixed in `daba0923e`.

`buildWorkspaceCanvasUrl(workspaceId, canvasId)` now lives in `canvasService.ts` next to `getCanvasUrl`, on the `config.slackFrontendUrl` base. `buildReleaseReportCanvasUrl` is deleted and both call sites go through the shared helper — this one (`commitAnalysisController.ts:623`) and `releaseReportService.ts:314`. So the route shape is in one place now.

Left `getCanvasUrl`'s `FRONTEND_URL` + hardcoded prod-host divergence alone, as you suggested — that's pre-existing and wants its own cleanup.

apps/dashboard/src/components/Chat/RenderMessageWithHTML/RenderMessageWithHTML.tsx
sumantop: Half of this is fixed in `daba0923e`; the other half I'm deliberately leaving.

**Fixed** — the full-page fallback is now:

```ts
void navigate(`${url.pathname}${url.search}${url.hash}`)
```

so query and hash survive the navigation the interception newly captures. Good catch; that was a real regression against default browser navigation.

**Not doing here** — carrying params through on the `#canvas=` overlay branch. The overlay still goes to `${location.pathname}#canvas=${canvasId}`, so a `?blockId=…` is dropped when the link opens in-channel. The reason is that there's nothing on the other end to receive it: `ChatView.tsx:133` parses the hash with `/#canvas=([a-zA-Z0-9-]+)/` and pulls out a canvas id and nothing else. Threading `blockId` through would mean inventing a hash format plus the overlay-side scroll-to-block that reads it — a real feature, not a fallback fix, and wider than this PR.

Filing that with the bare-form `canvasUrl` follow-up. Today it costs nothing, since as you note no query-bearing canvas links are emitted into message content yet.

sumantop:
Replies to the review — W01 and W02 are answered on their own threads; this covers **W03** and the summary.

### W03 — proof of testing

Partly addressed, and I'd rather be straight about the part that isn't.

**Done:** the template is filled in — type of change ticked, description with the ticket and the three fixes, areas touched, and a testing section that says what was actually exercised. Canvas links were checked in the running app: both the bare `/chat/canvas/:id` and the workspace-prefixed `/:workspaceId/chat/canvas/:id` shape now open as an in-channel overlay. Before the `{...props}` ordering fix neither was intercepted, which is exactly the failure you point out a smoke test would have caught.

**Still open:** no recording or screenshot is attached yet, and two of the three behavior changes are *not* manually confirmed — the `RELEASE_SYNC` timeline line, and the deduped `migrationCount`. Both need a release whose `schema.prisma` was touched across several commits plus a hotfix sync to trigger, which I haven't set up locally. I've marked that explicitly in the testing section rather than ticking boxes I can't stand behind, so please don't read the filled-in template as the gate being met. Happy to hold the merge until that evidence is up.

### Summary

W01 and W02 are both fixed in `daba0923e` — one shared `buildWorkspaceCanvasUrl` in `canvasService.ts`, and the same-origin fallback now preserves query and hash. One caveat on W02: I fixed the fallback but deliberately did not carry params through the `#canvas=` overlay branch, since nothing reads them on the other end yet. Reasoning is on that thread.

Also worth flagging for anyone reading the diff: `origin/main` was merged in at `7d75a838` to clear a conflict in `canvasService.ts`. It was an adjacent-edit collision, not a real one — the new URL helper landed next to the `LooseInline` type that #1534 expanded. Both sides kept, local CI mirror green across all eight legs.

The bare-form `canvasUrl` at controller ~553/677 stays deferred, as noted — thanks for not re-flagging it.

apps/dashboard/src/components/Chat/RenderMessageWithHTML/RenderMessageWithHTML.tsx
sumantop: Correcting my earlier reply on this thread, and closing it out properly.

I said carrying params through the `#canvas=` branch "costs nothing today, since no query-bearing canvas links are emitted into message content". That was right about query strings and **wrong about the workspace prefix** — which the same branch also drops. I hadn't tested it when I wrote that.

**What I found.** Driving the running app, a canvas link into *another* workspace was intercepted and opened as an overlay in the workspace already on screen:

- clicked: `/cmsodac…/chat/canvas/1dd809ec…`
- landed: `/cmsobix…/chat/dir/cmtlfw…#canvas=1dd809ec…`
- result: **"Canvas not found or access denied"**

That's a regression this PR introduced, not a pre-existing quirk. At merge-base the guard was `url.pathname.startsWith('/chat/canvas/')`, which the workspace-prefixed shape failed, so those links were never intercepted and the browser navigated to the right workspace. Generalising the interception to all same-origin canvas links caught them — and then discarded the one segment that said where to look.

**Fix** (`9eb129795`): `parseInternalXyneLink` already detects the workspace segment in order to `shift()` past it, so it now returns it rather than throwing it away, and `CanvasLink` only takes the overlay when the link belongs to the workspace already open:

```ts
const isSameWorkspace = !linkWorkspaceId || linkWorkspaceId === workspaceId;
if (url.origin === window.location.origin && isSameWorkspace) { … }
```

A cross-workspace link now falls through to plain browser navigation — exactly the pre-PR behaviour. A bare link carries no workspace, so it stays same-workspace and keeps opening as an overlay.

Two things worth recording from testing it:

1. My first attempt put the check *inside* the `preventDefault()` block and routed the cross-workspace case through `navigate()`. That produced `/cmsobix…/cmsodac…/chat/canvas/…` — the router scopes paths to the current workspace, so it doubled the prefix. Hence the guard sits on the interception itself. The same reasoning applies to the fallback, which now hands the router a path with the workspace stripped since the router re-adds it.
2. Verified by instrumenting the branch decision rather than eyeballing URLs: prefixed-same-workspace → `isSameWorkspace: true` → overlay; bare → no `linkWorkspaceId` → overlay; cross-workspace → `isSameWorkspace: false` → handed to the browser.

Local CI mirror green on `9eb129795` across all eight legs.

Still not carrying `blockId` through the overlay branch, for the original reason: `ChatView.tsx:133` parses the hash as `/#canvas=([a-zA-Z0-9-]+)/` and takes a canvas id and nothing else, so there is no consumer for it yet. That one stays a follow-up — but it's a missing feature, not a broken link, which is the distinction I got wrong the first time.

apps/backend/src/controllers/commitAnalysisController.ts
sumantop: Following up — the deferral on this is gone too, so W01 is now fully closed.

`daba0923e` fixed the site flagged here, but two siblings in the same file still built the URL inline, in the bare un-prefixed form that 404s on a hard load (`commitAnalysisController.ts` ~556 and ~713 — the `canvasUrl` on the analysis summary message). Those were the ones the summary called "consciously deferred to a follow-up ticket". Leaving them made the recommendation here half-applied: a shared builder existed, and two call sites still didn't use it, which is exactly the "fix it in N places" problem the finding was about.

`ad3ec9925` routes both through `buildWorkspaceCanvasUrl`. The second one lives in `postToParentTicket`, which had no workspace id in scope — its `workspace` parameter is the project key — so it now takes `workspaceId` explicitly (single call site, threaded from `params.workspaceId`).

There is now **no inline canvas-URL construction left in the controller**: all four sites go through the one helper, and the summary link is workspace-prefixed, so it no longer 404s on a hard load.

Not touching `confluenceImportService`'s private `buildWorkspaceCanvasUrl` — despite the matching name it builds a different route (`/chat/dir/canvas/:id`, relative, no host), so it isn't the same builder and folding it in would change behaviour unrelated to this PR.

Local CI mirror green on `ad3ec9925` across all eight legs.

sumantop:
**W03 is now addressed.** Earlier I said the template was filled in but the evidence was not, and that two of the three behaviours were unverified. That is no longer true — the PR body's testing section has been rewritten with a full end-to-end run, and the two gaps I flagged are closed.

The run was done on a **freshly wiped dev DB** and driven through the UI as a **newly created non-admin user** (granted LISTPROJECTS + RELEASE-MANAGER), not as the seeded admin: project -> channel -> repository connected against the real GitHub token -> 4 real PRs -> release -> hotfix merge.

- **Canvas generated** — `[CanvasService] Created release analysis canvas c5016726-… (section=main)`, posted to the release channel.
- **`RELEASE_SYNC` line** — after the hotfix merge, the release ticket's message section carries `Hotfix synced — release analysis canvas updated <link>`, `msgType: SYSTEM`, `metadata: {"activityType":"RELEASE_SYNC","isAutomation":true,"isTicketActivity":true}`. This is the item that previously rested on an artifact from an earlier run; it is now reproduced live.
- **Overlay** — clicking that link stays in the channel (`…#canvas=c5016726-…`) and mounts the canvas.
- **Migration dedupe** — worth recording, because the obvious test does not actually test it: the fixture writes a *distinct* migration file per PR, so raw count == distinct count and the dedupe never runs. I raised a second PR touching a migration file an earlier PR already touched (`0101_paye-101.sql`, PRs #22 and #27). Over that range the Migrations tab shows that file spanning **3 commits** while the count reads **4** — distinct files, not the 6 raw file×commit rows.

Two things found while doing this, neither introduced by this PR but both worth knowing:

1. `ResourceProtectedRoute` defaults `minAccess` to `ADMIN`, and `AppRoot.tsx` passes no `minAccess` for `LISTPROJECTS` (:1688) or `PROJECTS` (:1584). Granting **WRITE** through User Management therefore does nothing visible — the user is redirected to the AI home with no error, which reads as a broken page rather than a permissions problem.
2. `Create Release` exists only in the Release-Manager project view (tabs *Boards / Repositories / Releases*); the `/listProjects` view of the same project has no such action.

Still not exercised, and I would rather say so than imply full coverage: the same-workspace fallback branch in `CanvasLink` (query/hash preservation). It only fires on a route with no `:channelId` param, which this data set could not produce — verified by reading, not by clicking.

apps/backend/src/services/releaseReports/releaseReportService.ts
```diff
@@ -33,11 +33,6 @@ interface PublishReleaseReportInput {
 
 const ENV_VAR_REGEX = /^\s*([A-Za-z][A-Za-z0-9_]*)\s*[=:]/gm;
 
-function buildReleaseReportCanvasUrl(workspaceId: string, canvasId: string): string {
```
ameer030, line 36: Isnt this function getting used any where else?!

apps/backend/src/services/canvasService.ts
```diff
@@ -311,6 +312,16 @@ export function getCanvasUrl(canvasId: string, workspaceId?: string): string {
   return `${frontendUrl}${path}`;
 }
 
+/**
+ * Workspace-scoped canvas deep link on the Slack-facing frontend base URL. The
+ * workspace prefix is required — the bare `/chat/canvas/:id` route 404s on a
+ * hard load.
+ */
+export function buildWorkspaceCanvasUrl(workspaceId: string, canvasId: string): string {
+  const frontendUrl = (config.slackFrontendUrl ?? '').replace(/\/+$/, '');
```
ameer030: should this be slack fe url?!

apps/backend/src/services/releaseReports/releaseReportService.ts
sumantop: No — it was module-private (never exported), and its only caller on main was `publishReleaseReport` in this same file (line 319), which now goes through the shared `buildWorkspaceCanvasUrl` in `canvasService.ts`. `git grep buildReleaseReportCanvasUrl` on the branch returns nothing, so the delete is safe.

apps/backend/src/services/canvasService.ts
sumantop: Good call — switched to `config.frontendUrl` in `3fdd94af4a`. This is an in-app link rendered by the dashboard, and `CanvasLink` only opens the overlay when `url.origin === window.location.origin`, so the app's own origin (`FRONTEND_URL`, the same value `env.ts` uses for CORS own-origin, and required non-empty by Joi) is the right base. `SLACK_FRONTEND_URL` allows `''` and is meant for links pushed to Slack. The controller's hotfix-sync guard now checks `config.frontendUrl` to match.

No behaviour change today — every env (`.env.example`, `.env.test`, `sandbox.sh`) sets both to the same value. The old `buildReleaseReportCanvasUrl` was on the Slack base, so this also corrects that. Left the ticket-URL builders on `slackFrontendUrl` alone (pre-existing, same fix wants its own pass). Backend typecheck green.

sumantop: Follow-up `e19e5710c6`: added a note on `buildWorkspaceCanvasUrl` saying the ticket links (`/chat/:channelId?tab=tickets…`) intentionally stay on `slackFrontendUrl` for now, so the split isn't mistaken for an oversight.

apps/dashboard/src/components/Chat/RenderMessageWithHTML/RenderMessageWithHTML.tsx
```diff
@@ -260,17 +272,27 @@ const CanvasLink = ({
...
       } else {
-        // Fallback to full page navigation
-        void navigate(url.pathname);
+        // Fallback to full page navigation, keeping any query/hash the link
+        // carries. The router re-adds the workspace, so hand it the path
+        // without one.
+        const routerPath = linkWorkspaceId
+          ? url.pathname.slice(`/${linkWorkspaceId}`.length)
+          : url.pathname;
+        void navigate(`${routerPath}${url.search}${url.hash}`);
```
ameer030, line 294: 🔴 **B01 — the full-page fallback lands on the app's own 404 screen**

When a canvas link is clicked outside a channel route (no `:channelId` param — e.g. My Tickets, Activity, or a release/ticket view), this branch strips the workspace prefix and calls `navigate('/chat/canvas/…')`. React Router resolves that absolute path from the router root, and nothing re-adds the workspace: every canvas route lives under `/:workspaceId/chat/…`, so the URL matches only the root-level `*` catch-all → `NotFoundScreen`. Before this PR these prefixed links were not intercepted at all (the overwritten `onClick`), so the browser hard-loaded the working workspace-prefixed URL — this branch turns a working click into an in-app 404, for both link shapes.

- Here (`RenderMessageWithHTML.tsx:292-295`): for a same-workspace prefixed link, `routerPath = url.pathname.slice('/' + linkWorkspaceId.length)` → `'/chat/canvas/<id>'`, handed root-absolute to `navigate`.
- The comment's premise — "The router re-adds the workspace" — doesn't exist: `AppRoot.tsx` mounts all canvas routes only under `/:workspaceId/chat/…` (lines 1240-1247, 1349-1356, 1406-1422; the `*` at line 1450 lives *inside* the `chat` subtree and can't be reached from `/chat/canvas/x`). `/chat/canvas/x` falls through to the root `path: '*'` → `<NotFoundScreen/>` (`AppRoot.tsx:2111-2112`). The app's own idiom confirms the workspace is prepended by hand, not by the router — `CanvasRedirectPage.tsx:30-32` (`toWorkspaceRoute`).
- Reachable: `shouldReplaceWithSemanticLabel` is true for raw/autolinked URLs (text === href), and this PR's own RELEASE_SYNC line ("Hotfix synced — release analysis canvas updated <url>") is exactly that shape. Rendered from any message surface without a `:channelId` param, the click takes this branch.
- The PR body already notes this branch was "verified by reading, not by clicking" — unfortunately the reading is built on the false premise above.
- Bare links are affected the same way (`routerPath = url.pathname` is already workspace-less), so the fallback never produces a working navigation for any same-workspace click outside a channel.

</details>

Keep the workspace prefix — the link's own path is already the correct absolute route:

```ts
const routerPath = linkWorkspaceId
  ? url.pathname // already /:workspaceId/chat/canvas/:id
  : `/${workspaceId ?? ''}${url.pathname}`; // bare form: prepend current workspace
void navigate(`${routerPath}${url.search}${url.hash}`);
```

(or mirror the `toWorkspaceRoute` idiom from `CanvasRedirectPage.tsx`.)

</details>

rules: BUG-C28
agent: omp
run: paarvai-reviews/4
</details>

apps/backend/src/controllers/commitAnalysisController.ts
```diff
@@ -611,6 +615,39 @@ export class CommitAnalysisController {
         },
       });
 
+      // The canvas updates in place, so a hotfix sync posts an activity line (via the
+      // canonical helper, for the right SYSTEM/isTicketActivity invariants) pointing at it.
+      if (hotfixSync) {
+        try {
+          const canvasLink =
+            canvasId && config.frontendUrl
```
ameer030: 🔵 **N02 — the "frontend URL is not configured" notice is unreachable**

`config.frontendUrl` can never be falsy: `FRONTEND_URL` has a Joi default (`http://localhost:5173`) and empty strings fail boot validation, so `canvasId && config.frontendUrl` reduces to `canvasId` and this copy can never be shown.

- `apps/backend/src/config/env.ts:170` — `FRONTEND_URL: Joi.string().default('http://localhost:5173')`, no `.allow('')`; `env.ts:589-592` aborts boot on validation error.
- (After `3fdd94af` this guard moved off `SLACK_FRONTEND_URL`, which *does* allow `''` — that base made the branch reachable; the current one doesn't.)

</details>

Either drop the `config.frontendUrl` half of the guard (and the "not configured" copy), or keep the guard and tweak the comment so a future reader doesn't hunt for a config state that can't exist.

</details>

rules: STYLE-C14
agent: omp
run: paarvai-reviews/4
</details>

ameer030 (commented):
🔵 **N01 — `canvasService.ts` now has two canvas-URL builders on the same env var, with inverted argument order**

`getCanvasUrl(canvasId, workspaceId?)` reads `process.env.FRONTEND_URL` directly with a hardcoded production fallback; the new `buildWorkspaceCanvasUrl(workspaceId, canvasId)` reads the same variable through config. Same base, same path shape, opposite parameter order — easy to pick the wrong one.

- `canvasService.ts:309` `getCanvasUrl` — `process.env.FRONTEND_URL || 'https://spaces.xyne.juspay.net'` (pre-existing, incl. the direct env read that bypasses the config module)
- `canvasService.ts:324` `buildWorkspaceCanvasUrl` — `(config.frontendUrl ?? '').replace(/\/+$/, '')` (added by this PR)
- The ameer thread on the inline duplication was fixed well by extracting this helper; this is just the residual: the two exports now sit ten lines apart doing near-identical work.

</details>

Whenever the follow-up touches these call sites: have `getCanvasUrl` delegate to `buildWorkspaceCanvasUrl` (or vice versa) so one function owns the path shape and the env read; drop the hardcoded-host fallback while there. No need to hold this PR for it.

</details>

rules: ARCH-C03, CI-C06
agent: omp
run: paarvai-reviews/4
</details>

ameer030 (commented):
🟡 **W01 — the follow-up ticket should name every remaining bare-canvas-link emitter, not just the summary message**

The PR body defers "canvasUrl on the summary message still builds the bare un-prefixed form" to a follow-up ticket — good. But the same hard-load-404 shape is emitted from at least five other places, and the ticket as described doesn't cover them, so they risk being left behind.

Bare / un-prefixed emitters at `e19e5710c`:

- `apps/backend/src/services/callDocumentService.ts:2148, :2300, :2409` — `getCanvasUrl(canvasId)` with no workspaceId (PRD / detailed-summary links)
- `apps/backend/src/services/transcriptService.ts:1437` — `getCanvasUrl(notesCanvasId)`
- `apps/backend/src/controllers/jiraMigrationController.ts:1950` — `getCanvasUrl(canvasId)`
- `apps/backend/src/workflows/storage/db-storage.ts:3261` — `getCanvasUrl(canvasId)` (knowledge canvas)
- `apps/backend/src/controllers/releaseNotesController.ts:135` — inline `${config.slackFrontendUrl}/chat/canvas/${canvasIdResult}` (bare form *and* on the Slack base, so it additionally misses the overlay's same-origin check)

Contrast with `canvasController.ts:93,453,534`, which already pass `req.user?.workspaceId` into `getCanvasUrl`. All pre-existing — this is about the follow-up's scope, not this diff.

</details>

Widen the follow-up ticket's description to enumerate these five call sites (or point them at `buildWorkspaceCanvasUrl` / `getCanvasUrl(id, workspaceId)` in the same pass), so the bare form stops being emitted anywhere.

</details>

rules: PROC-C11
agent: omp
run: paarvai-reviews/4
</details>

ameer030:
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 1 |
| 🟡 Better to fix | 1 |
| 🔵 Nit | 2 |

### 🔁 Re-check
- 💬 **Existing threads:** 6 fixed ✅ · 0 still open — verified against `e19e5710c` (all six remain GitHub-unresolved but every claimed fix is in the head commit; ameer's W02 overlay-params half is a justified deferral — no consumer reads them)
- ✅ **Verified OK:** `countDistinctMigrationFiles` dedupes correctly for both GitHub (`/commit/<sha>`) and Bitbucket (`/commits/<sha>#<path>`) URL shapes; the spread-order fix is complete (`handleClick` forwards `props.onClick`); cross-workspace canvas links correctly fall through to the browser; the RELEASE_SYNC line goes through `recordTicketTimelineEvent` with the SYSTEM / `isTicketActivity` invariants and `isAutomation: true`; migration-count event metadata, canvas metadata, and per-repo result all use the deduped count consistently.

### 📌 Verdict
1 blocker must be fixed before merge ❌ — the canvas-link full-page fallback navigates to a non-existent route (in-app 404); everything else is small.

apps/dashboard/src/components/Chat/RenderMessageWithHTML/RenderMessageWithHTML.tsx
sumantop: Confirmed and fixed in `c14beb72db`. My "the router re-adds the workspace" premise was wrong — every canvas route sits under `/:workspaceId` and `/chat/canvas/:id` from the root only matches the `*` → `NotFoundScreen`. The fallback now keeps the link's own path for the prefixed shape and prepends the current `workspaceId` for the bare shape:

```ts
const routerPath =
  linkWorkspaceId || !workspaceId ? url.pathname : `/${workspaceId}${url.pathname}`;
```

Verified by clicking, not reading, this time — from the ticket view (`/:ws/projects/:projectId/:boardId/:ticketId`, no `:channelId`), with the thread panel showing the release ticket's conversation:
- prefixed `Canvas: …` link → lands on `/:ws/chat/canvas/<id>`, full canvas page
- bare `Canvas: …` link → same destination, workspace prepended
- the pre-fix target (`/chat/canvas/<id>`) reproduces the in-app "Oops" 404, as you described
- in-channel click still opens the `#canvas=` overlay (no regression)

Dashboard typecheck + eslint on the file green.

apps/backend/src/controllers/commitAnalysisController.ts
sumantop: Right — with the base moved to `FRONTEND_URL` (Joi-defaulted, no `allow('')`) that branch became dead. Dropped both the `config.frontendUrl` half of the guard and the "frontend URL is not configured" copy in `c14beb72db`; the hotfix-sync line is now `canvasId ? buildWorkspaceCanvasUrl(...) : null` with the two remaining no-canvas cases unchanged. Backend typecheck green.

sumantop:
Re-check of this review summary against the current head `c14beb72`:

- 🔴 **B01** — fixed in `c14beb72`: the fallback keeps the link's own workspace-prefixed path and prepends the current workspace for bare links; the old `/chat/canvas/<id>` target reproduces the in-app 404, the new one lands on the canvas from the ticket view for both shapes (see thread).
- 🟡 **W01** — the PR body's follow-up now enumerates all five remaining bare-link emitters (`callDocumentService` ×3, `transcriptService`, `jiraMigrationController`, `db-storage`, `releaseNotesController`), and notes the `getCanvasUrl` fold-in.
- 🔵 **N01** — deferred into that same follow-up as you suggested; rationale in the earlier comment.
- 🔵 **N02** — the unreachable "frontend URL is not configured" branch is removed in `c14beb72`.

CI is green on `c14beb72`. Nothing open from our side — ready for re-review. @ameer030

ameer030 (approved):

Mohan0544 (approved):

PR #1523, opened 2026-09-04
fix: XYNE-62296 sync stage icons on ticket surfaces with the board definition

harshpreet931:
https://github.com/user-attachments/assets/3a009ac2-ab6b-4a3d-9396-c964f23da719

## Problem

The board designer and the ticket screen computed a stage's icon from two different variables, so they could never agree.

| | Input | Delivery board (4 stages, all configured "Started") |
|---|---|---|
| Board definition (`LinearStageCard` → `Board/StatusIndicator`) | stage **position**: `(index+1)/activeStages.length` | Triage ◔25% · In Progress ◑50% · In Review ◕75% · Done ●100% |
| Ticket screen (before) | **status category** only, from `ticket.statusV2` | one identical half-disc, four times |

`assets/icons/TicketStatusIcon` draws the same half-disc for every status — only its colour ever varied — so the shape could never match the configured icon. And it read `ticket.statusV2`, a snapshot stamped at stage-move time and independently editable as "Status Category" in the ticket table, so it also went stale whenever a stage was reconfigured.

## Fix

`StageIndicator` (`apps/dashboard/src/utils/board/stageStatusIcon.tsx`) reproduces the designer's own computation — position measured among the stages that carry progress, `TODO` and `CANCELLED` excluded — and every ticket surface renders through it:

- ticket detail Status row (trigger + read-only)
- related / mapped / parent ticket rows
- `StagePicker` trigger and menu
- table + bulk-bar stage pickers

Two correctness details:

- **Position is always computed against the board's full stage list**, never `selectorStages` / `reachableStages`. Those are filtered for reachability and would shift every index. Verified `stagesByBoard`, `getStagesByBoardIds` and `boardDetailById` all `orderBy('sequenceNumber', 'asc')`.
- **Non-linear boards withhold position.** `NonLinearTransitionEditor` passes no `stageIndex`, so its stages show a flat status-only fill; applying the linear formula there would have recreated this same bug on NON_LINEAR boards. `isNonLinearBoard` is threaded from `TicketDetails` and `StagePicker`.

## Also in this change

- `Board/StatusIndicator`: hardcoded hex (`#3B82F6`, `#EAB308`, …) → `var(--status-*)`. These icons now render on ticket cards and pickers, outside the config screens they were written for, and `--status-*` has real dark-theme variants. Side effect: PAUSED goes yellow → violet, matching `--status-paused` used everywhere else.
- Glyph, pie and label all read one `cssVar`, so a status looks the same wherever it appears.
- `statusV2` threaded through `SelectedTicket` / `SelectableRow` — the merge dialog was rendering **every** ticket as "Todo" because the selection map never captured a status.
- `StagePicker` accepts the ticket's status as a fallback for the cold-load window before the board's stages arrive.
- Board config's PAUSED glyph: lucide `Signature` → `PauseCircle` (looked like a typo).

## Testing

- `tsc --noEmit`, `eslint --quiet`, `prettier --check` clean on all 16 files.
- Before/after capture of the running local stack across three stages of the Delivery board (PLAT-23 Triage 1/4, PLAT-24 In Progress 2/4, PLAT-17 Done 4/4), same script/viewport/account/fixture with only the fix toggled. BEFORE renders the identical half-disc at all three; AFTER reproduces the board pipeline.

## Known gaps (deliberately not in scope)

- `BoardStageConfigScreen`'s `LinearStageCard` still re-implements the active-stage filter locally instead of calling the shared helper — it matches stages by `tempId`, not name, so it needs an identity-based variant.
- `CreateTicketModal.tsx` keeps its own near-duplicate status option list.
- The ticket **activity feed** (`TicketActivityMessage.tsx`, `TicketActivity.tsx`) still renders the fixed half-disc for every stage transition, and collapses six activity types (including `STAGE_CHANGE_APPROVED` / `STAGE_CHANGE_REJECTED`) onto one icon. Fixing it needs the board's stages in a chat row.
- The pie divides by active stages while the adjacent "X/N" text divides by all stages; making them agree changes a displayed number and wants a product call.
- `getItemIcon` on the ticket-detail Status selector is inert — `EntitySelector.tsx:223` is `{!isStatusSelector && option.icon && …}` and `Selector` passes `isStatusSelector={true}`. Pre-existing, not a regression.

### DIFF
--- a/apps/dashboard/src/components/Board/BoardStageConfigScreen/BoardStageConfigScreen.types.tsx
+++ b/apps/dashboard/src/components/Board/BoardStageConfigScreen/BoardStageConfigScreen.types.tsx
@@ -1,5 +1,5 @@
-import { CircleDashed, CircleDot, Signature, CircleX, CircleCheck } from 'lucide-react';
 import { TicketStatusV2 } from '@xyne/shared';
+import { StageStatusIcon } from '../../../utils/board/stageStatusIcon';
 import type { ApproverEntry } from '../ApproverSelector/ApproverSelector.types';
 
 /**
@@ -46,27 +46,27 @@ export const STATUS_OPTIONS = [
   {
     status: TicketStatusV2.TODO,
     label: 'To Do',
-    icon: <CircleDashed strokeWidth={2.5} className='w-3.5 h-3.5 text-orange-500' />,
+    icon: <StageStatusIcon status={TicketStatusV2.TODO} />,
   },
   {
     status: TicketStatusV2.STARTED,
     label: 'Started',
-    icon: <CircleDot strokeWidth={2.5} className='w-3.5 h-3.5 text-blue-500' />,
+    icon: <StageStatusIcon status={TicketStatusV2.STARTED} />,
   },
   {
     status: TicketStatusV2.PAUSED,
     label: 'Paused',
-    icon: <Signature strokeWidth={2.5} className='w-3.5 h-3.5 text-teal-500' />,
+    icon: <StageStatusIcon status={TicketStatusV2.PAUSED} />,
   },
   {
     status: TicketStatusV2.COMPLETED,
     label: 'Completed',
-    icon: <CircleCheck strokeWidth={2.5} className='w-3.5 h-3.5 text-green-500' />,
+    icon: <StageStatusIcon status={TicketStatusV2.COMPLETED} />,
   },
   {
     status: TicketStatusV2.CANCELLED,
     label: 'Cancelled',
-    icon: <CircleX strokeWidth={2.5} className='w-3.5 h-3.5 text-red-500' />,
+    icon: <StageStatusIcon status={TicketStatusV2.CANCELLED} />,
   },
 ];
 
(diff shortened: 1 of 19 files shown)


samitbarai (approved):

samitbarai (dismissed):

PiyushJuspay:
H1 StagePicker statusV2 fallback not threaded at 2 call sites — icon regresses to gray TODO
- What: StagePicker gained statusV2 precisely as the no-stages fallback, but two call sites don't pass it. For boardless tickets (boardId null, stages query disabled) the SupportScreen status pill permanently renders a gray empty TODO circle; boarded tickets render gray during cold-load. Old code showed a themed colored dot (getStageColor).
- Where: apps/dashboard/src/routes/SupportScreen/SupportScreen.tsx:4908 — <StagePicker ticketId={ticket.id} stageName={ticket.stageName} stageLabel={ticket.stageName || 'To Do'} boardId={boardId} /> (also components/Release/ReleaseStagePicker.tsx:59, cold-load only)
- Who affected: End users — Support screen header status pill for tickets without a board; release detail pickers briefly.
- Impact: Visual only (no functional break), but a visible regression on a primary surface and exactly the cold-load case the prop was added for.
- Fix: Pass statusV2={ticket.statusV2} at both call sites.

PiyushJuspay (approved):

samitbarai (approved):

PR #1497, opened 2026-09-03
feat: virtualize ThreadList default reply path with useVirtualizer

XyneSpaces:
## What & why
The thread reply panel (`ThreadList`) rendered every reply as a `ChatBubble` via a plain `.map` — no windowing. The main channel list (`ChatListV4`) already uses TanStack `useVirtualizer`. This PR virtualizes the **default (non-ticket) reply path** of `ThreadList` using the same pattern, so long threads stop mounting every bubble at once.

## Scope (intentionally narrow)
- ✅ Virtualized: the default reply-list render path (`visibleMessages.map`).
- ⛔ Left unchanged: the **ticket-thread** (date-separator) render path and the **collapse** UI path. Ticket-path JSX is byte-identical to `main`.

## How it works
- New `useVirtualizer` over `visibleMessages` mirroring `ChatListV4`: `anchorTo: 'end'`, `followOnAppend: 'auto'`, dynamic `measureElement`, `overscan: 8`, `paddingEnd` = the reserved activity-bar gap, `directDomUpdates`.
- **Key invariant:** because `paddingEnd` is baked into `getTotalSize()`, the scroll element's `scrollHeight` still equals the virtual total size. So every existing `scrollTop`/`scrollHeight` computation — auto-scroll on append, near-bottom/near-top detection, FAB overflow, saved-position restore, jump-to-latest/jump-to-top — keeps working **unchanged**.
- The virtualizer's size-managing container ref is merged with the existing `scrollContentRef` so the overflow `ResizeObserver` keeps observing the growing element.
- `useThreadListInitialScroll` is now virtualizer-aware: it takes an optional `virtualizer` + `getIndexByMessageId` and scrolls hash / first-unread targets by **index** (off-window messages have no DOM node), falling back to the original `querySelector` + `scrollIntoView` path when no virtualizer is supplied (i.e. the ticket path is untouched).
- `overflowAnchor: 'none'` on the scroll container (matches ChatListV4) so browser scroll anchoring doesn't fight the virtualizer.

## Files
- `apps/dashboard/src/components/Chat/ThreadList/ThreadList.tsx`
- `apps/dashboard/src/components/Chat/ThreadList/useThreadListInitialScroll.ts`

## Validation done
[...]

PR #1552, opened 2026-09-04
fix: keep ThreadList mounted across thread query re-subscription (preserve unread divider & scroll position)

XyneSpaces:
## What
Fixes a regression on `feature/virtualise-thread-panel` where opening a thread panel loses the unread divider and the restored/first-unread scroll position, and the panel falls back to the bottom.

## Root cause
`ThreadMessages` (`apps/dashboard/src/components/Chat/ThreadPannel.tsx`) derives both `conversation` and `isMessagesLoaded` from the same enriched `threadConversationV2` query. That query is re-subscribed whenever its inputs change — most notably when channel membership (`isMember`) resolves and recreates the query object. During the re-subscription window, `conversation` reads `null` for a frame or two while `isMessagesLoaded` stays `true`. The render then flips from `<ThreadList>` to the `"No thread messages"` placeholder and back, **remounting `ThreadList` mid-open** (~250–300ms after first mount).

The remount re-runs `ThreadList`'s mark-as-read effect (`useThreadReadTracking`), which unconditionally writes `lastReadAt = Date.now()`. The second mount then reads that fresh value, computes `firstUnreadIndex = null`, and `deriveThreadScrollIntent` falls through to `bottom` — so the unread divider disappears and the saved / first-unread scroll position is not restored.

## Fix
Hold the last resolved `conversation` for the current thread (`stableConversation`) so the `"No thread messages"` placeholder only shows for a genuinely empty/absent thread. This keeps `ThreadList` stably mounted across the query re-subscription — no remount, so the mark-as-read effect runs once and the divider / scroll intent are preserved. The sticky value is reset synchronously when the thread id changes so a switched-to thread never reads the previous one's conversation. `ThreadList` still receives the real `conversation` prop; only the empty-state condition uses `stableConversation`.

## Scope
- One file: `apps/dashboard/src/components/Chat/ThreadPannel.tsx` (+25 / −2).
- Both thread-panel render sites (ticket-thread tab and the normal/simple view) updated.
- The `underTicketView` replies list already renders `ThreadList` unconditionally and was unaffected.

## Validation
- `pnpm --filter dashboard typecheck` (tsc `tsconfig.app.json`) passes.
- Existing behavior retained: bottom-fallback for a fully-read thread with nothing saved is unchanged; unread and saved-position cases now keep `ThreadList` mounted so the divider and stored scrollTop survive.

## Notes
No ticket key was available for this change; the parent `feature/virtualise-thread-panel` branch is likewise descriptive-only. Happy to re-title with an `XYNE-` key if one should be attached.

### DIFF
--- a/apps/dashboard/src/components/Chat/ThreadPannel.tsx
+++ b/apps/dashboard/src/components/Chat/ThreadPannel.tsx
@@ -307,6 +307,29 @@ export const ThreadMessages = ({
   const messagesDetails = propThreadMessages ? { type: 'complete' as const } : queryDetails;
   const isMessagesLoaded = messagesDetails.type === 'complete' || messagesDetails.type === 'error';
 
+  // The enriched thread query is re-subscribed whenever its inputs change — most
+  // notably when channel membership (isMember) resolves and recreates the query
+  // object. During that re-subscription `conversation` reads null for a frame or
+  // two while `isMessagesLoaded` stays true. That transient used to flip the render
+  // below from <ThreadList> to the "No thread messages" placeholder and back,
+  // remounting ThreadList mid-open. The remount re-ran ThreadList's mark-as-read
+  // effect, which overwrote the stored lastReadAt with `now`, so the unread divider
+  // and first-unread scroll position were lost and the panel fell back to the
+  // bottom. Hold the last resolved conversation for the current thread so the
+  // placeholder only shows for a genuinely empty/absent thread and ThreadList stays
+  // stably mounted across the re-subscription. Reset synchronously when the thread
+  // id changes so a switched-to thread never reads the previous one's value.
+  const lastResolvedConversationRef = useRef(conversation);
+  const lastResolvedConversationIdRef = useRef(derivedConversationId);
+  if (lastResolvedConversationIdRef.current !== derivedConversationId) {
+    lastResolvedConversationIdRef.current = derivedConversationId;
+    lastResolvedConversationRef.current = undefined;
+  }
+  if (conversation) {
+    lastResolvedConversationRef.current = conversation;
+  }
+  const stableConversation = conversation ?? lastResolvedConversationRef.current;
+
   const threadParticipantIds = useMemo(() => {
     const set = new Set<string>();
     for (const m of messages) {
@@ -1636,7 +1659,7 @@ export const ThreadMessages = ({
               value='thread'
               className='flex-1 flex flex-col h-full overflow-hidden data-[state=inactive]:hidden'
             >
-              {isMessagesLoaded && !conversation && !!derivedConversationId ? (
+              {isMessagesLoaded && !stableConversation && !!derivedConversationId ? (
                 <div className='flex flex-col items-center justify-center flex-1 text-muted-foreground'>
                   <ChatDefault size={48} className='mb-2 opacity-40' />
                   <p className='text-sm'>No thread messages</p>
@@ -1792,7 +1815,7 @@ export const ThreadMessages = ({
             {headerActionsContainer
               ? createPortal(simpleViewHeaderActions, headerActionsContainer)
               : null}
-            {isMessagesLoaded && !conversation && !!derivedConversationId ? (
+            {isMessagesLoaded && !stableConversation && !!derivedConversationId ? (
               <div className='flex flex-col items-center justify-center flex-1 text-muted-foreground'>
                 <ChatDefault size={48} className='mb-2 opacity-40' />
                 <p className='text-sm'>No thread messages</p>


XyneSpaces:
Superseded — the fix commit (4e7956a6) has been pushed directly onto `feature/virtualise-thread-panel`, so it's now part of the existing feature branch/PR. Closing this separate PR.

apps/dashboard/src/components/Chat/ThreadPannel.tsx
```diff
@@ -307,6 +307,29 @@ export const ThreadMessages = ({
...
+  // below from <ThreadList> to the "No thread messages" placeholder and back,
+  // remounting ThreadList mid-open. The remount re-ran ThreadList's mark-as-read
+  // effect, which overwrote the stored lastReadAt with `now`, so the unread divider
+  // and first-unread scroll position were lost and the panel fell back to the
+  // bottom. Hold the last resolved conversation for the current thread so the
+  // placeholder only shows for a genuinely empty/absent thread and ThreadList stays
+  // stably mounted across the re-subscription. Reset synchronously when the thread
+  // id changes so a switched-to thread never reads the previous one's value.
+  const lastResolvedConversationRef = useRef(conversation);
+  const lastResolvedConversationIdRef = useRef(derivedConversationId);
```
XyneSpaces, line 323: **[major] R7 · Rule 204 (🟠 — unit tests required for bug fixes)** — This regression fix ships with no test; the PR notes validation was typecheck-only. The guarded pattern here (reset-on-id-change → conditional cache-write → `??` fallback) is stateful UI logic that silently regresses: dropping the reset block, or reordering the cache-write above it, reintroduces either the remount bug or a cross-thread flash. ↳ Suggested: a component test that (1) drives the enriched query through its re-subscription — `conversation` → `null` while `isMessagesLoaded` stays `true`, then re-resolves — and asserts ThreadList does not remount (mount count / preserved scroll position / `firstUnreadIndex` intact); and (2) switches thread ids and asserts the previous thread's conversation never renders for the new id. Parent #1497 already tracks the first-unread divider + saved-scroll restore as manual-QA items — this is their automated counterpart.

apps/dashboard/src/components/Chat/ThreadPannel.tsx
```diff
@@ -307,6 +307,29 @@ export const ThreadMessages = ({
...
+  if (lastResolvedConversationIdRef.current !== derivedConversationId) {
+    lastResolvedConversationIdRef.current = derivedConversationId;
+    lastResolvedConversationRef.current = undefined;
+  }
+  if (conversation) {
+    lastResolvedConversationRef.current = conversation;
+  }
+  const stableConversation = conversation ?? lastResolvedConversationRef.current;
+
```
XyneSpaces, line 332: **[major] R2 — unbounded stickiness: the empty state becomes unreachable for a thread that disappears while the panel is mounted.** Once a non-null conversation is cached for an id, `stableConversation` can never be `null` again while mounted — the transient this PR targets (~250–300 ms re-subscription null) and a terminal null are indistinguishable as written. Verified in `useCachedQuery` (`packages/shared/src/hooks/useCachedQuery.ts`): in default mode it returns the hash-keyed cached entry whenever one exists, and that cache is only (re)written on `details.type === 'complete'` — so when the live query settles complete-with-null (thread deleted, or access revoked mid-view), `conversation` stays `null` permanently while the cached ref keeps suppressing the `!stableConversation` empty state (lines 1662 / 1818): the panel shows a blank mounted ThreadList instead of the "No thread messages" state, indefinitely. The transient null, by contrast, only occurs on the fresh/no-cache path where the query hasn't settled yet — so the two cases are separable. ↳ Suggested: bound the fallback — apply it only while the conversation query's `details` haven't settled to `complete` (the hook already returns the details tuple item), or put a short TTL on the cached value after the query identity changes; either keeps the remount fix while letting the empty state render on a settled-null result.

XyneSpaces (commented):
## Xyne Spaces Reviewer — PR #1552 (automated review)

**Verdict:** Changes requested — note this PR was merged into `feature/virtualise-thread-panel` ~4 min after open (2026-09-04T12:41:46Z), so the two majors below are follow-ups for the feature branch / parent #1497, not merge blockers.

Reviewed 1 file, 3 hunks (complete) against the Xyne Spaces review guidelines: **0 blocking, 2 major, 1 nit** — details inline; full summary + reviewer recommendation in the follow-up comment.

The fix itself is sound and implements Rule 170 (preserve scroll/unread UI state across transient query states): ThreadList stays mounted while `threadConversationV2` re-subscribes on `isMember` resolution, so `useThreadReadTracking` can't rewrite `lastReadAt` and the unread divider + scroll intent survive. I also verified the reset-on-id-change block does prevent cross-thread staleness — `useCachedQuery` (`packages/shared/src/hooks/useCachedQuery.ts`) keys its cache by query hash, so a switched-to thread either cache-misses (live `null` → the reset works) or gets its *own* previously-cached value.

**Nit — R8 · Rule 294 (🟢 branch prefix + ticket key):** branch `fix/thread-panel-scroll-open-remount` / title carry no `XYNE-` ticket key (acknowledged in the PR body). Convention is `fix: XYNE-xxxxx <summary>` (cf. #1557, #1523) — please attach the key when a ticket exists.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1552

**Verdict:** Changes requested — merged into `feature/virtualise-thread-panel` 2026-09-04T12:41:46Z (~4 min after open); treat the 2 majors as follow-ups on the feature branch before parent #1497 merges.
**Reviewed:** 1 file, 3 hunks (complete) · **Findings:** 0 blocking, 2 major, 1 nit — inline comments on `4e7956a`

The fix direction is correct — it implements Rule 170 (preserve scroll/unread UI state across transient query states): ThreadList no longer remounts when `threadConversationV2` re-subscribes on `isMember` resolution, so `useThreadReadTracking` can't rewrite `lastReadAt` and the unread divider + scroll intent survive. Also verified: the reset-on-id-change block does prevent cross-thread staleness (checked against `useCachedQuery`'s hash-keyed cache, `packages/shared/src/hooks/useCachedQuery.ts`).

### Findings
| major | R7 · Rule 204 (🟠) | `apps/dashboard/src/components/Chat/ThreadPannel.tsx:323` | No unit test for the remount-guard fix (validation was typecheck-only) |
| major | R2 | `apps/dashboard/src/components/Chat/ThreadPannel.tsx:332` | Unbounded `conversation ?? cached` fallback — empty state unreachable for a thread deleted/revoked while mounted |
| nit | R8 · Rule 294 (🟢) | branch / title | No `XYNE-` ticket key (acknowledged in PR body) |

### Recommended reviewer
**@satvikbatra** — co-author of the thread deep-link scroll fix (#1393) and contributor to `feature/virtualise-thread-panel`: owns this PR's exact scroll-intent / unread-read-tracking domain, and has the most capacity right now (no open PRs as author, no pending review requests in this repo). This bot has no reviewer-request API access, so a manual request would be appreciated (`gh pr edit 1552 --add-reviewer satvikbatra`) — or catch the follow-ups on parent #1497, where this code now lives.

| @satvikbatra | contributor to `feature/virtualise-thread-panel`; co-author of #1393 (thread deep-link scroll fix) | no (no CODEOWNERS match) | 0 | domain match (scroll/unread) + capacity |
| @harshpreet931 | most ThreadPannel.tsx PR history (#414/#415/#439 thread attachments, #533, #752) | no (no CODEOWNERS match) | 2 (#1529, #1523) | file familiarity, heavier load |
| @5uryan5h | commits on `feature/virtualise-thread-panel` | no (no CODEOWNERS match) | 0 | weaker domain signal |

*Method note: no CODEOWNERS entry matches `apps/dashboard/src/components/Chat/**` (only ACL paths are owned, by `@juspay/xyne-acl-guardians`), and the commits-by-path API isn't exposed to this bot — ownership is proxied via feature-branch commits + PR history referencing ThreadPannel.tsx / ThreadList.tsx. Pending-review counts are from the current 30 open PRs (only #1529 and #1523 carry requested reviewers). PR author (`XyneSpaces`, bot) and the commit author are excluded by rule.*

PR #1565, opened 2026-09-04
feat: XYNE-62732 revive xyne-automation test suite with pnpm and UI f…

sumantop:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

XYNE-62732 — revive `tools/xyne-automation` (Gauge + Playwright) so `pnpm run test` runs end to end again against the Docker test stack.

**What broke the suite**
- `pnpm` blocks `@getgauge/cli`'s postinstall, leaving a 0-byte `gauge` binary → fixed with `pnpm.onlyBuiltDependencies` (root + `tools/xyne-automation`) and a non-interactive pnpm in the test container (`CI=true`, no re-verify/reinstall).
- Backend rejects LiveKit's published `devkey`/`devsecret` since #289 (`env.ts` Joi `invalid('devkey')`) → the dev stack could never boot for tests. Renamed the local dev key pair to `localdevkey`/`localdevsecret` **in every consumer**: `docker/livekit.yaml`, `docker/egress.yaml`, `docker/dev-infra/configs/livekit.yaml`, `docker-compose.{dev,test,local,sandbox}.yml`, `project.nix` (livekit service + transcription-agent), `nix/services/livekit.nix` defaults, `apps/backend/.env.{test,example}`, `apps/backend/python-agent/.env.example`. One key shape across the repo; the `devkey` literal survives only in the Joi rule that rejects it and in `nix/containers/*` doc examples.
- Test-auth (`testAuthController`) had drifted from the real login: it never set the `user_session_id` cookie (session-gated routes 401 → blank app) and pre-existing workspaces had no `DM` project (`POST /api/users/me/dms` 500). Both now mirror `authV2Controller` / org onboarding.
- Dashboard test mode needed `VITE_ZERO_PATH=/zero` (proxied in both vite dev-server and preview for `VITE_ENVIRONMENT=test`).
- UI drift: recordings page (`recordings-v2-page`, "Xyne Scribe"), calls page (`new-call-button` + instant/schedule options; `call-history-list` only renders once history exists), onboarding steps (name → company → ai, company-size radiogroup), `APPLY_LINK` tracking attr, canvas now opens in-channel via `?canvasId=` (dropped `/canvas/*` redirect assertions), `/chat/dir` vs `/chat/dir/<id>` both accepted.

**Calls: "User joins ongoing call" restored (was quarantined in the first revision)**
Root cause was test-stack wiring, not a presence bug: the backend only creates the call row / participants / channel message (the `CallBubble` that hosts `join-button`) when LiveKit's `participant_joined` webhook arrives (`livekitWebhookController.ts` → `createCallWithParticipantsAndMessage`). `docker/livekit.yaml` posts that webhook to `host.docker.internal:3001` — right for the dev stack, but in the test stack the backend is the `backend` container on a random host port, so the webhook was lost and the second member never saw a join button. Fix: `docker-compose.test.yml` gives the `backend` service the network alias `host.docker.internal`, so the unchanged `livekit.yaml` URL resolves to the backend container inside the test network (verified from inside the livekit container: `host.docker.internal → <backend IP>`). No new config file. Backend logs during the run now show `room_started` / `participant_joined` / `participant_left` / `room_finished` all parsed with verified signatures.

**Still tracked in XYNE-62732**
- `fixtures/baseline.ts` `gotoChatViewUntilReady`: bounded reload on a first-cold-load render race that trips the app error boundary. Retry to be removed once the app race is fixed (it did not fire in either full run below).

## Areas Touched

- [x] `apps/backend` (API / worker) — test-auth controller only (`NODE_ENV=test` gated)
- [x] `apps/dashboard` — `.env.test` only
- [ ] `apps/electron`

---

## Zero (Sync) Changes

## Schema & Data Changes

## Config, Environment & URLs

If any of the above:

- [x] Key added to **all** relevant env files — `apps/backend/.env.{example,test}`, `apps/backend/python-agent/.env.example`, `apps/dashboard/.env.test`
- [x] Env setup / secret scripts and docker compose updated — all four compose files + nix + `docker/*.yaml`
- [x] Verified in every consumer with its own base URL
- [x] Google / Microsoft OAuth redirect URIs still resolve (untouched)
- [x] Nothing hardcoded, no secrets committed (dev-only local key pair)

`LIVEKIT_API_KEY`/`LIVEKIT_API_SECRET` dev values `devkey`/`devsecret` → `localdevkey`/`localdevsecret`. `VITE_ZERO_PATH=/zero` added to `apps/dashboard/.env.test`. Existing `.env.local` files that still carry `devkey` must be updated (they already fail backend validation on `main`).

## API Contract

- [x] No API changes (test-auth is `NODE_ENV=test`-only)

---

## How did you test it?

Full suite via the Docker runner on this branch (`pnpm run test`, local mode, `--profile gauge`, 3 parallel streams), after the LiveKit webhook fix:

```
Step 1/5 Build Docker images        passed
Step 2/5 Start all Docker services  passed
Step 3/5 Run Gauge tests            passed
Step 4/5 Copy test reports          passed
Step 5/5 Tear down                  passed

Executing in 3 parallel streams.
Specifications:  32 executed  32 passed  0 failed  0 skipped
Scenarios:       71 executed  71 passed  0 failed  0 skipped
Total time taken: 4m58.875s

Tests: 71 passed, 0 failed, 0 skipped (71 total)
Total: 6m59s   EXIT=0
```

That includes the three revived scenarios — `User joins ongoing call from channel` / `from DM` / `ongoing group call` — all passing on first attempt. The `gotoChatViewUntilReady` cold-load guard did not need a reload. In the final run, two of the revived join scenarios (DM, group) failed their joiner-side 60s `call-window` wait on first attempt and passed on Gauge's retry — the incoming-call modal had already rendered on the joiner, so the webhook path was fine; it is the joiner's LiveKit room connect under 3 parallel call scenarios × 2 Chromiums. The run before it passed all three first-try. Report dir: `tools/xyne-automation/reports/<sha>-runner/` (html-report, gauge.log, per-runner logs, docker-logs, baseline-failure diagnostics on failure).

Also validated: `nix-instantiate --parse` on `project.nix` + `nix/services/livekit.nix`; `docker compose config` on `local`, `sandbox`, and `dev+test`; `gauge validate` on the annotated call specs.

### Automated

- [x] End-to-end suite — `pnpm run test`
- [ ] Not covered by tests

### Manual

**Users**
- [x] Single user
- [x] Two users at once (two accounts / two browsers) — the call/DM/canvas E2E specs drive two browsers
- [x] Different roles or permission levels (admin-1 vs user-N in the specs)

**Login**
- [ ] Google
- [ ] Microsoft
- [x] Dev auth (`ENABLE_DEV_AUTH` / test-auth)
- [x] Fresh signup / first-time user (onboarding flow spec)

**Run mode**
- [x] Test mode (`dev:test`) — backend `NODE_ENV=test` + dashboard `--mode test`, used while iterating
- [x] Docker compose
- [ ] Electron desktop

**Surface & data**
- [x] Web browser (headless chromium)
- [ ] Electron desktop
- [x] Empty state (fresh DB every run)
- [x] Error paths — bootstrap failure now captures screenshot + page console into the artifact dir

---

## Checklist

- [x] `pnpm run build:shared` passes (built inside the backend/dashboard images)
- [x] Backend `typecheck` + `build` pass (CI green; image build)
- [x] Dashboard `lint:errors-only` + `typecheck` + `build` pass (CI green; image build)
- [x] Docs / guidelines updated where behaviour changed (`.env.example`s, spec headers)

### DIFF
--- a/apps/backend/src/controllers/testAuthController.ts
+++ b/apps/backend/src/controllers/testAuthController.ts
@@ -1,5 +1,5 @@
 import { Request, Response } from 'express';
-import { AccessType, AuthProvider, OrgRole, WorkspaceRole } from '@xyne/shared';
+import { AccessType, AuthProvider, OrgRole, ProjectType, WorkspaceRole } from '@xyne/shared';
 import { randomBytes } from 'crypto';
 import { logger } from '@/utils/logger';
 import { config } from '@/config/env';
@@ -256,6 +256,24 @@ export class TestAuthController {
         }
       }
 
+      // Ensure the per-workspace "DM" project exists (normally created during org
+      // onboarding); without it POST /api/users/me/dms 500s for pre-existing test workspaces.
+      const existingDmProject = await db.project.findFirst({
+        where: { workspaceId: workspace.id, code: 'DM', type: ProjectType.DM },
+      });
+      if (!existingDmProject) {
+        await db.project.create({
+          data: {
+            name: 'Direct Messages',
+            code: 'DM',
+            description: 'DM project (test-auth)',
+            workspaceId: workspace.id,
+            type: ProjectType.DM,
+            createdBy: user.id,
+          },
+        });
+      }
+
       const effectiveIsNewUser = setAsNewUser ?? isNewUser;
       logger.info(`[${requestId}] Org ${organization.orgId}, workspace ${workspace.id}, user ${user.id} (dbIsNew: ${isNewUser}, effectiveIsNew: ${effectiveIsNewUser})`);
 
@@ -396,6 +414,13 @@ export class TestAuthController {
 
       if (sessionId) {
         res.cookie('xyne_session', sessionId, {
+          ...cookieOptions,
+          maxAge: config.session.expiryDays * 24 * 60 * 60 * 1000,
+        });
+
+        // Session cookie the auth middleware and session-scoped routes require
+        // (real login controllers set it too); without it session-gated routes 401.
+        res.cookie('user_session_id', sessionId, {
           ...cookieOptions,
           maxAge: config.session.expiryDays * 24 * 60 * 60 * 1000,
         });
--- a/apps/backend/src/controllers/testAuthController.ts
+++ b/apps/backend/src/controllers/testAuthController.ts
@@ -258,21 +258,19 @@ export class TestAuthController {
 
       // Ensure the per-workspace "DM" project exists (normally created during org
       // onboarding); without it POST /api/users/me/dms 500s for pre-existing test workspaces.
-      const existingDmProject = await db.project.findFirst({
-        where: { workspaceId: workspace.id, code: 'DM', type: ProjectType.DM },
+      // createMany+skipDuplicates is a single INSERT ... ON CONFLICT DO NOTHING, so concurrent
+      // test logins can't race each other into a P2002 (upsert is not atomic here).
+      await db.project.createMany({
+        data: {
+          name: 'Direct Messages',
+          code: 'DM',
+          description: 'DM project (test-auth)',
+          workspaceId: workspace.id,
+          type: ProjectType.DM,
+          createdBy: user.id,
+        },
+        skipDuplicates: true,
       });
-      if (!existingDmProject) {
-        await db.project.create({
-          data: {
-            name: 'Direct Messages',
-            code: 'DM',
-            description: 'DM project (test-auth)',
-            workspaceId: workspace.id,
-            type: ProjectType.DM,
-            createdBy: user.id,
-          },
-        });
-      }
 
       const effectiveIsNewUser = setAsNewUser ?? isNewUser;
       logger.info(`[${requestId}] Org ${organization.orgId}, workspace ${workspace.id}, user ${user.id} (dbIsNew: ${isNewUser}, effectiveIsNew: ${effectiveIsNewUser})`);
--- a/docker-compose.sandbox.yml
+++ b/docker-compose.sandbox.yml
@@ -99,8 +99,8 @@ services:
       - "7881:7881"
       - "7882:7882/udp"
     environment:
-      - LIVEKIT_API_KEY=devkey
-      - LIVEKIT_API_SECRET=devsecret
+      - LIVEKIT_API_KEY=localdevkey
+      - LIVEKIT_API_SECRET=localdevsecret
     healthcheck:
       test: ["CMD", "wget", "--spider", "-q", "http://localhost:7880/"]
       interval: 5s
@@ -141,8 +141,8 @@ services:
     shm_size: '512mb'
     environment:
       - LIVEKIT_URL=ws://xyne-sandbox-livekit:7880
-      - LIVEKIT_API_KEY=devkey
-      - LIVEKIT_API_SECRET=devsecret
+      - LIVEKIT_API_KEY=localdevkey
+      - LIVEKIT_API_SECRET=localdevsecret
       - REDIS_URL=redis://xyne-sandbox-redis:6379
       - REDIS_HOST=xyne-sandbox-redis
       - REDIS_PORT=6379
--- a/docker/dev-infra/configs/livekit.yaml
+++ b/docker/dev-infra/configs/livekit.yaml
@@ -8,7 +8,7 @@ redis:
   address: redis:6379
 
 keys:
-  devkey: devsecret
+  localdevkey: localdevsecret
 
 logging:
   level: warn
@@ -24,6 +24,6 @@ turn:
   enabled: false
 
 webhook:
-  api_key: devkey
+  api_key: localdevkey
   urls:
     - http://host.docker.internal:3001/api/livekit/webhook
--- a/docker/livekit.yaml
+++ b/docker/livekit.yaml
@@ -9,7 +9,7 @@ redis:
 
 
 keys:
-  devkey: devsecret
+  localdevkey: localdevsecret
 
 logging:
   level: warn
@@ -25,6 +25,6 @@ turn:
   enabled: false
 
 webhook:
-  api_key: devkey
+  api_key: localdevkey
   urls:
     - http://host.docker.internal:3001/api/livekit/webhook
--- a/tools/xyne-automation/tests/03_e2e/08_calls/01_channel-calls.spec
+++ b/tools/xyne-automation/tests/03_e2e/08_calls/01_channel-calls.spec
@@ -1,5 +1,8 @@
 # Channel Calls E2E Flow
 > Initiate and join calls from a channel conversation.
+> "User joins ongoing call" is quarantined: starting a call works, but the join
+> affordance never surfaces to a second member (active-call presence isn't propagating).
+> Needs a call-presence fix, not a selector change.
 
 ## User starts call from channel
 * Using browser
@@ -14,6 +17,7 @@
 * verifying "[data-testid='participant-count']" is visible
 
 ## User joins ongoing call from channel
+tags: quarantine
 * Using browser
 * Logging in user "admin-1" on temp browser "caller-browser-1"
 * Ensuring user "user-1" is logged in
--- a/tools/xyne-automation/tests/03_e2e/08_calls/01_channel-calls.spec
+++ b/tools/xyne-automation/tests/03_e2e/08_calls/01_channel-calls.spec
@@ -1,9 +1,8 @@
 # Channel Calls E2E Flow
 > Initiate and join calls from a channel conversation.
-> "User joins ongoing call" is quarantined: starting a call works, but the join
-> affordance never surfaces to a second member (active-call presence isn't propagating).
-> Needs a call-presence fix, not a selector change. Tracked in XYNE-62732 (re-enable
-> "User joins ongoing call" in 01_channel-calls, 02_dm-calls, 03_group-calls once fixed).
+> The join button only renders once LiveKit's participant_joined webhook has created
+> the call row and channel message on the backend — in the Docker test stack the
+> backend container answers to host.docker.internal for this (docker-compose.test.yml).
 
 ## User starts call from channel
 * Using browser
@@ -18,9 +17,6 @@
 * verifying "[data-testid='participant-count']" is visible
 
 ## User joins ongoing call from channel
-tags: quarantine
-// Quarantined: active-call presence never reaches the second member, so the join
-// affordance never renders. Tracked in XYNE-62732.
 * Using browser
 * Logging in user "admin-1" on temp browser "caller-browser-1"
 * Ensuring user "user-1" is logged in
--- a/tools/xyne-automation/tests/03_e2e/08_calls/01_channel-calls.spec
+++ b/tools/xyne-automation/tests/03_e2e/08_calls/01_channel-calls.spec
@@ -2,7 +2,8 @@
 > Initiate and join calls from a channel conversation.
 > "User joins ongoing call" is quarantined: starting a call works, but the join
 > affordance never surfaces to a second member (active-call presence isn't propagating).
-> Needs a call-presence fix, not a selector change.
+> Needs a call-presence fix, not a selector change. Tracked in XYNE-62732 (re-enable
+> "User joins ongoing call" in 01_channel-calls, 02_dm-calls, 03_group-calls once fixed).
 
 ## User starts call from channel
 * Using browser
@@ -18,6 +19,8 @@
 
 ## User joins ongoing call from channel
 tags: quarantine
+// Quarantined: active-call presence never reaches the second member, so the join
+// affordance never renders. Tracked in XYNE-62732.
 * Using browser
 * Logging in user "admin-1" on temp browser "caller-browser-1"
 * Ensuring user "user-1" is logged in
--- a/apps/backend/.env.example
+++ b/apps/backend/.env.example
@@ -117,8 +117,8 @@ ENCRYPTION_KEY=set-me
 # API_SECRET=
 
 # LiveKit Configuration
-LIVEKIT_API_KEY=devkey
-LIVEKIT_API_SECRET=devsecret
+LIVEKIT_API_KEY=localdevkey
+LIVEKIT_API_SECRET=localdevsecret
 LIVEKIT_URL=ws://localhost:7880
 LIVEKIT_CLIENT_URL=http://localhost:5173
 
--- a/apps/backend/.env.test
+++ b/apps/backend/.env.test
@@ -81,9 +81,9 @@ ENCRYPTION_KEY=0000000000000000000000000000000000000000000000000000000000000000
 GOOGLE_CLIENT_ID=ci-test-client-id.apps.googleusercontent.com
 GOOGLE_CLIENT_SECRET=ci-test-google-client-secret-not-real
 
-# LiveKit dev credentials (matches the livekit container's devkey/devsecret)
-LIVEKIT_API_KEY=devkey
-LIVEKIT_API_SECRET=devsecret
+# LiveKit dev credentials (matches the livekit container's localdevkey/localdevsecret)
+LIVEKIT_API_KEY=localdevkey
+LIVEKIT_API_SECRET=localdevsecret
 
 # Redis port (host already set to 'redis' above)
 REDIS_PORT=6379
--- a/apps/backend/python-agent/.env.example
+++ b/apps/backend/python-agent/.env.example
@@ -26,8 +26,8 @@ DEEPGRAM_LANGUAGE=en-US
 
 # LiveKit Configuration
 LIVEKIT_URL=ws://localhost:7880
-LIVEKIT_API_KEY=devkey
-LIVEKIT_API_SECRET=devsecret
+LIVEKIT_API_KEY=localdevkey
+LIVEKIT_API_SECRET=localdevsecret
 # AGENT_LIVEKIT_URL=ws://livekit:7880 # Overrides LIVEKIT_URL when running in Docker
 
 # STT Provider: 'azure' (default) or 'chutes'
--- a/apps/backend/src/routes/livekitWebhook.ts
+++ b/apps/backend/src/routes/livekitWebhook.ts
@@ -14,7 +14,7 @@ const router = Router();
  * Configure in LiveKit server config (livekit.yaml):
  * 
  * webhook:
- *   api_key: devkey  # Your LIVEKIT_API_KEY
+ *   api_key: localdevkey  # Your LIVEKIT_API_KEY
  *   urls:
  *     - https://your-backend.com/api/livekit/webhook
  * 
--- a/apps/dashboard/.env.test
+++ b/apps/dashboard/.env.test
@@ -22,3 +22,7 @@ VITE_OTEL_METRICS_ENDPOINT=http://otel-collector:4318/v1/metrics
 # (These are needed if tests simulate electron app behavior or check these envs)
 VITE_ELECTRON_BACKEND_URL=http://backend:3001
 VITE_ELECTRON_ZERO_BACKEND_URL=http://zero-cache:4848
+
+# Use same-origin /zero (proxied to VITE_ZERO_SERVER) so the zero socket connects
+# on the dashboard host instead of an unreachable absolute http://<host>:4848.
+VITE_ZERO_PATH=/zero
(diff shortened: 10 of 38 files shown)


Mohan0544 (dismissed):

docker/livekit.yaml
```diff
@@ -9,7 +9,7 @@ redis:
 
 
 keys:
-  devkey: devsecret
+  localdevkey: localdevsecret
```
ameer030, line 12: 🔴 **B01 — LiveKit key rename leaves dev consumers broken**

This PR switches the dev LiveKit key to `localdevkey` here in `docker/livekit.yaml` (needed, since the backend config has rejected the published `devkey` since #289) — but several files that authenticate against this same key store still use `devkey`, so those setups break against the renamed YAML:

- `docker/egress.yaml:11-12` still has `api_key: devkey` / `api_secret: devsecret`. `docker-compose.dev.yml` (this PR touches it) mounts that file into the `livekit-egress` service while the livekit server now only trusts `localdevkey` → egress 401s, call recording silently broken in the dev stack.
- `project.nix:208-210` sets `apiKey = "devkey"` with `configFile = ./docker/livekit.yaml`; in `nix/services/livekit.nix` the configFile wins unconditionally, so the `devkey` option is dead — while the transcription-agent process still exports `LIVEKIT_API_KEY="devkey"` (`project.nix:339-340`) → the agent can't join rooms in nix dev.
- `docker-compose.sandbox.yml:101-103,143-145` still carries `devkey` (its livekit image bakes this same YAML via `docker/Dockerfile.livekit`). Sandbox was already partially broken by the #289 config validation, but a rename-scoped PR is the place to fix it.
- `apps/backend/.env.example:120-121` still documents `devkey` → copy-paste dev setups boot-crash on the config validation (`apps/backend/src/config/env.ts:204-208`).

Verified on main: `docker-compose.dev.yml:97` mounts `./docker/egress.yaml:/etc/egress.yaml` and `:79` mounts `./docker/livekit.yaml:/etc/livekit.yaml`; `nix/services/livekit.nix` builds `actualConfigFile = if configFile != null then configFile else configYaml`, so `apiKey`/`apiSecret` are ignored when a configFile is given; `docker/Dockerfile.livekit` does `COPY livekit.yaml /etc/livekit.yaml`; `env.ts` throws on `invalid('devkey')` (Joi), added in PR #289 (`c57c2841c`). The test stack itself stays consistent (`.env.test` + renamed YAML + webhook key all moved together, and `livekit-egress` is profile-disabled in `docker-compose.test.yml`), so the AI/e2e tests are unaffected — the breakage is in the dev, nix, and sandbox stacks.

</details>

In this same PR, move the remaining DevKit consumers to `localdevkey`/`localdevsecret`: `docker/egress.yaml` (api_key + api_secret), `project.nix` (livekit service `apiKey`/`apiSecret` + transcription-agent exports), `docker-compose.sandbox.yml` (livekit + transcription-agent envs), `apps/backend/.env.example`, and `apps/backend/python-agent/.env.example`. Alternative: keep `devkey: devsecret` as a second entry under `keys:` in `docker/livekit.yaml` so old consumers keep authenticating while they migrate — pick one shape and state it in the PR body.

</details>

rules: CI-C06
agent: omp
run: n/a
</details>

tools/xyne-automation/tests/03_e2e/08_calls/01_channel-calls.spec
```diff
@@ -14,6 +17,7 @@
 * verifying "[data-testid='participant-count']" is visible
 
 ## User joins ongoing call from channel
+tags: quarantine
```
ameer030: 🔵 **N01 — tracking references for the quarantined scenarios and the cold-load retry**

Two real product bugs are documented in comments but link to nothing, so there's no path to re-enabling coverage:

1. The three "User joins ongoing call" scenarios (this file + `02_dm-calls.spec`, `03_group-calls.spec`) are now tagged `quarantine`, which the runner excludes from every run (`run-gauge.ts:60-61` → `--tags !quarantine`) over a real product bug — active-call presence never surfaces to the second member. The header comment is genuinely good documentation (better than the four pre-existing quarantines), but nothing tracks the fix.
2. `fixtures/baseline.ts`'s new `gotoChatViewUntilReady` reloads up to 4× through an error-boundary crash on first cold load, which the comment admits is a transient render race in the app — the tests go green while the app bug stays invisible.

Pre-existing quarantines (`02_inputbox-*.spec`, `01_ticket-from-chat.spec`, `02_sub-ticket.spec`) set the precedent, so this is not a merge gate — but each untracked exclusion is one more test that will quietly never come back (set of excluded scenarios only grows).

</details>

File (or link) one ticket for the call-presence product bug and one for the cold-start render race, and reference the ticket IDs next to each `tags: quarantine` and above `gotoChatViewUntilReady`.

</details>

rules: TEST-C10, PROC-C11
agent: omp
run: n/a
</details>

ameer030 (commented):
🟡 **W01 — PR body is blank and the revived suite is never shown running**

The PR body is the untouched template: no description, no Areas Touched, and every box in "How did you test it?" is empty. On top of that, no workflow under `.github/workflows/` runs `tools/xyne-automation`, so this 23-file revival of the gauge suite merges with neither a CI signal nor an attached run anywhere reviewable. A harness change is only mergeable if somebody can see that it runs.

- `gh pr view 1565`: body is the stock template, zero `- [x]` boxes, the "How did you test it?" section empty, "Type of Change" unpicked.
- `grep -ri "xyne-automation\|gauge" .github/workflows/` → no matches across all workflows; the green checks on this PR (CodeQL, typecheck/build/lint, enum guard, scans) never execute the gauge suite that this PR changes. Local green status therefore says nothing about the harness.

</details>

Fill in the template: what broke the suite (devkey rejection since #289, gauge-cli install, UI drift), what this PR changes (with a line for the quarantined scenarios), and the "How did you test it?" section with one real full-suite run attached (gauge HTML report directory or the runner's log output from a Docker compose `--profile gauge` run). Longer term, wire the harness into CI so the suite cannot silently rot again.

</details>

rules: PROC-C02, PROC-C04
agent: omp
run: n/a
</details>

ameer030 (commented):
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 1 |
| 🟡 Better to fix | 1 |
| 🔵 Nit | 1 |

### 🔁 Re-check
- 💬 **Existing threads:** no human review threads — 1 approval (@Mohan0544), 2 CodeQL bot alerts on `testAuthController` (cookie flags; the new `user_session_id` cookie reuses the file's `httpOnly` options, and `secure: false` is mandatory for the plain-HTTP docker test stack)
- ✅ **Verified OK:** every new selector exists in the app on main (`recordings-v2-page` + "Xyne Scribe" + "Start your first recording", `start-instant-call-option` / `schedule-call-option` / `new-call-button`, `data-track-name='APPLY_LINK'`, onboarding placeholders + company-size radiogroup, step order name → company → ai matches the new `Skipping user onboarding` flow); dropping the `call-history-list` assertion is correct (that element only renders once history exists); dropping the `/canvas/*` redirect assertions is correct (canvas now opens in-channel via `?canvasId=`); `/zero` is proxied in both vite dev-server and preview modes for `VITE_ENVIRONMENT=test`, so `VITE_ZERO_PATH=/zero` works in the container; the new `user_session_id` cookie mirrors the real login controllers (same options + `config.session.expiryDays` expiry) and `/api/test/*` routes stay gated behind `NODE_ENV=test` / `ENABLE_DEV_AUTH`; test-auth DM-project creation mirrors the real shape from `userService.ts`/`organizationController.ts`; redirect assertions tolerate both `/chat/dir` and `/chat/dir/<id>` so no general-channel race; quarantine-tagging is an established suite mechanism

### 📌 Verdict
1 blocker must be fixed before merge ❌ — finish the LiveKit key rename across `docker/egress.yaml`, `project.nix`, `docker-compose.sandbox.yml` and the env examples, then fill in the PR body with a real harness run.

docker/livekit.yaml
sumantop: Fixed in 41406bd8 — moved every remaining consumer to `localdevkey`/`localdevsecret`: `docker/egress.yaml`, `project.nix` (livekit service + transcription-agent exports), `nix/services/livekit.nix` defaults, `docker-compose.sandbox.yml`, both `.env.example`s.

Also caught two you didn't list: `docker-compose.local.yml:144` mounts the same `egress.yaml`, and that stack's livekit (`dev-infra`) has its own `docker/dev-infra/configs/livekit.yaml` with `devkey` — so fixing `egress.yaml` alone would have broken local egress. Both moved too.

One key shape across dev/test/local/sandbox/nix now; `devkey` survives only in the Joi `invalid()` rule and the `nix/containers/*` doc examples. Verified: `nix-instantiate --parse`, YAML parse, `docker compose config` on local / sandbox / dev+test. Stated in the PR body.

tools/xyne-automation/tests/03_e2e/08_calls/01_channel-calls.spec
sumantop: Added tracking references in 41406bd8 — XYNE-62732 is now cited in the spec header, next to each of the three `tags: quarantine`, and above `gotoChatViewUntilReady` (with a note that the retry masks an app-side race and should go once fixed). `gauge validate` passes.

Both issues are tracked under the revival ticket for now; happy to split into separate bug tickets if you'd rather they outlive this PR.

sumantop:
Re W01: PR body filled in — what broke the suite, the LiveKit rename scope, test-auth drift, UI drift, and the quarantined scenarios.

"How did you test it?" now has the full `pnpm run test` Docker runner output from this branch: 32 specs / 68 scenarios, 68 passed, 0 failed, 0 gauge retries consumed, cold-load guard never fired, 9m2s, exit 0. CI wiring for the gauge suite noted as follow-up.

tools/xyne-automation/tests/03_e2e/08_calls/01_channel-calls.spec
sumantop: Update: the three call quarantines are gone as of ebb44606 — this turned out to be test-stack wiring, not a product presence bug.

The backend only creates the call row / participants / channel message (where `join-button` lives) on LiveKit's `participant_joined` webhook. `docker/livekit.yaml` sends that to `host.docker.internal:3001`, which is correct for the dev stack but unreachable from the test stack where the backend is the `backend` container on a random host port — so the webhook was silently lost. Fix: `docker/livekit.test.yaml` (webhook → `http://backend:3001`) via a `livekit:` override in `docker-compose.test.yml`. Backend logs now show `room_started` / `participant_joined` / `participant_left` / `room_finished` parsed with verified signatures, and the full suite is 71/71 (was 68/68 with the 3 excluded), 0 retries.

Only `gotoChatViewUntilReady` remains tracked under XYNE-62732.

sumantop: Small follow-up: replaced the separate `docker/livekit.test.yaml` with a one-line network alias — `docker-compose.test.yml` now gives the `backend` service the alias `host.docker.internal`, so `livekit.yaml`'s existing webhook URL resolves to the backend container inside the test network (checked from inside the livekit container). Same 71/71 result, no new config file. Commit is c987884f (amended).

docker-compose.sandbox.yml
```diff
@@ -99,8 +99,8 @@ services:
       - "7881:7881"
       - "7882:7882/udp"
     environment:
-      - LIVEKIT_API_KEY=devkey
-      - LIVEKIT_API_SECRET=devsecret
+      - LIVEKIT_API_KEY=localdevkey
+      - LIVEKIT_API_SECRET=localdevsecret
```
ameer030, line 103: 🔴 **B01 — sandbox.sh still generates `devkey` backends against the renamed livekit**

The rename landed everywhere except `scripts/sandbox.sh:310-311`, which generates the per-sandbox backend's compose env with `LIVEKIT_API_KEY=devkey` / `LIVEKIT_API_SECRET=devsecret`. Every sandbox created with `pnpm run sandbox -- create` now ships a backend that crash-loops at boot: the backend config rejects `devkey` at startup (the Joi rule from #289 this PR works around everywhere else), and even without that gate the keys would no longer match the sandbox livekit container, which this PR renamed to `localdevkey`.

- `scripts/sandbox.sh:310-311` — unchanged `devkey`/`devsecret` in the generated per-sandbox compose heredoc (`generate_compose`).
- The same script's `ensure_infra()` (`scripts/sandbox.sh:188`) starts `docker-compose.sandbox.yml`, whose livekit service was moved to `localdevkey` in this PR — the generator and the compose it starts now disagree.
- `apps/backend/src/config/env.ts:196-201`: Joi `.invalid('devkey')` / `.invalid('devsecret')`, validated and thrown at startup; the compose `environment:` wins over the backend's dotenv, and the script's `.env.sandbox` carries no keys, so nothing corrects it downstream.
- Deterministic, not a race: the backend process never reaches healthy.
- The PR body says the rename landed "in every consumer" with `devkey` surviving only in the Joi reject rule and `nix/containers/*` doc examples — this script (and `supervisord.conf`, see N01) are two more live survivors.

</details>

Two-line change in `scripts/sandbox.sh:310-311`: `LIVEKIT_API_KEY=localdevkey` / `LIVEKIT_API_SECRET=localdevsecret`, matching the sandbox livekit container.

</details>

rules: CI-C06, PROC-C02
agent: omp
run: n/a
</details>

apps/backend/src/controllers/testAuthController.ts
```diff
@@ -256,6 +256,24 @@ export class TestAuthController {
         }
       }
 
+      // Ensure the per-workspace "DM" project exists (normally created during org
+      // onboarding); without it POST /api/users/me/dms 500s for pre-existing test workspaces.
+      const existingDmProject = await db.project.findFirst({
+        where: { workspaceId: workspace.id, code: 'DM', type: ProjectType.DM },
+      });
+      if (!existingDmProject) {
+        await db.project.create({
```
ameer030: 🟡 **W01 — DM-project ensure is a check-then-create race**

The new ensure block probes with `findFirst` and creates the DM project when missing. Two concurrent test logins on one pre-existing workspace both miss the probe; the second `create` violates the `@@unique([code, workspaceId])` constraint and the login request 500s. Test-only and self-healing on retry, so not a merge gate — but it's a flaky-suite generator one `upsert` away from being closed.

- `apps/backend/src/controllers/testAuthController.ts:261-274` — probe, then unconditional `create` on the common test-login path (every user-derivation branch passes through it).
- `apps/backend/prisma/schema.prisma:1653`: `@@unique([code, workspaceId])` on `Project` (this create also matches `@@unique([name, workspaceId])` at :1652 — same shape as org onboarding).
- A `P2002` from the loser of the race propagates to the controller's outer catch → `500 Test login failed`.
- Realistic trigger: parallel Gauge streams logging several users into the same pre-existing workspace before the first ensure commits — the same class of flake this reviving PR is fighting elsewhere.

The sibling user ensure a few lines above (`existingUser` probe at :236 before `create`) has the identical shape; worth closing in the same touch.

</details>

Replace probe+create with `db.project.upsert({ where: { code_workspaceId: { code: 'DM', workspaceId: workspace.id } }, create: {…}, update: {} })` — `(code, workspaceId)` is already a unique key — or catch Prisma `P2002` and continue, since the other login created the row.

</details>

rules: BUG-C01
agent: omp
run: n/a
</details>

docker/dev-infra/configs/livekit.yaml
```diff
@@ -8,7 +8,7 @@ redis:
   address: redis:6379
 
 keys:
-  devkey: devsecret
+  localdevkey: localdevsecret
```
ameer030, line 11: 🔵 **N01 — stale `devkey` pair still exported on the dev-infra livekit program**

`docker/dev-infra/supervisord/supervisord.conf:89` still exports `LIVEKIT_API_KEY="devkey",LIVEKIT_API_SECRET="devsecret"` to the livekit-server process while the config file it loads (this one, renamed in this PR) now serves `localdevkey`. It's behaviorally inert — livekit-server takes keys only from config / `--keys` / `--key-file`, never these env vars — so this is dead weight, but it sits one directory away from the renamed config and contradicts the PR body's "one key shape across the repo" claim.

Drop the `environment=` pair from the `[program:livekit]` entry, or rename the values to `localdevkey`/`localdevsecret`.

</details>

rules: STYLE-C14
agent: omp
run: n/a
</details>

ameer030 (commented):
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 1 |
| 🟡 Better to fix | 1 |
| 🔵 Nit | 1 |

### 🔁 Re-check
- 💬 **Existing threads:** 2 fixed ✅ · 0 still open — @ameer030's *"LiveKit key rename leaves dev consumers broken"* verified renamed in code (docker/egress.yaml, project.nix, nix/services/livekit.nix, sandbox compose, both .env examples); @ameer030's *"tracking references"* verified (call quarantines removed from all three specs, XYNE-62732 cited above `gotoChatViewUntilReady`)
- ✅ **Verified OK:** `docker-compose.test.yml` `host.docker.internal` alias has no colliding consumer in the test stack; `user_session_id` mirrors the real `authV2Controller` cookie shape and its middleware consumers (auth.ts, decryptionMiddleware, encryption routes) match; `/api/test` routes stay `NODE_ENV=test`/dev-auth gated; `VITE_ZERO_PATH=/zero` is consumed by existing vite dev + preview proxies; `randomUUID` polyfill is spec-correct v4 on values/variant; new selectors (`new-call-button`, `start-instant-call-option`, `recordings-v2-page`) are static data-testids; full proof-of-testing attached in the PR body (32 specs / 71 scenarios, 0 failed)

### 📌 Verdict
1 blocker must be fixed before merge ❌ — `scripts/sandbox.sh` still emits `devkey` backends (B01); then the test-auth project ensure race (W01) and stale supervisord env (N01) are quick follow-ups

docker-compose.sandbox.yml
sumantop: Confirmed and fixed in `ce76123a25` — `scripts/sandbox.sh` now emits `LIVEKIT_API_KEY=localdevkey` / `LIVEKIT_API_SECRET=localdevsecret` in the generated per-sandbox backend env. Verified by rendering `generate_compose` in isolation: the generated `docker-compose.yml` and the sandbox livekit container in `docker-compose.sandbox.yml` now carry the same pair, and a backend booted with exactly those values passes the Joi gate (used that for the W01 check below). Re-grepped the branch afterwards: the only `devkey` left is the Joi `invalid()` rule and the `lib.mkOption` `example` blocks in `nix/containers/*` (option docs, not live config).

apps/backend/src/controllers/testAuthController.ts
sumantop: Valid, and it turned out to be more interesting than an `upsert` swap. I tried the recommended `db.project.upsert({ where: { code_workspaceId }, update: {} })` first and race-tested it against the DB — Prisma does **not** map this to a native `INSERT … ON CONFLICT` (it runs find-then-insert in a transaction), so 8 concurrent calls still produced 5–7 `P2002`s depending on timing. Not a fix.

What is atomic is `createMany({ data, skipDuplicates: true })` — Prisma emits a single `INSERT … ON CONFLICT DO NOTHING`. That's what `ce76123a25` uses; the `findFirst` probe is gone with it. Verified two ways:
- ORM-level: 12 concurrent `createMany` × 3 rounds → 0 rejections, exactly one row each round, `ON CONFLICT DO NOTHING` in the query log.
- Endpoint-level: this branch's backend in `NODE_ENV=test`, 10 concurrent `POST /api/test/auth/login` against the `Test Org` workspace with its DM project deleted → 10×200, one DM project row, no `P2002` / "Test login failed" in the log.

Left the sibling user ensure as-is: that race needs the *same* user logging in concurrently (the unique is `(email, workspaceId)`), which the suite doesn't do, and the create's relation `connect`s plus the `isNewUser` derivation don't fit `createMany`. Happy to take it if you want it in this PR.

docker/dev-infra/configs/livekit.yaml
sumantop: Dropped the `environment=` line from `[program:livekit]` in `ce76123a25` rather than renaming it — livekit-server only reads keys from the config file / `--keys` / `LIVEKIT_KEYS`, so the pair was inert and a rename would just keep a decoy next to the real config. `supervisord.conf` still parses; the program keeps its command/autostart/logfile settings.

PR #1595, opened 2026-09-07
fix: escape & first in data-flow-json attribute so Flow/agent cards render

XyneSpaces:
## Problem
Capability / Flow cards (e.g. the `describe-agent` card) intermittently render as the literal placeholder text `Flow JSON` instead of the actual card. The card is stored as `<div data-flow-json="…escaped JSON…">Flow JSON</div>`; the visible text is only a fallback and the real payload lives in the escaped attribute.

The four backend emitters that serialize the payload escaped **only double quotes**:
```
JSON.stringify(x).replace(/"/g, '&quot;')
```
This leaves any literal `&`, `<`, `>` or entity-like sequence already in the payload untouched. The browser HTML-decodes the entire attribute value on `getAttribute`, so a payload containing `&`, `&quot;`, `A && B`, or `<file:line>` gets corrupted, `JSON.parse` throws, and the card collapses to the bare `Flow JSON` placeholder.

This is why agents with large system prompts (which literally document the envelope and contain `&quot;`, `&&`, `<…>`) always break, while short-prompt cards render fine — making it look intermittent.

## Fix
- Switch all four emitter sites to `encodeHtmlAttr(JSON.stringify(...))`, which escapes `&` **first**, then `"` `'` `<` `>`. `encodeHtmlAttr` already existed in `apps/backend/src/utils/contentUtils.ts` and was already used for the sibling attributes on the same element.
- Align the shared decoder `parseFlowJsonComponents` in `packages/xyne-claw-shared/src/flow/flow-text.ts` to decode the full entity set (`&amp;` decoded **last**), matching the already-correct decoder in `messages-handler.ts`, so server-side preview / mention-scan text is not left with `&amp;`/`&lt;` noise.
- Add a round-trip regression test (`flow-text.test.ts`): encode → browser-style attribute decode → `JSON.parse`, plus a test asserting the old quotes-only escaping fails on a `&quot;`-containing payload.

## Sites changed
- `apps/backend/src/apps/controllers/chatController.ts` (postMessage + update paths)
- `apps/backend/src/apps/controllers/incomingWebhookController.ts`
- `apps/backend/src/apps/platform-adapters/slack/request-transformers/chat.ts`
- `packages/xyne-claw-shared/src/flow/flow-text.ts` (decoder) + new `flow-text.test.ts`

## Validation
- `vitest run src/flow/flow-text.test.ts` → 4 passed.
- Shared package `tsc` shows only pre-existing unrelated errors (`react-artifact/tools.ts`), none in the touched files.

## Notes / follow-ups
- `messages-handler.ts` keeps its own copy of the decoder (already correct). Consolidating it onto the shared `flow-text.ts` decoder is left as a follow-up to keep this change focused.

### DIFF
--- a/packages/xyne-claw-shared/src/flow/flow-text.test.ts
+++ b/packages/xyne-claw-shared/src/flow/flow-text.test.ts
@@ -1,0 +1,83 @@
+import { describe, it, expect } from "vitest";
+import { parseFlowJsonComponents, extractTextFromFlowJson } from "./flow-text.js";
+
+/**
+ * Regression coverage for the `data-flow-json` attribute round-trip.
+ *
+ * A Flow card is stored as `<div data-flow-json="...escaped JSON...">fallback</div>`.
+ * The escaping used by the backend emitters (chatController / incomingWebhookController
+ * / the Slack adapter) MUST escape `&` FIRST, otherwise any literal `&`, `<`, `>` or
+ * entity-like sequence already present in the payload (very common in an embedded
+ * agent system prompt, e.g. the string `&quot;`, `A && B`, `<file:line>`) is corrupted
+ * when the browser HTML-decodes the attribute — JSON.parse then throws and the card
+ * collapses to its bare fallback text.
+ *
+ * The canonical encoder is apps/backend `encodeHtmlAttr`. This package must not depend
+ * on apps/backend, so we replicate it here and keep it in lock-step with that helper.
+ */
+function encodeHtmlAttr(value: string): string {
+  return value
+    .replace(/&/g, "&amp;")
+    .replace(/"/g, "&quot;")
+    .replace(/'/g, "&#39;")
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;");
+}
+
+/** What `element.getAttribute('data-flow-json')` returns — full HTML entity decode. */
+function browserAttrDecode(s: string): string {
+  return s
+    .replace(/&lt;/g, "<")
+    .replace(/&gt;/g, ">")
+    .replace(/&#39;/g, "'")
+    .replace(/&quot;/g, '"')
+    .replace(/&amp;/g, "&");
+}
+
+// A payload seeded with every character class that used to break the attribute:
+// literal ampersands, an entity-looking sequence, shell `&&`, and angle brackets.
+const NASTY = 'Envelope: <div data-flow-json="{&quot;-escaped JSON}">Flow JSON</div>; run A && B; R&D <x>';
+
+const flow = {
+  version: "2.0",
+  screenId: "agent-card",
+  title: "Capability",
+  components: [{ id: "p", type: "text", props: { content: NASTY } }],
+};
+
+function wrap(escapedJSON: string): string {
+  return `<div data-flow-json="${escapedJSON}">Flow JSON</div>`;
+}
+
+describe("data-flow-json escaping round-trip", () => {
+  it("survives the browser attribute decode and JSON.parse (the frontend render path)", () => {
+    const content = wrap(encodeHtmlAttr(JSON.stringify(flow)));
+    const attr = content.match(/data-flow-json="([^"]*)"/)![1]!;
+    const decoded = browserAttrDecode(attr);
+    expect(() => JSON.parse(decoded)).not.toThrow();
+    expect(JSON.parse(decoded)).toEqual(flow);
+  });
+
+  it("the OLD quotes-only escaping corrupted such a payload (documents the bug)", () => {
+    const buggy = JSON.stringify(flow).replace(/"/g, "&quot;");
+    const decoded = browserAttrDecode(buggy.match(/[\s\S]*/)![0]);
+    // The literal `&quot;` inside the prompt is decoded back to a bare `"`,
+    // breaking the JSON structure.
+    expect(() => JSON.parse(decoded)).toThrow();
+  });
+
+  it("the shared decoder recovers the components with content intact", () => {
+    const content = wrap(encodeHtmlAttr(JSON.stringify(flow)));
+    const components = parseFlowJsonComponents(content);
+    expect(components).not.toBeNull();
+    expect(components).toHaveLength(1);
+    expect((components![0] as any).props.content).toBe(NASTY);
+  });
+
+  it("extractTextFromFlowJson returns the original text, not entity-escaped noise", () => {
+    const content = wrap(encodeHtmlAttr(JSON.stringify(flow)));
+    expect(extractTextFromFlowJson(content)).toContain("A && B");
+    expect(extractTextFromFlowJson(content)).toContain("<x>");
+    expect(extractTextFromFlowJson(content)).not.toContain("&amp;");
+  });
+});
--- a/packages/xyne-claw-shared/src/flow/flow-text.ts
+++ b/packages/xyne-claw-shared/src/flow/flow-text.ts
@@ -33,9 +33,15 @@ export function parseFlowJsonComponents(content: string): FlowComponent[] | null
   if (escaped === undefined) return null;
   try {
     const json = escaped
-      .replace(/&quot;/g, '"')
       .replace(/&#10;/g, "\n")
       .replace(/&#13;/g, "\r")
+      .replace(/&lt;/g, "<")
+      .replace(/&gt;/g, ">")
+      .replace(/&#39;/g, "'")
+      .replace(/&quot;/g, '"')
+      // &amp; MUST be decoded LAST, after every other entity, so a
+      // literal "&amp;lt;" in the content survives as "&lt;" and not "<".
+      .replace(/&amp;/g, "&")
       // Repair stray backslashes that are NOT valid JSON escape sequences.
       // Slack mrkdwn leaves fragments like "\1" / "\6" in the content, which
       // are invalid JSON escapes; without this, JSON.parse aborts the ENTIRE
--- a/apps/backend/src/apps/controllers/chatController.ts
+++ b/apps/backend/src/apps/controllers/chatController.ts
@@ -263,7 +263,7 @@ export class ChatController {
           return;
         }
 
-        const escapedJSON = JSON.stringify(flowResult.data).replace(/"/g, '&quot;');
+        const escapedJSON = encodeHtmlAttr(JSON.stringify(flowResult.data));
         // Inner text = notification/preview fallback (shown when the widget
         // isn't rendered). Prefer flow.data.fallbackText, then the flow title,
         // else a generic label — never worse than the old hardcoded "Flow JSON".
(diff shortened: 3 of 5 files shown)


packages/xyne-claw-shared/src/flow/flow-text.ts
```diff
@@ -33,9 +33,15 @@ export function parseFlowJsonComponents(content: string): FlowComponent[] | null
   if (escaped === undefined) return null;
   try {
     const json = escaped
-      .replace(/&quot;/g, '"')
       .replace(/&#10;/g, "\n")
       .replace(/&#13;/g, "\r")
+      .replace(/&lt;/g, "<")
+      .replace(/&gt;/g, ">")
+      .replace(/&#39;/g, "'")
+      .replace(/&quot;/g, '"')
```
XyneSpaces, line 41: **[major] R2 — sibling decoder not updated: `urlUtils.ts` still decodes only `&quot;`/`&#10;`/`&#13;`.** With this PR the emitters now write `&amp;`/`&lt;`/`&gt;`/`&#39;` into `data-flow-json`, but `extractFlowJsonLinks` in `apps/backend/src/utils/urlUtils.ts` (~L155) decodes just those three entities before `JSON.parse` — links extracted from every **new** flow message come back corrupted (e.g. a query string `?a=1&b=2` → `?a=1&amp;b=2`), and this feeds `zero/vespa-injection/core/mapper.ts` (Vespa link indexing). The old quotes-only emitters left `&`/`<`/`>` raw, so this path was correct before — this PR introduces the regression. Note: open PR #897 (same bug, same author) *does* include the urlUtils decode fix — worth folding in or superseding it.

↳ Fix: decode the same entity set with `&amp;` last in `urlUtils.ts` (or export this chain and reuse it there).

packages/xyne-claw-shared/src/flow/flow-text.ts
```diff
@@ -33,9 +33,15 @@ export function parseFlowJsonComponents(content: string): FlowComponent[] | null
   if (escaped === undefined) return null;
   try {
     const json = escaped
-      .replace(/&quot;/g, '"')
       .replace(/&#10;/g, "\n")
       .replace(/&#13;/g, "\r")
+      .replace(/&lt;/g, "<")
+      .replace(/&gt;/g, ">")
+      .replace(/&#39;/g, "'")
```
XyneSpaces, line 40: **[minor] R2 — `messages-handler.ts` decoder has no `&#39;` handling.** The PR notes say `messages-handler.ts` keeps an "already correct" copy — its `&amp;`-last ordering is right, but it decodes no apostrophe entity. Since the emitters now escape `'` → `&#39;`, apostrophes in flow text (`don't`) will surface as literal `&#39;` in that path's extracted texts.

↳ Fix: add `.replace(/&#39;/g, "'")` before the `&amp;` pass in `apps/backend/src/zero/side-effects/tables/messages-handler.ts`.

packages/xyne-claw-shared/src/flow/flow-text.ts
```diff
@@ -33,9 +33,15 @@ export function parseFlowJsonComponents(content: string): FlowComponent[] | null
...
+      .replace(/&quot;/g, '"')
+      // &amp; MUST be decoded LAST, after every other entity, so a
+      // literal "&amp;lt;" in the content survives as "&lt;" and not "<".
+      .replace(/&amp;/g, "&")
```
XyneSpaces, line 44: **[minor] R2 (pre-existing) — `commandController.ts` decodes `&amp;` too early.** `extractFlowJSONFromHtml` in `apps/backend/src/apps/controllers/commandController.ts` decodes `&quot;` → `&amp;` → `&lt;` → `&gt;` → `&#39;`/`&apos;`; decoding `&amp;` *before* `&lt;`/`&gt;` double-decodes entity-looking text (a legit `&lt;` in a payload — now encoded `&amp;lt;` by the new emitters — comes back as `<`). Not regressed by this PR (the old path had the same class of bug), but since this PR is exactly the round-trip cleanup, the sibling decoder should follow the same rule you encoded here: `&amp;` last.

packages/xyne-claw-shared/src/flow/flow-text.test.ts
```diff
@@ -0,0 +1,83 @@
...
+ * A Flow card is stored as `<div data-flow-json="...escaped JSON...">fallback</div>`.
+ * The escaping used by the backend emitters (chatController / incomingWebhookController
+ * / the Slack adapter) MUST escape `&` FIRST, otherwise any literal `&`, `<`, `>` or
+ * entity-like sequence already present in the payload (very common in an embedded
+ * agent system prompt, e.g. the string `&quot;`, `A && B`, `<file:line>`) is corrupted
+ * when the browser HTML-decodes the attribute — JSON.parse then throws and the card
+ * collapses to its bare fallback text.
+ *
+ * The canonical encoder is apps/backend `encodeHtmlAttr`. This package must not depend
+ * on apps/backend, so we replicate it here and keep it in lock-step with that helper.
```
XyneSpaces, line 16: **[major] R5 (🟡 per guidelines #263/#246) — Nth copy of the encode/decode pair.** The replication is acknowledged in this comment, but it makes the entity handling live in at least 7 places: `contentUtils.encodeHtmlAttr` (backend, canonical), this test copy, `flow-text.ts` decode chain, `slashCommandArtifact.ts`'s private `encodeHtmlAttribute`/`decodeHtmlAttribute` (an already-correct unexported pair in `packages/shared`), `commandController.ts`, `urlUtils.ts`, and `messages-handler.ts`. This PR's own misses (urlUtils entities, messages-handler `&#39;`, commandController ordering) are the drift cost made real.

↳ Suggested follow-up (not blocking this fix): export one `encodeFlowJsonAttribute`/`decodeFlowJsonAttribute` pair importable by both backend and `xyne-claw-shared` — note `xyne-claw-shared` deliberately avoids a runtime dep on `@xyne/shared`, so the pair likely belongs in `xyne-claw-shared` or a new leaf package — then delete the per-file copies.

packages/xyne-claw-shared/src/flow/flow-text.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    const content = wrap(encodeHtmlAttr(JSON.stringify(flow)));
+    const attr = content.match(/data-flow-json="([^"]*)"/)![1]!;
+    const decoded = browserAttrDecode(attr);
+    expect(() => JSON.parse(decoded)).not.toThrow();
+    expect(JSON.parse(decoded)).toEqual(flow);
+  });
+
+  it("the OLD quotes-only escaping corrupted such a payload (documents the bug)", () => {
+    const buggy = JSON.stringify(flow).replace(/"/g, "&quot;");
+    const decoded = browserAttrDecode(buggy.match(/[\s\S]*/)![0]);
```
XyneSpaces, line 63: **[nit] R8 — `buggy.match(/[\s\S]*/)![0]` is an identity.** `/[\s\S]*/` matches the entire string (zero or more of everything), so `[0]` is just `buggy`. Pass `buggy` straight to `browserAttrDecode`.

XyneSpaces (commented):
Xyne Spaces Reviewer — automated review of all 5 changed files / 9 hunks against the [review guidelines](https://github.com/juspay/xyne-spaces/blob/main/apps/xyne-claw/skills/xyne-spaces-review-guidelines/SKILL.md). 5 findings below (2 major, 2 minor, 1 nit) — full findings table and reviewer recommendation in the summary comment.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1595

**Verdict:** Changes requested — the four emitter switches are correct, but the `urlUtils.ts` sibling decoder regression should be fixed (or folded in from #897) before merge.
**Reviewed:** 5 files, 8 hunks · **Findings:** 0 blocking, 2 major, 2 minor, 1 nit

### Findings
| major | R2 | `flow-text.ts:41` → fix needed in `apps/backend/src/utils/urlUtils.ts` (~L155) | `extractFlowJsonLinks` still decodes only `&quot;`/`&#10;`/`&#13;`. With the new `encodeHtmlAttr` emitters, links from every **new** flow message extract corrupted (`?a=1&b=2` → `?a=1&amp;b=2`) and feed Vespa link indexing (`zero/vespa-injection/core/mapper.ts`). This path was correct before — regression introduced by this PR. Open PR #897 includes this fix. |
| major | R5 (🟡 #263/#246) | `flow-text.test.ts:16` | 7th copy of the encode/decode pair (contentUtils, this test replica, flow-text, slashCommandArtifact, commandController, urlUtils, messages-handler) — this PR's own missed decoders demonstrate the drift cost. Consolidate into one exported pair as a follow-up. |
| minor | R2 | `flow-text.ts:40` → `apps/backend/src/zero/side-effects/tables/messages-handler.ts` | That decoder lacks `&#39;` handling; emitters now escape `'` → `&#39;`, so apostrophes in flow text surface as literal `&#39;` there. |
| minor | R2 | `flow-text.ts:44` → `apps/backend/src/apps/controllers/commandController.ts` | `extractFlowJSONFromHtml` decodes `&amp;` *before* `&lt;`/`&gt;` → double-decodes entity-looking text (pre-existing, same bug class this PR fixes elsewhere). |
| nit | R8 | `flow-text.test.ts:63` | `buggy.match(/[\s\S]*/)![0]` is an identity — pass `buggy` directly. |

**What checks out:** all four emitter changes (`chatController` ×2, `incomingWebhookController`, Slack `chat.ts` transformer) and the `flow-text.ts` decoder ordering (`&amp;` last) are correct. `builder.ts`, `conversationService.ts`, and `bulkIngestConversationSlack.ts` were verified to not be emitters (guard/delegation only), so no emitter site was missed. Also note: **open PR #897** fixes the same bug and additionally patches `urlUtils.ts` — recommend folding that in or closing one of the two to avoid double-landing.

### Recommended reviewer
**@12-aryan** — authored 6 of the repo's flow/`data-flow-json` PRs (most of any non-bot author) and has no pending review requests.

| Candidate | Flow-domain PRs authored | CODEOWNER? | Pending review requests | Tie-breaker |
|-----------|--------------------------|------------|--------------------------|-------------|
| @12-aryan | 6 | no (`.github/CODEOWNERS` covers only ACL paths — none match this PR) | 0 | most flow-domain PRs |
| @harshpreet931 | 3 | no | 0 | — |
| @harsh-1923 | 2 | no | 0 | — |
| @KushagarGoel1 | 2 | no | 0 | — |
| @VishalM-23 | 1 | no | 0 | latest main committer (`230a480`) |

Workload note: no open PR in the repo currently carries a requested reviewer (verified across the open-PR list), so review load doesn't differentiate — familiarity wins.

Note: this bot doesn't have a reviewer-request API available, so the reviewer is @mentioned here instead of formally requested — @12-aryan could you pick this one up? Full details in the 5 inline review comments above.

PR #1597, opened 2026-09-07
fix: XYNE-62808 derive github credential owner from the token

ameer030:
## Problem

Repo cards under Project → Repos showed `READ REPOSITORY · proven` but `PUSH BRANCH · unavailable` and `CREATE PULL REQUEST · unavailable`, with overall status `ready` and no error message. The PAT was fine — `GET /repos/juspay/xyne-spaces` with it returns `permissions.push: true`.

`performRepositoryCheck` compared the stored `resourceOwner` against the repo owner and, on any mismatch, skipped the authenticated `inspectRepository` entirely and fell through to the anonymous one. `GitHubVcsAdapter.inspectRepository` computes `canPush = token ? repo.permissions?.push === true : false`, which is `false` without a token, so both write capabilities came back `UNAVAILABLE` while `READ_REPOSITORY` stayed `PROVEN` on a public repo.

The check was never a real fence:

- `resourceOwner` was free text typed by a workspace admin. `validateCredential` only did `GET /users/{resourceOwner}`, which proves the account exists, not that the token has any grant on it.
- Credentials are one-per-workspace-per-provider, so one typo degraded every GitHub repo in the workspace at once.
- The runtime write paths (`buildGitAuthentication`, `createDraftPullRequest`, `verifyRemoteCommit`) never compared `resourceOwner` to the repo owner. Push and PR creation were not fenced. Only the access check was.

## Change

- Deleted the `GITHUB_RESOURCE_OWNER_MISMATCH` branch. The authenticated inspect now runs whenever a token exists, and GitHub authorizes the call. A repo the token cannot see returns 404, which already flows through the existing `catch` into `fallbackError`.
- Removed `resourceOwner` from the connect payload, the shared zod schema, the credential record and the admin form. An admin now pastes a PAT and nothing else.
- The owner and the reachable repository count are derived instead of typed, and read live rather than stored: `repositoryReach(token)` on the adapter makes one `GET /user/repos?per_page=1` call at `listCredentials` time, taking the owner from the body and the count from the `Link` header's `rel="last"` page. Both are display-only and must never gate access. The credentials tab is opened rarely enough that one call per load beats keeping stored copies fresh.
- Dropped what the response no longer feeds: `fingerprint` and `attachedRepositoryCount` are gone from `listCredentials`, along with the `repositoryCountsByProvider` query that existed only to compute the latter and the now-unused `credentialFingerprint` helper. Nothing read the fingerprint, so it is no longer computed or stored either.
- Credential card now reads **GitHub account**, **Token for**, **Repos it can access**.

## Migration

None. No schema change, no backfill, no revalidation. The credential is an encrypted JSON blob inside `workflow.external_sources.credentials`, not a set of columns, and `parseStoredCredential` validates only `workspaceId`, `provider` and `revision`. Every existing blob is read as-is; its now-unused `resourceOwner` and `fingerprint` keys are ignored on read and dropped on the next save. `connectedCredential` no longer requires `resourceOwner`, so workspaces that have not revalidated keep working.

### Rollback constraint

Rolling *back* after this ships is not fully clean, in one narrow case. `configureCredential` builds its save object as a fresh literal, so a workspace that re-saves its PAT under this version gets a blob with no `resourceOwner`. The previous version's `connectedCredential` treats that as not connected.

Scope and severity:

- Only `configureCredential` drops the key. `revalidateCredential` and `invalidateWorkspaceCredential` save with `...row` / `...current`, which spread the object from `parseStoredCredential` and therefore carry `resourceOwner` forward at runtime. A workspace where nobody pastes a new PAT is unaffected, including after an unattended credential invalidation.
- Not a crash. `requireConnectedCredential` throws `AppError(..., 409)` and the error handler passes an `AppError`'s own status through, so affected calls return a 409 reading "Workspace GitHub credential is not connected". `listCredentials` does not go through that gate, so the credentials tab still loads.
- Affected paths under the old version: Start Work / `createDraftPullRequest`, `inspectPullRequest`, `revalidateCredential`.
- Self-healing: the admin re-saves the PAT under the old UI, which still has the resource-owner field.

Deliberately not dual-writing `resourceOwner` for a release. Say the word if you would rather the rollback path be clean.

## Testing

- `tsc --noEmit` clean on `apps/backend` and `apps/dashboard`.
- Pre-commit `dashboard:validate` passed (typecheck, prettier, eslint 0 errors).
- Verified against a live org-scoped PAT: card shows `Token for juspay`, `Repos it can access 622`, and the repo capabilities move from `unavailable` to `inferred`.

## Known limitation, not addressed here

The anonymous fallback swallows `fallbackError` whenever the anonymous read succeeds, so `deriveAccessStatus` reports `READY` with `errorMessage: null` and a capability downgrade on a public repo shows no reason. Pre-existing, tracked separately.

<details>
<summary>Debug</summary>

- Claude Code `session_01Q2QxUxqZacUx2SyFFxr5aj` — `8c2cc9a5e` — dropped the resourceOwner fence, derived the credential owner and reach from the token, and moved both to a live read.

</details>

### DIFF
--- a/apps/backend/src/sdlc/vcs/SdlcVcsService.ts
+++ b/apps/backend/src/sdlc/vcs/SdlcVcsService.ts
@@ -5,7 +5,6 @@ import { AppError } from '@/middleware/errorHandler';
 import { logger } from '@/utils/logger';
 import type { SdlcActor } from '../types';
 import { isSafeSdlcGitRef, requireSdlcBaseBranch } from '../sdlcRepositoryContext';
-import { credentialFingerprint } from './credentialEnvelope';
 import { GitHubVcsAdapter } from './GitHubVcsAdapter';
 import { SdlcVcsCredentialStore, type StoredSdlcVcsCredential } from './SdlcVcsCredentialStore';
 import { classifyRuntimeAccessFailure } from './accessCheckPolicy';
@@ -48,24 +47,24 @@ export class SdlcVcsService implements SdlcVcs {
   async listCredentials(actor: SdlcActor): Promise<unknown[]> {
     const user = await this.requireWorkspaceUser(actor);
     const rows = await this.credentialStore.list(this.prisma, actor.workspaceId);
-    const counts = await this.repositoryCountsByProvider(actor.workspaceId);
-    return rows.map((row) => ({
-      provider: row.provider,
-      status: row.status,
-      revision: row.revision,
-      identityLogin: row.identityLogin,
-      repositoryOwner: row.repositoryOwner,
-      repositoryCount: row.repositoryCount,
-      fingerprint: row.fingerprint,
-      validationStatus: row.validationStatus,
-      validatedAt: row.validatedAt,
-      validationErrorCode: row.validationErrorCode,
-      validationErrorMessage: row.validationErrorMessage,
-      disconnectedAt: row.disconnectedAt,
-      updatedAt: row.updatedAt,
-      attachedRepositoryCount: counts.get(row.provider) ?? 0,
-      canManage: user.role === 'OWNER' || user.role === 'ADMIN',
-    }));
+    return await Promise.all(
+      rows.map(async (row) => ({
+        provider: row.provider,
+        status: row.status,
+        revision: row.revision,
+        identityLogin: row.identityLogin,
+        ...(row.status === 'CONNECTED' && row.token
+          ? await this.adapter(row.provider).repositoryReach(row.token)
+          : { repositoryOwner: null, repositoryCount: null }),
+        validationStatus: row.validationStatus,
+        validatedAt: row.validatedAt,
+        validationErrorCode: row.validationErrorCode,
+        validationErrorMessage: row.validationErrorMessage,
+        disconnectedAt: row.disconnectedAt,
+        updatedAt: row.updatedAt,
+        canManage: user.role === 'OWNER' || user.role === 'ADMIN',
+      }))
+    );
   }
 
   async configureCredential(
@@ -95,9 +94,6 @@ export class SdlcVcsService implements SdlcVcs {
         token: input.token,
         revision,
         identityLogin: validation.identityLogin,
-        repositoryOwner: validation.repositoryOwner,
-        repositoryCount: validation.repositoryCount,
-        fingerprint: credentialFingerprint(input.token),
         validationStatus: 'VALID',
         validatedAt: now,
         validationErrorCode: null,
@@ -136,8 +132,6 @@ export class SdlcVcsService implements SdlcVcs {
           validationErrorCode: null,
           validationErrorMessage: null,
           identityLogin: validation.identityLogin,
-          repositoryOwner: validation.repositoryOwner,
-          repositoryCount: validation.repositoryCount,
           updatedBy: actor.userId,
           updatedAt: now,
         });
@@ -459,16 +453,18 @@ export class SdlcVcsService implements SdlcVcs {
     }
   }
 
-  async bootstrapSandboxCredential(binding: {
-    agentSlug: typeof SDLC_AGENT_SLUG;
-    repoId: string;
-    operation: 'CLONE' | 'PUSH' | 'INTERACTIVE';
-    sandboxId: string;
-    sandboxPublicKey: string;
-  } & (
-    | { executionId: string; sessionId: string }
-    | { interactiveGrant: string; conversationId: string }
-  )): Promise<SandboxCredentialEnvelope | null> {
+  async bootstrapSandboxCredential(
+    binding: {
+      agentSlug: typeof SDLC_AGENT_SLUG;
+      repoId: string;
+      operation: 'CLONE' | 'PUSH' | 'INTERACTIVE';
+      sandboxId: string;
+      sandboxPublicKey: string;
+    } & (
+      | { executionId: string; sessionId: string }
+      | { interactiveGrant: string; conversationId: string }
+    )
+  ): Promise<SandboxCredentialEnvelope | null> {
     const repo = await this.prisma.repo.findUnique({
       where: { id: binding.repoId },
       select: { id: true, workspaceId: true },
@@ -558,17 +554,19 @@ export class SdlcVcsService implements SdlcVcs {
     );
   }
 
-  async createDraftPullRequest(input: {
-    repoId: string;
-    title: string;
-    body: string;
-    head: string;
-    base: string;
-    commitHash: string;
-  } & (
-    | { executionId: string; sessionId: string }
-    | { interactiveGrant: string; conversationId: string }
-  )) {
+  async createDraftPullRequest(
+    input: {
+      repoId: string;
+      title: string;
+      body: string;
+      head: string;
+      base: string;
+      commitHash: string;
+    } & (
+      | { executionId: string; sessionId: string }
+      | { interactiveGrant: string; conversationId: string }
+    )
+  ) {
     const repo = await this.prisma.repo.findUnique({ where: { id: input.repoId } });
     if (!repo?.workspaceId) throw new AppError('SDLC repository not found', 404);
     let actor: SdlcActor;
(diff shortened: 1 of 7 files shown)


apps/backend/src/sdlc/vcs/SdlcVcsService.ts
```diff
@@ -94,7 +95,8 @@ export class SdlcVcsService implements SdlcVcs {
         token: input.token,
         revision,
         identityLogin: validation.identityLogin,
-        resourceOwner: validation.resourceOwner,
+        repositoryOwner: validation.repositoryOwner,
+        repositoryCount: validation.repositoryCount,
```
ameer030: 🔴 **B01 — rollback breaks SDLC GitHub access after any credential re-save**

The workspace credential is a JSON blob in `workflow.external_sources.credentials`. This PR renames its keys `resourceOwner` → `repositoryOwner`/`repositoryCount` with no read-side fallback or dual-write. Every save (configure, revalidate, the invalidation rewrite) produces a blob the previous backend version can parse but treats as "not connected", because old `connectedCredential` requires `resourceOwner` to be non-empty.

Consequence: if the deploy is rolled back — or an old replica is still serving during a rolling deploy — after any workspace re-saves its GitHub credential, that workspace's `createDraftPullRequest`, `inspectPullRequest`, and `revalidateCredential` all fail with 409 under the old code. The PR body's "Migration: None" only argues the forward direction (new code reading old blobs), which is fine.

<details><summary>Details</summary>

- Old parse tolerates new blobs: old `SdlcVcsCredentialStore.ts` only checks workspaceId/provider/revision — no strict schema.
- Old gate rejects them: old `SdlcVcsService.ts` (`connectedCredential`) returns null when `!row.resourceOwner`, so `requireConnectedCredential` throws 409.
- Old-code paths that hit that gate: `revalidateCredential`, `createDraftPullRequest`, `inspectPullRequest`.
- New code never dual-writes the old key: `resourceOwner` no longer appears anywhere at this revision.
- Blob rewrites happen on configure, revalidate, and the unattended invalidation save — an admin clicking nothing still re-keys the blob after a failed check.

</details>

<details><summary>Recommended fix</summary>

Dual-write for one release: keep writing `resourceOwner` (from the derived `repositoryOwner`) alongside the new keys, drop the old key in a tracked follow-up once this version is fully deployed everywhere. Alternatively, at minimum state the rollback constraint in the Migration section instead of "None".

</details>

rules: ARCH-C01, PROC-C04
</details>

ameer030 (commented):
🔵 **N01 — `listCredentials` still computes and returns fields the UI dropped**

The credential card no longer renders the fingerprint or the attached-repos tiles, but the backend `GET /sdlc/vcs/credentials` response still runs the per-provider repository count query on every call and returns both `fingerprint` and `attachedRepositoryCount`.

<details><summary>Details</summary>

- `SdlcVcsService.ts` (`listCredentials`): still calls `repositoryCountsByProvider(...)` — its only caller — and maps `fingerprint` + `attachedRepositoryCount` into each row.
- The only consumer of the route is `RepositoryCredentialsTab.tsx`, which this PR updates to drop both fields from its interface and rendering.

</details>

<details><summary>Recommended fix</summary>

Drop `fingerprint` and `attachedRepositoryCount` from the `listCredentials` mapping and delete `repositoryCountsByProvider` — or note why they're intentionally kept (e.g. a planned change-detection use of the fingerprint).

</details>

rules: STYLE-C14, PERF-C05
</details>

ameer030 (commented):
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 1 |
| 🟡 Better to fix | 0 |
| 🔵 Nit | 1 |

### 🔁 Re-check
- ✅ **Verified OK:** the fence removal changes no real authorization — `buildGitAuthentication`, `createDraftPullRequest`, and `verifyRemoteCommit` never consumed `resourceOwner` anywhere on main; the token's GitHub scope is the only gate, before and after. Anonymous-fallback semantics in `performRepositoryCheck` are unchanged, zod still validates the token shape at the route boundary, old credential blobs read fine under the new code (forward direction), and the PR body honestly names the remaining anonymous-fallback limitation.

### 📌 Verdict
1 blocker must be fixed before merge ❌ — the rollback hazard on the credential blob keys; the rest of the change is sound.

rules: ARCH-C01, PROC-C04, STYLE-C14, PERF-C05
run: 01a07bec-d795-70e8-bfd7-60f73808f7fa
</details>

apps/backend/src/sdlc/vcs/SdlcVcsService.ts
ameer030: Valid, and the rollback direction is now the only compatibility gap left. Two corrections to the scope, then where it stands.

**The unattended path does not re-key the blob.** `invalidateWorkspaceCredential` saves with `...current`, and `revalidateCredential` with `...row`. Both spread the object returned by `parseStoredCredential`, which is `{ id, ...parsed }` over the raw decrypted JSON — so `resourceOwner` is still present at runtime even though the type no longer declares it, and both paths carry it forward. Only `configureCredential` builds an explicit literal and therefore drops it.

So the exposure is: a workspace breaks under old code only if an admin pasted a new PAT while the new version was live. An admin clicking nothing stays intact.

**`repositoryOwner` and `repositoryCount` are no longer stored at all.** Pushed in a follow-up commit: they were display-only, so `validateCredential` is back to a single `GET /user` and the values are read live in `listCredentials` via a new `repositoryReach(token)` on the adapter. The blob shape is now exactly the previous one minus `resourceOwner` and `fingerprint`, so there are no new keys for an older parser to meet.

Forward compatibility is unaffected: `parseStoredCredential` validates only workspaceId, provider and revision, and the new `connectedCredential` needs only status, validationStatus and token, all present in every existing blob.

Leaving the dual-write out for now and treating the rollback constraint as documented rather than removed. Happy to add `resourceOwner: repositoryOwner` back to `configureCredential` for one release if you would rather the rollback path be clean.

ameer030:
**N01 — fixed** in `8c2cc9a5e`.

`listCredentials` no longer returns `fingerprint` or `attachedRepositoryCount`, and `repositoryCountsByProvider` is deleted along with them — it existed only to compute the latter, so the per-provider repository query is gone from every card load.

Went one step further than the recommendation: `fingerprint` is not just unreturned, it is no longer computed or stored either. Nothing read it, so the field is off `StoredSdlcVcsCredential`, the write in `configureCredential` is gone, and the `credentialFingerprint` helper in `credentialEnvelope.ts` is deleted as its only caller went with it. Existing blobs keep a stale `fingerprint` key that is ignored on read and dropped on the next save.

The same commit also stops persisting `repositoryOwner` and `repositoryCount`. Both were display-only, so `validateCredential` is back to a single `GET /user` and the values are read live in `listCredentials` through a new `repositoryReach(token)` on the adapter. The credentials tab is opened rarely enough that one GitHub call per load beats keeping stored copies fresh.

apps/backend/src/sdlc/vcs/SdlcVcsService.ts
ameer030: Re-verified at `8c2cc9a5e` — your correction holds up, and the fixup resolves this for me.

Checked every `credentialStore.save` call site: `configureCredential` is indeed the only fresh literal; `revalidateCredential` (both paths), `invalidateWorkspaceCredential`, and `disconnectCredential` all spread the parsed blob, so existing credentials carry `resourceOwner` forward indefinitely. Combined with reading the reach live instead of storing it, the blob never gains an old-code-incompatible shape except in the one case your "Rollback constraint" section now names.

Also confirming the cleanup lands: no `fingerprint` / `attachedRepositoryCount` / `repositoryCountsByProvider` / `credentialFingerprint` remnants anywhere, and the live `repositoryReach` call is display-only and fail-soft (nulls on any error), so the list path can't reject on a dead token.

Accepted as documented — no dual-write needed. An admin re-saving a PAT mid-rollout is an avoidable, self-healing case, and stating it in the PR body is the right call. Closing this blocker. ✅

ameer030 (commented):
## 📝 Xyne PR Review — Summary (re-review at `8c2cc9a5e`)

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 0 |
| 🟡 Better to fix | 0 |
| 🔵 Nit | 0 |

### 🔁 Re-check
- 💬 **Existing threads:** B01 accepted as documented ✅ (rollback constraint narrowed to one admin-initiated, self-healing case after the live-reach redesign) · N01 fixed in code ✅ (fingerprint, attachedRepositoryCount, repositoryCountsByProvider, credentialFingerprint all removed)
- ✅ **Verified OK:** save/revalidate/invalidate/disconnect paths spread the parsed blob and carry old keys forward — only a fresh `configureCredential` literal re-keys; `repositoryReach` is display-only, fail-soft, and leaks nothing (the stored token never lands in the response); the merged diff's remaining reflows are formatting-only.

### 📌 Verdict
No blockers — good to merge ✅

rules: ARCH-C01, STYLE-C14, PERF-C05
run: 01a07bec-d795-70e8-bfd7-60f73808f7fa
</details>

utkarsh-611 (approved):

Mohan0544 (approved):

PR #1613, opened 2026-09-07
feat: XYNE-62866 implement dev tickets search, sort & filters

sumantop:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
XYNE-62866 — dev-ticket search, sort & filters on the release detail screen, plus the release-side AI helpers that ride the same screen.

**Release detail → Dev Tickets tab**
- ag-grid table with quick search and status / type / hotfix / dev-owner filters, column picker ("Add column"), multi-select + CSV export. Column sorting is intentionally off (`sortable: false`) — sorting would scatter the per-repo group headers; the "sort" in the title refers to the stable (createdAt, id) ordering of the paged query.
- Column picks are persisted on the release board via `board.update({ devTicketColumns })` — merged server-side into the board's current metadata (fresh read, not the client's snapshot). Picker is only shown to users who can manage releases (`useCanManageRelease`, mirrors the backend `assertReleaseManageAccess`).
- **Accepted boundary:** search/filters run client-side over a window of the newest 1,000 rows (`ART_FILTER_LOAD_ALL`). Exact for releases ≤1,000 dev tickets; beyond that the amber banner states the cap and CSV export bypasses the window. Server-side filter args for `applicationReleaseTicketsByReleaseId` are the follow-up if a release outgrows this.
- Selection/toast/tracking extracted from `TicketTable` into shared `ui/DataGrid/gridSelection.tsx` (behaviour-identical).

**Dev ticket → Release tab** (`ThreadPannel`): a dev ticket that is part of a release shows its release rows; gated by the new Zero query `applicationReleaseTicketsByDevTicketId` (mirrored backend ↔ shared, workspace-pinned ACL).

**Backend**
- `POST /api/tickets/:ticketId/release-insights` — LLM release summary stored on the release ticket's metadata behind an in-flight flag (stale-window guard, 409 while running, flag always cleared on failure). Gated `RELEASE-MANAGER` WRITE / admin.
- `POST /api/commit-analysis/suggest-services` — suggests service → path-regex config from the repo tree (GitHub/Bitbucket). Suggested regexes are length-capped and rejected on nested quantifiers before they are stored; Bitbucket project key / repo slug are allow-list validated and encoded. Uses the org-wide VCS token — documented inline; repo scope is caller-supplied.
- `release_ai_model_name` CAC key (default `glm-flash-experimental`) for both agents.
- Release config wizard "Replace all" keeps the existing `id`/`boardId` for a same-named service, so save no longer deletes and recreates its board.

## Areas Touched

- [x] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

- [x] Adds/modifies a **mutator** — `board.update` gains optional `devTicketColumns`
- [x] Adds/modifies a **query or ACL** — `applicationReleaseTicketsByDevTicketId`

If any of the above:

- [ ] New tables exported in `createSchema()` and given a Query ACL registered in `query-acl-factory.ts` — no new tables
- [ ] Matching Prisma model + migration, client regenerated — n/a

## Schema & Data Changes

## Config, Environment & URLs

- [x] Feature flag or runtime config changed — CAC key `release_ai_model_name` (optional, defaulted in code; no env file changes)

## API Contract

- [ ] No API changes
- [x] New endpoint — `POST /api/tickets/:ticketId/release-insights`, `POST /api/commit-analysis/suggest-services`
- [ ] Breaking for dashboard / electron / external / bots

---

## How did you test it?
Screenshots of the Dev Tickets tab (search, filters, column picker, export) are in the comment below.

### Automated

- [ ] Claw tests
- [x] Not covered by tests — no automation spec for the release screens yet; review-round logic (metadata merge, replace-all id preservation, nested-quantifier guard) verified with small assert scripts against the source.

### Manual

**Users**
- [x] Single user
- [ ] Two users at once
- [ ] Different roles or permission levels — role gate (column picker / Create Release) added in the review round; covered by typecheck + code path shared with the backend gate, not yet clicked as a non-manager

**Login**

**Run mode**

**Surface & data**
- [x] Web browser — Dev Tickets tab: search, filters, column picker, export (screenshots in the comment below)
- [ ] Sync/resilience, empty state, error paths — not exercised manually

---

## Checklist

- [x] New deps added with `pnpm --filter <pkg> add <dep>` — none added
- [x] Commits follow `<type>: <TICKET-ID> <subject>`


sumantop:
image
image
image

apps/backend/src/controllers/releaseInsightsController.ts
```diff
@@ -0,0 +1,152 @@
...
+          ?.isGeneratingReleaseInsights === true
+      ) {
+        res.status(409).json({
+          success: false,
+          error: 'Insights are already being generated for this release',
+        });
+        return;
+      }
+
+      await this.ticketRepository.updateTicketMetadata(ticketId, {
```
devesh-juspay, line 69: Two related robustness notes on the in-flight flag:

1. **Stuck flag on crash/restart.** If the process dies between setting `isGeneratingReleaseInsights: true` and the `finally` (LLM calls can run long), the flag stays true forever — every later attempt 409s and the dashboard button is permanently stuck on "Analyzing…" with no way out short of a manual metadata edit. Suggest storing a timestamp (`insightsGenerationStartedAt`) and treating the flag as stale after ~5 minutes, so a retry can proceed.

2. **Check-then-set race.** The read at the top and the `updateTicketMetadata(..., true)` are not atomic, so two concurrent requests can both pass the 409 guard and double-spend the LLM call. Harmless output-wise (last write wins) but worth noting; a conditional update (set true only when currently unset) would close it if it ever matters.

apps/backend/src/routes/commitAnalysis.ts
```diff
@@ -162,4 +167,51 @@ router.post('/test-connection', authorizePrivilegedOrResource('RELEASE-MANAGER',
   }
 });
 
+// AI-suggest services + env/migration paths from the repo's file tree, to
+// prefill the release-config wizard. Supports GitHub and Bitbucket Server.
+router.post('/suggest-services', authorizePrivilegedOrResource('RELEASE-MANAGER', AccessType.WRITE), async (req: Request, res: Response): Promise<void> => {
```
devesh-juspay, line 176: Scope note, not a blocker: `repoUrl` is caller-supplied and the file listing runs with the org-wide GitHub/Bitbucket token, so anyone holding RELEASE-MANAGER WRITE can enumerate the full file tree of *any* repo that token can read — not just repos already configured as applications. The VCS host itself is pinned to config (good — no SSRF), this is purely about repo scope. Fine if that's an accepted trade-off for wizard prefill; worth a one-line comment saying so, or a check that the repoUrl matches an existing/being-configured application when you want to tighten it.

Related asymmetry: this endpoint requires RELEASE-MANAGER WRITE, but `/tickets/:id/release-insights` (which also spends LLM tokens and writes release metadata) only requires TICKETS WRITE — any ticket-writer can re-analyze any release. Consider gating insights the same way for consistency.

apps/backend/src/controllers/releaseInsightsController.ts
sumantop: Added insightsGenerationStartedAt; a flag older than 5 min is now treated as stale so a crashed run can be retried instead of wedging the button. Left the check-then-set as last-write-wins with a comment noting why.

apps/backend/src/routes/commitAnalysis.ts
sumantop: Added a comment documenting the repo scope as an intentional wizard-prefill trade-off (host pinned to config, no SSRF). Also moved /release-insights to RELEASE-MANAGER WRITE so it matches suggest/analyze.

apps/backend/src/controllers/releaseInsightsController.ts
devesh-juspay: Addressed in 796e3f755 — insightsGenerationStartedAt with a 5-min stale window (a legacy flag without the timestamp also counts as stale, nice), cleared on both success and finally. The check-then-set race stays open as discussed (bounded by the stale window, fine). Resolving.

apps/backend/src/routes/commitAnalysis.ts
devesh-juspay: Addressed in 796e3f755 — repo scope documented as intentional with the wizard-prefill rationale, and release-insights now gated by RELEASE-MANAGER WRITE matching the other AI actions. Resolving.

devesh-juspay (dismissed):

apps/dashboard/src/routes/ReleaseDetailScreen/ReleaseDetailScreen.tsx
```diff
@@ -208,27 +209,7 @@ function groupTimelineEvents(
...
-      )}
-      {counts.mig > 0 && (
-        <span className='px-1.5 py-0.5 rounded bg-amber-100 text-amber-700 dark:bg-amber-900/40 dark:text-amber-200'>
-          {counts.mig} mig
-        </span>
-      )}
-    </span>
-  );
-};
+const ART_FILTER_LOAD_ALL = 1000;
```
ameer030, line 213: 🟡 **W01 — dev-ticket search & filters only see the newest 1,000 rows**

This PR's headline feature — search, sort & filters on the dev-ticket table — runs entirely client-side. The moment any filter or search text is active, the Zero query jumps from a 26-row page to the newest 1,000 rows (`ART_FILTER_LOAD_ALL`) and the filter predicates run in the browser over that window. On a release with more than 1,000 dev tickets, older rows can never match, so the answer the UI shows is silently incomplete.

- `ReleaseDetailScreen.tsx:266` — `filtersLoadAll` flips on `onFiltersActiveChange`; `ReleaseDetailScreen.tsx:336-345` — query args become `limit: 1000, start: null`.
- `ReleaseDevTicketsTable.tsx:103-105` — `matchesReleaseDevTicketRow` runs client-side over the loaded window (`releaseDevTicketFilters.ts:30-45`).
- Mitigation present: the amber banner at `ReleaseDetailScreen.tsx:961-966` discloses the cap when the window is full, and CSV export bypasses the window. So exact for ≤1,000-row releases, approximate beyond.
- Checklist ARCH-C18: filtering for a limited/paged read belongs in the server/Zero query, never client-side after the limit.

</details>

Move at least the cheap predicates (status, type, hotfix, devOwner) into `applicationReleaseTicketsByReleaseId` query args so the limit applies after filtering; keep the free-text search client-side if Zero can't express it. If the capped window is the deliberate trade-off, note it in the PR body as the accepted boundary.

</details>

rules: ARCH-C18
agent: omp
run: n/a
</details>

apps/dashboard/src/routes/ReleaseDetailScreen/ReleaseDetailScreen.tsx
```diff
@@ -648,15 +670,17 @@ const ReleaseDetailScreen = (): ReactElement => {
...
+    const md = releaseBoard?.metadata;
     const base =
       md && typeof md === 'object' && !Array.isArray(md) ? (md as Record<string, unknown>) : {};
     void (async (): Promise<void> => {
       const res = await zero.mutate(
-        mutators.ticket.update({
-          id: releaseTicketId,
+        mutators.board.update({
+          boardId,
           metadata: { ...base, devTicketColumns: nextKeys },
```
ameer030: 🟡 **W02 — column prefs overwrite the whole release-board metadata from a stale snapshot**

`toggleColumn` sends `metadata: { ...base, devTicketColumns }` where `base` is the board metadata as the *local Zero replica last saw it*. The server mutator treats a caller-supplied `metadata` as a whole-column overwrite, so any other key written to the same board metadata between page load and the user's toggle is silently reverted — the exact shape the `etaManagement` server-side merge in the same mutator exists to avoid.

- `ReleaseDetailScreen.tsx:679-696` — base read from `releaseBoard?.metadata` (client replica), full blob shipped.
- `apps/backend/src/zero/mutators.ts` `board.update`: when ETA fields aren't set, `nextMetadata = metadata` verbatim — no server-side merge; the nearby comment says the etaManagement path merges server-side precisely because a raw overwrite "would drop sibling keys".
- Competing writers on the same blob exist: `BoardRolesConfigScreen` ETA toggles (`autoRecomputeEnabled` / `standardPathStageIds`) go through `board.update` — a column toggle concurrent with such an edit clobbers it.
- Old code did the same read-modify-write on the *ticket* metadata; the PR moves the pattern onto a blob shared by more writers.

</details>

Give `board.update` a narrow server-side merge for this key — a `devTicketColumns` arg merged into the freshly-read `board.metadata` inside the transaction (mirror `mergeBoardEtaManagement`) — instead of shipping the whole blob from the client.

</details>

rules: BUG-C14, BUG-C32
agent: omp
run: n/a
</details>

apps/dashboard/src/routes/ReleaseDetailScreen/ReleaseDetailScreen.tsx
```diff
@@ -648,15 +670,17 @@ const ReleaseDetailScreen = (): ReactElement => {
...
+    const boardId = releaseTicket?.boardId;
+    if (!boardId) return;
+    const md = releaseBoard?.metadata;
     const base =
       md && typeof md === 'object' && !Array.isArray(md) ? (md as Record<string, unknown>) : {};
     void (async (): Promise<void> => {
       const res = await zero.mutate(
-        mutators.ticket.update({
-          id: releaseTicketId,
+        mutators.board.update({
```
ameer030: 🟡 **W03 — column saving now fails for non-release-managers, but the control stays visible**

Column persistence moved from `mutators.ticket.update` (no role gate) to `mutators.board.update`, which gates **every** update on a `BoardType.RELEASE` board behind `assertReleaseManageAccess()` (ADMIN/OWNER or RELEASE-MANAGER WRITE). A plain project member who opens the Release Detail screen, clicks "Add column" and picks one gets a server rejection → "Failed to save columns" toast — the dropdown stays fully visible and clickable with no hint it can't work for them.

- `apps/backend/src/zero/mutators.ts:8337-8343` — `if (board.boardType === BoardType.RELEASE) { await assertReleaseManageAccess(); … }` fires before any mutation, including metadata-only.
- Old path (`origin/main` `toggleColumn`) used `mutators.ticket.update`, whose server mutator only gates FLOW stage transitions — plain members' picks saved fine.
- The "Add column" dropdown (`ReleaseDetailScreen.tsx:820+`) has no permission gate, and the toggle is optimistic — the user sees it flip, then roll back with a toast.
- Server-side fail-closed is correct; the regression is the UX: a previously-working per-viewer preference is now denied to lesser roles, and the affordance doesn't say so.

</details>

Either (a) hide/disable the Add-column trigger when the user lacks release-manage permission (mirror the server predicate client-side), or (b) keep the preference per-user at a scope everyone may write (localStorage keyed by board id) — it's a personal view preference, so (b) is arguably the truer semantics than board-wide metadata.

</details>

rules: SEC-C05
agent: omp
run: n/a
</details>

apps/dashboard/src/components/Release/ReleaseConfigWizard/useReleaseConfigForm.ts
```diff
@@ -167,6 +167,38 @@ export function useReleaseConfigForm({
...
+  const applyServiceSuggestions = useCallback(
+    (
+      services: { name: string; regex: string; envPaths: string[]; migrationPaths: string[] }[],
+      mode: 'replace' | 'add',
+    ) => {
+      if (services.length === 0) return;
+      setUserTouched(true);
+      const toRow = (service: (typeof services)[number]): ApplicationConfig => ({
+        ...EMPTY_APP,
+        id: uuidv4(),
```
ameer030: 🟡 **W04 — "Replace all" AI suggestions reseed fresh ids, so existing services get deleted & recreated**

In replace mode every suggestion row gets `id: uuidv4(), boardId: uuidv4()`, even when it names a service that's already configured. On save, `saveReleaseBoardConfig` deletes every persisted application missing from the payload — a same-named service counts as missing (its real id was replaced), so its application board (stages, approvers, PR-status mappings, form mappings) is torn down and recreated with only the 4 generic default stages.

- `useReleaseConfigForm.ts:186-190` — `toRow` seeds fresh uuids for both id and boardId; name-matching to preserved rows only happens in 'add' mode (`:195-199`).
- `apps/backend/src/zero/mutators.ts:7826-7841` — persisted apps absent from the payload are removed via `deleteApplicationBoard` (stages, stage_approvers, pr_status_mappings, form mappings, then the board); it throws only when the board still has tickets.
- Reachable: the "Suggest with AI" button and the "Replace all" radio render whenever `allowApplicationListChanges` is true, i.e. any multi-service config in edit mode (`ReleaseConfigWizard.tsx:863`).
- Existing containment: the throw on ticket-bearing boards makes mid-release configs error out loudly instead of losing data; silent loss hits ticket-free apps with customized stages. The destructive path also requires an explicit radio choice (default is 'add' when named rows exist) — but re-suggesting the *same* services and picking "Replace all" is the natural way to hit it.

</details>

In replace mode, name-match suggestions against existing rows (same normalized-name comparison 'add' mode uses) and carry the matched row's existing `id` + `boardId`; fresh-uuid only services that are genuinely new. "Replace all" then becomes update-in-place for matched services plus removal of unmatched ones.

</details>

rules: BUG-C19
agent: omp
run: n/a
</details>

apps/backend/src/services/bitbucketService.ts
```diff
@@ -37,6 +37,38 @@ export class BitbucketService {
...
+    const MAX = 10000;
+    let start = 0;
+    for (;;) {
+      const params = new URLSearchParams({ limit: String(PAGE), start: String(start) });
+      if (ref) params.set('at', ref);
+      const page = await this.makeRequest<{
+        values?: string[];
+        isLastPage?: boolean;
+        nextPageStart?: number | null;
+      }>(`/projects/${projectKey}/repos/${repositorySlug}/files?${params.toString()}`);
```
ameer030: 🔵 **N01 — Bitbucket project key / repo slug reach the request path unencoded (accept a strict allowlist instead)**

The new `listFilePaths` interpolates `projectKey`/`repositorySlug` raw into the request path, while the sibling `testRepoConnection` in the same flow `encodeURIComponent`s the same two values. In practice this is safe today — the call only runs after `testRepoConnection`'s encoded probe returns 200, and `fetch`'s URL parsing collapses `..` segments before anything is sent — but the asymmetry invites a future caller who skips the probe.

- `bitbucketService.ts:57` — `` `/projects/${projectKey}/repos/${repositorySlug}/files?…` `` unencoded.
- `repoInspector.ts:95` — identical values `encodeURIComponent`'d for the probe that gates this call.
- `parseBitbucketRepoUrl`'s clone-URL fallback takes the last two path segments with no character allowlist, so dot-only or %-encoded junk reaches here; encodeURIComponent alone would not neutralize `..`.
- Not re-raising the already-discussed org-wide-token scope on this route — that's accepted design; this is only about path construction.

</details>

Validate `projectKey`/`repoSlug` against Bitbucket's own shape (`^[A-Za-z0-9_][A-Za-z0-9_.-]*$`-class allowlist) either in `parseBitbucketRepoUrl` or at the top of `listFilePaths`, and drop the raw interpolation.

</details>

rules: SEC-C22
agent: omp
run: n/a
</details>

apps/backend/src/agents/release-service-suggest/index.ts
```diff
@@ -0,0 +1,168 @@
...
+  return `Repository file and directory listing:
+
+<paths>
+${paths.join('\n')}
+</paths>`;
+}
+
+function isValidRegex(source: string): boolean {
+  try {
+    new RegExp(source);
```
ameer030, line 113: 🔵 **N02 — AI-suggested regexes are only compile-checked before they're stored & run against every commit path**

`suggestReleaseServices` validates LLM regexes with `new RegExp(source)` alone — compilability, not cost — and the save mutator re-runs the same compile-only check. Accepted patterns then execute per changed file in release analysis. A catastrophic-backtracking pattern compiles fine and would hang the matching pass.

- `apps/backend/src/agents/release-service-suggest/index.ts:103-108` — `isValidRegex` is try/`new RegExp` only; no length cap, no nested-quantifier screening.
- Save path compile-validates the same way (`apps/backend/src/zero/mutators.ts:7560-7565`); execution is per file path via `new RegExp(applicationRegex).test(filePath)` in `versionReleaseMappingService.ts`.
- Mitigations: the wizard forces human review (checkbox + editable regex field) before save, and the LLM is prompted for anchored prefix patterns — odds are low, hence a nit.

</details>

Cap regex length (say 200 chars) and reject nested quantifiers / unbounded repetition in `sanitizeServices` (a safe-regex-style check), so a pathological suggestion can't be applied even by a user clicking apply-all blindly.

</details>

rules: SEC-C25
agent: omp
run: n/a
</details>

apps/backend/src/controllers/releaseInsightsController.ts
```diff
@@ -0,0 +1,173 @@
...
+        );
+
+        const releaseInsights = { generatedAt: new Date().toISOString(), stats, ...llm };
+        await this.ticketRepository.updateTicketMetadata(ticketId, {
+          releaseInsights,
+          isGeneratingReleaseInsights: false,
+          insightsGenerationStartedAt: null,
+        });
+        res.json({ success: true });
+      } finally {
```
ameer030, line 141: 🔵 **N03 — success path writes ticket metadata twice**

On a successful generate, the code first writes `{ releaseInsights, isGeneratingReleaseInsights: false, insightsGenerationStartedAt: null }`, and the unconditional `finally` immediately writes the flag-clear pair again — a second identical read-modify-write per request (one extra trip through `updateTicketMetadata`'s findUnique→merge→update).

- `releaseInsightsController.ts:157-163` — success-path metadata write.
- `releaseInsightsController.ts:165-175` — finally re-writes the same two keys; it's only needed on the failure path.
- Harmless but redundant, and each write is a non-transactional RMW of the whole blob.

</details>

Track a `completed` flag in the try block and skip the finally-clear when the inner write already cleared it.

</details>

rules: STYLE-C14
agent: omp
run: n/a
</details>

apps/dashboard/src/components/ui/DataGrid/gridSelection.tsx
```diff
@@ -0,0 +1,136 @@
...
+      const onSelectionChanged = (): void => setIsSelected(params.node.isSelected());
+      params.api.addEventListener('selectionChanged', onSelectionChanged);
+      return () => params.api.removeEventListener('selectionChanged', onSelectionChanged);
+    }, [params.api, params.node]);
+
+    if (!isSelectable(params.node)) return null;
+
+    return (
+      <button
+        className='flex items-center justify-center h-full w-full'
```
ameer030, line 105: 🔵 **N04 — shared selection renderer nests interactive elements inside a button**

The new shared `IndexCellRenderer` renders the deselect "✓" button and the hover `<input type="checkbox">` *inside* an outer `<button>` — interactive content nested in interactive content, which is invalid HTML and confuses assistive tech (the outer element "contains" two different actions). The markup is inherited from the old inline TicketTable renderers, but moving it into `ui/DataGrid` promotes it to the shared pattern future grids will copy.

- `apps/dashboard/src/components/ui/DataGrid/gridSelection.tsx:98-131` — outer `<button>` wraps both inner interactive elements.
- Consumed by TicketTable and ReleaseDevTicketsTable from this PR onward.

</details>

Make the outer element a plain `<div>` with the same hover handlers (it triggers no action of its own); the inner button/input stay valid siblings.

</details>

rules: STYLE-C05
agent: omp
run: n/a
</details>

apps/dashboard/src/routes/ReleaseDetailScreen/ReleaseDetailScreen.tsx
```diff
@@ -316,14 +299,19 @@ const ReleaseDetailScreen = (): ReactElement => {
...
+  // Column config lives on the release board so it persists across all its releases.
+  const [releaseBoards] = useCachedQuery(
+    queries.boardsByIds({ boardIds: releaseTicket?.boardId ? [releaseTicket.boardId] : [] }),
+    { enabled: !!releaseTicket?.boardId },
+  );
+  const releaseBoard = releaseBoards?.[0];
+
   const persistedColumnKeys = useMemo(() => {
-    const md = releaseTicket?.metadata;
+    const md = releaseBoard?.metadata;
```
ameer030: 🔵 **N05 — previously saved column picks are silently orphaned**

Column selections used to persist on `releaseTicket.metadata.devTicketColumns`; this PR reads only `releaseBoard.metadata.devTicketColumns` with no read-side fallback. After deploy, every user finds their picked columns reset to empty on every release, one time, with no explanation.

- `ReleaseDetailScreen.tsx:317-327` — `persistedColumnKeys` reads the board key only.
- Old read (`origin/main`) read the ticket key. Preference-only impact — nothing corrupts, and board-level persistence is a real improvement (columns now carry across releases of the board). ARCH-C01 just favors a read-side fallback or an explicit note.

</details>

One-line dual-read for a release or two: `board.metadata.devTicketColumns ?? ticket.metadata.devTicketColumns`; or call the one-time reset out in the PR body so nobody files it as a bug.

</details>

rules: ARCH-C01
agent: omp
run: n/a
</details>

apps/dashboard/src/components/Release/ReleaseDevTicketsTable/ReleaseDevTicketsTable.tsx
```diff
@@ -0,0 +1,176 @@
...
+      <div className='overflow-hidden rounded-lg border border-border'>
+        <AgGridReact<GridRow>
+          getRowId={params => params.data.id}
+          rowData={gridRows}
+          columnDefs={columnDefs}
+          theme={theme}
+          domLayout='autoHeight'
+          rowHeight={44}
+          headerHeight={44}
+          rowSelection={{
```
ameer030: 🔵 **N06 — fresh ag-grid option objects each render**

`ReleaseDevTicketsTable` re-creates `rowSelection={{…}}`, `getRowId`, `isRowSelectable` and `isFullWidthRow` on every render, so each search keystroke hands ag-grid new-identity option objects to re-process. Cost is small at this table's size, and TicketTable ships the same shape — but since this PR introduces the shared `gridSelection.tsx` abstraction, hoisting these to module constants keeps the new table cheaper than the old one.

- `ReleaseDevTicketsTable.tsx:146-169` — inline objects/lambdas as props to `AgGridReact`; `columnDefs` and the cancelled-stage callback *are* correctly memoized/ref-stabilized, so this is the only churny prop set left.

</details>

Hoist `rowSelection`, `getRowId`, `isRowSelectable`, `isFullWidthRow` to module-level constants.

</details>

rules: PERF-C07
agent: omp
run: n/a
</details>

ameer030 (commented):
🔴 **B01 — PR description is the untouched template**

The entire PR body is the unfilled template: no Description, no ticket link, no Areas Touched checked, no test section (answer to "How did you test it?"), and the HTML placeholder comments (`<!-- What changed, and why -->`) survive verbatim. On a 35-file, ~2.5k-line feature (search/sort/filters + two new AI agents + two new authenticated endpoints) reviewers are forced to reconstruct intent from the diff.

- `gh pr view 1613 --json body`: every section is its default; all checkboxes unchecked, including "Reviewed my own diff" and "Added / updated tests".
- Checklist PROC-C15 declares blank sections and surviving placeholder comments a review failure; PROC-C02 requires declared scope; PROC-C04 requires precise verification statements.

</details>

Fill the template before merge: what changed and why + XYNE-62866 link, Areas Touched (backend, dashboard, packages/shared), the API-contract section (two new POST routes), "How did you test it" (manual run of the wizard suggest + release insights + the new table), and the "Not covered by tests — why?" note since no test files ship.

</details>

rules: PROC-C15, PROC-C02, PROC-C04
agent: omp
run: n/a
</details>

ameer030:
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 1 |
| 🟡 Better to fix | 4 |
| 🔵 Nit | 6 |

### 🔁 Re-check
- 💬 **Existing threads:** 2 fixed ✅ · 0 still open — @devesh-juspay's in-flight-flag robustness (stale-window + 409 verified in head) and the org-wide-token scope note (now documented inline; `/release-insights` re-gated to RELEASE-MANAGER WRITE)
- ✅ **Verified OK:** new Zero query mirrored byte-identically (backend ↔ shared) with `@@index([ticketId])` and a workspaceId-pinned ACL · both new routes workspace-scope their ids (ticket 404 / projectId 404 before use) · LLM outputs zod-validated with array caps, services human-reviewed before save · VCS hosts pinned to config, outbound dials carry 30s deadlines · TicketTable → shared `gridSelection` extraction is behavior-identical (cap, toast, tracking names) · release-metrics roll-up fully guard-railed against malformed/legacy metadata

### 📌 Verdict
1 blocker must be fixed before merge ❌ — fill in the PR template (description, scope, how-tested); the 4 better-to-fix items (client-side filter cap, board-metadata overwrite, role-gated column saving, Replace-all id reseeding) deserve answers before merge too.

apps/dashboard/src/routes/ReleaseDetailScreen/ReleaseDetailScreen.tsx
sumantop: Accepted boundary, now stated in the PR body: the search/filters are exact up to 1,000 dev tickets per release; beyond that the amber banner discloses the cap and CSV export bypasses the window. Pushing status/type/hotfix/owner predicates into `applicationReleaseTicketsByReleaseId` args is the follow-up if a release actually outgrows this — no current release is close, so I'm not taking the query-shape change in this PR.

apps/dashboard/src/routes/ReleaseDetailScreen/ReleaseDetailScreen.tsx
sumantop: Valid — `toggleColumn` was sending the whole `metadata` blob from the client snapshot. Fixed in 2a70995342: `board.update` takes an optional `devTicketColumns: string[]` (shared + backend mirror) and merges it into the freshly read `board.metadata` server-side, same pattern as the ETA args; `ReleaseDetailScreen` now sends only `{ boardId, devTicketColumns }`. Sibling keys (`etaManagement` etc.) survive a stale client.

apps/dashboard/src/routes/ReleaseDetailScreen/ReleaseDetailScreen.tsx
sumantop: Valid. 2a70995342 adds `useCanManageRelease()` (admin/owner or RELEASE-MANAGER WRITE/ADMIN — same rule as the backend `assertReleaseManageAccess`) in `hooks/usePermissions.ts`; the Add-column dropdown renders only when it's true. `ProjectDetailScreen`'s inline Create-Release predicate now uses the same hook, so the two gates can't drift.

apps/dashboard/src/components/Release/ReleaseConfigWizard/useReleaseConfigForm.ts
sumantop: Valid — `replace` reseeded `id`/`boardId` for every row, and save deletes apps missing from the payload, so an unchanged service got its board torn down and recreated. 2a70995342: `toRow(service, existing?)` keeps `existing.id`/`existing.boardId` for a same-named (trimmed, case-insensitive) service in replace mode; new services still get fresh ids; add mode unchanged.

apps/backend/src/services/bitbucketService.ts
sumantop: Done in 2a70995342: `listFilePaths` validates project key and repo slug against `^[A-Za-z0-9_][A-Za-z0-9_.-]*$` and `encodeURIComponent`s them before building the path.

apps/backend/src/agents/release-service-suggest/index.ts
sumantop: Done in 2a70995342: `isValidRegex` now also rejects sources longer than 200 chars and any group that is quantified and contains a quantifier (`(a+)+`, `(.*)*`, `((x|y){2,})+`), via a small paren-stack walk that skips escaped chars. Checked against the real source: `^(foo|bar)+/`, `^(a|b)*c`, `^\\(x+\\)+$` still pass; the three nested forms are rejected.

apps/backend/src/controllers/releaseInsightsController.ts
sumantop: Done in 2a70995342: the success write is the only write on the happy path; `finally` clears the in-flight flag only when `completed` is still false (failure/throw).

apps/dashboard/src/components/ui/DataGrid/gridSelection.tsx
sumantop: Done in 2a70995342: the outer wrapper in `gridSelection.tsx` is a `<div>`; the inner controls stay the interactive elements.

apps/dashboard/src/routes/ReleaseDetailScreen/ReleaseDetailScreen.tsx
sumantop: Done in 2a70995342: `persistedColumnKeys` reads `releaseBoard.metadata.devTicketColumns` first and falls back to the legacy `releaseTicket.metadata` location, so picks saved before this change still apply; the next toggle writes them to the board.

apps/dashboard/src/components/Release/ReleaseDevTicketsTable/ReleaseDevTicketsTable.tsx
sumantop: Done in 2a70995342: `rowSelection`, `getRowId`, `isRowSelectable`, `isFullWidthRow` are module-level constants in `ReleaseDevTicketsTable.tsx`.

sumantop:
Review round addressed in 2a70995342 — W02/W03/W04 + N01–N06 fixed as replied inline; W01 kept as the accepted 1,000-row boundary and documented in the PR body; B01: PR template filled (scope, Zero/API/config sections, what was and wasn't tested). Local CI mirror green: `@xyne/shared` build, backend `tsc`, dashboard lint / format:check / typecheck / build.

apps/dashboard/src/routes/ReleaseDetailScreen/ReleaseDetailScreen.tsx
```diff
@@ -1025,70 +932,34 @@ const ReleaseDetailScreen = (): ReactElement => {
...
+                      }
+                      baseRoute={baseRoute}
+                      releaseTicketXyneId={releaseTicket?.xyneId}
+                      releaseVersion={releaseVersion}
+                      onFiltersActiveChange={active => {
+                        setFiltersLoadAll(active);
+                        if (active) setArtCursorStack([null]);
+                      }}
+                    />
+                    {filtersLoadAll && (artPage?.length ?? 0) >= ART_FILTER_LOAD_ALL && (
```
ameer030, line 954: 🔵 **N01 — the "newest 1,000" banner can claim a release has more rows than it does**

When a release has *exactly* 1,000 dev tickets, load-all mode fetches `limit: 1000`, so the banner's "this release has more, so matches in older rows may not appear" is untrue — there are no older rows. The banner exists precisely to disclose the cap honestly, so at that boundary it misreports it.

- `ReleaseDetailScreen.tsx:213` — `ART_FILTER_LOAD_ALL = 1000`
- `ReleaseDetailScreen.tsx:341` — load-all mode fetches `limit: ART_FILTER_LOAD_ALL` (exactly 1,000; the paged mode uses `ART_PAGE_SIZE + 1` precisely to detect "has more")
- `ReleaseDetailScreen.tsx:954` — banner fires at `artPage.length >= ART_FILTER_LOAD_ALL`, which cannot distinguish "exactly 1,000 rows" from "truncated at 1,000"

</details>

Fetch `ART_FILTER_LOAD_ALL + 1` in load-all mode, slice to 1,000 for rendering, and gate the banner on the extra row:

```ts
limit: filtersLoadAll ? ART_FILTER_LOAD_ALL + 1 : ART_PAGE_SIZE + 1,
```

then `const artWindowTruncated = (artPage?.length ?? 0) > ART_FILTER_LOAD_ALL` and render rows from `artPage.slice(0, ART_FILTER_LOAD_ALL)` — the same over-fetch pattern the paged path already uses.

</details>

rules: TEST-C07
agent: omp
run: n/a
</details>

ameer030 (approved):
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 0 |
| 🟡 Better to fix | 0 |
| 🔵 Nit | 1 |

### 🔁 Re-check
- 💬 **Existing threads:** 11 fixed ✅ · 1 satisfied by reply · 0 still open — @devesh-juspay's in-flight-flag and repo-scope threads resolved; @ameer030's W01 (1,000-row client filter window) answered as a declared accepted boundary in the PR body with the amber banner + CSV bypass and a named follow-up (server-side filter args for `applicationReleaseTicketsByReleaseId`); all of @ameer030's W02–W04 and N01–N06 verified fixed at e2654bf10.
- ✅ **Verified OK:** tenant-scoped 404 + RELEASE-MANAGER WRITE gate on both new endpoints (`authorizePrivilegedOrResource` mirrors `assertReleaseManageAccess` and `useCanManageRelease`); insights in-flight flag claimed atomically via Serializable transaction with stale-window and `completed`-guarded cleanup; LLM output Zod-validated with length/nested-quantifier regex guards (SEC-C25); Bitbucket/GitHub hosts pinned to config — no SSRF surface (SEC-C19); `board.update` devTicketColumns merged into freshly-read metadata with exact client/server mutator mirror (ARCH-C10); `applicationReleaseTicketsByDevTicketId` mirrored backend ↔ shared under the existing workspace-pinned ART query ACL (SEC-C05/C06); grid selection extraction is behavior-identical in TicketTable; "Replace all" now preserves same-named app ids/boards.

### 📌 Verdict
No blockers — approving ✔; the one nit above only sharpens a boundary banner's honesty at exactly 1,000 rows.

Mohan0544 (approved):

santhoshJus (approved):

PR #1622, opened 2026-09-07
fix: prevent duplicate ticket activities from debounced auto-save

XyneSpaces:
## Problem
Editing a ticket's title and/or description from the ticket Details panel could emit many duplicate activity rows (~3 for a title change, ~15+ for a description change) instead of one activity per field change. The activity feed and desk metrics were polluted with redundant TITLE/DESCRIPTION entries for a single user edit.

## Root cause
The debounced auto-save effects in `apps/dashboard/src/components/Tickets/TicketDetails/TicketDetails.tsx` depend on the whole `ticket` object. Each auto-save mutation bumps the ticket's `updatedAt` (and a description save triggers further server-side ticket writes), which changes the `ticket` object the effect depends on and re-arms the 600ms debounce. Under sync latency the written value has not yet propagated back into `ticket`, so the existing guard `trimmed === ticket.title` is still false and the re-armed timer re-sends the same edit — emitting one activity per cycle. The backend mutator is correct; this is purely a client-side re-dispatch loop.

## Fix
Track the last value we actually dispatched a save for, per field, in a ref (`lastSavedTitleRef` / `lastSavedDescriptionRef`). The ref is set immediately before dispatch, and each effect skips re-dispatching a value identical to the last one already sent. This makes the auto-save idempotent per distinct value regardless of how many times `ticket` re-renders mid-save. +12 lines, single file.

## Testing (proof)
Reproduced the ticket Details flow end-to-end against a local dev stack (dashboard + zero-cache + backend) driven by Playwright:
- Edited the title once and the description once.
- Confirmed the `ticket_activities` table contains exactly **one** TITLE row and **one** DESCRIPTION row for the edit (old→new values correct).
A screen recording of the corrected behavior is attached in the originating thread.

## Risk / rollout
Frontend-only change, no schema/migration/backfill. No behavior change other than suppressing duplicate writes. Safe to roll out with a normal dashboard deploy.

### DIFF
--- a/apps/dashboard/src/components/Tickets/TicketDetails/TicketDetails.tsx
+++ b/apps/dashboard/src/components/Tickets/TicketDetails/TicketDetails.tsx
@@ -592,6 +592,14 @@ export const TicketDetails: React.FC<TicketDetailsProps> = ({
   const [titleValue, setTitleValue] = useState('');
   const [editingDescription, setEditingDescription] = useState(false);
   const [descriptionValue, setDescriptionValue] = useState('');
+  // Remember the last value we actually dispatched a save for, per field.
+  // Each auto-save mutation bumps the ticket's updatedAt (and description saves
+  // trigger further server-side ticket writes), which changes the `ticket`
+  // object this effect depends on and re-arms the debounce. Without this guard
+  // the re-armed timer re-sends the same edit before the written value has
+  // propagated back into `ticket`, emitting one activity per cycle.
+  const lastSavedTitleRef = useRef<string | null>(null);
+  const lastSavedDescriptionRef = useRef<string | null>(null);
   const [showTagDropdown, setShowTagDropdown] = useState(false);
   const [isSubTicketModalOpen, setIsSubTicketModalOpen] = useState(false);
   const [isCreateTicketModalOpen, setIsCreateTicketModalOpen] = useState(false);
@@ -1888,8 +1896,10 @@ export const TicketDetails: React.FC<TicketDetailsProps> = ({
     if (!ticket || !editingTitle || isEmailDeskTicket) return;
     const trimmed = titleValue.trim();
     if (!trimmed || trimmed === ticket.title) return;
+    if (trimmed === lastSavedTitleRef.current) return;
     const ticketId = ticket.id;
     const timeoutId = setTimeout(() => {
+      lastSavedTitleRef.current = trimmed;
       void applyTicketUpdate(
         { id: ticketId, title: trimmed, updatedAt: Date.now() },
         'Failed to update title',
@@ -1904,8 +1914,10 @@ export const TicketDetails: React.FC<TicketDetailsProps> = ({
     if (!ticket || !editingDescription) return;
     const next = descriptionValue.trim();
     if (next === ticket.description) return;
+    if (next === lastSavedDescriptionRef.current) return;
     const ticketId = ticket.id;
     const timeoutId = setTimeout(() => {
+      lastSavedDescriptionRef.current = next;
       void applyTicketUpdate(
         { id: ticketId, description: next, updatedAt: Date.now() },
         'Failed to update description',
--- a/apps/dashboard/src/components/Tickets/TicketDetails/TicketDetails.tsx
+++ b/apps/dashboard/src/components/Tickets/TicketDetails/TicketDetails.tsx
@@ -1886,6 +1886,15 @@ export const TicketDetails: React.FC<TicketDetailsProps> = ({
     [zero],
   );
 
+  // Reset the per-field save guards when the viewed ticket changes. The refs
+  // track the last value dispatched for THIS ticket; without this, saving "X" on
+  // ticket A would make the auto-save skip an identical "X" typed on ticket B,
+  // since TicketDetails is not keyed by ticket id at its call sites.
+  useEffect(() => {
+    lastSavedTitleRef.current = null;
+    lastSavedDescriptionRef.current = null;
+  }, [ticket?.id]);
+
   // Debounced auto-save while editing — persist the title as the user types
   // rather than only on blur/exit. The blur/Enter handler still flushes an
   // immediate save; leaving edit mode flips `editingTitle`, whose cleanup clears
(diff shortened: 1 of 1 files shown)


apps/dashboard/src/components/Tickets/TicketDetails/TicketDetails.tsx
```diff
@@ -592,6 +592,14 @@ export const TicketDetails: React.FC<TicketDetailsProps> = ({
   const [titleValue, setTitleValue] = useState('');
   const [editingDescription, setEditingDescription] = useState(false);
   const [descriptionValue, setDescriptionValue] = useState('');
+  // Remember the last value we actually dispatched a save for, per field.
+  // Each auto-save mutation bumps the ticket's updatedAt (and description saves
+  // trigger further server-side ticket writes), which changes the `ticket`
+  // object this effect depends on and re-arms the debounce. Without this guard
+  // the re-armed timer re-sends the same edit before the written value has
+  // propagated back into `ticket`, emitting one activity per cycle.
+  const lastSavedTitleRef = useRef<string | null>(null);
```
PiyushJuspay, line 601: These refs are never cleared when ticketId changes, and TicketDetails isn't keyed by ticket id at any call site. So if the user saves "X" on ticket A and later types the same "X" on ticket B, the auto-save silently skips it (line 1899). Please reset both refs on ticket change

apps/dashboard/src/components/Tickets/TicketDetails/TicketDetails.tsx
```diff
@@ -1888,8 +1896,10 @@ export const TicketDetails: React.FC<TicketDetailsProps> = ({
     if (!ticket || !editingTitle || isEmailDeskTicket) return;
     const trimmed = titleValue.trim();
     if (!trimmed || trimmed === ticket.title) return;
+    if (trimmed === lastSavedTitleRef.current) return;
     const ticketId = ticket.id;
     const timeoutId = setTimeout(() => {
+      lastSavedTitleRef.current = trimmed;
```
PiyushJuspay, line 1914: H2 Ref marks the value as saved before we know the save worked
- What: The ref is set before the mutation runs (void, result ignored). If the server rejects the save or the network fails, the ref still holds the failed value, and the auto-save will never retry that exact text.
- Where: apps/dashboard/src/components/Tickets/TicketDetails/TicketDetails.tsx:1902 — lastSavedTitleRef.current = trimmed;
    void applyTicketUpdate( (same for description at line 1920)
- Who affected: End users whose save is rejected — this file's own comment (lines 1862-1866) documents that server-side rules reject some updates (e.g. Euler board constraints).
- Impact: After one error toast, the user's text stays unsaved until they blur/press Enter. Mostly recovered by the flush, but it's the exact case the ref should distinguish.
- Fix: applyTicketUpdate already returns a boolean — set the ref only when it resolves true (or clear the ref on failure).

PiyushJuspay (dismissed):

apps/dashboard/src/components/Tickets/TicketDetails/TicketDetails.tsx
```diff
@@ -1888,12 +1905,21 @@ export const TicketDetails: React.FC<TicketDetailsProps> = ({
     if (!ticket || !editingTitle || isEmailDeskTicket) return;
     const trimmed = titleValue.trim();
     if (!trimmed || trimmed === ticket.title) return;
+    if (trimmed === lastSavedTitleRef.current) return;
```
XyneSpaces, line 1908: **[major] R4 — duplicate-processing window: the blur/Enter flush doesn't consult the new last-saved guards.**

`handleSaveTitle` / `handleSaveDescription` still dispatch on `titleValue !== ticket.title` / `descriptionValue !== ticket.description` alone — and those are the Zero-synced values that lag an in-flight debounced save (the same sync latency this PR's root-cause section describes). Sequence: type "X" → debounce fires at +600 ms (`lastSavedTitleRef = "X"`, save in flight) → blur lands before the write replicates back into `ticket` → the flush's `!== ticket.title` guard is still true → a second `applyTicketUpdate` with the same value → a second activity row. The effect comment's "cleanup clears any pending debounce, so there's no duplicate write" only covers blur **before** the timer fires — not blur **after** dispatch but **before** sync.

Suggested fix: give the flush handlers the same guard — e.g. in `handleSaveTitle`: `if (titleValue.trim() && titleValue.trim() !== ticket.title && titleValue.trim() !== lastSavedTitleRef.current)`, and symmetrically `next !== lastSavedDescriptionRef.current` in `handleSaveDescription`. Failure recovery is preserved: once a debounced save is rejected, the `.then` has already cleared the ref, so the flush still re-sends it; the flush only defers while that exact value is genuinely in flight. Worth extending the Playwright proof with a "pause >600 ms, then blur" case so this window stays covered.

XyneSpaces (commented):
## Xyne Spaces Reviewer — PR #1622

**Verdict:** Changes requested — 1 major (duplicate window on the blur/Enter flush path). The core re-arm-loop fix is sound, and both of @PiyushJuspay's review comments are correctly addressed.

**Reviewed:** 1 file, 4 hunks (+34/−2) at `b06e64a1` · **Findings:** 0 blocking, 1 major, 2 nits

### Findings
| major | R4 | `TicketDetails.tsx:1908` (inline) | blur/Enter flush (`handleSaveTitle` / `handleSaveDescription`) doesn't check the new `lastSaved*Ref` guards, so a blur landing between debounce dispatch and Zero sync still double-dispatches → a duplicate activity row — the very thing this PR eliminates on the debounce path |
| nit | R7 | — | autosave idempotency verified only by a manual Playwright run; no regression e2e for "one activity per field edit", and the manual pass didn't cover blur-after-debounce |
| nit | R8 | PR description | "Fix" section still says "+12 lines, single file" (now +34/−2) and predates the ticket-change reset and the failure-clearing `.then` added in response to review |

### Previous-review items — both addressed ✓
- **Refs reset on ticket change** (Piyush, line 601): the new `[ticket?.id]` effect resets both refs; follows the file's established per-ticketId reset pattern (e.g. the sub-ticket-menu reset effect).
- **Ref cleared when a save is rejected** (Piyush, line 1902): `.then((saved) => …)` clears the ref on `false`, guarded by the `=== trimmed` / `=== next` check so a newer in-flight value isn't clobbered. Verified `applyTicketUpdate` always resolves a boolean (full try/catch around the mutator call), so the `.then` without `.catch` cannot produce an unhandled rejection.

### Recommended reviewer
**@PiyushJuspay** — this PR is the direct response to his two review comments; he has line-level context on exactly these effects and is the cheapest correct verifier of the rework.

|-----------|-------------------------|------------|-----------------|-------------|
| @PiyushJuspay | TicketDetails.tsx — reviewed this PR's exact effects (authored both comments this fix addresses) | no — `.github/CODEOWNERS` covers only ACL paths, nothing under `apps/dashboard/**` | unknown¹ | most recently active — reviewed this PR ~2.5h ago |
| @harshpreet931 | TicketDetails.tsx — recent edits (#1389 format pass, #1523 status-row icons) | no | unknown¹ | — |
| @adithyaganesh1 | TicketDetails.tsx — #1313 (hideBackNav) | no | unknown¹ | — |

¹ The issue-search API available to this automation does lexical matching, not GitHub's `review-requested:` qualifier — spot-checking #1695 (matches that search yet has no requested reviewers) proved those counts unusable, so pending-review load is reported as unknown rather than invented. Ranking therefore leans on familiarity + recency, per the selection rules.

@PiyushJuspay — requesting your re-review here (this automation's toolset has no review-request API; this mention is the request).

No migration files in this diff — migration-safety review not triggered.

PR #1542, opened 2026-09-04
fix: XYNE-61836 app ACL fix

HarimohanSharma11:
Migration-Changes:
  - unique index on app table

Services-Requiring-Redeployment:
  - backend

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Root Cause 
  Fixed the app deplication issue , via fixing the ACL and also backfilling / renaming the conflicting apps. Also shifted from baseacl to wrapped acls on apps 

## Areas Touched

- [ ] `apps/backend` (API / worker)
- [ ] `apps/dashboard`
- [ ] `apps/electron`
- [ ] `apps/xyne-claw` / `apps/xyne-claw-auth`
- [ ] `packages/shared` or another shared package
- [ ] Infra (docker, scripts, nix, CI)

---

## Zero (Sync) Changes

- [ ] No Zero changes — **skip this section**
- [ ] Adds/modifies a **mutator**
- [ ] Adds/modifies the **schema** (tables, columns, relationships)
- [ ] Adds/modifies a **query or ACL**

If any of the above:

- [ ] Server (`apps/backend/src/zero/mutators.ts`) and client (`apps/dashboard/src/zero/mutators.ts`) mutators mirror each other — same namespace, name, and args schema
- [ ] No `Date.now()` / `uuid()` generated inside a mutator
- [ ] Optimistic client result matches the server result (no state flash on rebase)
- [ ] New tables exported in `createSchema()` and given a Query ACL registered in `query-acl-factory.ts`
- [ ] Matching Prisma model + migration, client regenerated

## Schema & Data Changes

- [ ] No schema changes — **skip this section**
- [ ] Prisma schema modified (`apps/backend/prisma` and/or `prisma-common`)
- [ ] Migration added
- [ ] Backfill / data migration included

If any of the above:

- [ ] Clients regenerated (`db:generate`, `db:common:generate`)
- [ ] No new Postgres enum or enum value — enums are frozen (`pnpm run test:enum`)
- [ ] Backward compatible with deployed code, or rollout order noted below
- [ ] Backfill is idempotent

## Config, Environment & URLs

- [ ] No config changes — **skip this section**
- [ ] Env variable added / renamed / removed
- [ ] **Base URL, host, port, or origin changed** (API, Zero, Vespa, Claw, LiveKit, LiteLLM, …)
[...]

PR #1626, opened 2026-09-08
fix: recover Electron main window from renderer crashes (white screen after idle)

XyneSpaces:
## Problem
Every morning after the machine sits idle overnight, the Electron app shows a blank white main window and only a full quit-and-reopen fixes it. Root cause: the **main window registered no crash/hang/load-failure recovery**. During long idle/sleep the OS reclaims the backgrounded renderer (and/or GPU) process; the renderer dies and nothing reloads it, so the window stays white. The Claw overlay (`claw-overlay-window.ts`) and recording-pill (`recording-pill-window.ts`) windows already self-reload on `render-process-gone` — the main window did not. Idle is the trigger, not the root cause.

## Fix
Add capped exponential-backoff recovery to `createMainWindow` in `apps/electron/src/window/manager.ts`, mirroring the overlay/pill windows:
- `render-process-gone` (ignoring `clean-exit`) → schedule reload
- `did-fail-load` (main frame only, ignoring `ERR_ABORTED` / -3) → schedule reload
- `unresponsive` → schedule reload; `responsive` cancels a pending reload
- Backoff 1s → 2s → 4s (cap 8s), max 3 attempts; on exhaustion load the bundled `load-error.html` instead of leaving a blank window
- Reset the backoff counter on a successful `did-finish-load`
- Recovery reloads via the existing `loadApp()` path, so mTLS/enrollment checks are re-run exactly as on a normal launch

## Testing (POT)
Recorded a standalone Electron harness that reproduces the exact renderer-crash path (`forcefullyCrashRenderer()`):
- **Before** (no handler): renderer crash → window stays blank white (the reported bug)
- **After** (this handler): renderer crash → `render-process-gone` fires → window auto-reloads and recovers (render count #2, "Recovered automatically after renderer crash")
Video attached in the Spaces thread. `tsc --noEmit` passes.

## Risk / rollout
- Scoped to the main `BrowserWindow`; no schema/API/worker changes.
- Retry cap + backoff prevent reload storms; `unresponsive`-triggered reloads are cancelled by `responsive` if the renderer recovers on its own.
- No feature flag needed; behavior only activates on abnormal renderer termination.

### DIFF
--- a/apps/electron/src/window/manager.ts
+++ b/apps/electron/src/window/manager.ts
@@ -262,6 +262,55 @@ let isCompactMode = false;
 let isReloading = false;
 let normalBounds: { width: number; height: number } | null = null;
 
+// --- Main-window renderer crash / load-failure recovery ------------------
+// The overlay and recording-pill windows already self-reload when their
+// renderer dies; the main window previously did not, so an overnight renderer
+// crash (OS reclaiming a backgrounded renderer / GPU process after sleep) left
+// a blank white window that only a full app relaunch could clear. These
+// helpers reload the main window with capped exponential backoff.
+const MAX_RENDERER_RECOVERY_ATTEMPTS = 3;
+let rendererRecoveryAttempts = 0;
+let rendererRecoveryTimer: ReturnType<typeof setTimeout> | null = null;
+
+function resetRendererRecovery(): void {
+  rendererRecoveryAttempts = 0;
+  if (rendererRecoveryTimer) {
+    clearTimeout(rendererRecoveryTimer);
+    rendererRecoveryTimer = null;
+  }
+}
+
+function scheduleRendererRecovery(window: BrowserWindow, reason: string): void {
+  // A recovery is already queued; don't stack reloads.
+  if (rendererRecoveryTimer) return;
+
+  if (rendererRecoveryAttempts >= MAX_RENDERER_RECOVERY_ATTEMPTS) {
+    log.error(
+      `[WindowManager] Main-window recovery gave up after ${rendererRecoveryAttempts} attempts (${reason})`,
+    );
+    // Don't leave the user on a blank window — show the bundled error page.
+    const errorPage = path.join(__dirname, '..', '..', 'assets', 'load-error.html');
+    void window.loadFile(errorPage).catch(() => {});
+    return;
+  }
+
+  const attempt = ++rendererRecoveryAttempts;
+  const delayMs = Math.min(1000 * 2 ** (attempt - 1), 8000); // 1s, 2s, 4s (cap 8s)
+  log.warn(
+    `[WindowManager] Scheduling main-window recovery attempt ${attempt}/${MAX_RENDERER_RECOVERY_ATTEMPTS} in ${delayMs}ms (${reason})`,
+  );
+
+  rendererRecoveryTimer = setTimeout(() => {
+    rendererRecoveryTimer = null;
+    if (!window || window.isDestroyed()) return;
+    log.warn(`[WindowManager] Reloading main window (recovery attempt ${attempt}, ${reason})`);
+    void loadApp(window).catch((err) => {
+      log.error('[WindowManager] Recovery reload failed:', err);
+      scheduleRendererRecovery(window, reason);
+    });
+  }, delayMs);
+}
+
 export function getMainWindow(): BrowserWindow | null {
   return mainWindow;
 }
@@ -429,12 +478,52 @@ export async function createMainWindow(options?: { inactive?: boolean }): Promis
 
   // Register RBAC check before loading the app
   mainWindow.webContents.on('did-finish-load', async () => {
+    // A successful load means we're healthy again — clear any recovery backoff.
+    resetRendererRecovery();
     const bodyText = await mainWindow?.webContents.executeJavaScript('document.body.innerText').catch(() => '');
     if (bodyText && bodyText.includes('RBAC: access denied')) {
       const errorPage = path.join(__dirname, '..', '..', 'assets', 'vpn-error.html');
       void mainWindow?.loadFile(errorPage);
     }
   });
+
+  // Recover the main window from renderer death, hangs, and load failures.
+  // Without these, an overnight renderer crash leaves a blank white window that
+  // only quit-and-reopen can fix (see RCA).
+  mainWindow.webContents.on('render-process-gone', (_event, details) => {
+    if (details.reason === 'clean-exit') return;
+    log.error(
+      `[WindowManager] Main renderer gone: ${details.reason} (exitCode=${details.exitCode})`,
+    );
+    if (mainWindow && !mainWindow.isDestroyed()) {
+      scheduleRendererRecovery(mainWindow, `render-process-gone:${details.reason}`);
+    }
+  });
+
+  mainWindow.webContents.on('unresponsive', () => {
+    log.error('[WindowManager] Main window became unresponsive');
+    if (mainWindow && !mainWindow.isDestroyed()) {
+      scheduleRendererRecovery(mainWindow, 'unresponsive');
+    }
+  });
+
+  // If it recovers on its own before the reload fires, cancel the pending reload.
+  mainWindow.webContents.on('responsive', () => {
+    log.info('[WindowManager] Main window responsive again — cancelling pending recovery');
+    resetRendererRecovery();
+  });
+
+  mainWindow.webContents.on(
+    'did-fail-load',
+    (_event, errorCode, errorDescription, _validatedURL, isMainFrame) => {
+      // ERR_ABORTED (-3) fires on normal in-app navigations; ignore sub-frames too.
+      if (!isMainFrame || errorCode === -3) return;
+      log.error(`[WindowManager] Main frame load failed (${errorCode}): ${errorDescription}`);
+      if (mainWindow && !mainWindow.isDestroyed()) {
+        scheduleRendererRecovery(mainWindow, `did-fail-load:${errorCode}`);
+      }
+    },
+  );
 
   await loadApp(mainWindow);
 


apps/electron/src/window/manager.ts
```diff
@@ -429,13 +478,53 @@ export async function createMainWindow(options?: { inactive?: boolean }): Promis
 
   // Register RBAC check before loading the app
   mainWindow.webContents.on('did-finish-load', async () => {
+    // A successful load means we're healthy again — clear any recovery backoff.
+    resetRendererRecovery();
```
XyneSpaces, line 482: **[blocking] R2 — the recovery cap is defeated: `resetRendererRecovery()` fires on every `did-finish-load`, including intermediate pages.** `loadApp()` always loads the bundled `loading.html` first, and that page virtually always loads — so the counter is zeroed on every recovery cycle *before the real app URL is even attempted*. In a persistent-crash scenario (e.g. a bad deploy that crashes the renderer on every app load — exactly the case the 3-attempt cap exists for) the counter never reaches 3: the window reloads every ~1s + load time **forever**, re-running mTLS + full app load + the RBAC `executeJavaScript` check each cycle, and `load-error.html` is never shown. This contradicts the PR's stated "Retry cap + backoff prevent reload storms" — it produces precisely a reload storm in the scenario the cap targets. The `unresponsive`↔`responsive` flip-flop case also never escalates, for the same reason.

↳ FIX: reset only when the **real app page** finishes loading — e.g. reset inside `loadUrl`'s success path (where the `dashboardLoad` metric already fires) or gate this reset on `mainWindow.webContents.getURL()` matching the app origin. `loading.html` / error pages must not reset the backoff.

apps/electron/src/window/manager.ts
```diff
@@ -262,6 +262,55 @@ let isCompactMode = false;
 let isReloading = false;
 let normalBounds: { width: number; height: number } | null = null;
 
+// --- Main-window renderer crash / load-failure recovery ------------------
```
XyneSpaces, line 265: **[major] R5/R9 — duplicates open PR #347 (same bug, different mechanism).** PR #347 ("recover main window after renderer crashes", open since 2026-08-08) implements this same recovery as a `CrashRetryBudget` policy in `error-handler.ts`, *with unit tests* (`crash-recovery-policy.test.ts`). Landing this PR as-is leaves: overlay = immediate reload, pill = immediate reload, main = this backoff, plus a fourth competing policy sitting unmerged in #347 — four divergent answers to one question. Decide which supersedes which, close the other, and ideally extract one shared crash-recovery policy module all windows use.

apps/electron/src/window/manager.ts
```diff
@@ -262,6 +262,55 @@ let isCompactMode = false;
...
+  // A recovery is already queued; don't stack reloads.
+  if (rendererRecoveryTimer) return;
+
+  if (rendererRecoveryAttempts >= MAX_RENDERER_RECOVERY_ATTEMPTS) {
+    log.error(
+      `[WindowManager] Main-window recovery gave up after ${rendererRecoveryAttempts} attempts (${reason})`,
+    );
+    // Don't leave the user on a blank window — show the bundled error page.
+    const errorPage = path.join(__dirname, '..', '..', 'assets', 'load-error.html');
+    void window.loadFile(errorPage).catch(() => {});
```
XyneSpaces, line 293: **[minor] Rule 79 — silent catch.** `void window.loadFile(errorPage).catch(() => {})` swallows the failure to load the error page — the user's last-resort state. If this fails, we're back to a blank window with zero telemetry. Log it: `.catch((err) => log.error('[WindowManager] Failed to load load-error.html:', err))`.

apps/electron/src/window/manager.ts
```diff
@@ -262,6 +262,55 @@ let isCompactMode = false;
...
+    return;
+  }
+
+  const attempt = ++rendererRecoveryAttempts;
+  const delayMs = Math.min(1000 * 2 ** (attempt - 1), 8000); // 1s, 2s, 4s (cap 8s)
```
XyneSpaces, line 298: **[minor] R7 — backoff/cap policy has no automated test.** The POT covers the happy path (one crash → recover). The behavior that actually matters is the policy: 1s→2s→4s, cap 8s, give-up at 3, *no reset on intermediate pages*. That logic is pure and trivially extractable — pull it into a small module (as #347 did with `crash-recovery-policy.ts`) and unit-test the boundaries; guideline 204 asks for tests on behavior changes where feasible.

apps/electron/src/window/manager.ts
```diff
@@ -429,13 +478,53 @@ export async function createMainWindow(options?: { inactive?: boolean }): Promis
...
     if (bodyText && bodyText.includes('RBAC: access denied')) {
       const errorPage = path.join(__dirname, '..', '..', 'assets', 'vpn-error.html');
       void mainWindow?.loadFile(errorPage);
     }
   });
 
+  // Recover the main window from renderer death, hangs, and load failures.
+  // Without these, an overnight renderer crash leaves a blank white window that
+  // only quit-and-reopen can fix (see RCA).
+  mainWindow.webContents.on('render-process-gone', (_event, details) => {
```
XyneSpaces, line 493: **[nit, assumption] R2 — only renderer death is covered; the RCA also names GPU reclamation.** The description attributes the overnight white screen to the OS reclaiming the renderer *and/or GPU* process, but only `render-process-gone` is handled; the global `child-process-gone` handler in `error-handler.ts` only logs. A pure GPU-process death does **not** fire `render-process-gone` (Chromium restarts the GPU process itself; the window can stay blank until a repaint). Assumption — depends on fleet telemetry: #347's data shows `render_process_gone reason=crashed`, which supports the renderer hypothesis; if GPU-only deaths also show up, consider handling `child-process-gone` with `details.type === 'GPU'` (force repaint/reload).

XyneSpaces (commented):
Reviewed 1 file (2 hunks, +89) against the Xyne Spaces review guidelines. 1 blocking, 1 major, 2 minor, 1 nit — details inline and in the summary comment. Verified OK (not re-flagged): `load-error.html` asset exists, `isDestroyed()` guards in the timer callback, per-webContents handler registration (no duplicates), `clean-exit` and `ERR_ABORTED` (-3) filters, existing `log`/`path`/`BrowserWindow` imports. No migration files in this PR — migration gate not applicable.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1626

**Verdict:** Changes requested — 1 blocking issue (recovery cap is self-defeating)
**Reviewed:** 1 file, 2 hunks (+89/−0) · **Findings:** 1 blocking, 1 major, 2 minor, 1 nit

### Findings
| blocking | R2 | apps/electron/src/window/manager.ts:482 | `resetRendererRecovery()` fires on every `did-finish-load` — including the intermediate `loading.html` that `loadApp()` always loads first — so the attempts counter resets on every cycle. Persistent crash = infinite ~1s-cadence reload loop; the 3-attempt cap never trips and `load-error.html` is unreachable. Contradicts the stated "retry cap + backoff prevent reload storms". Fix: reset only on successful **app-URL** load (e.g. `loadUrl` success path / match `getURL()` against the app origin). |
| major | R5/R9 | apps/electron/src/window/manager.ts:265 | Duplicates open PR #347 (same bug, `CrashRetryBudget` in error-handler.ts, has unit tests). Four divergent recovery policies if both live — reconcile/close one, ideally extract a shared policy module. |
| minor | Rule 79 | apps/electron/src/window/manager.ts:293 | `void window.loadFile(errorPage).catch(() => {})` silently swallows error-page load failure — log it. |
| minor | R7 | apps/electron/src/window/manager.ts:298 | Backoff/cap policy has no automated test (POT covers the happy path only); the pure logic is extractable and testable. |
| nit | R2 (assumption) | apps/electron/src/window/manager.ts:493 | RCA names GPU reclamation too, but only `render-process-gone` is handled; a pure GPU death wouldn't trigger recovery. Telemetry in #347 supports the renderer hypothesis. |

**Verified OK (not flagged):** `load-error.html` asset exists; `isDestroyed()` guards in the timer callback; handlers registered once per webContents; `clean-exit` and `ERR_ABORTED` (−3) filters correct; existing `log`/`path`/`BrowserWindow` imports reused. No migration files — migration gate n/a.

### Recommended reviewer
**@harshpreet931** — owns the most relevant history for this exact file: #1470 (link/navigation policy in `apps/electron/src/window/manager.ts`, the same file this PR changes) and #1394 (recording-pill window — the sibling recovery code this PR mirrors).

| @harshpreet931 (Harshpreet Singh) | `window/manager.ts` via #1470; `recording-pill-window.ts` via #1394 | no — CODEOWNERS has no `apps/electron/**` entry | unverifiable via search (qualifier dropped; 0 in sampled open PRs) | active in Spaces Sep 7 |
| @HarimohanSharma11 (Harimohan Sharma) | adjacent electron (#1585 native domain conflict, #1542 ACL) | no | 0 confirmed | active in Spaces Sep 7 |

Rationale: familiarity wins — Harshpreet has direct commits in this file and in the sibling window this PR mirrors; workload is effectively equal (both ~zero pending review requests in this repo), both active yesterday. Harimohan is the natural second reviewer if you want two.

**Note on the review request:** this automation has no GitHub reviewer-request API tool available, so the formal "review requested" flag couldn't be set — the @mention above notifies on GitHub. Someone with `gh` access can run `gh pr edit 1626 --add-reviewer harshpreet931` to make it official.

### Ticket key
Yes — repo convention is `<type>: XYNE-NNNNN <subject>` (see #1624 `XYNE-62896`, #1623 `XYNE-57690`, #1585 `XYNE-56391`). Create the ticket, then update the PR title to e.g. `fix: XYNE-NNNNN recover Electron main window from renderer crashes` (branch rename optional but `fix/*` shape is already correct).

### Answers to the handoff's other question
The POT harness covers the single-crash happy path well; the follow-up worth doing is a second scenario where the app URL *persistently* crashes the renderer — with the current code that loop never terminates (see blocking finding), and once fixed it should land on `load-error.html` after 3 attempts.

XyneSpaces:
👋 Review request — tagging the assigned reviewers to please review this PR:
- @harshpreet931 (primary — owns window/manager.ts history)
- @HarimohanSharma11 (secondary — recent Electron work)

There's 1 blocking finding (renderer-recovery cap resets on the intermediate loading.html load, so the 3-attempt cap never trips). Please take a look when you get a chance. Thanks!

XyneSpaces:
Linked to **XYNE-55230 — Fix electron white-screen and theme sync**: https://app.spaces.xyne.juspay.net/chat/dir/cmi39e2jj00fex66cheedyjc8/cmsei5oq30ft7nzmqubrc99q9

This PR covers the **white-screen half** of XYNE-55230; the theme-sync work remains a follow-up under the same ticket.

Root-cause note: production log analysis for the affected user shows the white screen is driven primarily by a Zero connectivity/hydration dead-end after overnight idle (socket dropped while the window is hidden → "Client not found" on wake → boot queries never complete → blank), not by a renderer-process crash. The renderer-crash / did-fail-load / unresponsive recovery handlers added here are a defensive safety net for the main window (which previously had none, unlike the overlay/pill windows). A separate follow-up is recommended to harden the dashboard Zero-resume path (InitialStateLoader.tsx / ZeroProvider.tsx).

XyneSpaces:
Review requested 🙏 — @harshpreet931 (primary) and @HarimohanSharma11 (secondary), could you please review this PR?

@harshpreet931 you own the strongest history for this change (#1470 touched `window/manager.ts`; #1394 built the recording-pill window this PR mirrors). Note the current blocking review feedback: `resetRendererRecovery()` is called on every `did-finish-load` including the intermediate loading page, so the 3-attempt cap never trips and a persistent renderer crash would auto-reload indefinitely — the reset should only fire on a successful app-URL load. There's also overlap with PR #347 (CrashRetryBudget in error-handler.ts) to reconcile.

harshpreet931 (approved):

PR #1658, opened 2026-09-08
feat: XYNE-62464 redirect-xyne-links-to-electron-app

OmThakur2511:
  - dashboard

## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Xyne links from anywhere else should open electron app

https://github.com/user-attachments/assets/3aac9cd7-a1fa-4d5c-b69f-ea653582aaac

https://github.com/user-attachments/assets/c15ed3a0-4822-4478-a2a9-f7b3c3029fa2


apps/dashboard/src/main.tsx
```diff
@@ -115,4 +116,9 @@ if ('serviceWorker' in navigator && !isCustomProtocol) {
 
 globalClickTracker.initialize();
 
+// If this is a shared link opened in a plain browser, offer to hand it off to
+// the desktop app (Slack-style). No-op inside Electron. Runs before render so
+// the interstitial appears without a flash of the underlying app.
+maybeOpenInDesktopApp();
+
 ReactDOM.createRoot(document.getElementById('root')!).render(<App />);
```
sithaarth, line 126: `maybeOpenInDesktopApp()` shows the interstitial but React still fully mounts `<App />` right after — the whole app boots (auth, network, router) underneath the overlay for browser-only users who never leave. Comment says "runs before render so the interstitial appears without a flash", but render is not gated on the handoff. If the intent is to avoid booting the app during handoff, early-return / defer `createRoot` while the overlay is active. Otherwise reword the comment — it implies render is deferred when it isn't.

apps/dashboard/src/utils/openInDesktopApp.ts
```diff
@@ -0,0 +1,379 @@
...
+      onOpen: () => triggerOpen(false),
+      onContinue: () => dismiss(),
+    });
+
+    document.addEventListener('visibilitychange', onVisibility);
+    document.body.appendChild(controls.overlay);
+
+    // Best-effort silent auto-attempt on load; the interstitial stays either way
+    // so the user can click "Open Xyne Spaces" or "Continue in browser".
+    triggerOpen(true);
```
sithaarth, line 437: Auto-attempt via hidden iframe fires on page load with no user gesture. Many browsers block or throttle custom-scheme navigation without a gesture, and where it isn't blocked the OS "Open Xyne Spaces?" prompt appears automatically — then the user can click "Open Xyne Spaces" and get a second prompt. Consider dropping the auto-iframe and only attempting on the button click (the gesture path is already the reliable one per your own comment on `triggerOpen`).

apps/dashboard/src/utils/openInDesktopApp.ts
```diff
@@ -0,0 +1,379 @@
...
+  try {
+    if (window.sessionStorage.getItem(ATTEMPTED_KEY) === 'true') return;
+  } catch {
+    // sessionStorage unavailable — fall through and attempt once.
+  }
+
+  if (!isDeepLinkableRoute(window.location.pathname)) return;
+
+  try {
+    window.sessionStorage.setItem(ATTEMPTED_KEY, 'true');
```
sithaarth, line 355: `ATTEMPTED_KEY` is in `sessionStorage`, so browser-only users (app not installed) get the full interstitial on every new tab/session for every shared link, with no way to detect the app's absence. Consider a persistent `localStorage` "always continue in browser" opt-out (set when they click Continue) so repeat visitors aren't nagged — this is what Slack does.

apps/dashboard/src/utils/openInDesktopApp.ts
```diff
@@ -0,0 +1,379 @@
...
+        // A real user gesture makes the OS "Open Xyne Spaces?" prompt reliable.
+        window.location.href = deepLink;
+      }
+    };
+
+    let controls: OverlayControls | null = null;
+
+    const onVisibility = (): void => {
+      // Confirmed handoff — the app took foreground shortly after an attempt.
+      if (document.visibilityState === 'hidden' && Date.now() < armedUntil && phase !== 'opened') {
```
sithaarth, line 389: False-positive risk: any tab-hidden within `HANDOFF_ARM_MS` (8s) after an attempt is read as a successful handoff, including an unrelated manual tab switch / Cmd-Tab. It's documented as an accepted tradeoff, but 8s is generous — if the app isn't installed, backgrounding the tab for any reason flips the UI to "Opened in Xyne Spaces", which is misleading. Consider shortening, or pairing with a `blur`/`pagehide` heuristic.

apps/dashboard/src/utils/openInDesktopApp.ts
```diff
@@ -0,0 +1,379 @@
...
+function isDeepLinkableRoute(pathname: string): boolean {
+  const segments = pathname.split('/').filter(Boolean);
+  const [firstSegment] = segments;
+  if (!firstSegment) return false; // root '/'
+  if (EXCLUDED_FIRST_SEGMENTS.has(firstSegment)) return false;
+  return segments.length >= 2; // needs a sub-route beneath the workspace
+}
+
+/** Custom scheme for the current environment (mirrors electron `config.ts`). */
+function getDeepLinkScheme(hostname: string): string {
```
sithaarth, line 73: Scheme heuristic matches the three known hostnames in electron `config.ts`. It silently falls back to `xyne-spaces` (prod) for any other host — PR-preview / ephemeral deploy domains would build a prod deep link that the installed prod app then tries to route. Fine if no such domains serve this bundle; worth a quick confirm.

apps/dashboard/src/utils/openInDesktopApp.ts
```diff
@@ -0,0 +1,379 @@
...
+        attemptViaIframe(deepLink);
+      } else {
+        // A real user gesture makes the OS "Open Xyne Spaces?" prompt reliable.
+        window.location.href = deepLink;
+      }
+    };
+
+    let controls: OverlayControls | null = null;
+
+    const onVisibility = (): void => {
```
sithaarth, line 387: The modal overlay has no keyboard affordance — no Escape-to-dismiss and no focus trap. Since it sits at `zIndex 2147483647` over the app, keyboard users can't close it. Add an Escape handler wired to `onContinue` and move focus to the primary button on mount.

apps/electron/src/services/deep-links.ts
```diff
@@ -51,8 +51,11 @@ function isSafeDeepLinkPath(pathStr: string): boolean {
     if (/[\u0000-\u001F\u007F]/.test(c)) return false;     // control chars
     if (/[a-zA-Z][a-zA-Z0-9+.\-]*:/.test(c)) return false;     // embedded scheme (http:, javascript:, data:)
   }
-  // Conservative in-app route charset (path + query only).
-  if (!/^\/[A-Za-z0-9\-._~/?=&%]*$/.test(pathStr)) return false;
+  // Conservative in-app route charset (path + query + hash fragment).
+  // '#' is allowed so shared thread links keep their anchor
```
sithaarth, line 60: Allowing `#` is fine, but note the embedded-scheme guard `/[a-zA-Z][a-zA-Z0-9+.\-]*:/` runs over the whole string including the fragment, so any legit fragment containing a colon (e.g. a timestamp `createdAt=2026-09-09T10:00`) is rejected. Your example fragment has none, so no break today — just flagging that colons anywhere in the hash will drop the link.

sithaarth (commented):
Review — clean, well-commented feature. Scheme detection matches electron config.ts (localhost=dev, sandbox, else prod ✓). Left inline notes: main ones are the always-mounted App under the overlay, per-session re-prompting for browser-only users, and the auto-iframe firing an OS prompt without a user gesture. None blocking.

sithaarth (dismissed):

apps/dashboard/src/main.tsx
OmThakur2511: done

apps/dashboard/src/utils/openInDesktopApp.ts
OmThakur2511: slack is not doing this, but i have added a support to continue in browser now, pls check now.

apps/dashboard/src/utils/openInDesktopApp.ts
OmThakur2511: added this

sithaarth (approved):

samitbarai (approved):

PR #1708, opened 2026-09-09
feat: per-subagent MCP credentials (claw-auth foundation)

XyneSpaces:
## Summary
Adds a **per-subagent MCP credential** layer so a subagent can pin its own MCP credentials, resolved with high (optionally non-overridable) priority above the existing agent → user → global cascade.

**Motivation (Nammayatri):** one agent needs distinct MIDs/credentials per subagent — e.g. different creds for payouts vs. payment links. Today a subagent inherits the parent agent's pinned instance via `mcpInstanceMap`, so it cannot hold its own identity.

## What this PR does (reviewable core)
- **Schema:** new `SubagentMcpConnection` model + back-refs on `McpServer` and `SubagentDefinition`; two new `AgentAuditEvent` values (`SUBAGENT_MCP_UPDATED`, `SUBAGENT_MCP_DELETED`).
- **Migrations:** additive `CREATE TABLE subagent_mcp_connections` (unique on `(subagentDefinitionId, mcpServerId, slug)`, FKs `ON DELETE CASCADE`), plus a separate `ALTER TYPE ... ADD VALUE` migration so the enum add runs in its own transaction (older-Postgres safe).
- **Resolver (`credentials-loader.ts`):** new subagent branch resolves **above** the agent/user/global cascade. When `nonOverridable=true` (default) it short-circuits so no lower-priority credential can shadow it; a soft pin (`false`) falls through on a miss or undecryptable row. A non-overridable-but-undecryptable row returns `null` rather than silently falling back to a weaker identity.
- **claw-auth routes:** `GET/POST/DELETE /subagents/:name/mcp/connections`, gated by `canEditSubagent`, with `SUBAGENT_MCP_UPDATED`/`SUBAGENT_MCP_DELETED` audit logs. Built-in (code-defined) subagents are rejected — they have no DB row to pin against.
- **Call path:** `/mcp/call` and `/actions/sign` forward an optional `subagentId` into the resolver.
- **xyne-claw:** `loadMcpToolsForUser` accepts an optional `subagentId` and forwards it in the `/mcp/call` body (symmetric plumbing hook; undefined = unchanged main-agent behaviour).

## Security notes
- Credential CRUD requires **edit rights** on the subagent (`canEditSubagent`: owner, claw admin, or EDITOR share).
- Creds are encrypted at rest with `CONFIG.encryptionKey` (same `encrypt`/`decrypt` as `AgentMcpConnection`).
- Non-overridable pins are the core guarantee for the per-MID requirement: a lower-priority (agent/user/global) credential can never shadow a subagent's pinned identity.

## Follow-up (flagged, NOT in this PR)
Wiring `buildSubagentTools` to construct **per-subagent MCP tool groups** bound to `subagentId` is a high-blast-radius runtime path (it rebuilds the subagent tool palette's `execute()` closures) and can't be fully integration-tested here. The plumbing hook above makes that a localized follow-up. Until then the `subagentId` field is threaded end-to-end on the claw-auth side but only set once the runtime constructs subagent-scoped tools.

## Validation
- `prisma generate` + typecheck must run in CI: the authoring sandbox has **no network** to fetch the Prisma engine, so the client could not be regenerated locally. With the current (stale) client, the only new type error was a scope bug that is now fixed; all remaining `tsc` errors are the pre-existing ungenerated-client cascade also present on the base branch.
- Suggested checks: apply both migrations on a scratch DB; unit-test the resolver's subagent branch (hit, non-overridable short-circuit, soft-pin fallthrough, undecryptable non-overridable → null); route tests for ACL + audit-log emission.

### DIFF
--- a/apps/xyne-claw-auth/backend/prisma/schema.prisma
+++ b/apps/xyne-claw-auth/backend/prisma/schema.prisma
@@ -1505,6 +1505,11 @@ enum AgentAuditEvent {
   SUBAGENT_CREATED
   SUBAGENT_UPDATED
   MCP_SERVER_CREATED
+  // Subagent-level MCP credential writes. Added by migration
+  // 20260909120000_subagent_mcp_connections (ALTER TYPE ADD VALUE) —
+  // kept here so a fresh `prisma generate` accepts them.
+  SUBAGENT_MCP_UPDATED
+  SUBAGENT_MCP_DELETED
 }
 
 model AgentAuditLog {
@@ -1691,6 +1696,7 @@ model McpServer {
   agentConnections     AgentMcpConnection[]
   globalCredentials    GlobalMcpCredentials[]
   editRequests         McpConnectorEditRequest[]
+  subagentConnections  SubagentMcpConnection[]
 
   @@unique([type])
   @@index([enabled])
@@ -1922,6 +1928,7 @@ model SubagentDefinition {
   org              Organization    @relation(fields: [orgId], references: [id], onDelete: Restrict)
   skills           SubagentSkill[]
   shares           SubagentShare[]
+  mcpConnections   SubagentMcpConnection[]
 
   @@unique([orgId, name])
   @@index([orgId])
@@ -1955,6 +1962,44 @@ model SubagentSkill {
 
   @@unique([subagentDefinitionId, skillId])
   @@map("subagent_skills")
+}
+
+// Per-subagent MCP credential. Companion to AgentMcpConnection, but owned by a
+// SubagentDefinition instead of an Agent. When present it resolves ABOVE the
+// agent/user/global cascade in credentials-loader. When `nonOverridable` is
+// true (the default) it SHORT-CIRCUITS that cascade — no user or agent
+// credential can shadow it — which is the whole point of the feature: a
+// subagent that must always authenticate as a specific identity (e.g. one
+// Nammayatri MID per subagent). `slug` mirrors AgentMcpConnection so one
+// subagent can hold several instances of the same server type.
+model SubagentMcpConnection {
+  id                   String             @id @default(cuid())
+  subagentDefinitionId String
+  subagent             SubagentDefinition @relation(fields: [subagentDefinitionId], references: [id], onDelete: Cascade)
+  mcpServerId          String
+  mcpServer            McpServer          @relation(fields: [mcpServerId], references: [id], onDelete: Cascade)
+  /// Stable short identifier for this instance, scoped to (subagent, server).
+  /// Matched against SubagentDefinition.mcpInstanceMap at runtime. Defaults to
+  /// 'default'. Allowed chars: lowercase alnum + hyphen.
+  slug                 String             @default("default")
+  /// Optional human-readable label shown in the subagent editor.
+  displayName          String?
+  /// When true, this credential short-circuits the agent/user/global cascade
+  /// so it cannot be overridden by any lower-priority connection. When false
+  /// it is a soft pin: preferred, but normal fallback still applies if the
+  /// decrypt/lookup yields nothing.
+  nonOverridable       Boolean            @default(true)
+  encryptedCreds       String
+  iv                   String
+  authTag              String
+  /// User who pasted the creds — used in audit logs.
+  createdByUserId      String?
+  createdAt            DateTime           @default(now())
+  updatedAt            DateTime           @updatedAt
+
+  @@unique([subagentDefinitionId, mcpServerId, slug])
+  @@index([subagentDefinitionId])
+  @@map("subagent_mcp_connections")
 }
 
 model PendingMemoryReview {
--- a/apps/xyne-claw-auth/backend/prisma/schema.prisma
+++ b/apps/xyne-claw-auth/backend/prisma/schema.prisma
@@ -1506,7 +1506,7 @@ enum AgentAuditEvent {
   SUBAGENT_UPDATED
   MCP_SERVER_CREATED
   // Subagent-level MCP credential writes. Added by migration
-  // 20260909120000_subagent_mcp_connections (ALTER TYPE ADD VALUE) —
+  // 20260909120500_subagent_mcp_audit_events (ALTER TYPE ADD VALUE) —
   // kept here so a fresh `prisma generate` accepts them.
   SUBAGENT_MCP_UPDATED
   SUBAGENT_MCP_DELETED
--- a/apps/xyne-claw-auth/backend/src/lib/credentials-loader.ts
+++ b/apps/xyne-claw-auth/backend/src/lib/credentials-loader.ts
@@ -45,7 +45,7 @@ import { createLogger } from "../logger.js";
 const log = createLogger("credentials-loader");
 
 export interface EffectiveCredentials {
-  source: "agent" | "user" | "global";
+  source: "subagent" | "agent" | "user" | "global";
   connectionId: string;
   credentials: Record<string, unknown>;
   /** True iff the user owns this connection (admin/health code uses this to
@@ -142,6 +142,10 @@ export async function loadEffectiveCredentials(
   agentSlug?: string,
   instanceSlug?: string,
   agentOrgId?: string,
+  // SubagentDefinition.id when the tool call originates from a subagent's
+  // nested run. When set, a SubagentMcpConnection is consulted ABOVE the
+  // agent/user/global cascade below; a `nonOverridable` row short-circuits it.
+  subagentId?: string,
 ): Promise<EffectiveCredentials | null> {
   // google / microsoft: per-user OAuth connectors (never agent-pinned or
   // global), executed as claw-auth-hosted stdio MCP servers. Resolve a fresh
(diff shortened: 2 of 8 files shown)


apps/xyne-claw-auth/backend/src/routes/mcp.ts
```diff
@@ -1347,12 +1352,15 @@ router.post("/:sessionId/mcp/call", async (req: Request<{ sessionId: string }>,
     const strictAgentToolsConfig = isStrictAgentToolsEnabled()
       ? await withSurfaceDefaultToolsConfig(sessionAgentTools?.toolsConfig, req.params.sessionId, spacesAppId)
       : undefined;
-    const { serverType, tool, params, permission, backendId } = req.body as {
+    const { serverType, tool, params, permission, backendId, subagentId } = req.body as {
```
XyneSpaces, line 1355: **[blocking] R1 — body-supplied `subagentId` selects credentials with no validation.** `subagentId` is read straight from `req.body` here (and again in `/actions/sign` at `mcp.ts:2133` via `body.subagentId ?? body.pendingAction?.subagentId`) and forwarded into the credential resolver. Every other identity input on these routes — `userId`, `agentSlug`, `sessionAgentOrgId` — is derived from the authenticated session; this one is caller-supplied. Since a `nonOverridable` row short-circuits the entire cascade (credentials-loader), any S2S-authenticated session that reaches these endpoints can force credential resolution to a subagent pin it doesn't own — another org's or another agent's pinned identity. No claw-runtime caller passes `subagentId` yet (the `buildSubagentTools` wiring is explicitly follow-up per the PR description), so the exposure is latent — but the endpoint contract lands in this PR and the wiring will build directly on it. Suggested fix: validate before use — resolve the subagent scoped to the session's org and to the session agent's enabled subagents (`toolsConfig.subagents` / `subagentToolRefs`), and reject or ignore on mismatch. See the companion comment on the resolver query for the scoping fix.

apps/xyne-claw-auth/backend/src/lib/credentials-loader.ts
```diff
@@ -243,6 +247,57 @@ export async function loadEffectiveCredentials(
...
+  //    is preferred but falls through to the normal cascade on a miss. This
+  //    sits AFTER the OAuth / live-session short-circuits above by design:
+  //    those server types (google, xyne-dashboard, xyne-spaces-app-tools) have
+  //    no per-row secret to pin, so a subagent pin only applies to
+  //    secret-backed servers.
+  if (subagentId) {
+    let subConn = null;
+    if (instanceSlug) {
+      subConn = await prisma.subagentMcpConnection.findFirst({
+        where: { subagentDefinitionId: subagentId, mcpServer: { type: serverType }, slug: instanceSlug },
```
XyneSpaces: **[blocking] R1 — the subagent lookup is not org-scoped.** All three `findFirst` variants in this block key only on the caller-supplied `subagentDefinitionId` (+ server type/slug) with no `orgId` constraint — unlike the write side, which resolves via `findByName(name, getOrgId(req))` + `canEditSubagent` (routes/subagents.ts). `agentOrgId` is already a parameter of this function, so the fix is cheap: add the relation filter, e.g. `where: { subagentDefinitionId: subagentId, subagent: { orgId: agentOrgId }, mcpServer: { type: serverType }, ... }` across all three variants, and skip the pin when `agentOrgId` is undefined (direct admin/health calls). That way a body-supplied `subagentId` from another org can never force a credential even if the route-level check is missed.

apps/xyne-claw-auth/backend/src/routes/subagents.ts
```diff
@@ -414,4 +422,169 @@ router.delete("/:name/shares/:userId", asyncHandler(async (req: Request, res: Re
...
+
+    await prisma.subagentMcpConnection.delete({
+      where: {
+        subagentDefinitionId_mcpServerId_slug: {
+          subagentDefinitionId: subagent.id,
+          mcpServerId: server.id,
+          slug: instanceSlug,
+        },
+      },
+    }).catch(() => undefined);
```
XyneSpaces: **[major] R2/R6 — DELETE swallows every error, not just "not found".** `.catch(() => undefined)` turns any failure (DB outage, connection drop, constraint error) into `{ deleted: true }` plus a `SUBAGENT_MCP_DELETED` audit row — the response and the audit trail then claim a removal that never happened. For a `nonOverridable` pin this is the worst case: the credential stays active and keeps forcing identity at resolution time while the admin believes it's gone. Suggested fix: catch `Prisma.P2025` specifically and return 404 / `deleted: false`; let everything else propagate through `asyncHandler` so it surfaces as a 5xx with a log line.

apps/xyne-claw-auth/backend/src/lib/credentials-loader.ts
```diff
@@ -243,6 +247,57 @@ export async function loadEffectiveCredentials(
...
+  //    originates from a subagent's nested run). When the matched row is
+  //    `nonOverridable` (the default) it SHORT-CIRCUITS the cascade: no agent,
+  //    user, or global credential can shadow it — the subagent always
+  //    authenticates as this pinned identity. A soft pin (nonOverridable=false)
+  //    is preferred but falls through to the normal cascade on a miss. This
+  //    sits AFTER the OAuth / live-session short-circuits above by design:
+  //    those server types (google, xyne-dashboard, xyne-spaces-app-tools) have
+  //    no per-row secret to pin, so a subagent pin only applies to
+  //    secret-backed servers.
+  if (subagentId) {
```
XyneSpaces, line 261: **[major] R7 — the new resolution semantics land untested.** The precedence change in this block is the core contract the follow-up runtime wiring builds on: a subagent pin resolves above agent/user/global, `nonOverridable` short-circuits the cascade, and a decrypt failure on a non-overridable pin returns `null` instead of falling through to a weaker identity. No tests accompany it (the PR description defers "route tests for ACL + audit-log emission" to a follow-up). Suggested: land at least resolver-level tests with this PR — subagent pin beats agent+user creds; soft pin (`nonOverridable=false`) falls through on a miss; non-overridable decrypt failure refuses to fall through; slug fallback (`instanceSlug` → `default` → oldest row). Regressions here silently change which identity a tool call authenticates as.

apps/xyne-claw-auth/backend/prisma/schema.prisma
```diff
@@ -1505,6 +1505,11 @@ enum AgentAuditEvent {
   SUBAGENT_CREATED
   SUBAGENT_UPDATED
   MCP_SERVER_CREATED
+  // Subagent-level MCP credential writes. Added by migration
+  // 20260909120000_subagent_mcp_connections (ALTER TYPE ADD VALUE) —
```
XyneSpaces: **[nit] R8 — comment cites the wrong migration.** The enum values are added by `20260909120500_subagent_mcp_audit_events` (its own header explains the separate-transaction pattern), but this comment points at `20260909120000_subagent_mcp_connections`, which only creates the table. Anyone tracing the enum later will follow the wrong pointer.

apps/xyne-claw-auth/backend/src/lib/credentials-loader.ts
```diff
@@ -142,6 +142,10 @@ export async function loadEffectiveCredentials(
   agentSlug?: string,
   instanceSlug?: string,
   agentOrgId?: string,
+  // SubagentDefinition.id when the tool call originates from a subagent's
+  // nested run. When set, a SubagentMcpConnection is consulted ABOVE the
+  // agent/user/global cascade below; a `nonOverridable` row short-circuits it.
+  subagentId?: string,
```
XyneSpaces, line 148: **[nit] R8 — sixth positional optional parameter.** `loadEffectiveCredentials` is now `(userId, serverType, agentSlug, instanceSlug, agentOrgId, subagentId)` and call sites already thread `undefined` placeholders (routes/mcp.ts:823). An options object for the trailing optionals (`{ agentSlug, instanceSlug, agentOrgId, subagentId }`) would keep future additions transposition-safe — fine to defer to a follow-up if you prefer.

XyneSpaces (commented):
Automated Xyne Spaces review — 6 findings (2 blocking, 2 major, 2 nits) posted as inline comments below. A summary comment with the reviewer recommendation follows on the PR.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1708

**Reviewed:** 7 files, 21 hunks @ `90a7c20` · **Findings:** 2 blocking, 2 major, 2 nits (6 inline comments)

### Findings
| blocking | R1 | apps/xyne-claw-auth/backend/src/routes/mcp.ts:1355 | `subagentId` read from `req.body` in `/mcp/call` (and `/actions/sign` at :2133) and used to select credentials with no org/agent validation — a `nonOverridable` pin can be force-selected cross-org; latent until runtime wiring, but the contract lands here |
| blocking | R1 | apps/xyne-claw-auth/backend/src/lib/credentials-loader.ts:265 | subagent `findFirst` not scoped to `agentOrgId` (write side is org-scoped via `getOrgId(req)` + `canEditSubagent`) |
| major | R2/R6 | apps/xyne-claw-auth/backend/src/routes/subagents.ts:577 | DELETE `.catch(() => undefined)` swallows all errors, still returns `deleted: true` + writes the audit row |
| major | R7 | apps/xyne-claw-auth/backend/src/lib/credentials-loader.ts:261 | new precedence + `nonOverridable` decrypt-failure semantics land with no tests (listed only as "suggested checks" in the description) |
| nit | R8 | apps/xyne-claw-auth/backend/prisma/schema.prisma:1509 | enum comment cites migration `20260909120000`, but the `ALTER TYPE` lives in `20260909120500` |
| nit | R8 | apps/xyne-claw-auth/backend/src/lib/credentials-loader.ts:148 | 6th positional optional param; options object would be transposition-safe |

**Top blocking issue:** the two R1 findings share one fix — scope the subagent-pin lookup (and the route-level check) to the session's org/agent so a body-supplied `subagentId` can never select another org's credential. Since no runtime caller passes `subagentId` yet (per the follow-up section), this can be fixed cheaply before the wiring builds on it.

### Recommended reviewer
**@anuragdvd** (Anurag Dwivedi) — owns the claw-auth credential-resolution area this PR modifies and has the lightest review load right now.

| @anuragdvd | claw-auth backend (resolver + routes) — per-file commit history wasn't computable with the available tooling; Spaces evidence: ran `/security-scan` + `/framework` scoped to `apps/xyne-claw-auth`, debugs claw-auth logs, drives the OAuth-removal/key-rotation threads | no — `.github/CODEOWNERS` only covers `apps/backend` ACL paths; none of the 7 changed files match | 0 — no open PR in this repo requests his review (only #1596 has a request, → Pradeesh333) | most active in claw engineering channels (444 vs 161 activity signals over the last 2 weeks) |
| satvikbatra | — (no claw-auth ownership signal found) | no | 0 formal requests, but mid-way through a large PR review (per Spaces, 9 Sep) | — |

Excluded: the PR author (XyneSpaces bot, branch pushed by karunrajk) and bot accounts.

⚠️ **Reviewer request:** this automation's GitHub toolset has no `request_reviewers` endpoint (and no shell for `gh pr edit --add-reviewer`), so the formal review request could not be filed via the API — @anuragdvd is tagged here instead, which notifies him on GitHub. Please add him via the Reviewers sidebar for the formal request.

PR #1740, opened 2026-09-10
fix: XYNE-63210 moving labels from channel level to user level

rahul1841:
image
image
image
image


Himanshvarma (approved):

satvikbatra (approved):

karunrajk (approved):

Himanshvarma (commented):
I have added this channel level label uniqueness checks in lot of places in this desk rules flow in backend we might also have to remove that check since now we are not having any channel level unique constraint as label are shown at user level from now. or are we still going with channel level label uniqueness constraint?

Himanshvarma:
image
no need to show this "you" also i have added that code to distinguish between labels that user created and all channel labels. if we are going to have channel level uniqueness for labels then we might have to show all labels here of that channel then only user will know what all label names already exist and what all i cannot create again

KushagarGoel1 (approved):

PR #1506, opened 2026-09-03
feat: XYNE always spill MCP results to file + forward via sandbox-copy-in contextPath

XyneSpaces:
## What
Makes the MCP tool-result → sandbox data path deterministic so the agent never has to re-type large blobs (root cause of the recurring bracket-count mismatches in the PI alerts bot).

Two small, connected changes:

1. **Always persist MCP results to a file + hand the agent the path.** `promoteIfOversized` gains a `forceFile` flag (passed `true` from the MCP execute path in `mcp.ts`). Even when a result fits inline, the raw bytes are written to `.context/tool-results/<server>-<tool>-<ts>.json` and the inline text gets a marker line pointing at the file. Over-cap results keep their existing preview+spill behaviour.
2. **Let the agent forward that file into a sandbox byte-for-byte.** `sandbox-copy-in` now accepts a `contextPath` (a file under this run's `.context` root) in addition to `skillPath`. Exactly one source is required; each is confined to its own root with the same `..`/absolute-path/credential rejection. The `.context` root is injected as `context.meta.contextRoot` in `routes/run.ts` (= `join(mcpOutputDir, ".context")`) and flows through the existing `toolMeta` spread in `custom-tools.ts`.

## Why
The agent was manually re-emitting large MCP data through `sandbox-write-file`, dropping/mismatching brackets. Now it forwards the whole result file server-side — no re-typing, byte-identical.

## Files
- `apps/xyne-claw/src/tool-output.ts` — `forceFile` param + always-write branch
- `apps/xyne-claw/src/mcp.ts` — pass `forceFile: true`
- `apps/xyne-claw/src/routes/run.ts` — inject `meta.contextRoot`
- `packages/xyne-claw-shared/src/tools/sandbox/tools.ts` — `sandbox-copy-in` accepts `contextPath`

## Testing
- `tsc --noEmit` clean on `packages/xyne-claw-shared`.
- `apps/xyne-claw` typecheck surfaces only pre-existing, unrelated errors (missing `zod`/`bullmq` modules, implicit-any in `main.ts`/`run-queue-worker.ts`/`pr-review-findings.ts`); none touch the four changed files.

PR #1761, opened 2026-09-10
fix: XYNE-sandbox stream sandbox-copy-in to avoid 413 on large tool-results

XyneSpaces:
## Problem
`sandbox-copy-in` (contextPath) read the whole source file and sent it as a single base64 JSON POST to the sandbox `/write` endpoint. For a large spilled MCP result (~486 KB raw → ~650 KB base64 body) the sandbox workspace server rejects it with `Error: request entity too large`, so forwarding a big tool-result into the sandbox for processing fails — exactly the PI Alerts exhaustive-table case.

## Fix
Switch the copy path from the one-shot `FilesystemModule.write()` to the existing `FilesystemModule.writeStream()`, which splits the source into bounded **256 KiB** chunks over the same `/write` endpoint and appends them server-side (`cat >> dest && rm part`). No single request is large, so the 413 goes away — and it needs no workspace-image change and no sandbox egress. This is the same pattern already used by the video-explainer composer and record-skill to fix the identical 413.

## Scope
- `packages/xyne-claw-shared/src/tools/sandbox/tools.ts`: `sandboxCopyIn.execute` now streams via `writeStream(createReadStream(sourceAbs), { chunkBytes: 256 * 1024 })`; removed the now-unused `readFile` import; added `createReadStream`.

## Validation
- `tsc --noEmit` passes for `xyne-claw-shared`.
- `vitest run` green for `copy-in.test.ts` (17) and `record-skill/streaming-write.test.ts` (2). The chunking mechanism itself is covered by `streaming-write.test`.

## Notes / follow-ups
- 256 KiB gives a clear margin under the observed ~650 KB failing body.
- Independent of this, server-side filtering on the Alerts MCP (filter by `metadata.classification` + column projection/pagination) would keep the payload small enough to never spill — the durable fix for that specific request.

### DIFF
--- a/packages/xyne-claw-shared/src/tools/sandbox/tools.ts
+++ b/packages/xyne-claw-shared/src/tools/sandbox/tools.ts
@@ -7,7 +7,7 @@ import { redactSecrets, redactAndStringify } from "./redact.js";
 import { rotateTemplate, isSameTemplateFamily } from "./template-rotation.js";
 import { formatSandboxUnavailable, isSandboxUnavailableDeferEnabled } from "./unavailable-signal.js";
 import { createLogger } from "../../logger.js";
-import { readFile } from "node:fs/promises";
+import { createReadStream } from "node:fs";
 import { resolve, join, sep } from "node:path";
 import {
   cleanupSdlcGitCredentialMaterial,
@@ -1235,9 +1235,20 @@ export const sandboxCopyIn: ToolDefinition = {
     if (roWrite) return roWrite;
 
     try {
-      const buf = await readFile(sourceAbs);
-      await session.files.write(destPath, buf);
-      return JSON.stringify({ sourcePath: relPath, destPath, bytes: buf.length, copied: true });
+      // Stream the file in bounded chunks instead of one base64 JSON POST.
+      // A whole-file write() base64-inflates the payload ~1.33x into a single
+      // /write body, which the sandbox workspace server rejects with
+      // "request entity too large" for large spilled MCP results (observed at
+      // ~486 KB raw -> ~650 KB body). writeStream reuses the same /write
+      // endpoint per chunk and appends server-side, so no single request is
+      // large and no workspace-image change is needed. 256 KiB keeps a clear
+      // margin under the observed cap.
+      const { bytesWritten } = await session.files.writeStream(
+        destPath,
+        createReadStream(sourceAbs),
+        { chunkBytes: 256 * 1024 },
+      );
+      return JSON.stringify({ sourcePath: relPath, destPath, bytes: bytesWritten, copied: true });
     } catch (err) {
       if (isStaleSessionError(err)) {
         evictSession(session);


packages/xyne-claw-shared/src/tools/sandbox/tools.ts
```diff
@@ -1235,9 +1235,20 @@ export const sandboxCopyIn: ToolDefinition = {
...
-      return JSON.stringify({ sourcePath: relPath, destPath, bytes: buf.length, copied: true });
+      // Stream the file in bounded chunks instead of one base64 JSON POST.
+      // A whole-file write() base64-inflates the payload ~1.33x into a single
+      // /write body, which the sandbox workspace server rejects with
+      // "request entity too large" for large spilled MCP results (observed at
+      // ~486 KB raw -> ~650 KB body). writeStream reuses the same /write
+      // endpoint per chunk and appends server-side, so no single request is
+      // large and no workspace-image change is needed. 256 KiB keeps a clear
+      // margin under the observed cap.
+      const { bytesWritten } = await session.files.writeStream(
```
XyneSpaces, line 1246: **[major] R7 — Tests**: The swapped happy path has no test. All 17 tests in `copy-in.test.ts` exercise validation branches (XOR skill/context path, path confinement, credential refusal, session lookup) — none reach the copy call, so nothing pins that `sandboxCopyIn` now calls `writeStream(destPath, createReadStream(sourceAbs), { chunkBytes: 256 * 1024 })`, echoes `bytesWritten` as `bytes`, or that the ENOENT error branch still maps correctly when a *stream* rejects rather than `readFile`. The PR notes `streaming-write.test` covers the chunking mechanism — that tests the SDK, not this tool's wiring of it. Suggested fix: one test mocking `session.files.writeStream` asserting the call args and the `{ bytes: bytesWritten, copied: true }` result shape, and one where it rejects (ENOENT-style) asserting the existing error branch. Repo precedent (#1506 round 2) treated missing happy-path/contract tests as major.

packages/xyne-claw-shared/src/tools/sandbox/tools.ts
```diff
@@ -1235,9 +1235,20 @@ export const sandboxCopyIn: ToolDefinition = {
...
+        destPath,
```
XyneSpaces, line 1247: **[nit] R1 — hardening (defense-in-depth, not an injection finding)**: `destPath` is agent-supplied, trimmed/defaulted but never path-validated (no confinement under `/workspace`, no allowlist), and it now flows into `writeStream`'s server-side shell commands (`: > dest`, `cat part >> dest`, `rm -f`). I verified the SDK's POSIX single-quoting is sound, so injection is contained — but per the guidelines' validation rules, a schema allowlist (e.g. `^/[\w./-]+$`) or explicit `/workspace` confinement would make the shell boundary structural instead of incidental to `quote()` staying correct.

packages/xyne-claw-shared/src/tools/sandbox/tools.ts
```diff
@@ -1235,9 +1235,20 @@ export const sandboxCopyIn: ToolDefinition = {
...
+        createReadStream(sourceAbs),
```
XyneSpaces, line 1248: **[nit] R2 — behavior change on failure (labeled assumption)**: `writeStream`'s contract is truncate-first (`: > dest`) and remove-dest-on-error (`rm -f part dest`), so a mid-copy failure now **destroys a pre-existing `destPath` file**. Assumption: the old `session.files.write()` left an existing dest untouched when it failed — if `write()` also truncated up-front this is moot; but if any caller relied on copy-in failing non-destructively over an existing file, that guarantee is gone with this swap. Worth a one-line confirmation either way.

packages/xyne-claw-shared/src/tools/sandbox/tools.ts
```diff
@@ -1235,9 +1235,20 @@ export const sandboxCopyIn: ToolDefinition = {
...
+        { chunkBytes: 256 * 1024 },
```
XyneSpaces, line 1249: **[nit] R5 — duplicated constant**: `256 * 1024` is now the third copy of this chunk size (sandbox-copy-in here, video-explainer composer, record-skill streaming-write — the PR description itself notes the pattern is shared). Consider exporting one constant (e.g. `SANDBOX_STREAM_CHUNK_BYTES`) from `@xyne/kata-sdk` or `xyne-claw-shared` so the "clear margin under the observed ~650 KB cap" rationale lives in exactly one place and can't drift across the three call sites.

XyneSpaces (commented):
Automated review of every hunk (1 file, 2 hunks) against the Xyne Spaces review guidelines. Findings are inline below; a summary comment with the reviewer recommendation follows.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1761

**Verdict:** Changes requested — 1 major (test gap on the swapped code path); the streaming approach itself is sound and verified.
**Reviewed:** 1 file, 2 hunks · **Findings:** 0 blocking, 1 major, 4 nits

### Findings
| major | R7 | `packages/xyne-claw-shared/src/tools/sandbox/tools.ts:1246` | Swapped happy path (writeStream call args, `bytes` echo, ENOENT mapping) has no test — all 17 `copy-in.test.ts` tests cover validation branches only |
| nit | R1 | `tools.ts:1247` | `destPath` is unvalidated yet now flows into `writeStream`'s server-side shell commands — SDK quoting verified sound, so this is defense-in-depth (schema allowlist / `/workspace` confinement) |
| nit | R2 | `tools.ts:1248` | Failed copy now deletes a pre-existing dest file (truncate-first + rm-on-error contract); old `write()` failure semantics labeled as an assumption inline |
| nit | R5 | `tools.ts:1249` | `256 * 1024` chunk size duplicated for the third time (copy-in, video-explainer composer, record-skill) — hoist a shared constant |
| nit | R8 | PR title | `XYNE-sandbox` is not a `<TICKET-ID>`; commit convention expects a real ticket id |

**Verified sound, no action needed:** `writeStream` truncates the dest before appending, so retries/re-runs cannot double-append; 256 KiB chunks keep each `/write` POST (~349 KB base64) well under the observed ~650 KB 413 cap; the identical pattern is already proven in prod by the video-explainer composer and record-skill; the SDK's POSIX shell quoting is correct.

### Recommended reviewer
**@KushagarGoel1** — recommended. Note: `.github/CODEOWNERS` does not cover `packages/xyne-claw-shared/**` (it only owns ACL paths), so there is no default reviewer for this change; the ranking below is built from pattern/file ownership and Spaces activity instead.

| @KushagarGoel1 | video-explainer composer — the same `writeStream` chunked-write pattern this PR reuses to fix the identical 413 | no (no CODEOWNERS entry) | not computable¹ | active on claw/sandbox in Spaces this week; requested security/coding reviews of his own claw PRs (#1442/#1462) |
| @anuragdvd | approved #1506 — the most recent change to this exact tool (`contextPath`), merged today | no | not computable¹ | strongest file familiarity, but is driving this branch (`feature/deploy-xyneclaw`) and carries the highest review-coordination load |
| @harsh-1923 | agent sandbox feature area (#530) | no | not computable¹ | flagged OOO in Spaces on 28 Aug |
| @sumantop | — | no | 0 | inactive on claw/sandbox topics in the last 2 weeks |

¹ The `review-requested:<login>` GitHub search qualifier returned inconsistent results in this environment (it matched #1761 itself, which has zero requested reviewers), so per-person pending-review counts can't be trusted; capacity was assessed from Spaces activity instead of invented numbers.

**Rationale:** @KushagarGoel1 built and shipped the exact streaming-write-over-`/write` pattern this PR applies (video-explainer, plus the sandbox hackathon setup this month), is active on claw topics this week, and is not the PR author. @anuragdvd knows this precise file best (approved its direct predecessor #1506 today) and is listed as the strongest file-familiarity alternate — but he's the one driving this branch and is already coordinating the heaviest review load. Harsh is possibly OOO; Sumant and Utkarsh aren't active in this area.

⚠️ This bot has no reviewer-request API surface, so the @mention above is the assignment signal — @anuragdvd, please add **@KushagarGoel1** as a reviewer on this PR via the GitHub UI.

anuragdvd (approved):

PR #1762, opened 2026-09-10
fix: XYNE-63232 emit MESSAGE_RECEIVED on message update

satvikbatra:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist

### DIFF
--- a/apps/backend/src/automations/engine/event-router.ts
+++ b/apps/backend/src/automations/engine/event-router.ts
@@ -6,9 +6,12 @@ import { currentUpstreamChain } from './automation-context-storage';
 import {
   AUTOMATION_WORKFLOW_TYPE,
   DESK_AUTOMATION_WORKFLOW_TYPE,
+  parseAutomationConfig,
   parseAutomationMetadata,
   triggerTypeToEventType,
 } from '../types/workflow-adapter';
+import { triggerRegistry } from '../triggers/trigger-registry';
+import type { TriggerType } from '../types/trigger-types';
 import { AutomationStatus, AutomationRunStatus } from '../types/status';
 import { automationQueue } from '../queue/automation.queue';
 import type { AutomationEvent } from '../types/automation-events';
@@ -33,13 +36,24 @@ class EventRouter {
     for (const workflow of candidates) {
       try {
         const metadata = parseAutomationMetadata(workflow.metadata);
+
+        const triggerImpl = triggerRegistry.has(event.type as TriggerType)
+          ? triggerRegistry.get(event.type as TriggerType)
+          : null;
+        const triggerConfig = (parseAutomationConfig(workflow.context).trigger.config ?? {}) as Record<
+          string,
+          unknown
+        >;
+        const data =
+          triggerImpl?.projectPayload?.(triggerConfig, payload as unknown as Record<string, unknown>) ??
+          payload;
         const initialContext = {
           automation: {
             id: workflow.id,
             workspaceId: workflow.workspaceId,
             createdById: metadata.createdById,
           },
-          trigger: { type: eventType, ...payload, data: payload },
+          trigger: { type: eventType, ...data, data },
           steps: {},
           __meta: { error: null, chain },
         };
--- a/apps/backend/src/automations/triggers/message-received.trigger.ts
+++ b/apps/backend/src/automations/triggers/message-received.trigger.ts
@@ -39,6 +39,12 @@ const MessageReceivedConfigSchema = z.object({
     .array(z.string())
     .optional()
     .describe('Only fire when at least one of these user groups is mentioned. Empty matches any group mention; omit entirely to not filter on group mentions.'),
+  fireOnEdit: z
+    .boolean()
+    .default(false)
+    .describe(
+      'Also fire when an existing message is edited into a match — e.g. an alert card whose status field is rewritten in place. Only the transition fires: an edit that leaves an already-matching message still matching is ignored. Needs a Content Contains value to be meaningful.',
+    ),
 });
 
 export const MessageReceivedOutputSchema = z.object({
@@ -62,6 +68,8 @@ export const MessageReceivedOutputSchema = z.object({
   conversationId: z.string(),
   msgType: z.nativeEnum(MessageType),
   deleted: z.boolean(),
+  isEdit: z.boolean().optional(),
+  previousContentMatched: z.boolean().optional(),
   mentionedUsers: z
     .array(
       z.object({
@@ -106,6 +114,23 @@ export class MessageReceivedTrigger extends BaseTrigger<typeof MessageReceivedCo
     return hydrateMessageReceivedPayload(payload as unknown as MessageReceivedEventPayload);
   }
 
+  /**
+   * Drop the pre-edit body and keep only whether it already satisfied this
+   * automation's content filter. The body itself is never persisted or logged.
+   */
+  override projectPayload(
+    config: Record<string, unknown>,
+    payload: Record<string, unknown>,
+  ): Record<string, unknown> {
+    const { previousContent, ...rest } = payload as { previousContent?: string };
+    if (previousContent === undefined) return rest;
+    const needle = (config as MessageReceivedConfig).contentContains;
+    return {
+      ...rest,
+      previousContentMatched: !!needle && previousContent.toLowerCase().includes(needle.toLowerCase()),
+    };
+  }
+
   override matchFilters(
     filter: Record<string, unknown>,
     payload: Record<string, unknown>,
@@ -115,6 +140,8 @@ export class MessageReceivedTrigger extends BaseTrigger<typeof MessageReceivedCo
 
     // The message was deleted before we got to run — nothing to act on.
     if (p.deleted) return false;
+    // Edits reach every candidate; only automations that opted in act on them.
+    if (p.isEdit && !cfg.fireOnEdit) return false;
     if (cfg.messageTypes && cfg.messageTypes.length > 0) {
       if (!cfg.messageTypes.includes(p.msgType)) return false;
     }
@@ -146,8 +173,11 @@ export class MessageReceivedTrigger extends BaseTrigger<typeof MessageReceivedCo
         // Decoded text and the stored blob: a filter written against a FlowJSON
         // card title or button label must keep firing after the decode.
         const needle = cfg.contentContains!.toLowerCase();
-        const contentPasses = [p.message.content, p.message.rawContent]
+        const nowMatches = [p.message.content, p.message.rawContent]
           .some(text => !!text && text.toLowerCase().includes(needle));
+        // On an edit only the transition counts — an edit that leaves an
+        // already-matching message still matching must not re-run the automation.
+        const contentPasses = p.isEdit ? nowMatches && !p.previousContentMatched : nowMatches;
         matchResults.push(contentPasses);
       }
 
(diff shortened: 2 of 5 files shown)


PiyushJuspay (dismissed):

apps/backend/src/automations/engine/event-router.ts
```diff
@@ -33,13 +36,28 @@ class EventRouter {
     for (const workflow of candidates) {
       try {
         const metadata = parseAutomationMetadata(workflow.metadata);
+
+        const triggerImpl = triggerRegistry.has(event.type as TriggerType)
```
XyneSpaces, line 40: **[major] R2 + R8 — registry lookup keyed on `event.type as TriggerType`.** Dispatch now depends on a string-level coincidence across three declarations: the event constant (`MESSAGE_RECEIVED_EVENT` in message-received.trigger.ts), the trigger's `readonly type`, and the registration key in `automations/index.ts`. `TriggerType` is just `string`, so the `as` cast proves nothing at compile time (review guidelines flag bare `as` casts), and `triggerTypeToEventType()` silently maps unknown strings to `NO_OP` — a drifted string would registry-miss here and quietly fall into the `?? payload` branch instead of failing loudly. I verified all three are `'MESSAGE_RECEIVED'` today, so this is robustness, not a live bug. Suggestion: resolve the impl from the workflow's own trigger type, or use `triggerRegistry.get()` + catch `UnknownTriggerError` so a misalignment is loud.

apps/backend/src/automations/engine/event-router.ts
```diff
@@ -33,13 +36,28 @@ class EventRouter {
...
+          ? triggerRegistry.get(event.type as TriggerType)
+          : null;
+        const triggerConfig = (parseAutomationConfig(workflow.context).trigger.config ?? {}) as Record<
+          string,
+          unknown
+        >;
+        // Edits only matter to automations that opted in — skip before enqueue.
+        if ((payload as { isEdit?: boolean }).isEdit && !triggerConfig.fireOnEdit) {
```
XyneSpaces: **[major] R5 — `fireOnEdit` gate duplicated from the trigger.** `MessageReceivedTrigger.matchFilters` already enforces `if (p.isEdit && !cfg.fireOnEdit) return false;` (message-received.trigger.ts:147). This router copy checks the same key against the raw JSON config (`parseAutomationConfig(...).trigger.config`) while the trigger reads it through its zod schema — two independent sources of truth for one behavior that drift apart on a rename or default change. It also bakes a message-trigger-specific concept (`isEdit` / `fireOnEdit`) into the generic per-workflow loop, and `parseAutomationConfig` adds a full-JSON parse per candidate on the busiest write path in the product. Suggestion: expose the skip decision on the trigger impl (e.g. `skipEvent(config, payload)`) and have the router delegate — the router stays generic and the gate lives where the schema lives.

apps/backend/src/automations/engine/event-router.ts
```diff
@@ -33,13 +36,28 @@ class EventRouter {
...
+          continue;
+        }
+        const data =
+          triggerImpl?.projectPayload?.(triggerConfig, payload as unknown as Record<string, unknown>) ??
```
XyneSpaces: **[nit] R1-adjacent + R8 — `?? payload` fallback can persist the pre-edit body.** When the registry lookup misses or a trigger impl doesn't define `projectPayload`, `data` falls back to the raw payload and is written into `workflow_execution_state.context` — for MESSAGE_RECEIVED that includes `previousContent`, the full pre-edit message body that `projectPayload` deliberately strips. Today the lookup is verified aligned so the happy path strips it, but a future drift turns this into silent content-at-rest. Suggestion: sanitize the fallback (or log when it triggers), and replace the double-cast `payload as unknown as Record<string, unknown>` with a typed helper — the guidelines flag double-casts precisely because they hide shape mismatches.

apps/backend/src/automations/triggers/message-received.trigger.ts
```diff
@@ -106,6 +114,26 @@ export class MessageReceivedTrigger extends BaseTrigger<typeof MessageReceivedCo
     return hydrateMessageReceivedPayload(payload as unknown as MessageReceivedEventPayload);
   }
 
+  /**
+   * Drop the pre-edit body and keep only whether it already satisfied this
+   * automation's content filter. The body itself is never persisted or logged.
+   */
+  override projectPayload(
```
XyneSpaces, line 121: **[major] R7 — no tests for the new edit behavior.** `projectPayload` (strips `previousContent`, computes `previousContentMatched`), the transition rule in `matchFilters` (`contentPasses = p.isEdit ? nowMatches && !p.previousContentMatched : nowMatches`), the router's `fireOnEdit` gate, and the emit-on-update in messages-handler are all new and untested — and the first two are pure functions, trivially testable. Minimum cases: (a) edit newly matches the needle → fires; (b) edit keeps matching → does not re-fire; (c) `fireOnEdit: false` → no run created; (d) encoded FlowJSON card — needle matches the decoded text on both old and new content (the decode symmetry this PR carefully preserves).

apps/backend/src/zero/side-effects/tables/messages-handler.ts
```diff
@@ -2173,7 +2173,28 @@ export class MessagesSideEffectHandler extends BaseSideEffectHandler {
...
+        conversationId: previousValue.conversationId,
+        channelId: previousValue.channelId,
+        msgType: previousValue.msgType as MessageType,
+        userId: previousValue.senderId,
+        isEdit: true,
+        previousContent: previousValue.content,
+      });
+    }
+
+    if (currentMessage.edited && currentMessage.content !== previousValue.content && !currentMessage.isDeleted) {
```
XyneSpaces, line 2197: **[minor] R2 (please verify) — `else if` became a standalone `if`.** The MESSAGE_EDITED notification block used to be the `else` arm of the branch above it (conversation-move handling, judging by its trailing `conversationId` / `isThreadReply` arguments — not fully visible in this diff, so labeling this an assumption). As a standalone `if`, an update that matches both that branch *and* `currentMessage.edited && content-changed` now also fires the edit notifications where it previously didn't. If the two are mutually exclusive by construction, a one-line comment saying so would prevent the next reader from re-asking; otherwise restore the `else`.

XyneSpaces (commented):
Automated review by **Xyne Spaces Reviewer** — 5 inline findings: 0 blocking · 3 major · 1 minor · 1 nit. Full verdict and reviewer recommendation in the top-level summary comment.

Verified sound along the way: registry alignment (`MESSAGE_RECEIVED_EVENT` = trigger `type` = registration key = `'MESSAGE_RECEIVED'`, registered in `automations/index.ts`), `previousContent` stripped before persistence, decode-symmetric needle matching on old + new content, pre-update row used for `previousContent`, and `fireOnEdit` defaulting to `false` (existing automations unaffected).

XyneSpaces:
## Xyne Spaces Reviewer — PR #1762

**Reviewed:** 6 files, 11 hunks · **Findings:** 0 blocking · 3 major · 1 minor · 1 nit

### Verified sound (worth saying)
- **Registry alignment is real:** `MESSAGE_RECEIVED_EVENT` = `MessageReceivedTrigger.type` = the registration key in `automations/index.ts` = `'MESSAGE_RECEIVED'` — the new `triggerRegistry.has(event.type)` lookup resolves and `projectPayload` runs.
- **`previousContent` never persists on the happy path:** `projectPayload` strips it and keeps only `previousContentMatched`; nothing logs it.
- **Decode symmetry preserved:** `prevTexts = [previousContent, toReadableMessageContent(previousContent)]` tests {raw, decoded} of the old body, mirroring `[message.content, message.rawContent]` on the new one — an encoded FlowJSON card can't read as a non-match on either side.
- **Pre-update row used:** conversationService dispatches `onUpdate` with the old row's shape (old content, old thread-reply flag) — `previousContent` is genuinely the pre-edit body.
- **Backward compatible:** `fireOnEdit` defaults to `false`; existing automations see no behavior change until they opt in.
- **No double-emit:** the Zero-path edit and the service-path edit converge on the same `onUpdate` gate (content diff + `!isThreadReply` + `!isDeleted`), and the dispatch is fire-and-forget with a `.catch` log, matching the side-effect dispatch guideline.

### Findings
| major | R2+R8 | `automations/engine/event-router.ts:40` | Registry lookup keyed on `event.type as TriggerType` — string-level alignment across 3 declarations masked by an `as` cast; drift falls silently into the `?? payload` branch |
| major | R5 | `automations/engine/event-router.ts:48` | `fireOnEdit` gate duplicated from `matchFilters` (raw JSON config vs zod-typed), message-specific key hardcoded in the generic router loop, full-JSON `parseAutomationConfig` per candidate on the message hot path |
| major | R7 | `automations/triggers/message-received.trigger.ts:121` | No tests for `projectPayload`, the edit-transition rule, the `fireOnEdit` gate, or the emit-on-update path |
| minor | R2 | `zero/side-effects/tables/messages-handler.ts:2197` | `else if (currentMessage.edited …)` became a standalone `if` — edit notifications can now fire when the preceding branch also matched (assumption — preceding branch not fully visible in diff) |
| nit | R1-adj, R8 | `automations/engine/event-router.ts:52` | `?? payload` fallback would persist the pre-edit body into `workflow_execution_state.context` if the registry ever misses; double-cast on `payload` |

### Recommended reviewer
**@Revanthvenkatpasupuleti** — reworked this PR's core file (`event-router.ts`) in #536 (pre-filter, hydration checks, match-function exports) and owns the surrounding automations engine; no pending review requests observed.

| @Revanthvenkatpasupuleti | `event-router.ts` (reworked in #536), automations engine area | no — no CODEOWNERS match | 0 observed* | authored the pre-filter/hydration design this PR builds on |
| @santhoshJus | — (recent #1749 is message-md builders, not these files) | no — no CODEOWNERS match | heavy* | most recently active in Spaces (today) |

\* No `.github/CODEOWNERS` entry matches the changed paths (the repo's CODEOWNERS covers ACL paths only), so candidates come from recent-commit authorship on `main`. GitHub's `review-requested:` search fell back to lexical mode this session, so per-person counts are unreliable; on the 40 most-recent open PRs, none has any pending requested reviewer.

⚠️ This session has no GitHub reviewer-request tool available, so I could not formally add @Revanthvenkatpasupuleti as a requested reviewer — the @mention here is the request. Please add him via `gh pr edit 1762 --add-reviewer Revanthvenkatpasupuleti` or the GitHub UI.

PiyushJuspay (approved):

Mohan0544 (approved):

PR #1779, opened 2026-09-11
fix: XYNE-63195 testing test cases

5uryan5h:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [x] CI/CD

## Description

XYNE-63195 — run the revived `tools/xyne-automation` Gauge+Playwright suite in CI (`ci / Automation tests`) and make it pass on the self-hosted `ci-cluster` runners. Builds on #1565 (its three follow-up commits are cherry-picked here). The suite runs the full Docker test stack (`docker-compose.dev.yml` + `docker-compose.test.yml`) and executes 32 specs / 71 scenarios; head is green with zero retries consumed.

**CI wiring (`.github/workflows/ci.yml`)** — new `test` job pinned to `ci-cluster-github-runners` (needs Docker + network access to `SPACES_URL`); prints container states + healthcheck probe output + service logs inline; pushes `html-report` and `docker-logs` to a Spaces channel as two archives (the endpoint caps `files` at 10 and Gauge's report needs its css/fonts/images siblings). The push is skipped when the secrets are empty (fork PRs) and fails visibly on a rejected upload. `secrets: inherit` added to `CI-pr.yml` for the job's secrets; the reusable-workflow ref is `@main`.

**Making the stack work behind the runner's egress proxy** — the runners sit behind squid and Docker injects `HTTP(S)_PROXY` into every container:
- healthchecks: `wget -Y off` / `curl --noproxy '*'` (busybox wget has no `no_proxy`; `localhost` probes were being sent to the proxy → "livekit/fake-gcs unhealthy")
- `NO_PROXY`/`no_proxy` for in-stack hostnames on backend, dashboard, livekit and the automation container (Chromium routed `dashboard:5173` via the proxy; LiveKit's Go client proxied its `participant_joined` webhook)
- redis via `mirror.gcr.io` (ECR Public rate-limits the shared runner IP); `start_period` on infra healthchecks

**Test-harness fixes** — browser contexts opt out of the "open in desktop app" interstitial (#1658) that intercepted clicks; the three call "starts call" scenarios now hang up, so the join scenarios never meet a stale call (that was the first-attempt failure hidden behind Gauge retries); idempotent link-toolbar step; `collect-ci-artifacts.sh` resolves compose files from its own location and records healthcheck probe output; gauge binary chosen by build arch in the automation Dockerfile.

**Backend (behavioural, benefits production)** — `callRepository.handleRoomFinished` now clears `conversation.callId` only when it still points at the room that finished. A stale room's `room_finished` can land after a newer call has started in the same conversation and was wiping the live call's pointer (observed in the test stack; the same sequence happens when a caller's device drops mid-call and a new call starts within LiveKit's `empty_timeout`).

**Test-auth (test mode only)** — sets the canonical `user_session_id` cookie; ensures the per-workspace DM project exists via a race-safe `createMany … skipDuplicates`.

**LiveKit dev credentials** — `devkey`/`devsecret` → `localdevkey`/`localdevsecret` everywhere (`docker/livekit.yaml`, egress, dev-infra config, supervisord, all compose files, `sandbox.sh`, nix service defaults, both `.env.example`s, nix doc examples). The backend rejects the published `devkey` at boot since #289, so every consumer had to move together.

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

## Schema & Data Changes

## Config, Environment & URLs

- [x] Key added to **all** relevant env files — `LIVEKIT_API_KEY`/`LIVEKIT_API_SECRET` values renamed in `apps/backend/.env.{example,test}`, `apps/backend/python-agent/.env.example`, every compose file, `scripts/sandbox.sh`, nix
- [x] Env setup / secret scripts and docker compose updated
- [x] Verified in every consumer — livekit server config, egress, transcription agent exports, sandbox generator; test stack boots and passes the suite
- [x] Google / Microsoft OAuth redirect URIs still resolve (untouched)
- [x] Nothing hardcoded, no secrets committed (CI secrets: `TESTCASES_JWT`, `TESTCASES_CHANNELID`, `SPACES_URL` read from repo secrets)

`LIVEKIT_API_KEY=localdevkey`, `LIVEKIT_API_SECRET=localdevsecret` (dev/test/sandbox/nix only — value rename, no new keys). Test-stack-only: `NO_PROXY`/`no_proxy` on backend/dashboard/livekit/automation services in `docker-compose.test.yml`.

## API Contract

- [x] No API changes

---

## How did you test it?

**CI** — `ci / Automation tests` on head: 32/32 specs, 71/71 scenarios, 0 failed, **0 retries consumed** (verified from the Gauge log, not just the tally), ~17 min. Earlier runs on this branch document each fix landing: proxy-bypassed healthchecks → stack healthy; `NO_PROXY` → first page load; interstitial opt-out → fixture passes; LiveKit `NO_PROXY` → join-call scenarios; hang-ups → no retries.

**Local** — full CI mirror (backend build, dashboard lint/format/typecheck/build, framework lint, claw typecheck, automation `tsc`, compose configs, gitleaks) green under Node 22. Docker runner locally with `--parallel=1` (CI's sequential order): reproduced the stale-call first-attempt failure, then all 7 call scenarios first-attempt after the hang-ups; backend log confirmed the stale `room_finished` arriving after the new call and no longer clearing `conversation.callId`. Automation image built and gauge verified on both `linux/arm64` and `linux/amd64`.

### Automated

- [ ] Backend unit tests
- [ ] Claw tests
- [x] End-to-end suite — `pnpm run test` (Docker runner, local and CI)
- [x] Added / updated tests for this change (call-spec hang-ups, link-toolbar step, interstitial opt-out)
- [ ] Not covered by tests

### Manual

**Users**
- [x] Two users at once — the join-call scenarios drive a caller and a callee in separate browser contexts

**Login**
- [x] Dev auth — test-auth (`NODE_ENV=test`) path; 10 concurrent test logins verified against the DM-project ensure

**Run mode**
- [x] Test mode (`dev:test`)
- [x] Docker compose

**Sync & resilience** — n/a (no Zero changes)

**Surface & data**
- [x] Web browser (headless Chromium via Playwright)
- [x] Error paths — proxy-blocked probes/pages, stale-call race, rate-limited image pulls all reproduced and fixed

---

## Checklist

- [x] Reviewed my own diff; scoped to one thing (CI for the automation suite + what it took to make it pass; the one backend change is declared above)

### DIFF
--- a/.github/workflows/CI-pr.yml
+++ b/.github/workflows/CI-pr.yml
@@ -23,6 +23,7 @@ concurrency:
 
 jobs:
   ci:
-    uses: juspay/xyne-spaces/.github/workflows/ci.yml@main
+    uses: juspay/xyne-spaces/.github/workflows/ci.yml@fix/testcases-run-v2
+    secrets: inherit
     permissions:
       contents: read
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,8 +154,20 @@ jobs:
           REPORT_HTML=$(find tools/xyne-automation/reports -type f -path '*/html-report/index.html' -printf '%T@ %p\n' \
             | sort -n | tail -n1 | cut -d' ' -f2-)
 
-          if [ -z "$REPORT_HTML" ]; then
-            echo "::error::No html-report/index.html found under tools/xyne-automation/reports"
+          DOCKER_LOGS_DIR=$(find tools/xyne-automation/reports -maxdepth 2 -type d -name docker-logs -printf '%T@ %p\n' \
+            | sort -n | tail -n1 | cut -d' ' -f2-)
+          UPLOAD_ARGS=()
+          if [ -n "$DOCKER_LOGS_DIR" ]; then
+            for log in "$DOCKER_LOGS_DIR"/*.log; do
+              [ -e "$log" ] && UPLOAD_ARGS+=(-F "files=@${log};type=text/plain")
+            done
+          fi
+          if [ -n "$REPORT_HTML" ]; then
+            UPLOAD_ARGS+=(-F "files=@${REPORT_HTML};type=text/html")
+          fi
+
+          if [ ${#UPLOAD_ARGS[@]} -eq 0 ]; then
+            echo "::error::Nothing to push — no html-report and no docker-logs found"
             exit 1
           fi
 
@@ -163,7 +175,7 @@ jobs:
             -H "Authorization: Bearer $JWT" \
             -F "channelId=$CHANNELID" \
             -F "markdownText=CI automation test report — commit ${GITHUB_SHA::9}" \
-            -F "files=@${REPORT_HTML};type=text/html"
+            "${UPLOAD_ARGS[@]}"
 
   # Framework lint — self-contained (no workspace deps).
   framework-lint:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,7 +158,7 @@ jobs:
             | sort -n | tail -n1 | cut -d' ' -f2-)
           UPLOAD_ARGS=()
           if [ -n "$DOCKER_LOGS_DIR" ]; then
-            for log in "$DOCKER_LOGS_DIR"/*.log; do
+            for log in "$DOCKER_LOGS_DIR"/*.log "$DOCKER_LOGS_DIR"/*.txt; do
               [ -e "$log" ] && UPLOAD_ARGS+=(-F "files=@${log};type=text/plain")
             done
           fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,6 +143,13 @@ jobs:
       - name: Run automation tests
         run: pnpm run test
 
+      - name: Show container states and logs
+        if: always()
+        run: |
+          find tools/xyne-automation/reports -type f \( -name '*.log' -o -name '*.txt' \) | sort | while read -r f; do
+            echo "::group::$f"; tail -n 200 "$f"; echo "::endgroup::"
+          done
+
       - name: Push report to Spaces
         if: always()
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
   # ---------------------------------------------------------------------------
   backend:
     name: Backend typecheck + build
-    runs-on: ${{ vars.USE_SELF_HOSTED_CI == 'true' && 'ci-cluster-github-runners-mini' || 'ubuntu-latest' }}
+    runs-on: ${{ vars.USE_SELF_HOSTED_CI == 'false' && 'ci-cluster-github-runners-mini' || 'ubuntu-latest' }}
     # tsc over backend's type surface (incl. generated Prisma/Zero schemas)
     # exceeds Node's ~2GB default heap on the runner.
     env:
@@ -65,7 +65,7 @@ jobs:
   # ---------------------------------------------------------------------------
   dashboard:
     name: Dashboard ${{ matrix.task }}
-    runs-on: ${{ vars.USE_SELF_HOSTED_CI != 'true' && 'ubuntu-latest' || matrix.task == 'build' && 'ci-cluster-github-runners-mini-max' || 'ci-cluster-github-runners-mini' }}
+    runs-on: ${{ vars.USE_SELF_HOSTED_CI != 'false' && 'ubuntu-latest' || matrix.task == 'build' && 'ci-cluster-github-runners-mini-max' || 'ci-cluster-github-runners-mini' }}
     strategy:
       fail-fast: false
       matrix:
@@ -116,6 +116,54 @@ jobs:
       - name: Build
         if: matrix.task == 'build'
         run: pnpm --filter xyne-spaces-dashboard run build
+
+  test:
+    name: Automation tests
+    runs-on: ci-cluster-github-runners
+    env:
+      TEST_ENV: test
+      BROWSER: chromium
+    steps:
+      - uses: actions/checkout@v5
+
+      - uses: pnpm/action-setup@v4
+
+      - uses: actions/setup-node@v5
+        if: vars.USE_SELF_HOSTED_CI == 'true'
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          cache: pnpm
+
+      - name: Install dependencies
+        run: pnpm install ${{ env.PNPM_FLAGS }} --filter xyne-automation...
+
+      - name: Install Playwright browser
+        run: pnpm --filter xyne-automation exec playwright install chromium
+
+      - name: Run automation tests
+        run: pnpm run test
+
+      - name: Push report to Spaces
+        if: always()
+        env:
+          JWT: ${{ secrets.TESTCASES_JWT }}
+          CHANNELID: ${{ secrets.TESTCASES_CHANNELID }}
+          SPACES_URL: ${{ secrets.SPACES_URL }}
+        run: |
+          set -euo pipefail
+          REPORT_HTML=$(find tools/xyne-automation/reports -type f -path '*/html-report/index.html' -printf '%T@ %p\n' \
+            | sort -n | tail -n1 | cut -d' ' -f2-)
+
+          if [ -z "$REPORT_HTML" ]; then
+            echo "::error::No html-report/index.html found under tools/xyne-automation/reports"
+            exit 1
+          fi
+
+          curl -X POST "${SPACES_URL%/}/api/apps/files/filesUpload" \
+            -H "Authorization: Bearer $JWT" \
+            -F "channelId=$CHANNELID" \
+            -F "markdownText=CI automation test report — commit ${GITHUB_SHA::9}" \
+            -F "files=@${REPORT_HTML};type=text/html"
 
   # Framework lint — self-contained (no workspace deps).
   framework-lint:
--- a/apps/backend/src/controllers/testAuthController.ts
+++ b/apps/backend/src/controllers/testAuthController.ts
@@ -1,5 +1,5 @@
 import { Request, Response } from 'express';
-import { AccessType, AuthProvider, OrgRole, WorkspaceRole } from '@xyne/shared';
+import { AccessType, AuthProvider, OrgRole, ProjectType, WorkspaceRole } from '@xyne/shared';
 import { randomBytes } from 'crypto';
 import { logger } from '@/utils/logger';
 import { config } from '@/config/env';
@@ -256,6 +256,24 @@ export class TestAuthController {
         }
       }
 
+      // Ensure the per-workspace "DM" project exists (normally created during org
+      // onboarding); without it POST /api/users/me/dms 500s for pre-existing test workspaces.
+      const existingDmProject = await db.project.findFirst({
+        where: { workspaceId: workspace.id, code: 'DM', type: ProjectType.DM },
+      });
+      if (!existingDmProject) {
+        await db.project.create({
+          data: {
+            name: 'Direct Messages',
+            code: 'DM',
+            description: 'DM project (test-auth)',
+            workspaceId: workspace.id,
+            type: ProjectType.DM,
+            createdBy: user.id,
+          },
+        });
+      }
+
       const effectiveIsNewUser = setAsNewUser ?? isNewUser;
       logger.info(`[${requestId}] Org ${organization.orgId}, workspace ${workspace.id}, user ${user.id} (dbIsNew: ${isNewUser}, effectiveIsNew: ${effectiveIsNewUser})`);
 
@@ -396,6 +414,13 @@ export class TestAuthController {
 
       if (sessionId) {
         res.cookie('xyne_session', sessionId, {
+          ...cookieOptions,
+          maxAge: config.session.expiryDays * 24 * 60 * 60 * 1000,
+        });
+
+        // Session cookie the auth middleware and session-scoped routes require
+        // (real login controllers set it too); without it session-gated routes 401.
+        res.cookie('user_session_id', sessionId, {
           ...cookieOptions,
           maxAge: config.session.expiryDays * 24 * 60 * 60 * 1000,
         });
--- a/apps/backend/src/controllers/testAuthController.ts
+++ b/apps/backend/src/controllers/testAuthController.ts
@@ -258,21 +258,19 @@ export class TestAuthController {
 
       // Ensure the per-workspace "DM" project exists (normally created during org
       // onboarding); without it POST /api/users/me/dms 500s for pre-existing test workspaces.
-      const existingDmProject = await db.project.findFirst({
-        where: { workspaceId: workspace.id, code: 'DM', type: ProjectType.DM },
+      // createMany+skipDuplicates is a single INSERT ... ON CONFLICT DO NOTHING, so concurrent
+      // test logins can't race each other into a P2002 (upsert is not atomic here).
+      await db.project.createMany({
+        data: {
+          name: 'Direct Messages',
+          code: 'DM',
+          description: 'DM project (test-auth)',
+          workspaceId: workspace.id,
+          type: ProjectType.DM,
+          createdBy: user.id,
+        },
+        skipDuplicates: true,
       });
-      if (!existingDmProject) {
-        await db.project.create({
-          data: {
-            name: 'Direct Messages',
-            code: 'DM',
-            description: 'DM project (test-auth)',
-            workspaceId: workspace.id,
-            type: ProjectType.DM,
-            createdBy: user.id,
-          },
-        });
-      }
 
       const effectiveIsNewUser = setAsNewUser ?? isNewUser;
       logger.info(`[${requestId}] Org ${organization.orgId}, workspace ${workspace.id}, user ${user.id} (dbIsNew: ${isNewUser}, effectiveIsNew: ${effectiveIsNewUser})`);
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -140,6 +140,15 @@ services:
       - BACKEND_URL=http://backend:3001
       - DASHBOARD_URL=http://dashboard:5173
       - BROWSER=chromium
+      # Use the pre-baked image deps as-is: make pnpm non-interactive and stop it
+      # re-verifying/reinstalling (a reinstall re-blocks @getgauge/cli's build script).
+      - CI=true
+      - NPM_CONFIG_VERIFY_DEPS_BEFORE_RUN=false
+      - NPM_CONFIG_CONFIRM_MODULES_PURGE=false
+      # Container-scoped log-noise suppression for the test run.
+      - NPM_CONFIG_LOGLEVEL=error
+      - NODE_NO_WARNINGS=1
+      - COREPACK_ENABLE_DOWNLOAD_PROMPT=0
     depends_on:
       backend:
         condition: service_healthy
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -76,8 +76,8 @@ services:
   #     - DATABASE_URL=postgresql://xyne:xyne123@postgres:5432/xyne_dev_db
   #     - REDIS_URL=redis://redis:6379
   #     - LIVEKIT_URL=ws://livekit:7880
-  #     - LIVEKIT_API_KEY=devkey
-  #     - LIVEKIT_API_SECRET=devsecret
+  #     - LIVEKIT_API_KEY=localdevkey
+  #     - LIVEKIT_API_SECRET=localdevsecret
   #     - STORAGE_EMULATOR_HOST=http://fake-gcs:4443
   #     - YSWEET_URL=http://ysweet:8085
   #     - ZERO_CACHE_URL=http://zero-cache:4848
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -33,6 +33,11 @@ services:
       - "${BACKEND_BIND_PORT:-0}:3001"
     env_file:
       - ./apps/backend/.env.test   # Override with container hostnames
+    environment:
+      # The CI runner injects HTTP(S)_PROXY into every container (squid egress); in-stack
+      # hostnames must bypass it. Chromium reads the lower-case var, curl/gaxios/axios either.
+      - NO_PROXY=localhost,127.0.0.1,backend,dashboard,zero-cache,livekit,fake-gcs,ysweet,superposition,host.docker.internal
+      - no_proxy=localhost,127.0.0.1,backend,dashboard,zero-cache,livekit,fake-gcs,ysweet,superposition,host.docker.internal
     command: sh -c "/scripts/setup-fake-gcs-test.sh && pnpm run dev:test"
     depends_on:
       backend-migrate:
@@ -61,6 +66,11 @@ services:
     container_name: ${COMPOSE_PROJECT_NAME:-xyne}-dashboard
     env_file:
       - ./apps/dashboard/.env.test   # Override with container hostnames
+    environment:
+      # The CI runner injects HTTP(S)_PROXY into every container (squid egress); in-stack
+      # hostnames must bypass it. Chromium reads the lower-case var, curl/gaxios/axios either.
+      - NO_PROXY=localhost,127.0.0.1,backend,dashboard,zero-cache,livekit,fake-gcs,ysweet,superposition,host.docker.internal
+      - no_proxy=localhost,127.0.0.1,backend,dashboard,zero-cache,livekit,fake-gcs,ysweet,superposition,host.docker.internal
     ports:
       - "${DASHBOARD_BIND_PORT:-0}:5173"
     depends_on:
@@ -157,6 +167,10 @@ services:
       - NPM_CONFIG_LOGLEVEL=error
       - NODE_NO_WARNINGS=1
       - COREPACK_ENABLE_DOWNLOAD_PROMPT=0
+      # The CI runner injects HTTP(S)_PROXY into every container (squid egress); in-stack
+      # hostnames must bypass it. Chromium reads the lower-case var, curl/gaxios/axios either.
+      - NO_PROXY=localhost,127.0.0.1,backend,dashboard,zero-cache,livekit,fake-gcs,ysweet,superposition,host.docker.internal
+      - no_proxy=localhost,127.0.0.1,backend,dashboard,zero-cache,livekit,fake-gcs,ysweet,superposition,host.docker.internal
     depends_on:
       backend:
         condition: service_healthy
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -21,6 +21,14 @@ services:
       context: .
       dockerfile: apps/backend/Dockerfile.test
     container_name: ${COMPOSE_PROJECT_NAME:-xyne}-backend
+    # docker/livekit.yaml posts webhooks to host.docker.internal:3001 (the dev stack's
+    # host-run backend). Here the backend is this container, so answer to that name —
+    # without it participant_joined never lands, no call row/message is created, and a
+    # second member never sees the join button.
+    networks:
+      default:
+        aliases:
+          - host.docker.internal
     ports:
       - "${BACKEND_BIND_PORT:-0}:3001"
     env_file:
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -46,7 +46,7 @@ services:
       livekit:
         condition: service_healthy
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:3001/api/health"]
+      test: ["CMD", "curl", "-f", "--noproxy", "*", "http://localhost:3001/api/health"]
       interval: 5s
       timeout: 10s
       retries: 60
@@ -67,7 +67,7 @@ services:
       backend:
         condition: service_healthy
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:5173/"]
+      test: ["CMD", "curl", "-f", "--noproxy", "*", "http://localhost:5173/"]
       interval: 5s
       timeout: 5s
       retries: 24
@@ -103,7 +103,7 @@ services:
   #     livekit:
   #       condition: service_healthy
   #   healthcheck:
-  #     test: ["CMD", "curl", "-f", "http://localhost:3001/api/health"]
+  #     test: ["CMD", "curl", "-f", "--noproxy", "*", "http://localhost:3001/api/health"]
   #     interval: 10s
   #     timeout: 5s
   #     retries: 15
@@ -126,7 +126,7 @@ services:
   #     backend:
   #       condition: service_healthy
   #   healthcheck:
-  #     test: ["CMD", "curl", "-f", "http://localhost:5173/"]
+  #     test: ["CMD", "curl", "-f", "--noproxy", "*", "http://localhost:5173/"]
   #     interval: 10s
   #     timeout: 5s
   #     retries: 10
--- a/.github/workflows/CI-push.yml
+++ b/.github/workflows/CI-push.yml
@@ -21,6 +21,7 @@ concurrency:
 jobs:
   ci:
     uses: ./.github/workflows/ci.yml
+    secrets: inherit
 
   semantic-release:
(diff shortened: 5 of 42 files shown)


sumantop:
Pushed to this branch (on request from @sumantop, no history rewritten):

- **Cherry-picked the three follow-up commits from #1565** that this branch's base was missing (`e98b0d24`, `eaed2d79`, `26df087f`): the rest of the `localdevkey` rename across every consumer of the renamed `docker/livekit.yaml` (egress, sandbox/local compose, nix, both `.env.example`s, `sandbox.sh`, supervisord), the `host.docker.internal` alias on the test backend so LiveKit webhooks land in the test stack (un-quarantines the 3 join-call scenarios), and the race-safe test-auth DM-project ensure. Without these, the test backend boots with `devkey` against a livekit that only trusts `localdevkey`.
- **`e6031b64` — healthcheck `start_period`** for redis / livekit / fake-gcs / minio / zero-cache in `docker-compose.dev.yml`. The last three CI failures were `livekit is unhealthy` then `fake-gcs is unhealthy`: those checks are `interval 5s × retries 5` with no `start_period`, so a cold runner gets 25 s to come up before `up --wait` gives up (livekit: started 08:28:06 → unhealthy 08:28:33 in run 34577916640). Adding `depends_on: redis` only changed which container lost the race. `start_period` doesn't slow a healthy start — failed probes just don't count until it elapses — which is what postgres/superposition/ysweet/backend already have.

Watching CI on this branch and will keep fixing until it's green.

sumantop:
**Automation suite is green on CI as of `f060afd9`: 32/32 specs, 71/71 scenarios, 0 failed (18m50s).** First full pass on the `ci-cluster` runners.

What it took, for the record — every one of these was the runner's egress proxy or its environment, not the tests:
- `8cafc6dc` redis via `mirror.gcr.io` (ECR Public rate-limits the shared runner IP)
- `0fc830fe` `wget -Y off` / `curl --noproxy '*'` on the healthchecks — busybox wget has no `no_proxy`, so `localhost` probes were sent to squid and timed out → the "livekit/fake-gcs unhealthy" failures
- `902d78cb` `NO_PROXY`/`no_proxy` for in-stack hosts on backend/dashboard/automation — Linux Chromium was routing `dashboard:5173` through squid (403 page)
- `e66bd1f6` opt test contexts out of the new "open in desktop app" interstitial (#1658) that covered the page
- `f060afd9` same `NO_PROXY` for livekit (its Go client proxied the `participant_joined` webhook → 3 join-call scenarios), plus an idempotent link-toolbar step
- `7ffd23ae` log collection fixed + printed in the job (`docker inspect` health-probe output), which is what made the above diagnosable

**`ff3c194e` (just pushed):** the `runs-on` gating in `ci.yml` had been inverted (`USE_SELF_HOSTED_CI == 'false'` → self-hosted), sending build/lint/typecheck to the `mini`/`mini-max` runners where the Dashboard build takes 13–26 min and gets killed (2/2 runs). Restored main's expressions; those jobs are ~4 min on `ubuntu-latest`. The automation job keeps its own `ci-cluster-github-runners` pin.

One thing to do **before merge** (left alone on purpose): `CI-pr.yml` / `CI-push.yml` reference `ci.yml@fix/testcases-run-v2` so this branch exercises its own workflow — they need to go back to `@main` in the final commit, or CI breaks repo-wide once the branch is deleted.

Known soft spot to watch: `User joins ongoing call from DM` timed out once on `call-window` (60s) and passed on Gauge's retry — media negotiation on that runner; no action unless it repeats.

sumantop:
**`92f8009a` — the `User joins ongoing call from DM` first-attempt failure (was passing only on Gauge's retry, every run).**

Root cause, reproduced locally with `--parallel=1` (the sequential order CI uses):
1. `User starts call from DM` leaves user-1 in a call and ends by discarding the browser context — no hang-up. LiveKit keeps the room for `departure_timeout: 10` + `empty_timeout: 120`; the backend only clears `conversation.callId` on `room_finished`.
2. `User joins ongoing call from DM` runs seconds later. Because the baseline DM still has an active call, `useCallActions` (`hasActiveCallInChannel`) turns the caller's start button into **`JOIN_CALL` on the stale call** instead of `INITIATE_CALL`. user-2's incoming modal is for that stale call, and accepting it never yields a call window.
3. The retry ~90 s later lands after `empty_timeout` → fresh call → passes. Hence "fails first, passes on retry, every time". Locally it never showed because I'd run with parallel workers, so the two scenarios weren't back-to-back.

Two changes:
- **Test:** scenario 1 now hangs up (`end-call-button`, wait for `call-window` to disappear) — what a real user does. Sequential local run: 2/2 first attempt, 36 s (was 1m44s burning the 60 s timeout).
- **Backend (`callRepository.handleRoomFinished`):** the live backend log from the repro showed a real bug underneath — the stale room's `room_finished` arrived 3 s *after* the new call had started in the same DM and **unconditionally cleared `conversation.callId`**, wiping the pointer to the live call. Now `updateMany({ where: { conversationId, callId: callExternalId } })` — only clear it if it still points at the room that finished. Verified: same overlap in the re-run, no clear. One-line guard; happy to split it into its own PR if you'd rather keep this one CI-only.

Product note for a ticket: the user-facing version of (1)–(2) is real — caller's laptop closes mid-call, a teammate opens the DM within ~2 min, sees "incoming call", accepts, gets nothing. The test just happened to exercise it.

sumantop:
`cd6f447420` — the same stale-call fix for the group and channel specs. The `92f8009a` run passed the DM join first-attempt on CI, but `User joins ongoing group call` consumed a retry for the identical reason (its "starts call" scenario also never hung up; it had only been passing first-try by timing). All three "starts call" scenarios now end the call. Local sequential run of the whole calls folder: 7/7 first attempt, 1m15s, no retries.

.github/workflows/CI-pr.yml
```diff
@@ -23,6 +23,7 @@ concurrency:
 
 jobs:
   ci:
-    uses: juspay/xyne-spaces/.github/workflows/ci.yml@main
+    uses: juspay/xyne-spaces/.github/workflows/ci.yml@fix/testcases-run-v2
+    secrets: inherit
```
ameer030: 🔴 **B01 — this PR makes its own branch the CI that grades it, and hands it secrets**

This changes the reusable-workflow ref from `@main` to this PR's own branch (`@fix/testcases-run-v2`) and adds `secrets: inherit`. Whoever pushes to this branch now edits the very workflow that produces the green checks on this PR — and that workflow receives repo secrets, including the new `TESTCASES_JWT`. As @sumantop already noted in this thread's history, the ref must go back to `@main` before merge; once the branch is deleted, every other PR (and the merge queue — this workflow also fires on `merge_group`) fails instantly.

`.github/workflows/CI-pr.yml:26-27` — `uses: juspay/xyne-spaces/.github/workflows/ci.yml@fix/testcases-run-v2` + `secrets: inherit`. Main's copy of this file has no `secrets: inherit` and points at `@main`. CodeQL already flagged this exact line (unpinned non-immutable ref) — thread still open. The new `Automation tests` job in `ci.yml` consumes `TESTCASES_JWT` / `TESTCASES_CHANNELID` / `SPACES_URL`, which only reach it through this `secrets: inherit`, so the PR branch's workflow code currently controls code that handles those secrets. Note `secrets: inherit` itself is fine and needed — it is safe as soon as the callee is the base-branch copy again.
</details>

Revert the `uses:` line to `juspay/xyne-spaces/.github/workflows/ci.yml@main` (a `temp:` commit where the last commit before merge restores it works too — the point is it must not merge like this). Keep `secrets: inherit`.
</details>

rules: CI-C03, CI-C17
agent: omp
run: n/a
</details>

ameer030 (commented):
🔴 **B02 — PR body is the untouched template, and the diff quietly goes well beyond its title**

Every template section is still blank or holds only its placeholder comment — no Description, no Areas Touched checkboxes, no "How did you test it" evidence. Meanwhile the diff contains more than test-harness repairs, and none of the extras are declared anywhere in the body.

- Real backend race fix: `apps/backend/src/database/repositories/callRepository.ts` — `handleRoomFinished` now clears `conversation.callId` only when it still points at the finished room (a stale `room_finished` could wipe the pointer to a live call). Good fix, but it belongs in the body as a declared behavioral change — it benefits every production call, not just tests.
- Test-auth changes: `apps/backend/src/controllers/testAuthController.ts` — new `user_session_id` cookie + an ensured DM project.
- Repo-wide LiveKit credential rename `devkey` → `localdevkey` across compose files, env examples, nix, configs — an env/config change the declared "Environment" section should mention (boot code now rejects the old values, so any developer doc/example still carrying them breaks).
- The excellent step-by-step story lives only in PR comments. Reviewers and the release-notes parser read the template sections, not comment archaeology.
</details>

Fill the template: Description with the XYNE-63195 link and the true scope (test-harness CI wiring + the callRepository race fix + LiveKit key rename), tick the real Areas Touched (apps/backend, infra, tools), declare the LiveKit dev-key rename in Config/Environment, and paste/link the CI-green evidence from @sumantop's comments as the POT.
</details>

rules: PROC-C15, PROC-C02
agent: omp
run: n/a
</details>

ameer030 (commented):
🔵 **N01 — three nix examples still show the old LiveKit dev credentials**

The rename covered every executed path, but the teensy leftovers below are copy-paste traps: `env.ts` now rejects `devkey` at boot ("LIVEKIT_API_KEY must not be the published development key"), so anyone following these examples gets a backend that refuses to start.

- `nix/containers/README.md:83-84`
- `nix/containers/container.nix:75`
- `nix/containers/default.nix:47`

Rename `devkey`/`devsecret` → `localdevkey`/`localdevsecret` in those three example blocks.
</details>

rules: DOCS-C04
agent: omp
run: n/a
</details>

.github/workflows/CI-pr.yml
```diff
@@ -23,6 +23,7 @@
 
 jobs:
   ci:
-    uses: juspay/xyne-spaces/.github/workflows/ci.yml@main
+    uses: juspay/xyne-spaces/.github/workflows/ci.yml@fix/testcases-run-v2
```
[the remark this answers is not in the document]
ameer030: Confirmed — folded into review finding B01 above. This is worse than "unpinned": the ref points at this PR's own branch, so pushes to the branch rewrite the very workflow that grades the PR, and the added `secrets: inherit` hands repo secrets (incl. `TESTCASES_JWT`) to workflow code on that branch. It must return to `@main` before merge (as @sumantop already flagged); after the branch is deleted this also breaks every other PR and the merge queue, since the same workflow serves `pull_request` + `merge_group`. `secrets: inherit` can stay once the callee is the base-branch copy.

.github/workflows/ci.yml
```diff
@@ -115,6 +115,54 @@
...
+  test:
+    name: Automation tests
+    runs-on: ci-cluster-github-runners
+    env:
+      TEST_ENV: test
+      BROWSER: chromium
+    steps:
+      - uses: actions/checkout@v5
+
+      - uses: pnpm/action-setup@v4
```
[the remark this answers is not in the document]
ameer030: Still open at head `cd6f4474`. Context worth recording on this thread: the same unpinned `@v4`/`@v5` tag refs are the existing repo-wide convention for every other job on `main` (checkout, setup-node, pnpm/action-setup), so this new job isn't introducing a regression so much as inheriting the convention. Per the alert-dismissal policy this is a repo-policy call for the code owners: either dismiss this alert with a written justification, or take a single follow-up PR that pins every `uses:` ref in the workflows to commit SHAs at once — fixing it only in this new job while leaving the rest would be noise.

ameer030 (commented):
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 2 |
| 🟡 Better to fix | 0 |
| 🔵 Nit | 1 |

### 🔁 Re-check
- 💬 **Existing threads:** 0 fixed ✅ · 0 satisfied by reply · 2 still open — CodeQL's "unpinned reusable workflow ref" on `CI-pr.yml` (folded into finding B01 — must revert to `@main`) and CodeQL's "unpinned `pnpm/action-setup@v4`" on `ci.yml` (repo-wide convention; replied on-thread with dismiss-or-pin-all guidance)
- ✅ **Verified OK:** race-guarded `conversation.callId` clearing uses the DB-atomic conditional update (`updateMany` with the finished room's id in `WHERE`, side effect gated on row count); test-auth DM-project ensure is backed by `@@unique([code, workspaceId])` and matches onboarding's shape; the `devkey`→`localdevkey` rename covers every executed path (compose, env files, livekit/egress configs, nix livekit service, sandbox script); dashboard consumes `VITE_ZERO_PATH` (`config.ts:55`); every testid/key the specs touch exists (`recordings-v2-page`, `APPLY_LINK`, `composer-link-popover`, `xyne:openInBrowser`); Gauge's `--max-retries-count` guard is correct (flag now only passed when retries > 0); full automation suite green on CI at head (71/71 scenarios).

### 📌 Verdict
2 blockers must be fixed before merge ❌ — point `CI-pr.yml`'s reusable-workflow ref back at `@main` and fill the PR template; the nix-example nit can ride the same push.

.github/workflows/CI-pr.yml
sumantop: Done in `b735eea8` — `CI-pr.yml` is back on `ci.yml@main`; the only remaining diff against main on that file is `secrets: inherit`, kept as you said. Held it as the last workflow change on purpose: while the ref pointed at the branch, each push exercised the branch's own `ci.yml` (that's how the new Automation job was verified — 71/71, 0 retries at `f08e4be4`, and the report push actually landed both archives in the channel for the first time). From here on the branch runs main's workflow, so the Automation job won't appear on further pushes to this PR until it merges.

.github/workflows/CI-pr.yml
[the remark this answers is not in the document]
sumantop: Resolved by `b735eea8` — ref back to `@main`, so the CodeQL 'unpinned non-immutable ref' finding for this line goes away with it.

sumantop:
Replies to the review:

- **B02** — PR body filled in: true scope (CI wiring + what it took to pass on the proxied runners + the `handleRoomFinished` race guard declared as a behavioural change + the LiveKit dev-key rename under Config/Environment), Areas Touched, and the POT from the CI runs / local mirror / local sequential repro.
- **N01** — the three `nix/containers` examples renamed to `localdevkey`/`localdevsecret` in `b735eea8`; `grep devkey` on the branch now returns only the Joi `invalid()` guard.
- **B01** — answered on the thread; `CI-pr.yml` is on `@main` with `secrets: inherit` kept.

Also in this push (`b46abfed`): the report-push step from the earlier review follow-up assumed a `{"success":true}` body, but the endpoint's success shape is a `FILE_UPLOADED` event — the `f08e4be4` run uploaded both archives fine and then failed the step on that check. It now treats `FILE_UPLOADED` as success and only `success:false` / HTTP errors as failure; exercised in a Linux container against stubbed success / rejected / 500 / no-secrets responses.

sumantop:
`b0668e41aa` — re: the "can we make the report an .html" ask in the channel: Gauge's report is multi-page (index + 32 spec pages + css/fonts/images), so a bare `index.html` renders unstyled with dead links — that's why it was a `.tgz`. Instead the job now generates **`automation-report.html`**, a single self-contained page (no external assets) from Gauge's `json-report`: status + totals, failures up top with the failed step, error, and screenshot inline, scenarios that only passed after a retry called out, and passing specs collapsed. It's posted first in the Spaces message; the two archives stay for full detail. Verified locally with a deliberate failure and a retried scenario.

docker-compose.test.yml
```diff
@@ -21,10 +21,23 @@ services:
...
+          - host.docker.internal
     ports:
       - "${BACKEND_BIND_PORT:-0}:3001"
     env_file:
       - ./apps/backend/.env.test   # Override with container hostnames
+    environment:
+      # The CI runner injects HTTP(S)_PROXY into every container (squid egress); in-stack
+      # hostnames must bypass it. Chromium reads the lower-case var, curl/gaxios/axios either.
+      - NO_PROXY=localhost,127.0.0.1,backend,dashboard,zero-cache,livekit,fake-gcs,ysweet,superposition,host.docker.internal
+      - no_proxy=localhost,127.0.0.1,backend,dashboard,zero-cache,livekit,fake-gcs,ysweet,superposition,host.docker.internal
```
ameer030: 🟡 **W01 — the same NO_PROXY list is now pasted 8 times in this file**

Every service in the test stack needs to bypass the CI runner's squid proxy for in-stack hostnames. The identical two `NO_PROXY`/`no_proxy` lines (an 11-host list) are duplicated across four services (backend, livekit, dashboard, automation). The next in-stack hostname has to be added in 8 places — miss one and that service's in-stack traffic silently routes through the proxy and flakes the suite.

Identical `NO_PROXY`+`no_proxy` pairs at `docker-compose.test.yml:39-40` (backend), `:66-67` (livekit), `:79-80` (dashboard), `:179-180` (automation). The compose files use no YAML anchors or `x-` extension fields today, so nothing keeps the copies in sync.

</details>

Define the list once as an anchor and reference the alias. Aliases can't expand inside `key=value` list items, so switch the two lines to map form:

```yaml
x-test-stack-no-proxy: &test-stack-no-proxy
  localhost,127.0.0.1,backend,dashboard,zero-cache,livekit,fake-gcs,ysweet,superposition,host.docker.internal

services:
  backend:
    environment:
      NO_PROXY: *test-stack-no-proxy
      no_proxy: *test-stack-no-proxy
```

One hostname addition then edits one line instead of eight.

</details>

rules: BUG-C03
agent: omp
run: n/a

</details>

.github/workflows/ci.yml
```diff
@@ -117,6 +117,84 @@
...
+    # Always self-hosted (not gated on USE_SELF_HOSTED_CI): needs Docker for the compose
+    # stack and network access to SPACES_URL for the report push.
+    runs-on: ci-cluster-github-runners
+    env:
+      TEST_ENV: test
+      BROWSER: chromium
+    steps:
+      - uses: actions/checkout@v5
+
+      - uses: pnpm/action-setup@v4
```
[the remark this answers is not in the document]
ameer030: Re-reviewed at head `b0668e41`: this is a duplicate of the resolved sibling alert on the same lines. The `@v4`/`@v5` tag refs used by the new `test` job match the convention in every other job of `ci.yml` (backend :33-37, dashboard :76-80, framework-lint :214-218, xyne-claw :238-242, gitleaks :268, trivy :314) — the job adds no new pinning gap. Moving the whole repo to SHA pinning, or dismissing these alerts with written justification, is a code-owner policy call (PROC-C17); this thread should not block the PR. Recommend dismissing as repo-wide convention / tracked separately.

.github/workflows/CI-pr.yml
ameer030: Verified fixed at head `b0668e41`: `CI-pr.yml` runs `ci.yml@main` again (`b735eea8`) — the only remaining delta on that file is `secrets: inherit`, with the secrets scoped to the push step env as agreed in the thread. OK to resolve.

ameer030 (commented):
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 0 |
| 🟡 Better to fix | 1 |
| 🔵 Nit | 0 |

### 🔁 Re-check
- 💬 **Existing threads:** 2 fixed ✅ · 1 satisfied by reply · 1 still open — the duplicate CodeQL unpinned-tag alert on `ci.yml:131` (identical point was already resolved on the sibling alert; re-reviewed and left a dismissal recommendation for the code owners). @ameer030's B01 "branch-ref grading its own PR + secrets" is verified fixed at head (`b735eea8`, ref back on `@main`) — reply posted, thread safe to close.
- ✅ **Verified OK:** `handleRoomFinished` now clears `conversation.callId` only when it still points at the finished room — writer and clearer both use `call.externalId`, so the id spaces match and a stale `room_finished` can't wipe a live call's pointer. `04_calls-page.spec` swap is sound (`call-history-list` only renders when calls exist; the three replacement testids are always-rendered cards). Test-auth DM-project `createMany … skipDuplicates` is backed by `@@unique([code, workspaceId])` — genuinely race-safe. CI-pr reusable ref pinned to base branch `@main` per CI-C03; secrets scoped to the report-push step only. Full CI suite green.

### 📌 Verdict
No blockers — good to merge once the small `NO_PROXY` dedup above lands (or is consciously deferred); the one open CodeQL duplicate awaits a code-owner dismissal call, not code changes.

docker-compose.test.yml
sumantop: Done in `286509ad93` — one `x-test-stack-no-proxy` anchor at the top of `docker-compose.test.yml`, referenced as `NO_PROXY: *test-stack-no-proxy` / `no_proxy: *test-stack-no-proxy` in the four services (map form, as you suggested; the automation service's env list was converted to map form to allow it). Verified by diffing `docker compose config` output before/after: every service's rendered environment is byte-identical — the only difference is the `x-` key itself being echoed.

.github/workflows/ci.yml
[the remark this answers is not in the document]
sumantop: Agreed — the `test` job uses the same `@v4`/`@v5` tag refs as every other job in `ci.yml`, so this is the repo-wide convention rather than something this PR introduced. Leaving the SHA-pinning / dismissal decision to the code owners as you recommend; no code change from this PR.

tools/xyne-automation/scripts/build-summary-report.mjs
```diff
@@ -0,0 +1,120 @@
...
+  .counts{color:#656d76;margin-left:auto}.t{color:#8c959f;font-size:12px;margin-left:10px;font-variant-numeric:tabular-nums}
+  ul{list-style:none;margin:0 0 10px;padding:0 0 0 18px}li{padding:3px 0}li.failed{color:#a40e26}li.skipped{color:#9a6700}
+  .fail{background:#fff8f8;border:1px solid #ffcecb;border-radius:6px;padding:12px;margin:8px 0}.crumb{color:#656d76;margin-bottom:6px}.step{font-weight:600;color:#a40e26}
+  pre{white-space:pre-wrap;word-break:break-word;background:#f6f8fa;border-radius:6px;padding:10px;margin:8px 0;font-size:12px}.stack{max-height:280px;overflow:auto}
+  .shot{max-width:100%;border:1px solid #d0d7de;border-radius:6px;margin-top:6px}
+  details>summary{cursor:pointer}
+</style></head><body>
+<div class="hdr"><h1>Automation report</h1><span class="badge ${status}">${status.toUpperCase()}</span></div>
+<div class="muted">commit <code>${esc(meta.commitHash ?? '')}</code> · ${esc(result.timestamp)} · ${esc(meta.mode ?? '')} · tags <code>${esc(result.tags || '—')}</code></div>
```
ameer030, line 103: 🔵 **N01 — the report header always prints an empty "mode"**

`build-summary-report.mjs` reads `meta.mode` from `run-metadata.json`, but nothing ever writes that field to that file — so the header always renders `commit <hash> · <time> ·  · tags …` with a dangling empty segment.

- Read: `tools/xyne-automation/scripts/build-summary-report.mjs:14` (`run-metadata.json`) used at `:103` (`${esc(meta.mode ?? '')}`)
- Writer: `tools/xyne-automation/scripts/run-gauge.ts:356-373` (`initialRunMetadata`) — no `mode` key exists on `RunMetadata`
- The only place a `mode` is written is `tools/xyne-automation/scripts/runner/artifacts.ts:156`, and it goes to `runner-summary.json` — a different file the summary builder never reads

Cosmetic only: the report is otherwise correct, it just always shows the stray `·`.

</details>

Either drop the segment (`commit <code>…</code> · ${esc(result.timestamp)} · tags …`) or add `mode` to `initialRunMetadata` in `run-gauge.ts` so the header is truthful.

</details>

rules: STYLE-C14
agent: omp
run: paarvai-reviews/27

</details>

apps/backend/src/controllers/testAuthController.ts
```diff
@@ -399,6 +415,13 @@ export class TestAuthController {
           ...cookieOptions,
           maxAge: config.session.expiryDays * 24 * 60 * 60 * 1000,
         });
+
+        // Session cookie the auth middleware and session-scoped routes require
+        // (real login controllers set it too); without it session-gated routes 401.
+        res.cookie('user_session_id', sessionId, {
```
ameer030, line 422: 🔵 **N02 — test logout never clears the session cookie this PR adds**

Login now sets `user_session_id`, but `testLogout` still only clears `xyne_ws_*_token`, `xyne_last_workspace`, `xyne_session` and `is_new_user`. The real logout clears it too, so a test that logs out stays half-logged-in on session-scoped routes.

- Cookie set here: `apps/backend/src/controllers/testAuthController.ts:421`
- `testLogout` (same file, ~L467-481) clears the other four cookies but not `user_session_id`
- Real logout parity: `apps/backend/src/controllers/authV2Controller.ts:1477` — `res.clearCookie('user_session_id', { path: '/' })`

Automation mostly reuses fresh browser contexts, so impact is minor — it's just asymmetric with the cookie this same PR introduces.

</details>

Add one line to `testLogout`:

```ts
res.clearCookie('user_session_id', { path: '/' });
```

</details>

rules: BUG-C28
agent: omp
run: paarvai-reviews/27

</details>

docker-compose.test.yml
ameer030: Verified fixed at head `286509ad93`: `docker-compose.test.yml` now declares one `x-test-stack-no-proxy` anchor at the top and every service references `NO_PROXY: *test-stack-no-proxy` / `no_proxy: *test-stack-no-proxy`. The duplication is gone — nothing outstanding on this thread.

ameer030 (commented):
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 0 |
| 🟡 Better to fix | 0 |
| 🔵 Nit | 2 |

### 🔁 Re-check
- 💬 **Existing threads:** 5 fixed ✅ · 1 satisfied by reply · 0 still open — the CodeQL unpinned-tag alerts (`ci.yml` action refs) were settled as repo-wide convention with the pin/dismiss decision left to code owners; **B01** (`@main` ref + secrets), **B02** (PR body), prior **N01** (nix examples) and **W01** (`NO_PROXY` anchor) all verified fixed at `286509ad93`.
- ✅ **Verified OK:** the `conversation.callId` clear is now race-guarded in the UPDATE's WHERE clause with the same `call.externalId`/roomName id space the writers use (BUG-C01 satisfied, in-transaction, failure caught + logged) · `user_session_id` matches real login cookies and `/api/test` stays test-env-gated · DM-project `createMany…skipDuplicates` is genuinely atomic via the `@@unique([code,workspaceId])`/`[name,workspaceId]` constraints · LiveKit cred rename is complete (only the intentional `.invalid('devkey')` boot guard remains) · CI report-upload contract lines up end to end (`-runner` layout, probe paths, `/api/apps/files/filesUpload`) · every testid/track-name the updated specs click exists in the dashboard.

### 📌 Verdict
No blockers — good to merge after the two small nits above ✅

rules: STYLE-C14, BUG-C28, BUG-C01, SEC-C20, CI-C03, CI-C12, TEST-C13
agent: omp
run: paarvai-reviews/27

</details>

ameer030 (dismissed):

apps/backend/src/controllers/testAuthController.ts
Mohan0544: @sumantop if this is the case, clear session id as well.

Mohan0544: @sumantop is this comment closed?

sumantop:
`538697d3cb` — test-auth logout was incomplete: #1565 taught test-login to set the canonical `user_session_id` cookie, but test-logout never cleared it, nor revoked the session row / encryption key the way `authV2Controller` logout does. Result in test mode: after logout the browser still holds a valid session and every session-authenticated route keeps serving the previous user until the next login overwrites the cookie (found by hand — the suite hid it because each scenario starts a fresh browser context). Now mirrors the real logout. Verified against the test-mode backend: login → `/api/auth/me` 200 → logout → cookie cleared, replaying the old cookie → 401, session row `REVOKED`. Test-mode only (`/api/test/*`).

Mohan0544 (approved):

ameer030 (approved):

tools/xyne-automation/scripts/build-summary-report.mjs
ameer030: Re-checked at head `538697d3` — **this is still open in code**, though the thread is marked resolved.

- `tools/xyne-automation/scripts/build-summary-report.mjs:103` still renders `${esc(meta.mode ?? '')}` from `run-metadata.json`
- The only writer of that file is `run-gauge.ts` (`RunMetadata` RunMetadata interface, `writeRunMetadata`), and it still has **no `mode` field** — so the header keeps printing `commit <hash> · <time> ·  · tags …` with the dangling empty segment
- One nuance from the re-check: a `mode` value *is* written, but to `runner-summary.json` (`tools/xyne-automation/scripts/runner/artifacts.ts`), which the report never reads

Minimal fix: read the mode from `runner-summary.json` instead of `run-metadata.json` (one line), or drop the segment from the header. Leaving this thread flagged as unaddressed; if the empty segment is acceptable as-is, feel free to resolve with a note saying so.

apps/backend/src/controllers/testAuthController.ts
ameer030: @Mohan0544 Yes — this is closed in code at head `538697d3`.

`testLogout` (`apps/backend/src/controllers/testAuthController.ts:467-494`) now:

- revokes the session row (`this.userSessionService.revokeSession`, :476)
- revokes the session's encryption key (`getEncryptionProvider().revokeSessionKey(sessionId)`, :479)
- clears the `user_session_id` cookie (`res.clearCookie`, :482)

That mirrors the real logout in `authV2Controller.ts:1465-1492`, and both revokes are warn-and-continue, so a revoke hiccup can't fail the test logout itself. Verified against the head checkout.

ameer030 (commented):
## 📝 Xyne PR Review — Summary

Re-review at head `538697d3`. This pass found **no new issues**; it focused on verifying the fixes and re-checking the existing threads against the current code.

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 0 |
| 🟡 Better to fix | 0 |
| 🔵 Nit | 0 |

### 🔁 Re-check
- 💬 **Existing threads:** 3 fixed ✅ (CI ref back to `@main`, NO_PROXY anchor dedup, test logout session handling) · 1 satisfied by reply (CodeQL tag-pin alerts — repo-wide `@v4`/`@v5` convention; dismiss-or-pin-all is a code-owner policy call) · **1 still open** — the report-header empty `mode` nit: `build-summary-report.mjs` reads it from `run-metadata.json` where nothing writes it (it's actually written to `runner-summary.json`) — posted a follow-up on that thread
- ✅ **Verified OK:** `conversation.callId` guard matches the `roomName` domain exactly (set at call start, cleared only by the finishing room) · DM-project `createMany … skipDuplicates` is backed by both `@@unique` constraints and can't race or duplicate · LiveKit `localdevkey` rename is complete in every consumer (only the intentional Joi `invalid('devkey')` boot guards mention the old value) · `/api/test/*` stays gated to test/dev envs and the new `user_session_id` cookie mirrors the real login's flags · every changed spec selector exists in the app (`recordings-v2-page`, call options, `composer-link-popover`, `APPLY_LINK`, `nav-chat`, `end-call-button`) · `VITE_ZERO_PATH` has both a consumer and a preview proxy · CI callers pin `ci.yml@main` and the report push fails open-safely on fork PRs (no secrets → clean skip)
- 📝 Note for awareness (not a finding): after merge, `semantic-release` (`needs: ci`, pre-existing) will gate on the ~17-min automation suite on every main push.

### 📌 Verdict
Changes requested — no new findings, but one existing thread is still unaddressed in code (cosmetic `mode` field in the CI report header); see the follow-up on that thread ❌

rules: BUG-C01, BUG-C34, SEC-C20, SEC-C30, CI-C03, CI-C06, CI-C17, TEST-C13, TEST-C14, PROC-C17
agent: omp
run: 01a0a9c4-f697-71e4-914b-9ad6ebe6b917
</details>

PR #1780, opened 2026-09-11
feat: XYNE-63275 Hub Knowledge and Wiki as SDLC hub workflows (claw)

ameer030:
Claw half of XYNE-63275: SDLC **Hub Knowledge** and the **Wiki** as hub workflows. The backend and dashboard half is #1781 against `main`.

> **Deploy with #1781.** The Wiki page routes and the SDLC agent run context changed on both sides, so neither half works against the other's old version.

## What changes

| Area | Change |
|---|---|
| Wiki tools | The Wiki actions of `spaces-sdlc-mutate-artifact` and the Wiki listing of `spaces-sdlc-list-artifacts` call the backend's folder-tree page routes (`/api/internal/sdlc/wiki/pages/write` and `/list`). A run pinned to one repository writes only that repository's Wiki. |
| Old wiki runtime | Removed: the git context, begin checkpoint, verify sources and finalize commit tools, the wiki policy, `wiki_execution` bindings and the wiki webhook roles. |
| Run context | `sdlc-baseline-run-context` becomes `sdlc-agent-run-context`: interactive runs only, hub-scoped or pinned. Baseline draft injection and execution-bound (CLONE / PUSH) credentials are gone, so a run clones on its interactive grant. |
| Hub Knowledge | claw-auth fetches the hub's Hub Knowledge at SDLC agent run start (5 s timeout) and prepends it to the run's context. |
| `packages/shared/src/sdlc.ts` | Mirrors main's `sdlcAgentContextSchema` and its interactive-only `SDLC_AGENT_OPERATIONS`, which the run context parses. This is the only non-claw file here. |

## Testing

- `tsc --noEmit` clean on `apps/xyne-claw-auth/backend` and `packages/xyne-claw-shared`.
- `vitest`: `sdlc-agent-run-context.test.ts` (2 passing) in claw-auth, and the SDLC credential bootstrap and repository context tests in `xyne-claw-shared` (9 passing).
- Pre-commit hooks passed.

<details>
<summary>Debug</summary>

- Claude Code `session_014y72w6JamZwSNwsTa9hpJz` — `eba6e40c4` — split the claw slice out of the SDLC branch and rebased it onto feature/deploy-xyneclaw.
- Claude Code `325277dd-27ba-4ea0-b3b8-d90e34618aeb` — `2d671b69a` — synced the claw half of Hub Knowledge and the Wiki on hub workflows, and mirrored the shared agent context schema.

</details>

PR #1781, opened 2026-09-11
feat: XYNE-63275 Hub Knowledge and Wiki as SDLC hub workflows (non-claw)

ameer030:
Moves SDLC **Hub Knowledge** (formerly Repo Knowledge) and the **Wiki** onto seeded workflows-v2 hub workflows, and removes the SDLC worker.

> **Ships with #1780** (claw, into `feature/deploy-xyneclaw`). Deploy both together: the Wiki page routes and the SDLC agent run context changed on both sides.
>
> **Deploy prerequisites:** `ENABLE_WORKFLOWS_WORKER=true` wherever runs execute, or runs stay queued. After deploy, call `POST /api/sdlc/admin/backfill-workflows` so existing hubs get both workflows and the Wiki folder tree.

### Hub workflows

Every hub is seeded with two workflows, filed under `SDLC / <hub>` in its Workflows section.

| Workflow | Steps |
|---|---|
| Generate Hub Knowledge | Four `CREATE_SDLC_ARTIFACT` steps: System Overview, Code Map, Conventions, Run and Test. Each document covers every repository in the hub. |
| Generate Wiki | `SDLC_WIKI_PLAN` splits each repository's new first-parent commits into windows. A Generator and a Corrector (`SDLC_AGENT`) run per window. The Hub Wiki steps, shown as **Relationships**, run only once the hub has two repositories. |

- Wiki windows resume from run history (the resolved input of completed steps), so no extra state is stored.
- Hub Knowledge reaches every SDLC agent run through claw-auth run start, from `POST /api/internal/sdlc/agent/hub-knowledge`.
- `POST /api/sdlc/admin/reset-workflows` rewrites both configs. `POST /api/sdlc/admin/cleanup-legacy` (migration admin, dry run by default) removes what the old implementation left behind.

### Storage

- Wiki pages and Hub Knowledge documents are placed in a hub folder tree: `sdlc_folders` plus `HUB_ITEM` (parent to child) and `HUB_ITEM_SECONDARY` (scope folder to everything under it) edges. Clients cannot write these edges through Zero.
- `repos.sdlcSetupExecutionId` stays in the database, marked `@deprecated`, with no migration. Nothing writes it.

### Removed

- The SDLC worker, `sdlcQueue`, admission and the reconcile timer, with `ENABLE_SDLC_WORKER` and the five `SDLC_*` limits.
- The WORK and BASELINE paths, the old claw callback, and the old wiki runtime with its internal and public routes.

### Dashboard

- Wiki opens on a repository grid, then one repository's page tree (or Relationships), with breadcrumbs down to folders.
- Wiki and Hub Knowledge headers show a status icon and Start / Cancel.
- Sidebar: the Hub rows keep their height, folded sections sit at the bottom, and managing repositories moved to the Wiki grid's Add repository card.
- SDLC canvases drop the page-title block. Labels sit above the document's own title and show on hover.

### Safety

- SDLC agent steps act as the workflow's `sdlcAuthorUserId`, which only the server sets: whoever last created the workflow or changed its steps. A rerun with changed steps is author-only. Seed and reset pick a member of the hub.
- Only hub admins create Hub Knowledge documents, one per title. The Zero mutator grants Hub Knowledge edit rights only inside SDLC hubs.
- The Plan step checks that the hub belongs to the run's workspace.
- Claw joins the agent's last text messages into one reply, so a JSON-output step takes the JSON at the end of it instead of re-running the whole task.
- The SDLC steps register through an `sdlc` connector, as the workflow SDK on `main` requires.

## Testing

- `tsc --noEmit` clean on `apps/backend` and `apps/dashboard` after merging `main` (1.322.0).
- Pre-commit hooks passed on the merge commit: gitleaks, tenant-key guard, backend typecheck and build, dashboard lint and validate, schema-migration validation, Zero column parity, enum guard.
- Local runs against a one-repository hub: Generate Wiki exposed claw's joined replies, fixed here, and the fix accepts all five replies that were rejected. Neither workflow has been re-run end to end since.
- Not yet exercised through the UI: author stamping on edit or rerun, and admin-only Hub Knowledge creation.

<details>
<summary>Debug</summary>

- Claude Code `session_014y72w6JamZwSNwsTa9hpJz` — `adf0699c4` — reviewed the branch, fixed the deleted wiki debug route, the overlapping hub runs, the over-broad reset error handling and the folder-name ACL scope, then split the work into claw and non-claw PRs and rebased this half onto main.
- Claude Code `325277dd-27ba-4ea0-b3b8-d90e34618aeb` — `cef577478` — moved Hub Knowledge and the Wiki onto hub workflows, removed the SDLC worker, fixed the review findings and merged main.

</details>


apps/backend/src/zero/mutators.ts
```diff
@@ -9971,10 +9971,10 @@ export function createMutators(
...
             guestSharedRole === CanvasRole.EDITOR ||
             guestSharedRole === CanvasRole.OWNER;
-          const sdlcArtifact = await tx.run(
-            zql.sdlc_artifacts.where('artifactId', params.id).one(),
-          );
-          const isSdlcBaseline = isBaselineCanvasType(sdlcArtifact?.artifactType);
+          const canvasFolder = canvas.folderId
+            ? await tx.run(zql.canvas_folders.where('id', canvas.folderId).one())
+            : null;
+          const isSdlcBaseline = canvasFolder?.name === SDLC_REPO_KNOWLEDGE_FOLDER;
```
ameer030: 🔴 **B01 — folder-name edit gate lost its SDLC scope on the server mirror**

The rule "a hub admin may edit Repo Knowledge canvases they are not a participant on" used to be keyed on the SDLC artifact type. This PR re-derives it three times — and the three copies no longer agree:

- Here (the **authoritative server-side Zero mutator**): folder name only.
- `packages/shared/src/zero/mutators.ts:5962` (the client mirror of this exact mutator): folder name **and** `channel.type === ChannelType.SDLC`.
- `apps/backend/src/services/canvasAuthService.ts:233-242` (REST): folder name **and** SDLC channel type.

Since `SDLC_REPO_KNOWLEDGE_FOLDER` is the plain string `"Baseline"` — a folder name any user can set in any channel type — a channel ADMIN of a non-SDLC channel gains edit rights over every canvas in a folder that happens to be called "Baseline", even on canvases they are not a participant on. The client copy would reject the optimistic write, but a push sent straight to the Zero server is authorized by this (server) copy, which is the permissive one.

- Server mutator (authorizes all Zero pushes): `apps/backend/src/zero/mutators.ts:9974-9977`
- Client mirror (strict): `packages/shared/src/zero/mutators.ts:5960-5963`
- REST mirror (strict): `apps/backend/src/services/canvasAuthService.ts:233-242`
- Pre-PR, all of these gate on `isBaselineCanvasType(artifactType)`, which cannot be forged by naming a folder.
- No schema/uuid reservation on folder names: `canvasFolder.name` is free user input.
- Minor: this hunk also fetches the same `canvas_folders` row twice (`currentFolder` at ~:9951 and `canvasFolder` at ~:9975).

</details>

Mirror the client copy here: reuse the already-fetched `currentFolder`, fetch the channel once, and require `ChannelType.SDLC`, e.g.

```ts
const currentChannel = currentChannelId
  ? await tx.run(zql.channels.where('id', currentChannelId).one())
  : null;
const isSdlcBaseline =
  currentFolder?.name === SDLC_REPO_KNOWLEDGE_FOLDER &&
  currentChannel?.type === ChannelType.SDLC;
```

keep the two mirrored mutators byte-equivalent in predicate terms, and drop the duplicate folder fetch.

</details>

rules: SEC-C05, SEC-C06, ARCH-C10, BUG-C03
agent: omp
run: 2026-09-15T10-29-48-171Z_01a0a49d-9e0b-72ba-b793-b4b8153de644
</details>

apps/backend/src/app.ts
```diff
@@ -637,6 +639,8 @@ export class App {
       );
       res.json({ encryptedFields });
     });
+
+    this.app.use('/api/internal/sdlc/agent', validateS2SKey, sdlcAgentInternalRoutes);
```
ameer030: 🟡 **W01 — sdlcAgentInternalRoutes is mounted twice on the same path**

`app.ts:629` and this line both register `this.app.use('/api/internal/sdlc/agent', validateS2SKey, sdlcAgentInternalRoutes);`. Express will match the first, so the second is dead — but on a credential-bearing S2S route, having two identical mounts reads like an unresolved merge conflict and confuses route audits.

Keep exactly one of the two mounts (and double-check the neighboring `/api/internal/sdlc/wiki` + `/api/internal/sdlc/artifact-versions` mounts aren't accidentally reordered by the same edit).

</details>

rules: BUG-C28
agent: omp
run: 2026-09-15T10-29-48-171Z_01a0a49d-9e0b-72ba-b793-b4b8153de644
</details>

apps/backend/src/sdlc/seedHubWorkflow.ts
```diff
@@ -0,0 +1,249 @@
...
+    {
+      name: WORKFLOW_NAME,
+      config: buildRepoKnowledgeConfig({ channelId, repoIds }),
+      folderId,
+      metadata: { [SDLC_AUTHOR_METADATA_KEY]: actor.userId },
+      attributes: { workspaceId: actor.workspaceId, createdByUserId: actor.userId },
+    },
+  ).then((record) => record.id);
+
+  await db.sdlcEntityLink.create({
```
ameer030: 🟡 **W02 — seed is find-then-create; the losing concurrent seed leaves an orphaned workflow**

`seedHubWorkflow` checks for the marking edge (`:133`), creates the workflow (`:137`), then inserts the edge here (`:148`). The edge has a unique key (`sdlc_entity_links_channel_edge_key`, `schema.prisma:3358`), so two concurrent seeds of the same hub — e.g. `POST /api/sdlc/channels` racing `POST /api/sdlc/admin/backfill-workflows` — both pass the check, **both** create workflows, and the loser's link insert throws an unhandled unique-key error. The result is an orphan workflow row sitting in the hub's `SDLC / <hub>` folder, visible in the Workflows tab, never linked, and future seeds skip the hub because the winner's edge is already what's found.

Related: `routes/sdlc.ts:72-78` (`POST /channels`) catches any seed failure and only logs it — the channel is created and the caller gets a success response even though no workflow was seeded; the hub then sits in "No workflow configured" until someone runs the admin backfill.

Catch the unique violation on the edge insert (or `ensureLink`-style upsert) and return `{ status: 'skipped' }` — the other seed won, so nothing more to do. Consider surfacing the remaining genuine seed failure (workflow create error) back to the channel-create response as a soft warning so it doesn't need a log archaeology session.

</details>

rules: BUG-C01, BUG-C04
agent: omp
run: 2026-09-15T10-29-48-171Z_01a0a49d-9e0b-72ba-b793-b4b8153de644
</details>

apps/backend/src/sdlc/triggerHubKnowledgeRun.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (!link) return null;
+
+    const workflow = await db.workflow.findUnique({
+      where: { id: link.targetId },
+      select: { id: true },
+    });
+    if (!workflow) return null;
+
+    // Every hub repository fires this, and concurrent runs duplicate the same documents.
+    const inFlight = await db.workflowExecution.findFirst({
```
ameer030: 🟡 **W03 — run admission is check-then-act: concurrent wiki completions can double-fire a full-hub run**

Here the flow is `findFirst` an active run (`:22`) → `triggerManual` (`:33`). Every repository in a hub fires this hook when its Wiki run finishes; two Wiki completions landing in the same window both see "no in-flight run" and both start one. `triggerManual` in the SDK does no dedupe, and there is no active-run unique constraint on `workflow_executions`, so nothing downstream closes the race either. Two concurrent runs re-do a full hub of LLM work, and the step's own find-or-create-by-title instructions can emit duplicate documents — the exact failure the PR body already names ("concurrent find-or-create steps would duplicate documents"), so this window is known, just not yet closed.

Two extra notes: (a) `SDLC_ACTIVE_RUN_STATUSES` includes `PAUSED`, so a paused run currently blocks hub refresh indefinitely — worth a deliberate decision whether pause should hold the gate; (b) manual triggers from the Workflows UI/API bypass even this soft gate entirely.

Wrap check + trigger in a `hub-knowledge:<channelId>` lock from `apps/backend/src/utils/distributedLock.ts` (TTL comfortably above the check→enqueue span), which is the established pattern in this codebase for exactly this "serialize per entity" problem. Decide separately whether `PAUSED` belongs in the active set (skip-if-paused likely wants to trigger, or at least log differently).

</details>

rules: BUG-C01, BUG-C04
agent: omp
run: 2026-09-15T10-29-48-171Z_01a0a49d-9e0b-72ba-b793-b4b8153de644
</details>

apps/backend/src/workflowsV2/agents/sdlc-artifact-provider.ts
```diff
@@ -0,0 +1,234 @@
...
+  readonly name = 'SDLC Agent';
+  readonly configSchema = SdlcArtifactConfigSchema;
+
+  async dispatch(
+    stepConfig: SdlcArtifactConfig,
+    input: AgentRunInput,
+    ctx: StepExecutionContext,
+  ): Promise<{ externalRef: string }> {
+    const attrs = ctx.runtime.attributes as XyneResourceAttrs;
+    const actorUserId = readAuthorUserId(ctx.runtime.metadata) ?? attrs.createdByUserId;
```
ameer030: 🟡 **W04 — any workspace member can trigger/cancel the hub's workflow, and the run acts as the seed author**

Before this PR, the knowledge-run control (`POST /api/sdlc/repositories/:id/setup`) required hub ADMIN (`requireRepositoryRole(actor, repoId, true)`). The replacement control is the generic `POST /workflows-v2/workflows/:id/trigger` (+ `.../executions/:id/cancel`). The v2 `XyneWorkflowAuthorizer` grants full workflow actions, including `workflow:run`, to **any** same-workspace member (`sameWorkspace → WORKFLOW_ACTIONS`, `apps/backend/src/workflowsV2/authorizer.ts`), and seeded hub workflows carry only workspace attributes — so a workspace member who cannot even see the hub can start (or cancel) runs. The dashboard's `isAdmin` control hiding is client-side only.

The run then executes as the metadata-pinned seed author (here, `:95` — `readAuthorUserId` + the SDLC interactive grant issued for that user): identity, hub membership, and clone credentials derive from the author, not the triggerer. Service-account semantics per se are fine, but combined with the open trigger it means an outsider acting "as" a hub admin across hubs they have no membership in, with no in-flight dedupe on the manual path — real LLM/compute cost on top of the impersonation-shaped trail.

Gate run/cancel admission for SDLC-seeded workflows to hub members (the `sdlc-hub-<channelId>` folder-id prefix is deterministic, so the authorizer can recognize this class), or at minimum stamp the triggerer as the run's recorded starter and dedupe concurrent manual triggers per hub. If the current reach is deliberate, say so in the PR so reviewers sign off on it explicitly.

</details>

rules: SEC-C04, SEC-C05
agent: omp
run: 2026-09-15T10-29-48-171Z_01a0a49d-9e0b-72ba-b793-b4b8153de644
</details>

apps/backend/src/routes/sdlc.ts
```diff
@@ -350,8 +347,6 @@ router.delete(
   })
 );
 
-// One-off SDLC data migrations. Delete with src/sdlc/cleanup/ once every
-// environment has run them.
-router.use('/cleanup', sdlcCleanupRoutes);
+// router.use('/cleanup', sdlcCleanupRoutes);
```
ameer030, line 291: 🔵 **N01 — cleanup routes are commented out instead of removed**

This line is now `// router.use('/cleanup', sdlcCleanupRoutes);` with the import already gone. The whole `apps/backend/src/sdlc/cleanup/` directory is unmounted dead weight from this commit on, and its own header comment still claims it's "Mounted at /api/sdlc/cleanup". Parked-but-dead routes tend to get "resurrected" by later copy-paste; the old comment's own instruction ("Delete with src/sdlc/cleanup/") says this is the PR that should remove them.

Delete the commented-out `router.use` line and the `src/sdlc/cleanup/` directory in this PR.

</details>

rules: STYLE-C14, DOCS-C04
agent: omp
run: 2026-09-15T10-29-48-171Z_01a0a49d-9e0b-72ba-b793-b4b8153de644
</details>

apps/backend/src/sdlc/SdlcAgentContextService.ts
```diff
@@ -166,6 +176,84 @@ export class SdlcAgentContextService {
...
+      interactiveGrant: input.conversationId
+        ? issueSdlcInteractiveGrant(
+            {
+              agentSlug: SDLC_AGENT_SLUG,
+              workspaceId: actor.workspaceId,
+              repoIds,
+              actorUserId: actor.userId,
+              conversationId: input.conversationId,
+            },
+            process.env['INTERNAL_S2S_KEY'] || process.env['XYNE_CLAW_S2S_KEY'] || ''
```
ameer030: 🔵 **N02 — new `process.env` read for the signing key; config already exposes both**

This line reads `process.env['INTERNAL_S2S_KEY'] || process.env['XYNE_CLAW_S2S_KEY'] || ''` directly. Both keys are already in the central config module (`config.internalS2sKey`, `config.xyneClaw.s2sKey` in `env.ts`), and this new read duplicates the pre-existing one at `:158`. No security impact — `issueSdlcInteractiveGrant`'s `requireSecret` still throws on a blank key, so it fails closed — but the repo convention (and its config lint direction) is env-through-config, so the new copy should follow the config path rather than clone the legacy hack.

Use `config.internalS2sKey || config.xyneClaw.s2sKey` from `@/config/env` (keeping the same precedence), and ideally collapse the duplicate read at `:158` into it at the same time.

</details>

rules: CI-C06
agent: omp
run: 2026-09-15T10-29-48-171Z_01a0a49d-9e0b-72ba-b793-b4b8153de644
</details>

apps/dashboard/src/routes/SdlcScreen/SdlcScreen.tsx
```diff
@@ -1015,19 +995,34 @@ export default function SdlcScreen(): ReactElement {
...
+        items: { id: string; status: string; updatedAt?: string }[];
+      }>('/workflows-v2/executions', { params: { workflowId: hubWorkflowId, limit: 5 } });
+      return response.data.items.map(run => ({
+        id: run.id,
+        status: run.status,
+        updatedAt: run.updatedAt ? Date.parse(run.updatedAt) : null,
+      }));
+    },
+    enabled: Boolean(hubWorkflowId),
+    refetchInterval: 5 * 60_000,
```
ameer030: 🔵 **N03 — "Status reads from Zero, not REST — live, no polling" isn't what shipped**

The PR body says Repo Knowledge status now "reads from Zero, not REST — live, no polling." The hub→workflow link really does come from Zero. But run status is `useQuery` over `GET /workflows-v2/executions` with `refetchInterval: 5 * 60_000` here (`SdlcScreen.tsx:1016`) — i.e. REST, polled every five minutes. Users can sit up to 5 minutes on a stale "in progress"/"failed" view — a behavior claim this body asserts has gone away but has only been slowed down.

Either tighten the claim in the PR body ("status from REST, refreshed periodically"), or make the status honest: poll only while `state.phase === 'RUNNING'` (cheap, and matches the old behavior's intent).

</details>

rules: PROC-C02
agent: omp
run: 2026-09-15T10-29-48-171Z_01a0a49d-9e0b-72ba-b793-b4b8153de644
</details>

ameer030:
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 1 |
| 🟡 Better to fix | 4 |
| 🔵 Nit | 3 |

### 🔁 Re-check
- ✅ **Verified OK:** new admin endpoints (`backfill`/`reset-workflows`) are `authorize('SDLC', ADMIN)`-gated; agent slug is pinned in code (not builder config) per SEC-C33; the interactive grant keeps single-repo write authorisation while widened `repoIds` cover reads only; claw-artifact creation/update paths re-check hub membership and re-authorise every named repository server-side; mirror Zero queries for `getSdlcHubWorkflow` are byte-identical and the `sdlc_entity_links → workflow` relation ships in the Zero schema; tenants stay pinned on all new find/create paths (`workspaceId` + `channelId` scopes); all CI checks green.

### 📌 Verdict
1 blocker must be fixed before merge ❌ — the server-side Zero mutator gate (B01) must regain its SDLC channel-type scope to match its two mirrors; the rest are improvements and polish.

Excluded per review config: no-proof-of-testing finding, test-coverage finding.
agent: omp
run: 2026-09-15T10-29-48-171Z_01a0a49d-9e0b-72ba-b793-b4b8153de644
</details>

apps/backend/src/workflowsV2/router.ts
```diff
@@ -38,6 +40,37 @@
...
+  if (key === 'POST /workflows' || key === 'PUT /workflows/:id') {
+    const sent = body['metadata'];
+    const metadata = isPlainObject(sent) ? { ...sent } : {};
+    delete metadata[SDLC_AUTHOR_METADATA_KEY];
+    if (body['config']) metadata[SDLC_AUTHOR_METADATA_KEY] = ctx.userId;
+    if (sent !== undefined || body['config']) body['metadata'] = metadata;
+    return;
+  }
+
+  if (key === 'POST /executions/:execId/rerun' && body['configOverrides']) {
```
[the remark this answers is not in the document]
ameer030: Dismissed as a false positive. Without `configOverrides`, a rerun uses the workflow's stored steps, which any workspace member can already run through a trigger, so there is nothing to guard. The author check runs whenever a rerun changes the steps.

gyash1512 (approved):

Shailendra-03 (approved):

Mohan0544 (approved):

PR #1784, opened 2026-09-11
fix: XYNE-63312 KB-scoped Ask AI attachment + Vespa chunk access for public collections

PriyanshuC132:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
XYNE-63312

Two related KB/Ask-AI access fixes:

1. **Vespa chunk-level access for public KB collections with no scoping channel** — `kb-get-chunks`/`kb-search-within-doc`'s doc-fetch path only granted public-doc access via chat-channel participation. A KB collection created directly (not tied to a channel) never gets a `channelRef`, so the channel check could never pass and every non-owner/non-permissions user was permanently 403'd on collections their own Share dialog says are Public. Now grants access outright when the doc has no scoping channel, matching `collectionAccess.ts`'s `resolveCollectionAccess` (`isPrivate=false` already grants every workspace member implicit VIEWER with no channel gate).

2. **Attach scoped KB files/folders to Ask AI, with server-side dedup** — the dashboard's `toAttachedContext()` previously only built `channel`/`ticket`/`canvas`/`call`/`activity` items, so a file/folder/collection scoped from the composer's picker or the KB file viewer's Ask AI chip never reached the agent as structured context. Wired `fileScopes`/`folderScopes`/selected collections into the same `attached_context` shape the backend already resolves, added `folder` to the backend's accepted enum, fixed the KB file viewer's Ask AI button sending the row's `id` instead of `fileId` (backend now matches on either), and added server-side dedup so the backend's own folderIds/fileIds/collectionIds → attached_context conversion only fills gaps instead of double-appending what the dashboard already sent.

3. **Follow-up: drop the now-duplicate backend collection/folder resolution** — commit 2 above left both sides building the same collection/folder `attached_context` items: the client (via `toAttachedContext`) and the backend (via `collectionIds`/`folderIds` → DB lookup → dedup). Collections/folders don't need a server round-trip — the client already has the correct cuid + name — so this removes `collectionIds`/`folderIds` end-to-end (schema, controller resolution, dashboard hook/manager/worker plumbing) and unifies `composerContext.ts` (AIScreen) onto the same client-built `attachedContext` path `XyneAISidebar` already used. **Files are kept as a backend round-trip on purpose**: `fileScopes` can hold either the file's stable UUID (composer's own picker) or its cuid (KB file viewer's Ask AI chip), and only the backend can resolve which — so files still ride as bare `fileIds`, with a new `displayAttachedContext` override carrying the client-known title for the optimistic message pill only.

## Areas Touched

- [x] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

## Schema & Data Changes

## Config, Environment & URLs

## API Contract

- [ ] No API changes
- [ ] New endpoint

`attached_context[].type` enum gains `'folder'` (backward-compatible addition); auth check for public Vespa chunk access is now more permissive for channel-less public collections.

---

## How did you test it?
Verified the two cherry-picked commits apply cleanly against current `main` with no functional diff introduced during the port (only commit-message ticket IDs were rewritten). For the follow-up refactor commit, ran `pnpm run build:shared` + `apps/backend` `typecheck` + `apps/dashboard` `typecheck`/`eslint` on the touched files — clean (pre-existing, unrelated errors only, in files this PR doesn't touch: stale Prisma client / unbuilt `workflow-sdk` package).

### Automated

- [x] Not covered by tests 

### Manual

**Users**
- [ ] Single user
- [x] Different roles or permission levels (member vs admin, ACL boundaries)

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [x] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist

- [x] Backend `typecheck` passes (build not run)
- [x] Dashboard `lint:errors-only` (0 errors) + `typecheck` pass (build not run)

---

https://claude.ai/code/session_01DiC9UGa1PzG1GuQhyHEZE2

### DIFF
--- a/apps/backend/src/controllers/xyneAIControllerV2.ts
+++ b/apps/backend/src/controllers/xyneAIControllerV2.ts
@@ -68,7 +68,7 @@ const SelectionContextSchema = z
 // `collection` and `file` are appended below from top-level `collection_ids`
 // and `file_ids` so the dashboard can keep its existing payload shape.
 const AttachedContextItemSchema = z.object({
-  type: z.enum(['channel', 'ticket', 'canvas', 'call', 'activity', 'collection', 'file']),
+  type: z.enum(['channel', 'ticket', 'canvas', 'call', 'activity', 'collection', 'file', 'folder']),
   id: z.string().min(1),
   title: z.string().min(1),
   threadId: z.string().optional(),
@@ -581,20 +581,40 @@ export class XyneAIControllerV2 {
         }
       }
       if (effectiveFileIds && effectiveFileIds.length > 0) {
-        // The dashboard sends the stable `fileId` UUID, but the agent's
-        // kb-read-file expects CollectionItem.id (cuid). Resolve UUIDs to
-        // latest-version row ids in a single query.
+        // The dashboard SHOULD send the stable `fileId` UUID here, but not
+        // every picker does: the KB file viewer's own "Ask AI" button
+        // (FileViewerLayout.tsx, KnowledgeBaseV2Screen.tsx) sends
+        // CollectionItem.id (cuid) instead — that data simply isn't exposed
+        // as `fileId` at those call sites today. Rather than chase every
+        // caller, accept either id shape here: a cuid IS already the row id
+        // kb-read-file expects, so matching on `id` OR `fileId` (latest
+        // version only) resolves both conventions in one query.
         const items = await db.collectionItem.findMany({
-          where: { fileId: { in: effectiveFileIds }, isLatest: true, deletedAt: null },
+          where: {
+            OR: [{ fileId: { in: effectiveFileIds } }, { id: { in: effectiveFileIds } }],
+            isLatest: true,
+            deletedAt: null,
+          },
           select: { id: true, name: true },
         });
         for (const it of items) {
           kbAttachedContextItems.push({ type: 'file', id: it.id, title: it.name });
         }
       }
+      // The client may already have pre-built some of these (e.g. the
+      // dashboard's own toAttachedContext for file/folder/collection scopes)
+      // — kbAttachedContextItems is meant to FILL GAPS for callers that only
+      // send bare id arrays, not duplicate what's already there. Skip any
+      // (type, id) pair the client's own attached_context already covers.
+      const existingContextKeys = new Set(
+        (effectiveAttachedContext ?? []).map((item) => `${item.type}:${item.id}`),
+      );
+      const newKbAttachedContextItems = kbAttachedContextItems.filter(
+        (item) => !existingContextKeys.has(`${item.type}:${item.id}`),
+      );
       const mergedAttachedContext = [
         ...(effectiveAttachedContext ?? []),
-        ...kbAttachedContextItems,
+        ...newKbAttachedContextItems,
       ];
 
       try {
--- a/apps/backend/src/controllers/xyneAIControllerV2.ts
+++ b/apps/backend/src/controllers/xyneAIControllerV2.ts
@@ -65,8 +65,9 @@ const SelectionContextSchema = z
   );
 
 // Attached context item schema - for Add Context feature.
-// `collection` and `file` are appended below from top-level `collection_ids`
-// and `file_ids` so the dashboard can keep its existing payload shape.
+// `file` items are appended below, resolved from top-level `file_ids`, since
+// that resolution needs a DB round-trip the client can't do itself.
+// `collection`/`folder` items arrive as ordinary entries in this list.
 const AttachedContextItemSchema = z.object({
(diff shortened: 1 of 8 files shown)


apps/backend/src/controllers/xyneAIControllerV2.ts
```diff
@@ -542,59 +530,54 @@ export class XyneAIControllerV2 {
...
         for (const it of items) {
           kbAttachedContextItems.push({ type: 'file', id: it.id, title: it.name });
         }
       }
+      // A caller could in principle already have a matching 'file' entry in
+      // its own attached_context — kbAttachedContextItems is meant to FILL
+      // GAPS, not duplicate what's already there. Skip any (type, id) pair
+      // the client's own attached_context already covers.
+      const existingContextKeys = new Set(
+        (effectiveAttachedContext ?? []).map((item) => `${item.type}:${item.id}`),
```
VishalM-23: Can we use a type here?

apps/backend/src/controllers/xyneAIControllerV2.ts
```diff
@@ -542,59 +530,54 @@ export class XyneAIControllerV2 {
...
+        // (FileViewerLayout.tsx, KnowledgeBaseV2Screen.tsx) sends
+        // CollectionItem.id (cuid) instead — that data simply isn't exposed
+        // as `fileId` at those call sites today. Rather than chase every
+        // caller, accept either id shape here: a cuid IS already the row id
+        // kb-read-file expects, so matching on `id` OR `fileId` (latest
+        // version only) resolves both conventions in one query.
         const items = await db.collectionItem.findMany({
-          where: { fileId: { in: effectiveFileIds }, isLatest: true, deletedAt: null },
+          where: {
+            OR: [{ fileId: { in: effectiveFileIds } }, { id: { in: effectiveFileIds } }],
```
VishalM-23: Why do we need this?

apps/dashboard/src/components/Chat/XyneAISidebar/XyneAISidebar.tsx
```diff
@@ -799,13 +816,12 @@ const XyneAISidebar = ({
...
     canvasIds: selectedCanvases.map(c => c.id),
     callIds: [...selectedTranscripts.map(t => t.id), ...selectedRecordings.map(r => r.id)],
-    attachedContext: toAttachedContext({
-      channels: selectedChannels,
-      tickets: selectedTickets,
-      canvases: selectedCanvases,
-      transcripts: selectedTranscripts,
-      recordings: selectedRecordings,
-    }),
+    attachedContext: toAttachedContext(kbContextSelections),
```
VishalM-23, line 934: Why does it not have files: fileScopes ?

apps/backend/src/controllers/xyneAIControllerV2.ts
PriyanshuC132: Fixed in b2d96c7e4 — added an `AttachedContextItem` type (inferred from `AttachedContextItemSchema`) and used it on the `item` param instead of implicit `any`.

apps/backend/src/controllers/xyneAIControllerV2.ts
PriyanshuC132: The dashboard's KB file viewer "Ask AI" button (FileViewerLayout.tsx / KnowledgeBaseV2Screen.tsx) sends CollectionItem.id (cuid) here instead of the stable fileId UUID — that data isn't exposed as `fileId` at those call sites today. Rather than fix every caller, this accepts either id shape: a cuid is already the row id kb-read-file expects, so matching on `id` OR `fileId` (latest version only) resolves both conventions in one query without a client-side change.

apps/dashboard/src/components/Chat/XyneAISidebar/XyneAISidebar.tsx
PriyanshuC132: It's still sent — just not folded into attachedContext. A few lines up, `fileIds: fileScopes.map(f => f.id)` is passed as its own param (see comment at line ~794): the backend resolves the fileId(UUID)-vs-id(cuid) ambiguity server-side in xyneAIControllerV2.ts, which the client can't do without a DB round-trip. `displayAttachedContext` below does include `files: fileScopes` for the optimistic message pill.

VishalM-23 (approved):

mamtha-v (approved):

PR #1792, opened 2026-09-11
fix: unbind Redis cron on cancel/delete/pause even when bullSchedulerId is unset

XyneSpaces:
## Problem
When a scheduled cron job is cancelled, deleted, or paused, the handlers in `apps/xyne-claw-auth/backend/src/routes/scheduled-jobs.ts` only unbound the BullMQ job-scheduler from Redis when `row.bullSchedulerId` was set, and the cancel/delete paths swallowed `removeJobScheduler` errors with `.catch(() => {})`.

If `bullSchedulerId` was never persisted (e.g. the post-create `update({ bullSchedulerId })` write failed) or the Redis removal errored, the DB row flipped to `cancelled`/`paused` while the Redis cron entry kept firing — an orphaned schedule that keeps running the stale prompt, with no log. This is the root cause behind the credit-data-doctor "two prompts fired but the UI shows one schedule" incident (an old schedule that appeared cancelled kept firing alongside the new one).

## Fix
In the cancel, delete, and pause handlers:
- Fall back to the deterministic `cron-${row.id}` schedulerId when `row.bullSchedulerId` is unset — the same fallback the PATCH/reschedule and resume paths already use.
- Replace the silent `.catch(() => {})` with `log.warn`, so a failed Redis unbind is observable instead of masquerading as a clean cancel.

Behavior is unchanged for the common case (bullSchedulerId set, removal succeeds); this only closes the orphan path.

## Scope
- One file changed: `apps/xyne-claw-auth/backend/src/routes/scheduled-jobs.ts` (+30/-7).
- No schema, contract, or API changes.

## Validation
- Confirmed the update/resume paths already used the `?? cron-${row.id}` fallback; this brings cancel/delete/pause in line.
- Manual follow-up recommended: reconcile existing Redis job-schedulers on the `agent-scheduled-jobs` queue against `active` ScheduledJob rows to clean up any already-orphaned crons (pre-existing ones this fix does not retroactively remove).


apps/xyne-claw-auth/backend/src/routes/scheduled-jobs.ts
```diff
@@ -952,8 +956,18 @@ router.post(
...
       await cancelJob(row.bullJobId).catch(() => {});
     }
-    if (row.type === "cron" && row.bullSchedulerId) {
-      await cancelCronJob(row.bullSchedulerId).catch(() => {});
+    if (row.type === "cron") {
+      // Fall back to the deterministic schedulerId when bullSchedulerId was
+      // never persisted, and surface Redis failures instead of swallowing
+      // them — a silent no-op here left the cron firing while the row flipped
+      // to "cancelled" (the orphaned-schedule incident).
+      const schedulerId = row.bullSchedulerId ?? `cron-${row.id}`;
```
XyneSpaces, line 964: **[major 🟠] R7** — The new behavior here — the `bullSchedulerId ?? cron-${row.id}` fallback plus warn-and-continue-on-failure — ships with no test. This is exactly the regression class this PR fixes (an orphaned cron left behind by a silent failure), and this service has precedent for route-level tests (#266 `require-auth.test.ts`, #344 `tools.test.ts`).

Suggested: extract the unbind into a helper and cover: (a) fallback id used when `bullSchedulerId` is null, (b) persisted id used when set, (c) removal failure logs a `warn` while the request still succeeds, (d) `once` rows skip the cron unbind.

apps/xyne-claw-auth/backend/src/routes/scheduled-jobs.ts
```diff
@@ -952,8 +956,18 @@ router.post(
     if (row.type === "once" && row.bullJobId) {
       await cancelJob(row.bullJobId).catch(() => {});
```
XyneSpaces, line 957: **[minor 🟡] R6** — One `if` above the fixed cron path, the `once` path still swallows failures: `await cancelJob(row.bullJobId).catch(() => {})` — here, and again at line 1064 in the delete route. The cron path now logs removal failures (good), but a failed Redis cancel for one-shot jobs remains invisible.

Suggested: give the `once` path the same `.catch((err) => log.warn(...))` treatment so both paths are observable.

apps/xyne-claw-auth/backend/src/routes/scheduled-jobs.ts
```diff
@@ -825,10 +825,14 @@ router.post(
         );
       });
     }
-    if (row.type === "cron" && row.bullSchedulerId) {
-      await cancelCronJob(row.bullSchedulerId).catch((err) => {
+    if (row.type === "cron") {
+      // Fall back to the deterministic schedulerId when bullSchedulerId was
+      // never persisted (e.g. a create-time write failure); otherwise the
+      // Redis cron keeps firing after the row is paused.
+      const schedulerId = row.bullSchedulerId ?? `cron-${row.id}`;
```
XyneSpaces, line 832: **[minor 🟡] R5** — The fallback + catch/warn block is copy-pasted three times in this diff (here at :832, :964, :1070), and per the PR description PATCH/resume already carry the same pattern — roughly five sites that can drift independently; the next edit that gets one of them wrong re-creates the orphan.

Suggested: extract a shared `removeCronScheduler(row, action)` helper and call it from pause/cancel/delete/resume.

apps/xyne-claw-auth/backend/src/routes/scheduled-jobs.ts
```diff
@@ -1049,8 +1063,17 @@ router.delete("/:id", asyncHandler(async (req: Request<{ id: string }>, res: Res
   if (row.type === "once" && row.bullJobId) {
     await cancelJob(row.bullJobId).catch(() => {});
   }
-  if (row.type === "cron" && row.bullSchedulerId) {
-    await cancelCronJob(row.bullSchedulerId).catch(() => {});
+  if (row.type === "cron") {
+    // Fall back to the deterministic schedulerId when bullSchedulerId was
+    // never persisted, and log Redis failures — otherwise deleting the row
+    // leaves an orphaned cron that keeps firing the (now stale) prompt.
+    const schedulerId = row.bullSchedulerId ?? `cron-${row.id}`;
```
XyneSpaces, line 1070: **[minor 🟡] R2 (assumption to confirm)** — This removes exactly one scheduler id: `bullSchedulerId` when set, else the `cron-${row.id}` fallback — never both. If a job can ever be live under **both** ids (registered under the deterministic id before the post-create `update()` persisted `bullSchedulerId`, then re-registered under the persisted id on resume), one stays in Redis and the orphan survives this fix. That double-registration depends on code outside this diff — I could not verify the create path's registration id from the patch alone.

Suggested (safe either way): remove both ids deduped (a `Set` of `row.bullSchedulerId` and the `cron-${row.id}` fallback) — removing a non-existent id only produces the warn above. If create can only ever register under one id, close this with that confirmation.

XyneSpaces (commented):
Automated review of 9727e7b — 1 file, 3 hunks: **0 blocking / 1 major / 3 minor**. Findings inline; full summary + reviewer recommendation in the top-level comment. cc @anuragdvd

XyneSpaces:
## Xyne Spaces Reviewer — PR #1792

**Verdict:** Approve-worthy after addressing findings — **no blocking issues**. The fix direction is correct for the incident: an unset `bullSchedulerId` no longer skips the Redis unbind (deterministic `cron-${row.id}` fallback), and removal failures are now logged instead of silently swallowed.
**Reviewed:** `apps/xyne-claw-auth/backend/src/routes/scheduled-jobs.ts` — 1 file, 3 hunks (+30/−7) at `9727e7b` · **Findings:** 0 blocking, 1 major, 3 minor

### Findings
| major | R7 | scheduled-jobs.ts:964 | New fallback + warn-don't-fail behavior ships with no test (repo precedent: #266, #344) |
| minor | R6 | scheduled-jobs.ts:957 | `once` path still swallows `cancelJob` failures (also :1064 in the delete route) |
| minor | R5 | scheduled-jobs.ts:832 | Fallback + warn block duplicated ×3 in this diff (:832/:964/:1070, plus PATCH/resume per the PR description) — extract a helper |
| minor | R2 | scheduled-jobs.ts:1070 | Removes exactly one scheduler id (persisted OR fallback, never both) — double-live assumption flagged inline |

### Ops follow-up
This fix prevents *new* orphans but doesn't retroactively remove already-orphaned crons — the stale credit-data-doctor cron from the incident still needs a one-time manual removal, or it keeps firing alongside the legitimate schedule.

### Recommended reviewer
**@anuragdvd** — assigned via this mention (see note below).

| @anuragdvd | scheduled-jobs.ts — approved #1580 (last PR to touch this exact file, 2026-09-06); author/co-author across recent base-branch commits on this service | no (no CODEOWNERS entry for `apps/xyne-claw-auth/**`) | 350 open PRs w/ review requested | strongest familiarity; incident routed to him |
| @satvikbatra | base-branch committer, adjacent areas | no | 294 | — |
| @milinjp | base-branch committer, adjacent areas | no | 0 | weakest familiarity with this file |

Signals & caveats: the repo has **no CODEOWNERS entry** covering `apps/xyne-claw-auth/**` (only ACL paths are owned), so the eligibility set is file-history-based; commits-by-path couldn't be filtered directly (the available commits API has no path filter), so file ownership was derived from the last PR touching this exact file plus base-branch history; pending-review counts are open PRs with `review-requested:<login>` in this repo (includes automation-driven requests — a rough load signal). Familiarity is the dominant ranking signal: @anuragdvd reviewed the most recent change to this exact file and owns the incident this PR fixes, so he's the pick despite the highest pending count.

**Review-request note:** this run's GitHub toolset has no review-request API, so the reviewer couldn't be formally requested — @anuragdvd, please pick this up via this mention.

---
*4 inline comments posted on `9727e7b` (review [#5177907354](https://github.com/juspay/xyne-spaces/pull/1792#pullrequestreview-5177907354)).*

PR #1855, opened 2026-09-15
fix: XYNE-55777 sync user workload inside the assignment engine

XyneSpaces:
## Problem

Round-robin auto-assignment concentrated every new ticket on one user (XYNE-55777). `user_workload_mappings` is the only input to the least-loaded scorer, and it was never refreshed for the user that got picked, so that user stayed at zero load and kept winning.

The full-role path made it structural: `assignRoleDrivenSlots` skips the primary slot (`if (slot.roleId === primarySlot.roleId) continue;`) **before** `persist()` runs, and `syncUserWorkload` only ran inside `persist()` — so the one user who becomes `ticket.assignedTo` was the only user never synced. Separately, the apps create/update/email paths and the Zero full-role mutator never synced their pick at all.

## Approach (revised — previous revision of this PR did it differently)

The first revision added a `syncUserWorkload` call at each of the nine assignment call sites. That repeats the same call everywhere and any new call site silently reintroduces the bug. This revision instead syncs **once, inside the assignment engine, where the data is consumed**:

- `assignmentEngine` gains `loadWorkloadMappings(userGroupId, boardId, userIds)` — refreshes the candidates' rows from ticket truth, then reads them back.
- All three entry points — `evaluateAssignmentRule`, `evaluateAllRoles`, `evaluateRoleSlots` — now read workload through it (they each previously ran their own `userWorkloadMapping.findMany`).
- Result: anyone calling an `evaluate*` function gets an accurate scorer input by construction. No call site has to remember to sync, and the nine call-site syncs from the previous revision are gone.
- Best-effort by design: a refresh failure is logged and scoring falls back to the existing rows, since counts are recomputed from committed ticket state on the next evaluation.

Supporting change in `workloadUtils`: new `syncWorkloadForUsers(userIds, userGroupId, boardId, createdBy, userRoleIdsByUserId?)` is now the single implementation of the counting rules, and `syncUserWorkload` delegates to it. Counting semantics are unchanged (direct assignee OR role match against the project's `assignmentRoles`; `activeTasks` restricted to TODO/STARTED). The bulk form uses a fixed number of queries regardless of pool size and skips rows whose counts did not change, so the assignment hot path emits no no-op writes or Zero pokes.

This also resolves the review comment on the previous revision about calling `syncUserWorkload` from inside a Zero transaction at `zero/mutators.ts:809` — that call no longer exists. (`user_workload_mappings` is never written inside a Zero transaction, so the engine-side Prisma upsert has no lock interaction with one.)

## Second commit: pre-existing argument-order bug

`ticketService.asignUserGroupToTicket(ticketId, userId, groupId)` called `ticketRepository.assignUserGroupToTicket(ticketId, userId, groupId)`, but the repository signature is `(ticketId, groupId, updatedBy)`. Swapped arguments meant every apps `updateTicket` group change wrote the **actor's user id** into `ticket.userGroupId` (and the group id into `updatedBy`), dropping those tickets out of every group-scoped query including workload counting. One-line fix; drop the commit if you want it tracked separately.

## Verification (sandbox, real Postgres, seeded workspace)

All six auto-assignment flows, run against `main` and against this branch: 4-user group (u1–u3 primary role, u4 secondary), full-role board + plain board, cold workload rows, invoking the real controller methods, the real service, and the real Zero `ticket.update` mutator in-process.

The workload assertion "settles" by making **one real `evaluate*` call** (exactly what the next incoming ticket does) and then requiring every row to equal ticket truth — which is the proof that the engine refreshes by default.

| Flow | On `main` | With this PR |
|---|---|---|
| apps create ×6, full-role board | `u1→u1→u1→u1→u1→u1` (6/0/0) | `u1→u2→u3→u1→u2→u3` (2/2/2) |
| apps create ×6, plain board | all u1 (6/0/0/0) | rotation (2/2/2/0) |
| apps update group change (full-role + plain) | assigns u1, `userGroupId` written wrong | assigns + group persists + rows increment |
| service `assignTicketToGroup` ×6 | all u1 (12/0/0) | continues rotation (5/4/4) |
| Zero mutator `ticket.update` ×2 | u1, u1 | u3, u1 (least-loaded) |
| apps `createEmailTicket` ×2 | u1, u1 | assigns + syncs on both board types |
| workload rows vs ticket truth | **no rows at all** for any assignee | every row matches |

**main: 6/20 checks passed. This branch: 20/20.** `tsc --noEmit` clean on the changed files.

## Rollout / risk

- No schema change, no migration. Existing skewed rows self-correct the first time that group/board is evaluated.
- Behaviour change worth noting: rows are now refreshed at evaluation time rather than immediately after each assignment, so a row reflects the newest ticket from the next evaluation onward. Every pre-existing post-assignment sync (mutators, `ticketAssignmentService`, `emailService`, workers, `prTicketStatusSyncService`) is left in place, so read consumers keep their immediate updates.
- Added cost per assignment: one bounded refresh (a handful of queries) for the candidate pool, with unchanged rows skipped.


XyneSpaces (commented):
Automated review by Xyne Spaces Reviewer — 4 files / 8 hunks checked against the Xyne Spaces Review Guidelines. 6 findings: 0 blocking, 2 high, 2 medium, 2 nits. Summary + recommended reviewer posted as a PR comment.

apps/backend/src/zero/mutators.ts
```diff
@@ -801,6 +801,15 @@ async function assignFullRoles(
     }
 
     logger.info(`[AUTO-ASSIGN] Full role assignment complete for ticket ${ticketId}`);
+
+    // Sync the primary assignee's workload mapping. assignFullRolesToTicket syncs
+    // only the non-primary roles; ticket.assignedTo (the primary) was never synced
+    // (XYNE-55777), so the least-loaded scorer kept picking a stale-low user.
+    try {
+      await syncUserWorkload(primaryUserId, userGroupId, boardId, createdBy);
```
XyneSpaces: **[🟠 High] Zero mutator transaction — external DB work inside the tx** — `assignFullRoles` runs inside the Zero mutation transaction (it writes `tx.mutate.ticket_activities.insert` above), and `syncUserWorkload` is Prisma-backed external DB work (`db.board.findUnique` + upserts on the shared connection). The review guidelines are explicit: do not perform heavy operations inside a Zero mutator transaction — move them to `asyncTasks.push()` so they run after commit.

Beyond latency, two consistency hazards:
- if the Zero tx rolls back after this point (a later `tx.mutate` throws), the workload upsert has already committed on the Prisma connection → drift, the mirror image of the bug this PR fixes;
- the sync's Prisma connection cannot see this transaction's uncommitted writes (assumption — depends on what `syncUserWorkload` reads), so the row it writes may not reflect this very assignment.

↳ FIX: `asyncTasks.push(async () => { try { await syncUserWorkload(primaryUserId, userGroupId, boardId, createdBy); } catch (e) { logger.error(...); } });` — the post-commit pattern this file already uses elsewhere.

apps/backend/src/services/ticketAssignmentService.ts
```diff
@@ -528,6 +528,17 @@ export class TicketAssignmentService {
       }
     }
 
+    // Full-role path syncs the non-primary roles inside assignFullRolesToTicket,
```
XyneSpaces: **[🟠 High] R7 — behavior fix with no regression test** — this is the core behavior change (workload sync on the full-role path, the actual XYNE-55777 fix), but the diff contains no automated test. The sandbox verification (19/19 checks, real Postgres) is good evidence, but the harness under `apps/backend/scripts/x55777/` is not committed, so nothing prevents this exact skew from regressing silently later.

↳ FIX: commit the harness (or a trimmed service-level test) asserting that after `assignTicketToGroup` on a full-role board the primary assignee's `user_workload_mappings` row is synced, plus one test for the single-assignee path.

apps/backend/src/apps/controllers/ticketController.ts
```diff
@@ -842,6 +843,16 @@ export class TicketController {
         });
       }
 
+      // Sync workload mapping for the assigned user (single-assignee path).
+      if (userGroupId && resolvedAssignedTo && result.ticketId) {
+        try {
+          await syncUserWorkload(resolvedAssignedTo, userGroupId, boardId, userId);
```
XyneSpaces: **[🟡 Medium] Duplication — same block at 8 call sites** — this try/sync/log/catch block is copy-pasted at 8 sites: here and at 873, 1194, 1211, 2514, 2543 in this file, plus `ticketAssignmentService.ts:536` and `zero/mutators.ts:809`. The log formats also drift between sites (`[Ticket Assignment]` vs `[AUTO-ASSIGN]` vs no prefix).

↳ FIX: extract `syncUserWorkloadSafe(userId, userGroupId, boardId, actorId, contextLabel)` in `workloadUtils.ts` owning the try/catch + logging, and call it from all 8 sites.

apps/backend/src/services/ticketService.ts
```diff
@@ -179,7 +179,7 @@ export class TicketService {
 
   async asignUserGroupToTicket(ticketId: string, userId: string, groupId: string): Promise<void> {
     try {
-      await this.ticketRepository.assignUserGroupToTicket(ticketId, userId, groupId);
+      await this.ticketRepository.assignUserGroupToTicket(ticketId, groupId, userId);
```
XyneSpaces, line 182: **[🟡 Medium] R3 — no repair for already-corrupted rows** — the arg-order fix itself is correct (`ticketRepository.assignUserGroupToTicket(ticketId, groupId, updatedBy)` matches the repository signature). But per the PR description, pre-fix apps `updateTicket` group changes wrote the actor's **userId** into `ticket.userGroupId` (and the groupId into `updatedBy`), and those tickets are excluded from group-scoped pools and workload counting. They do **not** self-correct: the PR's "existing skewed rows self-correct on the next sync" claim covers `user_workload_mappings` rows, not `ticket.userGroupId`.

↳ FIX: add an idempotent data-repair migration (repoint `userGroupId` from the userId back to the intended group where recoverable via `updatedBy`), or state explicitly why the impacted rows are out of scope — with a count query (tickets whose `userGroupId` matches a user id) so the blast radius is known.

apps/backend/src/apps/controllers/ticketController.ts
```diff
@@ -1173,6 +1190,12 @@ export class TicketController {
                 data: { assignedTo: primaryUserId, updatedBy: userId, updatedAt: new Date() },
               });
               await syncConversationTicketMdFromPrismaTicket(prismaClient, updatedTicket);
+              try {
+                await syncUserWorkload(primaryUserId, effectiveGroupId!, targetBoardId!, userId);
```
XyneSpaces: **[🟢 Nit] Unguarded nullable args** — `effectiveGroupId!` / `targetBoardId!` here, and `ticket.boardId!` at 2514 and 2543: `boardId` is nullable on the email-ticket path, so a boardless ticket passes `null` into `syncUserWorkload` → `db.board.findUnique({ where: { id: null } })` throws a Prisma validation error → an error-level log per assignment instead of an intentional skip.

↳ FIX: guard the sync on `boardId`/`groupId` being present (`if (boardId && groupId) { … }`), or make `syncUserWorkload` no-op cleanly on null.

apps/backend/src/services/ticketAssignmentService.ts
```diff
@@ -528,6 +528,17 @@ export class TicketAssignmentService {
...
+    // but the primary assignee (ticket.assignedTo) was never synced, leaving its
+    // user_workload_mappings row stale (XYNE-55777). Sync it here as well.
+    if (usedFullRole) {
```
XyneSpaces: **[🟢 Nit] Collapsible with the existing sync** — this new `if (usedFullRole)` block mirrors the pre-existing `if (!usedFullRole)` sync just above; both bodies call `syncUserWorkload(assignedUserId, groupId, ticket.boardId, actorId)`. Since both branches set `assignedUserId`, one unconditional sync after the branch is equivalent and one less place for the two paths to drift.

↳ FIX: drop the `if (usedFullRole)` wrapper (or fold both into the suggested `syncUserWorkloadSafe` helper).

XyneSpaces:
## Xyne Spaces Reviewer — PR #1855

**Verdict:** Changes requested — 2 high findings (Zero-transaction sync, missing regression tests) should be addressed before merge; the fix direction itself is correct and the sandbox verification is thorough.
**Reviewed:** 4 files, 8 hunks · **Findings:** 0 blocking, 2 high, 2 medium, 2 nits (6 inline comments posted on `1d00fd2`)

### Findings
|---|---|---|---|
| 🟠 High | Zero mutator txn | `zero/mutators.ts:809` | `await syncUserWorkload(...)` runs inside the Zero mutation transaction (Prisma-backed external DB work) — move to the file's `asyncTasks.push()` post-commit pattern; rollback-after-sync causes drift |
| 🟠 High | R7 Tests | `services/ticketAssignmentService.ts:531` | Core behavior fix with no committed regression test; the sandbox harness (`apps/backend/scripts/x55777/`) is not in the diff |
| 🟡 Medium | Duplication | `apps/controllers/ticketController.ts:849` | Identical try/sync/log/catch block at 8 call sites — extract a `syncUserWorkloadSafe` helper in `workloadUtils.ts` |
| 🟡 Medium | R3 Data | `services/ticketService.ts:182` | No backfill/repair for rows already corrupted by the swapped-args bug (actor userId sitting in `ticket.userGroupId`) — these do not self-correct |
| 🟢 Nit | Null-safety | `apps/controllers/ticketController.ts:1194` | `effectiveGroupId!` / `targetBoardId!` / `ticket.boardId!` — a boardless ticket makes `syncUserWorkload` throw a Prisma validation error, logged per assignment |
| 🟢 Nit | Simplify | `services/ticketAssignmentService.ts:534` | New `if (usedFullRole)` sync mirrors the existing `if (!usedFullRole)` sync — collapse to one unconditional call |

✅ Verified correct: the `assignUserGroupToTicket(ticketId, groupId, updatedBy)` arg-order fix matches the repository signature. No migrations in the diff, so the migration-safety gate was not triggered.

### Recommended reviewer
**@AnkurAbraham** (Ankur Mohan Abraham — ankur.abraham@juspay.in) — assignee of XYNE-55777 (currently in the PR Review stage) who oversaw the RCA for the skewed assignment this PR fixes; not the PR author.

| Candidate | Relation to this PR | Eligible? | Pending reviews | Tie-breaker |
|---|---|---|---|---|
| @AnkurAbraham | XYNE-55777 assignee (PR Review stage), followed the RCA thread | yes (org MEMBER) | n/a* | only 4 open PRs authored, none recent |
| Utkarsh Pandey (utkarsh.pandey@juspay.in) | ACL guardian — CODEOWNERS `@juspay/xyne-acl-guardians` covers the changed `zero/mutators.ts` | GitHub login not resolvable from available data | n/a* | surfaced as codeowner-cc; could not be formally requested |
| @harsh-1923 (Harsh Sharma) | Named Zero-mutators reviewer on an earlier mutator fix | yes | n/a* | alternate for the mutators.ts finding |

\* The available search API does not honor the `review-requested:` qualifier (it falls back to lexical search over all open PRs), so per-person pending-review counts could not be computed; author-load and Spaces activity were used as the tie-breaker instead.

**Assignment note:** this agent has no review-request API or `gh` CLI available, so the review could not be formally requested via the API — @AnkurAbraham is @mentioned here instead (GitHub notifies on mention). To formalize: `gh pr edit 1855 --add-reviewer AnkurAbraham`. If Utkarsh (ACL guardian) should sign off on the `mutators.ts` change, he needs to be added manually.

XyneSpaces:
Branch was force-pushed and rewritten with a redesign. See the updated PR description for the revised approach.

apps/backend/src/utils/assignmentEngine.ts
```diff
@@ -187,6 +188,45 @@ async function filterMappingsToChannelParticipants(
...
+ * updated after an assignment freezes that user's load at its old value and the
+ * least-loaded pick keeps landing on them (XYNE-55777). Doing the refresh here —
+ * inside the engine, for exactly the users about to be scored — means every
+ * assignment path gets it by construction and no caller has to remember to sync
+ * after it persists an assignment.
+ *
+ * Best-effort by design: a refresh failure must never block an assignment, and
+ * the counts are recomputed from committed ticket state on the next evaluation.
+ */
+async function loadWorkloadMappings(
```
XyneSpaces, line 210: **[🟠 High · Rule 204] No committed tests for this rewrite.** This PR changes scorer input on every assignment path (`loadWorkloadMappings` refreshes before scoring), rewrites workload counting into bulk in-memory counting (`syncWorkloadForUsers`), and fixes two bugs (XYNE-55777 stale rows + the `asignUserGroupToTicket` argument order) — but the diff adds no test file, and a repo-wide code search finds zero test files importing `workloadUtils` or `assignmentEngine` (only production queues/services/workers/controllers), so nothing in CI guards the regression this PR fixes. Suggested: (1) unit test for `syncWorkloadForUsers` — skip-unchanged rows, direct-assignee ∪ role-match counting with dedupe, `provided` role-map passthrough; (2) integration test asserting two consecutive `evaluateAssignmentRule` calls rotate the least-loaded pick after the first assignment persists — the exact XYNE-55777 failure mode.

apps/backend/src/utils/assignmentEngine.ts
```diff
@@ -187,6 +188,45 @@ async function filterMappingsToChannelParticipants(
  * Returns: { assignedUserId } or { reason: "NO_ON_CALL_USERS" | "EXCLUDED_USER_ONLY_CANDIDATE" }
  * @param projectId - Optional project ID to scope workload calculation to boards in the same project only
  */
+/**
```
XyneSpaces, line 191: **[🟢 Nit · style] This inserted block detaches `evaluateAssignmentRule` from its JSDoc.** The `Returns: … / @param projectId` block ending at line 190 now visually documents `WORKLOAD_SYNC_ACTOR` / `loadWorkloadMappings` instead of the function below (line 230). Suggested: move the constant + helper below `evaluateAssignmentRule` (or above its JSDoc) so each doc block stays attached to its own symbol.

XyneSpaces (commented):
## Xyne Spaces Reviewer — PR #1855 (re-review @ 2d1d1ac)

**Verdict:** Changes requested — 0 blocking · 1 high · 3 medium · 1 nit
**Reviewed:** 3 files, 8 hunks (+185/−59) · every changed hunk checked · head `2d1d1ac` (force-pushed since the previous review @ `1d00fd26`)

### Re-review context
This revision replaces the previous one's per-call-site syncs with a single engine-side refresh (`loadWorkloadMappings` → `syncWorkloadForUsers`) — the architecture-level concern from the first review (sync inside the Zero mutation transaction) is resolved by design: `zero/mutators.ts` is untouched on this head, and every `evaluate*` / `assignFullRoles` caller is a service/controller/worker/queue outside a Zero mutation. The prior `ticketService.ts` note (data repair for rows written with swapped `userGroupId`) still applies — not reposted.

### Findings
|---|---|---|---|
| 🟠 high | 204 | apps/backend/src/utils/assignmentEngine.ts:210 | No committed tests for the engine-side refresh + bulk-count rewrite + arg-order fix |
| 🟡 medium | 102 | apps/backend/src/utils/workloadUtils.ts:155 | `getProjectAssignmentRoleIds` serialized ahead of a `Promise.all` that doesn't depend on it |
| 🟡 medium | 102 | apps/backend/src/utils/workloadUtils.ts:211 | Per-user upserts sequential in the loop despite independent composite keys |
| 🟡 medium | 263 | apps/backend/src/utils/workloadUtils.ts:88 | `resolveUserRoleIds` duplicates the `getUserRoleIds` role-union in the same file |
| 🟢 nit | style | apps/backend/src/utils/assignmentEngine.ts:191 | Inserted helpers detach `evaluateAssignmentRule`'s JSDoc from its function |

### Verified safe (context, no action)
- **Counting semantics preserved** — the in-memory count (direct assignee ∪ role-matched ticket, single pass with dedupe; `activeTasks` = statusV2 TODO/STARTED) reproduces the old per-user COUNT queries, including the empty-role-set OR branch, null `assignedTo`, and board+userGroup scoping of role-matched tickets.
- **No Zero-mutator exposure** — see re-review context.
- Note only: the board-wide `db.ticket.findMany` (minimal select) trades per-user COUNTs for one query whose row count scales with board size — fine for current board sizes, worth revisiting if very large boards appear.

### Recommended reviewer
**@AnkurAbraham** — owns ticket XYNE-55777 (stage: PR Review), active in the ticket/board domain (#1047 Created-By group-by, #1048 board-scoped user filters). This bot has no review-request API access, so the @mention here is the request — please add yourself as reviewer (or `gh pr edit 1855 --add-reviewer AnkurAbraham`).

| Candidate | Domain evidence | CODEOWNERS match | Pending reviews | Tie-breaker |
|---|---|---|---|---|
| @AnkurAbraham | XYNE-55777 owner (stage: PR Review); ticket-domain PRs #1047/#1048 | none (no entry covers these files) | 0 | ticket stage makes this his active task |
| @5uryan5h (Suryansh Srivastava) | authored the original workload-sync fix #513/#525 (XYNE-55809) that this PR refactors | none | 0 | directed this revision — natural consult, not sole reviewer |
| @architjuspay | board-config work in this repo | none | 0 | weaker workload-domain signal |

No `.github/CODEOWNERS` entry covers the three changed files (repo CODEOWNERS covers ACL paths only), so ranking comes from ticket ownership + domain history + Spaces context. Pending-review counts come from every open PR in the repo currently having an empty `requested_reviewers` list (0 for all candidates).

**Rationale:** Ankur owns the ticket at exactly this stage; Suryansh knows the workload code best but drove the redesign, so he's the second pair of eyes rather than the assigned reviewer.

apps/backend/src/utils/workloadUtils.ts
```diff
@@ -71,6 +71,78 @@ export async function syncUserWorkload(
...
+    userRoleIds !== undefined ? new Map([[userId, userRoleIds]]) : undefined,
+  );
+}
+
+/**
+ * All roleIds held by each of the given users, in one pair of queries.
+ * Pre-resolved entries in `provided` are used as-is (callers that already know a
+ * user's roles should pass them, exactly as the single-user path allowed).
+ */
+async function resolveUserRoleIds(
```
XyneSpaces, line 88: **[🟡 Medium · Rule 263] `resolveUserRoleIds` duplicates `getUserRoleIds` in this same file.** `getUserRoleIds` — still used by `syncUserWorkloadAllBoards` — keeps its own copy of the `userRoleMapping` ∪ `userGroupMapping` role union with the `isActive` filter, so the repo now has two implementations of "which roles does a user hold" that can drift. Suggested: keep one — `getUserRoleIds` delegating: `return (await resolveUserRoleIds([userId])).get(userId) ?? [];`.

apps/backend/src/utils/workloadUtils.ts
```diff
@@ -80,62 +152,87 @@ export async function syncUserWorkload(
     return;
   }
 
-  const [projectAssignmentRoleIds, resolvedUserRoleIds] = await Promise.all([
-    getProjectAssignmentRoleIds(board.projectId),
-    userRoleIds !== undefined ? Promise.resolve(userRoleIds) : getUserRoleIds(userId),
-  ]);
-  const workloadRoleIds = projectAssignmentRoleIds.filter(id => resolvedUserRoleIds.includes(id));
+  const projectAssignmentRoleIds = await getProjectAssignmentRoleIds(board.projectId);
```
XyneSpaces, line 155: **[🟡 Medium · Rule 102] `getProjectAssignmentRoleIds` is serialized ahead of a `Promise.all` that doesn't consume it.** Only the guarded `ticketAssignment.findMany` (and the role-match filtering below) uses `projectAssignmentRoleIds`, while the `Promise.all` over `resolveUserRoleIds` / ticket findMany / existing-rows read is independent of it — the pre-rewrite code resolved it inside a `Promise.all`. Suggested: fold it in — `const [projectAssignmentRoleIds, roleIdsByUserId, tickets, existingRows] = await Promise.all([...])` — to keep the assignment hot path at the fixed query-count/latency the design is aiming for.

apps/backend/src/utils/workloadUtils.ts
```diff
@@ -80,62 +152,87 @@ export async function syncUserWorkload(
...
+      }
+    }
+
+    const existing = existingByUserId.get(userId);
+    if (existing && existing.activeTasks === activeTasks && existing.totalTasks === totalTasks) {
+      continue;
+    }
+
+    // Writes the row of the assignee rather than the caller, so it runs above the caller's own scope.
+    await withWorkspaceScope(() =>
```
XyneSpaces, line 211: **[🟡 Medium · Rule 102] Per-user upserts run sequentially inside this loop.** Each write targets a distinct `userId_userGroupId_boardId` composite key and only rows whose counts changed reach this point, so the upserts are independent and can be issued concurrently: collect the changed `(userId, activeTasks, totalTasks)` triples and run the upserts under `Promise.all` (each `withWorkspaceScope` callback already re-establishes the AsyncLocalStorage context it needs). The read side is fixed-query-count; the write side currently still costs one round trip per changed user.

PiyushJuspay (approved):

AnkurAbraham (approved):

KushagarGoel1 (approved):

PR #1856, opened 2026-09-15
feat: XYNE-57204 ephemeral msg support for slack/xyne

HarimohanSharma11:
  - backend
  - dashboard

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


utkarsh-611 (approved):

utkarsh-611 (dismissed):

apps/dashboard/src/components/Chat/ChatList/ChatListV4.tsx
```diff
@@ -342,8 +343,14 @@ const ChatListV4: React.FC<ChatListProps> = ({
     );
   }, [conversationsWithPending, unreadsOnly, channelParticipation?.lastViewedAt]);
 
+  const ephemeralConversations = useEphemeralChannelConversations(channelId);
+  const conversationsWithEphemeral = useMemo(() => {
+    if (ephemeralConversations.length === 0) return filteredConversations;
+    return [...filteredConversations, ...ephemeralConversations];
```
ameer030, line 349: 🟡 **W01 — channel-level ephemeral cards open a dead thread**

A card posted to a channel (no `thread_ts` / `conversationId`) is rendered as a normal conversation row. Clicking it (or using the reply affordance, which BOT cards get) opens the thread view for that row's `conversationId` — but that id was synthesized at post time and names nothing in the database. The thread panel then shows an empty conversation, the card itself does not appear there, and any reply the user types fails server-side with "conversation or channel does not exist".

- Synthesized id: `apps/backend/src/apps/core/ephemeralDelivery.ts` — `const conversationId = args.conversationId ?? crypto.randomUUID()`; the payload even notes the client "must not try to open [it] as a thread", but nothing enforces that.
- Row construction: `useEphemeralChannelConversations` in `apps/dashboard/src/hooks/useEphemeralMessages.ts` builds a `Conversation` row carrying the synthetic id.
- Navigation has no existence guard: `handleOpenThread` in `ChatListV4.tsx:1138` calls `find()` and then navigates to `/chat/.../<conversationId>` regardless; BOT messages get `onReplyInThread` (`ChatBubble.tsx:1390-1412`).
- The card is invisible in the thread view: `useEphemeralThreadMessages` only returns cards with `isThreadReply === true` (`useEphemeralMessages.ts:241`), and channel cards carry `false`.
- Reply failure: the Zero messages insert ACL rejects inserts for a non-existent conversation (`apps/backend/src/zero/acl/tables/messages-acl.ts:19-25`).
</details>

Mark ephemeral rows (they already carry `metadata.ephemeral: true`) and suppress the thread-open affordance for them: in `ChatListItem`/`ChatBubble`, skip `onReplyInThread`/click navigation when the row's metadata has `ephemeral === true`, so the card renders read-only inline (matches Slack, where channel-level ephemeral messages have no thread affordance).
</details>

rules: BUG-C28
agent: omp
run: n/a
</details>

apps/backend/src/apps/controllers/chatController.ts
```diff
@@ -336,6 +384,144 @@ export class ChatController {
...
+      if (!targetUser) {
+        res.status(404).json({ error: `User not found: ${user}`, code: 'NOT_FOUND' });
+        return;
+      }
+
+      const resolvedChannelId = await resolveChannelId(channelId, conversationId, channelName);
+
+      // Verified token appId — authenticateApp sets req.auth from the validated
+      // token, so this is safe to sign into a capability.
+      const appId = (req as any).auth?.appId ?? (req.body as Record<string, unknown>).appId;
```
ameer030, line 442: 🔵 **N01 — appId body fallback contradicts the comment above it**

The comment says the appId is the verified one from `authenticateApp` "so this is safe to sign into a capability", but the expression still falls back to `req.body.appId` — a client-supplied value. Today that fallback is unreachable because this controller is only mounted behind `authenticateApp`, which always sets `req.auth.appId` before calling `next()`. The risk is latent: this value is what the flow capability token signs, so if the route is ever mounted differently (like `/apps/flow/action` under user auth), a body-supplied appId would be trusted silently.

- `apps/backend/src/apps/controllers/chatController.ts:442` — `const appId = (req as any).auth?.appId ?? (req.body as Record<string, unknown>).appId;`
- `apps/backend/src/apps/middelware/authenticator.ts` sets `(req as any).auth = { appId: verifiedResult.data.appId, … }` on every success path; all failure branches return.
- Mounting: `apps/backend/src/apps/routes/apps.ts:83` — `router.use("/chat", authenticateApp, chatRoutes)`.
</details>

Drop the fallback and fail closed:

```ts
const appId = (req as any).auth?.appId;
if (flow && (typeof appId !== 'string' || !appId)) { …400… }
```

The identity that gets signed into the token then has exactly one possible source — the verified token — which is what the comment already promises.
</details>

rules: SEC-C04
agent: omp
run: n/a
</details>

apps/backend/src/apps/platform-adapters/slack/request-transformers/chat.ts
```diff
@@ -107,6 +120,63 @@ export async function transformPostMessage(
...
+	);
+
+	return {
+		channelId: slackReq.channel,
+		userId: context.userId,
+		recipientId: slackReq.recipientId,
+		flow,
+		content,
+		isMarkdown,
+		conversationId: slackReq.thread_ts,
```
ameer030, line 175: 🔵 **N02 — dead `conversationId` field carries the wrong value**

`transformPostEphemeral` puts `slackReq.thread_ts` into an args field named `conversationId`. A Slack `thread_ts` is not a Xyne conversation id, and the only consumer (`chatPostEphemeral`) never reads it — it uses the properly resolved `threadResolution.conversationId` instead. The field is dead on arrival and mislabeled, inviting a future caller to wire it straight through.

- Set here: `apps/backend/src/apps/platform-adapters/slack/request-transformers/chat.ts:175` (also added to the `PostEphemeralArgs` interface above).
- Sole consumer: `chatPostEphemeral` in `apps/backend/src/apps/platform-adapters/slack/controller.ts:476-512` — passes `threadResolution.conversationId` to `deliverEphemeralMessage`, never `args.conversationId`.
</details>

Delete the field from both the interface and the return value — the resolved conversation id already has exactly one source of truth in the controller.
</details>

rules: STYLE-C14
agent: omp
run: n/a
</details>

apps/dashboard/src/hooks/useEphemeralMessages.ts
```diff
@@ -0,0 +1,283 @@
...
+          visibleTo: m.visibleTo ?? userId,
+          isSent: true,
+          metadata: { ...(m.metadata ?? {}), ephemeral: true },
+        }),
+        parent_message_md: null,
+        doNotPostToChannel: null,
+        createdAt: timestamp,
+        initialMessageAttachments: [],
+        initialMessageNudgeCounts: [],
+      } as unknown as Conversation;
```
ameer030, line 219: 🔵 **N03 — hand-built rows hidden behind `as unknown` casts**

The ephemeral conversation/message rows are built as inline object literals and then forced through `as unknown as Conversation` (and `as typeof base` in `ThreadPannel`). These objects pretend to be real synced rows, so any future field added to `Conversation`/message shapes will silently miss these literals — the compiler can no longer help.

- `apps/dashboard/src/hooks/useEphemeralMessages.ts:219` — `} as unknown as Conversation;` (and line 182, `EMPTY as unknown as Conversation[]`)
- `apps/dashboard/src/components/Chat/ThreadPannel.tsx:329` — `ephemeralThreadMessages as typeof base`
</details>

Give the builders a real return type: derive the row shape from the same types the Zero schema/`Conversation` uses (or a small `EphemeralConversation` subtype) so excess/missing fields are caught at build time instead of at the cast.
</details>

rules: STYLE-C11
agent: omp
run: n/a
</details>

ameer030 (commented):
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 0 |
| 🟡 Better to fix | 1 |
| 🔵 Nit | 3 |

### 🔁 Re-check
- ✅ **Verified OK:** flow capability token — HMAC-SHA256 with `timingSafeEqual` (length-guarded), audience + version binding, bound to userId + messageId, 30-min TTL, re-minted on screen advance and stripped before outbound dispatch (`flowToken.ts`, `flowController.ts`); appId resolution never trusts the client on the action path; recipient channel-participation checked in the shared `deliverEphemeralMessage` before broadcast; both entry points (native API + Slack adapter) route through that one shared delivery function behind `requirePermission('chat:write')` + channel-access validation; flow content HTML-attribute escaping (`encodeHtmlAttr` + quote-escaped `data-flow-json`) and markdown sanitization match the persisted-flow path; user-room routing confirmed (`user:<id>` emit shape matches the client's listener); Slack response shape is faithful to Slack's documented `chat.postEphemeral`.

### 📌 Verdict
No blockers — good to merge after the small fixes above (W01 is the only one with real user impact: dead-end threads from channel-level cards).

apps/dashboard/src/components/Chat/ChatBubble/ChatBubble.tsx
```diff
@@ -270,6 +270,26 @@ export const ChatBubble: React.FC<ChatBubbleProps> = ({
...
+  // the single gate for every entry point — hover toolbar, mobile long-press, the
+  // reply-count strip and the bubble action all read it — so removing it once
+  // closes all of them and cannot be missed when another is added.
+  //
+  // Matches Slack, where a channel-level ephemeral message has no thread
+  // affordance at all.
+  const replies = ((): typeof repliesProp => {
+    if (!repliesProp || metadata?.['__xyneEphemeral'] !== true) return repliesProp;
+    const withoutThread = { ...repliesProp };
+    delete withoutThread.onOpenThread;
```
ameer030, line 289: 🟡 **W01 — "Reply in Thread" is still offered for these cards on mobile, and tapping it does nothing**

On desktop the fix is good: dropping `onOpenThread` removes the thread affordance everywhere it was reachable. But the mobile actions drawer (long-press a message) checks only that the `replies` object exists — not that it still contains the handler. After the strip, the object still exists, so the drawer still shows "Reply in Thread", and tapping it calls `replies?.onOpenThread?.(e)` — which is now `undefined`, so nothing happens.

- The strip (ChatBubble.tsx:286-291) deletes only the `onOpenThread` key — `replies` itself stays truthy (replyCount + lastActivityAt remain).
- The drawer spread (ChatBubble.tsx:1424-1433) is gated on `replies && …`; the hover toolbar (`ChatBubble.tsx:1092`) correctly gates on `replies?.onOpenThread` instead.
- `MessageActionsDrawer.tsx:222-228` renders the row whenever `onReplyInThread` is provided — so mobile gets a visible, dead button.
- Every other entry point verified closed: hover toolbar, mobile bubble-click (`shouldEnableMobileThreadOpen` requires `replies?.onOpenThread`, and `handleMobileBubbleThreadOpen` early-returns without it), the reply-count strip (early-returns when replyCount is 0), and chat row click (there is none outside `replies`).

</details>

Give the drawer spread the same predicate as the hover toolbar:

```ts
{...(replies?.onOpenThread &&
  (!isSystemMessage || isTicketCreationMessage || isCallMessage) &&
  !isShowInChannel &&
  (!isMessageDeleted || context === 'channel') && {
    onReplyInThread: ...
```

</details>

rules: BUG-C28, DOCS-C04
agent: omp
run: n/a
</details>

apps/dashboard/src/hooks/useEphemeralMessages.ts
```diff
@@ -0,0 +1,283 @@
...
+    // first time anything reconnects.
+    void websocketService
+      .connect()
+      .then(() => {
+        if (!active) return;
+        websocketService.on('ephemeral_message', add);
+        websocketService.on('open_screen_message', add);
+        websocketService.on('connect', handleConnect);
+      })
+      .catch(() => {});
```
ameer030, line 153: 🟡 **W02 — the socket listeners attach exactly once; a later socket rebuild silently turns ephemeral messaging off for the rest of the tab**

The bridge registers `ephemeral_message` / `open_screen_message` one time, inside `websocketService.connect().then(...)`. But the socket client is a shared service: from a not-connected state, `connect()` calls `removeAllListeners()` and builds a **new** socket — and other owners call `connect()` later in the session. Two verified paths: NotificationHandler re-runs its `[user]` effect (disconnect + reconnect) on an in-tab account switch, and Kanban counts call `connect()` whenever the connection drops. After either, the bridge's handlers are attached to a socket that no longer exists: cards keep arriving from the server, but nothing in the tab hears them, and the popup and inline cards simply stop working with no error.

Separately, if the very first handshake fails, `.catch(() => {})` swallows it: socket.io's own reconnection may recover the transport later, but `.then` never reruns, so the handlers are never attached — again with no telltale.

- Attach site: `useEphemeralMessages.ts:143-153` — effect deps `[]`, handlers registered once in `.then(...)`; the registered `connect` handler only clears the store, it does not re-attach the message listeners.
- `socketClient.ts:29-58`: `connect()` early-returns only when connected/in-flight; otherwise it disconnects, `removeAllListeners()`, and creates a new `io()`.
- Other `connect()` callers: `NotificationHandler.tsx` (`[user]` effect cleanup calls `websocketService.disconnect()`, body reconnects) and `useKanbanCounts.ts:629-637` (`!isConnectedToServer()` → `connect()`).
- The code comment says registration inside `connect()` is deliberate — but that only protects against the bridge's own call, not against every other component sharing the singleton.

</details>

Attach the two message listeners inside the `connect` event handler itself so they are re-registered on every fresh socket:

```ts
const attach = (): void => {
  websocketService.on('ephemeral_message', add);
  websocketService.on('open_screen_message', add);
};
…
websocketService.on('connect', () => { clearAll(); attach(); });
…attaching `connect` + one initial `attach()` after the first connect succeeds…
```

And log in the `.catch` (or retry with a short backoff) so a dead feature is diagnosable instead of silent.

</details>

rules: BUG-C25, ARCH-C06
agent: omp
run: n/a
</details>

apps/dashboard/src/components/flowUI/FlowRenderer.tsx
```diff
@@ -368,15 +401,61 @@ export const FlowRenderer: React.FC<FlowRendererProps> = ({
...
   return (
     <FlowContext.Provider value={contextValue}>
-      <div className={compact ? 'flow-ui-compact' : 'flow-ui-container max-w-2xl w-full'}>
-        {validatedFlow.title && (
-          <h2 className={compact ? 'text-sm font-semibold mb-2' : 'text-base font-semibold mb-3'}>
-            {validatedFlow.title}
-          </h2>
-        )}
-        {validatedFlow.components.map(renderComponent)}
+      <div className='flow-ui-container max-w-2xl w-full flex flex-col gap-4'>
```
ameer030, line 456: 🟡 **W03 — existing inline flow cards are restyled everywhere, while the comment says they are "untouched"**

The comment above this branch promises that inline rendering "(compact=false, the channel) is untouched", but the markup change applies to every inline flow card already in the product, not just the new popup: the container gains `flex flex-col gap-4`, and the title loses its `mb-2`/`mb-3`. That changes the spacing between the title and the first component and inserts a 16px gap between *every* pair of components — where previously sibling margins collapsed — and flex-column stretches children to full width.

If the restyle is deliberate (it may well look better), it just needs to be owned: say so in the PR body with a screenshot. If it isn't, restore the old non-compact markup and leave the new layout to the compact branch only.

- Pre-PR non-compact render (base): `<div className='flow-ui-container max-w-2xl w-full'>` with the `h2` carrying `mb-2`/`mb-3`.
- Now (FlowRenderer.tsx:456): `flow-ui-container max-w-2xl w-full flex flex-col gap-4`; title at :415-419 with no margin class — comment at :428 claims the branch is untouched.
- Reach confirmed: `RenderMessageWithHTML.tsx:1116` mounts `FlowScreenManager` without `compact`, so every existing inline flow card takes this path.

</details>

Either restore the exact pre-change non-compact markup:

```tsx
<div className='flow-ui-container max-w-2xl w-full'>
  {validatedFlow.title && (
    <h2 className='text-base font-semibold mb-3'>{validatedFlow.title}</h2>
  )}
  {components.map(renderComponent)}
</div>
```

…or flag the restyle as intentional in the PR description with a before/after screenshot.

</details>

rules: DOCS-C04, STYLE-C10
agent: omp
run: n/a
</details>

apps/backend/src/apps/controllers/chatController.ts
```diff
@@ -336,6 +384,144 @@ export class ChatController {
...
+        return;
+      }
+
+      const targetUser = await repositories.users.findById(user);
+      if (!targetUser) {
+        res.status(404).json({ error: `User not found: ${user}`, code: 'NOT_FOUND' });
+        return;
+      }
+
+      const resolvedChannelId = await resolveChannelId(channelId, conversationId, channelName);
```
ameer030, line 438: 🟡 **W04 — a caller-supplied `conversationId` is never checked against the resolved channel on this door**

When the caller passes both `channelId` and `conversationId`, `resolveChannelId` returns the channel immediately without looking at the conversation, and the body's `conversationId` is forwarded verbatim. So an app with access to channel A can post an interactive ephemeral card that is addressed as a *thread reply* in a conversation belonging to channel B — the bot and recipient participation checks all ran against channel A. The Slack front door rejects this exact case (`thread_not_found` unless the thread's channel matches), so the two doors the PR says cannot drift apart already have: a multi-app Slack client hitting the native API gets different placement rules depending on which door it uses.

Impact is bounded (the caller needs to know a foreign conversation id, and nothing is persisted): at worst the recipient never sees the card, or sees an app's interactive card inside an unrelated thread.

- `channelUtils.ts:18-21` — `resolveChannelId` short-circuits on `channelId` and never consults `conversationId`.
- `validateChannelAccessForPost` (`middelware/channelValidation.ts`) validates only the channel; no conversation↔channel pairing check.
- `chatController.ts:463` forwards the unverified `conversationId`; `ephemeralDelivery.ts` then broadcasts it with `isThreadReply: true`.
- Contrast: `platform-adapters/slack/controller.ts:110-131` (`resolveSlackThreadConversationId`) rejects unless `parentConversation.channelId === channelId`.
- Header of `ephemeralDelivery.ts`: "this and the Slack adapter's chat.postEphemeral cannot drift apart on authorization" — the pair-check is the drift.
- Sibling note: persisted `postMessage` has the same inversion (`findOrCreateConversation` ignores `channelId` when `conversationId` is set) — pre-existing, so no need to scope-creep, but don't copy it here.

</details>

When `conversationId` is provided, load it once and reject when it doesn't belong to the resolved channel — mirroring the Slack door:

```ts
if (conversationId) {
  const convo = await repositories.conversations.findById(conversationId);
  if (!convo || convo.channelId !== resolvedChannelId) {
    res.status(404).json({ error: `Conversation not found: ${conversationId}`, code: 'NOT_FOUND' });
    return;
  }
}
```

Even better: put the check inside `deliverEphemeralMessage` itself, so both doors share it the same way they share the recipient-participation check.

</details>

rules: SEC-C01, SEC-C07
agent: omp
run: n/a
</details>

apps/backend/src/apps/controllers/chatController.ts
```diff
@@ -64,6 +65,53 @@ const PostMessageBodySchema = ChatActionBodySchema.extend({
...
+    touched: z.record(z.boolean()),
+    errors: z.record(z.string()),
+    submitting: z.boolean(),
+    submitted: z.boolean(),
+    history: z.array(z.string()),
+    loadingComponentIds: z.array(z.string()).optional(),
+  }),
+});
+
+const PostEphemeralBodySchema = ChatActionBodySchema.extend({
```
ameer030, line 94: 🔵 **N01 — `uploadedFiles` is accepted and silently dropped**

Because this schema extends `ChatActionBodySchema`, callers can pass `uploadedFiles` (the ids returned by filesUpload). `postEphemeral` never reads the field — nothing is persisted, so there is genuinely nothing to attach them to — and still answers 201. An app posting files learns nothing about them not arriving; the response reports a delivered message that can never contain them.

`attachments` *is* consumed (in `processMessageContent`), so this is only about the `uploadedFiles` member.

- Declared: `ChatActionBodySchema` (`chatController.ts:28-36`) → inherited by `PostEphemeralBodySchema`.
- Dropped: `postEphemeral` destructures the body without `uploadedFiles`; `deliverEphemeralMessage` has no such parameter. Contrast: `postMessage` consumes it at `chatController.ts:274`.

</details>

Either omit it from the schema (`ChatActionBodySchema.omit({ uploadedFiles: true, userId: true }).extend({…})`), or reject with 400 when `uploadedFiles?.length` is non-zero — failing loudly beats an untruthful 201.

</details>

rules: BUG-C28
agent: omp
run: n/a
</details>

apps/dashboard/src/hooks/useEphemeralMessages.ts
```diff
@@ -0,0 +1,283 @@
...
+  senderId: string;
+  senderName?: string;
+  content: string;
+  /** Present when the card carries a flow — the popup renders this directly. */
+  flowJSON?: FlowDefinition;
+  msgType?: string;
+  createdAt: Date | number | string;
+  metadata?: Record<string, unknown>;
+  visibleTo?: string | null;
+  messageDelivery?: 'EPHEMERAL' | 'OPENSCREEN';
```
ameer030, line 42: 🔵 **N02 — same delivery-mode union defined twice in the same PR**

The PR adds `MESSAGE_DELIVERY` / `MessageDelivery` to `@xyne/shared`, and then this hook redeclares the domain inline as a string-literal union (`'EPHEMERAL' | 'OPENSCREEN'`) and hand-mirrors the server payload as `EphemeralMessage`/`EphemeralEvent`. The shared export exists to be the single source of the wire contract; this local copy will compile cleanly the day a third delivery mode is added and start silently mis-filtering.

- Shared: `packages/shared/src/validation/flowSchema.ts:1076-1082` (`MESSAGE_DELIVERY`, `MessageDelivery`).
- Local copy: `useEphemeralMessages.ts:42` — `messageDelivery?: 'EPHEMERAL' | 'OPENSCREEN';`.
- The wire shape itself (`messageId`, `isThreadReply`, `visibleTo`, …) is duplicated against `ephemeralDelivery.ts:160-183`.

</details>

Export the ephemeral event payload type from `@xyne/shared` next to `MESSAGE_DELIVERY` (server `ephemeralDelivery.ts` then emits `satisfies` it) and use `MessageDelivery` here; at minimum, type `messageDelivery?: MessageDelivery`.

</details>

rules: STYLE-C01, STYLE-C13
agent: omp
run: n/a
</details>

apps/dashboard/src/components/flowUI/FlowRenderer.tsx
```diff
@@ -368,15 +401,61 @@ export const FlowRenderer: React.FC<FlowRendererProps> = ({
...
   }
 
+  // A trailing row or button is the screen's action bar; anything else at the
+  // end is ordinary content and stays in the body.
+  const components = validatedFlow.components;
+  const last = components.length > 0 ? components[components.length - 1] : undefined;
+  // A trailing `row` is only an action bar if it actually holds buttons. A row of
+  // two text columns at the end of a screen is content, and pulling it into the
+  // footer would put a divider above it and pin it for no reason.
+  const isActionBar =
```
ameer030, line 411: 🔵 **N03 — two compact-popup edge glitches: stale "scroll for more" chevron, and an empty footer when the action bar is hidden**

Two small render edges in the new compact mode:

1. The ResizeObserver watches the body's children once per `validatedFlow`. Components whose `hidden` expression depends on form values appear/disappear as the user types without changing `validatedFlow`, so added children are never observed and removed ones never re-measured — the chevron can stick on (content fully visible) or off (content cut off) until the user happens to scroll.
2. When the trailing action-bar row itself is conditionally hidden, `renderComponent(last)` returns null — but the `border-t pt-3 mt-3` footer wrapper still renders, leaving an empty padded divider at the bottom of the popup.

- Observer effect: `FlowRenderer.tsx:365-386`, deps `[validatedFlow]`; `isVisible` evaluation reads `state.values`, so visibility flips land without re-running the effect.
- `isActionBar` (`:411-413`) doesn't account for a hidden last node; the footer wrapper (`:441-443`) renders regardless of `renderComponent(last)` being null.

</details>

- Re-measure on every commit that can change heights: run `measure()` from a layout effect on `[validatedFlow, state]` in addition to scroll (cheap, no MutationObserver needed).
- Skip the footer wrapper when the trailing node is hidden: reuse the `isVisible` check before the action-bar branch (or render `{renderComponent(last) && <div …/>}`-style gating so a null child collapses the strip).

</details>

rules: PERF-C07, STYLE-C14
agent: omp
run: n/a
</details>

apps/dashboard/src/components/Chat/ChatList/ChatListV4.tsx
ameer030: Re-checked at head `d5890ad`: **largely fixed, one residual**.

The new `ChatBubble` strip does close the desktop paths — verified every thread-open route reads the stripped `replies.onOpenThread`: hover toolbar (gated at `ChatBubble.tsx:1092`), mobile bubble-tap (`shouldEnableMobileThreadOpen` requires it), keyboard shortcuts (get no handler), and the reply-count strip (gets the stripped object and also returns null when `replyCount` is 0). Chat rows have no other click path to a thread, and the SDLC panel's override is unreachable for these rows.

Residual on mobile: the long-press actions drawer is gated on `replies` truthiness, not on `onOpenThread`, so it still shows "Reply in Thread" for ephemeral cards — and the tap now no-ops. Filed as the new W01 comment above with the one-line fix (same predicate as the hover toolbar).

apps/backend/src/apps/controllers/chatController.ts
ameer030: Re-checked at head `d5890ad`: **still open** — `chatController.ts:442` is unchanged (`const appId = (req as any).auth?.appId ?? (req.body as Record<string, unknown>).appId;`). The fallback is still dead today (only mount is behind `authenticateApp`, which always sets `req.auth.appId` before `next()`), so it stays a nit — but the comment above the line still promises the value is the verified one, which the expression doesn't guarantee.

apps/backend/src/apps/platform-adapters/slack/request-transformers/chat.ts
ameer030: Re-checked at head `d5890ad`: **still open** — `transformPostEphemeral` still sets `conversationId: slackReq.thread_ts` (`request-transformers/chat.ts:175`), and `chatPostEphemeral` still passes only the properly resolved `threadResolution.conversationId` to `deliverEphemeralMessage`, never `args.conversationId`. The field remains dead and wrongly typed.

apps/dashboard/src/hooks/useEphemeralMessages.ts
ameer030: Re-checked at head `d5890ad`: **still open** — the casts are unchanged: `EMPTY as unknown as Conversation[]` (`useEphemeralMessages.ts:182`), the row literal `as unknown as Conversation` (`:219`), and `ephemeralThreadMessages as typeof base` in `ThreadPannel.tsx`. No regression, just unresolved from the last round.

ameer030:
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 0 |
| 🟡 Better to fix | 4 |
| 🔵 Nit | 3 |

### 🔁 Re-check
- 💬 **Existing threads:** 1 partially fixed (the dead-thread issue — residual re-filed as W01 above) · 3 still open — "appId body fallback" (`chatController.ts:442`), "dead `conversationId` field" (`request-transformers/chat.ts:175`), "hand-built rows behind casts" (`useEphemeralMessages.ts`)
- ✅ **Verified OK:** the new capability token (HMAC-SHA256 over a canonical payload, timing-safe compare behind a length pre-check, own audience string, bound to user + card, 30-min TTL, secret read at call time); the action dispatch path (token stripped before anything reaches the app's webhook, payload serialized once and signed with the correct app's secret, appId taken only from the verified token, webhook URL SSRF-guarded); recipient gating (must be a participant of the resolved channel, on both doors; user-room broadcast reaches only the recipient's sockets); Slack-door thread↔channel cross-check; route/middleware parity with the `chat.postMessage` sibling (same auth, permission, and channel validation); HTML/flow content sanitization unchanged from postMessage; cross-version behavior (old dashboards ignore the new socket events; an old backend plus new dashboard simply never fires).

### 📌 Verdict
No blockers — but 3 earlier comments are still open and the W01 fix has a mobile residual: close the open threads and the 4 better-to-fix items above before merge ❌

ameer030 (approved):

Mohan0544 (approved):

utkarsh-611 (approved):

PR #1858, opened 2026-09-15
feat: XYNE-63456 SDLC hubs without a repository (non-claw)

adarshr001-stack:
Hubs can now be created with no repository, and Ask AI works in them.

Hub without a repository
- repoIds optional on hub creation; hub-scoped schemas (tracks, artifact types, claw artifacts, discussions) accept repoId or channelId
- SdlcHubService.requireHubScope resolves a hub by repoId or channelId; artifacts store repoId null when the hub has none, and a channel-only write on a single-repo hub still records that repository
- SdlcScreen renders without a repository: Wiki and Repo Knowledge hidden, folders read from the hub channel, docs/tracks open, right panel fixed
- hub-addressed link routes (/sdlc/channels/:channelId/links)

Ask AI without a repository
- hub-mode Ask AI context and prompt (no sandbox/code/PR/baseline steps)
- claw routes for tracks, artifact types and links accept channelId, as the claw SDLC tools send it
- SdlcScreen opens Ask AI on any hub, pinning a repository only if present

Correct hub when a repository is shared by several hubs
- the open hub is passed through Ask AI -> claw -> run context, so the agent works in that hub instead of the actor's oldest one
- linkContext resolves the hub from the link's source entity

  - backend
  - dashboard

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description
POT: https://drive.google.com/drive/folders/1d-f7u7pM2VrlHBatrg4jl_2ZRTxg9iwb

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


apps/backend/src/routes/sdlc.ts
```diff
@@ -213,10 +213,13 @@ router.get(
     const conversationId =
       typeof req.query.conversationId === 'string' ? req.query.conversationId.trim() : '';
     if (!conversationId) throw new AppError('conversationId is required', 400);
+    const channelId =
+      typeof req.query.channelId === 'string' ? req.query.channelId.trim() : '';
```
ameer030, line 260: 🔴 **B01 — The pinned hub never reaches claw: SDLC tools still run in the actor's OLDEST hub, not the one being viewed**

This PR adds a `channelId` here so Ask AI can pin the open hub on a repository shared by several hubs. The pin travels dashboard → controller → clawAgentService → xyne-claw-auth — and is then thrown away at the last hop: xyne-claw-auth's `resolveSdlcRepositoryForUser` never appends it to this `/context` call. The server therefore resolves the actor's **oldest** hub membership (its original behavior), and that value becomes the run's trusted SDLC `channelId` — which **overrides** whatever hub the model passes to the tools. Net effect for the exact case this PR targets (repo R in hubs A and B, user chatting in B, member of both): `spaces-sdlc-list-artifact-types` / `list-tracks` read hub A, and `spaces-sdlc-mutate-artifact` creates the artifact in hub A — the wrong hub, silently. Nothing downstream corrects it.

Chain, verified at this head:

- Dashboard pins the open hub: `openSdlcAssistant` sends `researchContext: { type: 'repository', id, name, channelId: channel.id }` (`apps/dashboard/src/routes/SdlcScreen/SdlcScreen.tsx:1372-1378`).
- Controller rewrites it to the authorized hub: `apps/backend/src/controllers/xyneAIControllerV2.ts:386-388`.
- `clawAgentService.ts:538` forwards the whole object: `...(request.researchContext && { researchContext: request.researchContext })`.
- Backend accepts the hint: `apps/backend/src/routes/sdlc.ts:216-224` → `getRepositoryRunContext(actor, repoId, conversationId, channelId)` → `requireRepositoryRole(actor, repoId, false, channelId)` (`apps/backend/src/sdlc/SdlcHubService.ts:610-621`, membership pick at `:1976-1995`, `orderBy createdAt asc` + `memberships[0]` when no hint).

The break:

- xyne-claw-auth's `ResearchRepositoryContext` declares only `{ type, id }`: `apps/xyne-claw-auth/backend/src/lib/sdlc-repository-context.ts:15-18`.
- The fetch URL omits the param entirely: `…/repositories/${id}/context?conversationId=${…}` (`sdlc-repository-context.ts:37-41`). Both call sites ride it — `routes/agent-chat.ts:1150`, `routes/run-stream.ts:474-481`, which even casts the body to `{ type?: unknown; id?: unknown }`.
- The resolved `agentContext` is stored as the run's trusted `sdlcContext` (`run-stream.ts:944-947`, `agent-chat.ts:1560-1561`).
- `packages/xyne-claw-shared/src/sdlc/bindings.ts:18-24` (`trustedSdlcToolBindings`) prefers `context.channelId` and binds it into every `trustedBinding: 'hub' | 'repository'` tool: `mutateArtifact`, `listTracks`, `createTrack`, `listArtifactTypes` (`registry.ts:43,51-53`).
- `apps/xyne-claw/src/mcp.ts:96-101` (`applyTrustedMcpBindings`) is `{ ...params, ...bindings }` — the trusted (oldest-hub) channelId **overrides** the model's hub-B channelId.
- Backend `requireHubScope` honors the injected channelId, so listings come from hub A, `hubOf(folderId)` resolves hub A folders, and the write lands in hub A.

`(apps/xyne-claw-auth is not among this PR's 15 changed files.)`
</details>

In `apps/xyne-claw-auth/backend/src/lib/sdlc-repository-context.ts`:

1. Add `channelId?: unknown` to `ResearchRepositoryContext`.
2. After the existing `type === 'repository'` check, extract `const channelId = typeof researchContext.channelId === 'string' && researchContext.channelId.trim() ? researchContext.channelId.trim() : '';`.
3. Append `&channelId=${encodeURIComponent(channelId)}` to the `/context` URL when present.

Single function serves both agent-chat and run-stream. The backend already fail-closes on a foreign/unknown channelId, so no new trust is granted.
</details>

rules: BUG-C03, STYLE-C13
agent: omp
run: n/a
</details>

apps/dashboard/src/routes/SdlcScreen/SdlcScreen.tsx
```diff
@@ -3042,7 +3087,9 @@ export default function SdlcScreen(): ReactElement {
                 <div className='min-h-0 flex-1 overflow-auto bg-background p-7'>
                   {section === 'overview' && (
                     <section>
-                      <h1 className='mb-5 text-2xl font-semibold tracking-tight'>{repo.name}</h1>
+                      <h1 className='mb-5 text-2xl font-semibold tracking-tight'>
+                        {repo?.name ?? channel.name}
+                      </h1>
                       {showAccessWarning && (
```
ameer030, line 2928: 🔴 **B02 — Repo-less hubs always show a false "Repository connection needs attention" banner**

This card was correctly wrapped in `{repo && …}` — but the amber warning banner directly above it was not. On every hub without a repository, `accessCapabilities` is an empty array, so `readReady`/`writeReady` are `false` and `showAccessWarning` is always `true`. Every repo-less hub — this PR's headline scenario — opens its overview with a big amber GitHub-access error that is never true. Previously unreachable because the screen early-returned when `!repo`.

- Derivation: `apps/dashboard/src/routes/SdlcScreen/SdlcScreen.tsx:1081-1095` — `accessCapabilities = repo && Array.isArray(...) ? ... : []` → `capabilityReady('READ_REPOSITORY', …)` is `false` → `showAccessWarning = !readReady || !writeReady` = `true`.
- Render: `SdlcScreen.tsx:3093-3103` — `{showAccessWarning && (…)}` sits in the always-rendered overview section, outside every `repo` guard; the banner text comes from the `readReady === false` branch of `accessWarning`.
- The `if (!repo) return (…)` early return that made this unreachable was removed by this PR.
</details>

Gate the warning on the repository existing, e.g. in the derivation:

```ts
const showAccessWarning = Boolean(repo) && (!readReady || !writeReady);
```

(Or wrap the banner render in `{repo && showAccessWarning && …}`.)
</details>

rules: BUG-C06
agent: omp
run: n/a
</details>

apps/backend/src/sdlc/types.ts
```diff
@@ -83,7 +84,7 @@ export interface SdlcHub {
     actor: SdlcActor,
     input: UpdateSdlcBaselineDraftInput
   ): Promise<SdlcArtifact>;
-  listTracks(actor: SdlcActor, repoId: string, channelId?: string): Promise<unknown>;
+  listTracks(actor: SdlcActor, repoId: string | undefined, channelId?: string): Promise<unknown>;
   createTrack(actor: SdlcActor, input: CreateSdlcTrackInput): Promise<unknown>;
   linkContext(actor: SdlcActor, repoId: string, input: CreateSdlcLinkInput): Promise<SdlcLink>;
```
ameer030: 🟡 **W01 — The `SdlcHub` interface still declares the old `linkContext`/`unlinkContext` signatures**

`getRepositoryRunContext` and `listTracks` in this same interface were updated for the new hub addressing, but these two were missed: they still require a non-optional `repoId: string` and have no `channelId` parameter, while the implementation — and every route calling it — now passes `undefined` and a fourth argument. It compiles today only because routes instantiate the concrete `SdlcHubService` (and TypeScript methods are bivariant), so the exported contract quietly lies to any future consumer or mock.

- Stale declarations: `apps/backend/src/sdlc/types.ts:89-90` — `linkContext(actor, repoId: string, input)` / `unlinkContext(actor, repoId: string, linkId)`.
- Real implementations: `apps/backend/src/sdlc/SdlcHubService.ts:1536-1540` (`repoId: string | undefined`, plus `channelId?: string`) and `:1743-1747`.
- Call sites that rely on the new shape: `apps/backend/src/routes/sdlc.ts:362-368,375` (passes `undefined`; the channel route), `apps/backend/src/routes/sdlcClaw.ts:68-73`.
</details>

In `apps/backend/src/sdlc/types.ts`:

```ts
linkContext(
  actor: SdlcActor,
  repoId: string | undefined,
  input: CreateSdlcLinkInput,
  channelId?: string
): Promise<SdlcLink>;
unlinkContext(actor: SdlcActor, repoId: string | undefined, linkId: string): Promise<void>;
```
</details>

rules: STYLE-C13
agent: omp
run: n/a
</details>

apps/backend/src/routes/sdlc.ts
```diff
@@ -350,6 +353,30 @@ router.delete(
   })
 );
 
+// Hub-addressed twins of the two above, for a hub that covers no repository. The
+// link is placed by its channel either way; the repository was only ever the route.
+router.post(
+  '/channels/:channelId/links',
+  route(async (req, res) => {
```
ameer030: 🔵 **N01 — The `:channelId` in this URL is decorative — the link can land in a different hub than the URL names**

`linkContext` now scopes the link from the **source entity's** hub whenever the source has one, so a caller-supplied `channelId` is only a fallback. `POST /sdlc/channels/HUB_B/links` with a source canvas in HUB_A silently writes the link into HUB_A. Authorization is enforced on the resolved hub (`requireChannelRole`), so nothing is exposed — but a route that names one hub and writes into another lies about its scope and will confuse anyone auditing it. The DELETE twin ignores the param entirely.

- Fallback-only use: `apps/backend/src/sdlc/SdlcHubService.ts:1544-1559` — `sourceChannelId ? { channelId: sourceChannelId } : { …repoId, channelId }`.
- DELETE ignores it: `apps/backend/src/routes/sdlc.ts:374-375` — `unlinkContext(actor, undefined, req.params.linkId)`.
</details>

Either 400/404 when a caller-supplied `channelId` differs from the resolved hub, or drop the param from the route (address by `/links` + body fields only).
</details>

rules: SEC-C01
agent: omp
run: n/a
</details>

packages/shared/src/sdlc.ts
```diff
@@ -781,26 +793,54 @@ export const createSdlcLinkSchema = z.object({
...
-  name: z.string().trim().min(1).max(80),
-});
+export const renameSdlcArtifactTypeSchema = z
+  .object({
+    repoId: z.string().min(1).optional(),
+    channelId: z.string().min(1).optional(),
+    folderId: z.string().min(1),
+    name: z.string().trim().min(1).max(80),
+  })
+  .superRefine((value, ctx) => {
```
ameer030: 🔵 **N02 — `channelId` satisfies this schema's own guard, then is silently ignored**

This superRefine accepts `channelId` as an alternative to `repoId` — but the PATCH route never forwards it: it calls `renameArtifactType(actor, input.repoId, input.folderId, input.name)`, and the hub is derived from the folder. Authorization is safe (the folder's hub is re-checked), but a caller can pass the gate with a field that then does nothing — exactly the kind of accepted-but-unread input that misleads the next person wiring a client.

- Schema: `packages/shared/src/sdlc.ts:829-843` (`renameSdlcArtifactTypeSchema` superRefine).
- Route: `apps/backend/src/routes/sdlcClaw.ts` PATCH `/artifact-types/:folderId` — `input.channelId` never read.
- Service derives the hub via `hubOf('canvasFolder', folderId)`: `apps/backend/src/sdlc/SdlcHubService.ts:1707-1720`.
</details>

Forward `input.channelId` as a fourth argument and 404 when it disagrees with the folder's hub — or drop `channelId` from `renameSdlcArtifactTypeSchema` until it is read.
</details>

rules: SEC-C24
agent: omp
run: n/a
</details>

packages/shared/src/sdlc.ts
```diff
@@ -781,26 +793,54 @@ export const createSdlcLinkSchema = z.object({
...
-  channelId: z.string().min(1).optional(),
-  name: z.string().trim().min(1).max(80),
-});
+export const createSdlcArtifactTypeSchema = z
+  .object({
+    repoId: z.string().min(1).optional(),
+    channelId: z.string().min(1).optional(),
+    name: z.string().trim().min(1).max(80),
+  })
+  .superRefine((value, ctx) => {
```
ameer030: 🔵 **N03 — Four copies of the same repoId-or-channelId refinement (two of them byte-identical)**

This same "repoId or channelId required" superRefine block appears four times in this file now: artifact creation (`:700`), tracks (`:803`), artifact types (`:819`) and rename (`:836`) — the last two are identical including the message string. The next wording change has to be remembered four times.

- `packages/shared/src/sdlc.ts:698-711`, `:803-810`, `:819-833`, `:836-850`.
</details>

One shared helper, e.g.

```ts
const requireRepoIdOrChannelId = (label: string) =>
  (value: { repoId?: string; channelId?: string }, ctx: z.RefinementCtx) => {
    if (!value.repoId && !value.channelId)
      ctx.addIssue({ code: z.ZodIssueCode.custom, message: `${label} require either a repoId or a channelId` });
  };
```

used by all four schemas.
</details>

rules: BUG-C03
agent: omp
run: n/a
</details>

apps/backend/src/sdlc/SdlcHubService.ts
```diff
@@ -799,34 +801,39 @@ export class SdlcHubService implements SdlcHub {
...
       : await this.prisma.canvasFolder.findFirst({
-          where: { channelId: repo.channelId, name: 'Baseline' },
+          where: { channelId: channelId, name: 'Baseline' },
           select: { id: true },
         });
     if (!folder) throw new AppError('Artifact type folder not found', 409);
 
     let baselineGenerationCommit: string | null = null;
     if (input.kind === 'BASELINE') {
+      if (!repo) throw new AppError('Baseline artifacts require a repository', 400);
```
ameer030: 🔵 **N04 — Duplicate unreachable baseline guard**

The identical `if (!repo) throw new AppError('Baseline artifacts require a repository', 400)` already ran at line 819, and `repo` is a `const` from the same destructure, so this second check can never fire.

- First (effective) guard: `apps/backend/src/sdlc/SdlcHubService.ts:817-820`.
- Unreachable twin: `apps/backend/src/sdlc/SdlcHubService.ts:836`.
</details>

Delete this copy.
</details>

rules: STYLE-C14
agent: omp
run: n/a
</details>

apps/backend/src/controllers/xyneAIControllerV2.ts
```diff
@@ -517,6 +523,47 @@ export class XyneAIControllerV2 {
...
         });
+      } else if (sdlcChannelId) {
+        const hub = await db.channel.findFirst({
+          where: { id: sdlcChannelId, type: ChannelType.SDLC },
+          select: { id: true, workspaceId: true },
+        });
+        if (hub) {
+          const contextLinks = await db.sdlcEntityLink.findMany({
+            where: { channelId: sdlcChannelId, relationType: 'CONTEXT' },
+            orderBy: { createdAt: 'desc' },
```
ameer030: 🔵 **N05 — Three independent lookups awaited one after another**

The hub lookup, this `contextLinks` fetch, and the `selectedCanvas` fetch all depend only on `sdlcChannelId` / `effectiveCanvasId` — not on each other — yet they run serially on a chat-request path. Only `resolveAuthorizedSdlcLinkedContext` genuinely needs the hub row (for its `workspaceId`). Two avoidable round trips of user-perceived latency.

- `apps/backend/src/controllers/xyneAIControllerV2.ts:527-552` — `db.channel.findFirst` → `db.sdlcEntityLink.findMany` → `db.canvas.findFirst`, sequential.
</details>

```ts
const [hub, contextLinks, selectedCanvas] = await Promise.all([
  db.channel.findFirst({…}),
  db.sdlcEntityLink.findMany({…}),
  effectiveCanvasId ? db.canvas.findFirst({…}) : Promise.resolve(undefined),
]);
```
</details>

rules: PERF-C16
agent: omp
run: n/a
</details>

apps/backend/src/sdlc/SdlcHubService.ts
```diff
@@ -799,34 +801,39 @@ export class SdlcHubService implements SdlcHub {
...
 
     const folder = input.folderId
       ? await this.prisma.canvasFolder.findFirst({
-          where: { id: input.folderId, channelId: repo.channelId },
+          where: { id: input.folderId, channelId: channelId },
           select: { id: true },
         })
       : await this.prisma.canvasFolder.findFirst({
-          where: { channelId: repo.channelId, name: 'Baseline' },
+          where: { channelId: channelId, name: 'Baseline' },
```
ameer030: 🔵 **N07 — `channelId: channelId` shorthand churn right after destructuring, and a `repo` local that isn't a repository**

Two small readability papercuts in this hunk:

1. `const { channelId, projectId, repo } = await this.requireHubScope(…)` is immediately followed by `where: { channelId: channelId, … }` / `projectId: projectId` in about eight places — the file itself already uses property shorthand elsewhere (`where: { id, channelId }` at `:2061-2064`), so the longhand reads like a sed rename.
2. In `linkContext` (`:1551`) and `unlinkContext` (`:1743`) the `requireHubScope` result is named `repo` and used as `repo.channelId` — while everywhere else in this file `repo` is an actual repository row (sibling call sites like `listTracks` already destructure the same result as `scope`).

- Shorthand sites: `apps/backend/src/sdlc/SdlcHubService.ts:825,829,922,924,941,953,988,995`.
- Misnomer: `apps/backend/src/sdlc/SdlcHubService.ts:1551` (`const repo = await this.requireHubScope(…)`) and `:1743-1752`.
</details>

Use shorthand in the `where` clauses, and rename the local to `scope` in `linkContext`/`unlinkContext` to match the other call sites.
</details>

rules: STYLE-C14
agent: omp
run: n/a
</details>

ameer030 (commented):
🟡 **W02 — You can create an empty hub, but you can never detach into one (guard + comment both stale)**

The creation side of this PR made zero-repository hubs fully legal — the schema drops `min(1)`, the screen renders them, the dialog's own screenshot shows one. But `removeChannelRepository` still throws `409 "A hub must keep at least one repository"` when removing the last repo, and its reason is the now-false comment *"With no repositories a hub renders nothing and cannot be deleted"*. Worse, the edit dialog now allows saving with **zero** repos selected (`disabled={!editing && …}`), so a user who deselects the last one gets a partial apply and then a 409 toast. The new redirect effect in SdlcScreen even cites "the repository being detached" as a reachable situation — the server forbids it.

- Guard + stale justification: `apps/backend/src/sdlc/SdlcHubService.ts:264-276` — `if (remaining <= 1) throw new AppError('A hub must keep at least one repository', 409)`.
- Creation allows empty: `packages/shared/src/sdlc.ts:276-280` (`repoIds` without `.min(1)`); `SdlcScreen.tsx:533-536` redirect comment names the detached case.
- Dialog: `apps/dashboard/src/routes/SdlcScreen/SdlcHubDialog.tsx:255-258` — save enabled at zero repos when editing.
</details>

Decide the policy once, and make all four places agree:

- **Empty hubs are allowed (the natural reading of this PR):** delete the `remaining <= 1` guard and rewrite the comment to state the new invariant.
- **Detach past one is intentionally forbidden:** keep the guard, fix the comment to say so, and re-disable save-at-zero-repos in the edit dialog (`repoIds.length === 0` back in the disabled condition when editing).
</details>

rules: DOCS-C04, BUG-C03
agent: omp
run: n/a
</details>

ameer030 (commented):
🟡 **W03 — The server-derived research context loses the hub pin the pinned path now carries**

Right above this branch you attach `channelId` when the **caller** pinned a repository. But when the caller sent no research context and the server derives the hub's repository itself, the derived object is built without `channelId` — so these two paths into the same downstream disagree about the hub pinning this PR exists to provide. Any caller that supplies `channelIds` without a pinned repository (web worker / API clients) silently reverts to the actor's oldest hub.

- Pinned path (gets the pin): `apps/backend/src/controllers/xyneAIControllerV2.ts:386-388`.
- Derived path (drops it): `xyneAIControllerV2.ts:417-422` — `effectiveResearchContext = { type: 'repository', id: sdlcRepo.id, name: sdlcRepo.name }`, `sdlcChannelId` in scope but not attached.
- The dashboard happens to always pin, so this is latent for its flow, but it is the fallback once the claw-side forwarding (B01) is fixed — the derived path would then be the one losing the pin.
</details>

```ts
effectiveResearchContext = {
  type: 'repository',
  id: sdlcRepo.id,
  name: sdlcRepo.name,
  ...(sdlcChannelId && { channelId: sdlcChannelId }),
};
```
</details>

rules: BUG-C03
agent: omp
run: n/a
</details>

ameer030 (commented):
🟡 **W04 — A track created on a repo-less hub is never opened**

The guard above was correctly relaxed to `if (!channel) return`, and track creation itself works hub-scoped — but this leftover `if (repoId)` means the track is created and the dialog closes **without** navigating to it on the exact hubs this PR adds. On repo-backed hubs the same action auto-opens the new track. Same button, two outcomes, split on repository presence for no functional reason.

- `apps/dashboard/src/routes/SdlcScreen/SdlcScreen.tsx:1750-1752` — `if (repoId) { navigateWithinSdlc('/sdlc/${channelId}/tracks', '?track=…') }`.
- The condition was vacuous before (the old `if (!repo || !channel) return` made it unreachable-false); now it false-paths exactly on repo-less hubs.
- Tracks render and `?track=` selection work fine without a repo — `openTrack` no longer requires it, and the tracks section is not in `REPO_DERIVED_SECTIONS`.
</details>

Delete the `if (repoId)` guard and navigate unconditionally.
</details>

rules: BUG-C03
agent: omp
run: n/a
</details>

ameer030 (commented):
🔵 **N06 — `sdlcChatLayout` still computes a `panelIds` field nobody reads**

SdlcScreen now derives the ResizableGroup's panel ids itself (`sdlcRightPanelIds(showRightPanel ? rightPanelMode : 'closed')`) and only ever reads `activeTab` and `panelOpen` off the layout object. The `panelIds` member of the returned object is dead — a second source of truth for the panel contract, so the next layout change will edit the wrong copy first.

- Dead field: `apps/dashboard/src/routes/SdlcScreen/sdlcChatPolicy.ts:29-42` (`panelIds` in the return value).
- Only consumers: `SdlcScreen.tsx:945` (`chatLayout.activeTab`), `:949` (`chatLayout.panelOpen`); panel ids come from `SdlcScreen.tsx:3070`.
</details>

Drop `panelIds` from `sdlcChatLayout`'s return value and its constant table.
</details>

rules: STYLE-C14
agent: omp
run: n/a
</details>

ameer030 (commented):
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 2 |
| 🟡 Better to fix | 4 |
| 🔵 Nit | 6 |

### 🔁 Re-check
- ✅ **Verified OK:** every hub-addressed write re-authorizes the actor against the resolved hub (`requireHubScope`/`requireChannelRole`: workspace-scoped, membership + admin-role) — no path reachable by a non-member was found; the link hub derived from the source entity is verified by `requireLinkSource`, and callers cannot forge `channelId` in the link row (explicit override after the spread); claw routes fail closed when both ids are absent; `?channelId=` on `/repositories/:repoId/context` can only narrow to hubs the actor belongs to; the claw tool schemas (`spaces-sdlc-mutate-artifact`, `spaces-create-ticket`) already accept the hub-addressed shape the new prompts teach; the `debuggerOpen` `Boolean()` fix and the `panelIds` panel-slot fix are correct; CI green.

### 📌 Verdict
2 blockers must be fixed before merge ❌ — the Ask-AI hub pin is silently dropped before it reaches claw (B01) and every repo-less hub shows a false GitHub-access error banner (B02).

rules: BUG-C03, BUG-C06, STYLE-C13, STYLE-C14, DOCS-C04, SEC-C01, SEC-C24, PERF-C16
agent: omp
run: n/a
</details>

PR #1865, opened 2026-09-15
fix: XYNE-63489 Implement Xyne AI and ticket tracking events

shubhamdhamija-juspay:
  - dashboard

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


apps/dashboard/src/services/Analytics/trackSource.ts
```diff
@@ -0,0 +1,26 @@
+import { NavigationType } from 'react-router-dom';
```
XyneSpaces, line 1: **[🟠 High · R7 / review guideline #204] No tests for the new tracking surface.** This PR adds three new analytics modules (`trackSource.ts`, `xyneAiTracking.ts`, `ticketTracking.ts`) and instruments ~60 files, but the diff contains zero test files. The metadata builders here (e.g. `lengthBucket`, the source maps) are pure functions — ideal unit-test targets. Suggested fix: add unit tests for the new modules (at minimum `lengthBucket` boundaries and source-map lookups) before merge.

apps/dashboard/src/components/AIScreen/AIComposer.tsx
```diff
@@ -977,7 +977,7 @@ export const AIComposer = forwardRef<AIComposerHandle, AIComposerProps>(function
                   aria-label='Send'
                   title='Send'
                   data-track-category='XyneAI'
-                  data-track-name='SEND_MESSAGE'
+                  data-track-name='SEND_BUTTON_CLICK'
```
XyneSpaces: **[🟡 Medium · R2] Analytics event renamed: `SEND_MESSAGE` → `SEND_BUTTON_CLICK`.** Any dashboard, funnel, or alert keyed on the old event name silently stops matching once this ships, and the PR description doesn't note the rename. Suggested fix: confirm no downstream consumers depend on `SEND_MESSAGE` (or emit both names for a transition window), and note the cutover in the PR description.

apps/dashboard/src/components/Chat/XyneAISidebar/components/XyneAIInputBox.tsx
```diff
@@ -1975,7 +1975,7 @@ export const XyneAIInputBox = forwardRef<XyneAIInputBoxHandle, XyneAIInputBoxPro
                           : 'bg-muted text-muted-foreground cursor-not-allowed'
                     }`}
                     data-track-category='XyneAI'
-                    data-track-name={isStreaming ? 'ABORT_MESSAGE' : 'SUBMIT_MESSAGE'}
+                    data-track-name={isStreaming ? 'ABORT_MESSAGE' : 'SEND_BUTTON_CLICK'}
```
XyneSpaces: **[🟡 Medium · R2] Same rename on the panel side: `SUBMIT_MESSAGE` → `SEND_BUTTON_CLICK`.** The two surfaces previously used *different* send-button names (`SEND_MESSAGE` on /ai, `SUBMIT_MESSAGE` in the panel) — unifying is good, but both historical series end here, so trend charts over send-button clicks will show an artificial cliff. Suggested fix: note both old names in the PR description so the analytics owner can stitch the series.

apps/dashboard/src/components/Chat/XyneAISidebar/XyneAISidebar.tsx
```diff
@@ -293,6 +292,60 @@ const XyneAISidebar = ({
...
+  // The one exception is AppRoot's `visible={false}` copy, kept mounted in a
+  // hidden div so an in-flight stream survives closing the drawer: nobody is
+  // looking at it, so it must not count as an open (or a close).
+  // `source` is whichever surface dispatched OPEN (machine `openSource`), or
+  // the prop for embedded instances. The otel ask_ai_opened counter rides the
+  // same effect so every entry point counts, not just the two that used to
+  // call it directly. No label: the channel name for a DM is a person's name.
+  const openedAtRef = useRef<number>(Date.now());
+  const messagesSentWhileOpenRef = useRef(0);
+  useEffect(() => {
```
XyneSpaces: **[🟡 Medium · review guideline #242] Mount-only open/close tracking is fragile, and the `eslint-disable` suppresses exactly the warning that would catch the regression.** The effect runs once on mount (`[]` deps) and early-returns when `!visible`. That's correct for the two-instance pattern AppRoot uses today (conditional copy while open + hidden keep-alive copy — verified in AppRoot), but if a future caller keeps *one* instance mounted and toggles `visible` instead of mounting/unmounting, `XYNE_AI_OPENED`/`XYNE_AI_CLOSED` silently never fire. The `react-hooks/exhaustive-deps` disable on this effect hides that class of break. Suggested fix: read visibility inside the effect from a ref/subscription rather than capturing it at mount, or cover the mount invariant with a test.

apps/dashboard/src/components/Chat/ChatInput/ChatInput.tsx
```diff
@@ -1233,9 +1255,16 @@ const ChatInputInner = forwardRef<InputBoxHandle, ChatInputProps>(
                           toast.error('Failed to create ticket', {
                             description: 'Please try again or contact support.',
                           });
+                          globalClickTracker.trackManualEvent(
+                            'Tickets',
+                            'CREATE_TICKET_FAILED',
+                            undefined,
+                            { source: 'chat_composer', errorKind: 'server' },
```
XyneSpaces: **[⚪ Nitpick · R6] `errorKind` hardcoded to `'server'`.** This catch also sees network/abort failures, but they'd all be labelled `server`, skewing the failure breakdown. `Summary.tsx` in this same PR derives `'http'` vs `'network'` from the error — worth doing the same here.

apps/dashboard/src/components/Chat/ChatInput/ChatInput.tsx
```diff
@@ -1217,7 +1223,23 @@ const ChatInputInner = forwardRef<InputBoxHandle, ChatInputProps>(
                             ...(conversationId && { sourceConversationId: conversationId }),
                           };
 
-                          await createTicket(ticketPayload);
+                          const created = await createTicket(ticketPayload);
+                          // No modal on this path, so the create outcome is reported here.
+                          globalClickTracker.trackManualEvent(
```
XyneSpaces: **[🟢 Low · review guideline #263] Second emit site for `CREATE_TICKET_SUCCEEDED` with a hand-built payload.** The modal path reports through the new `ticketTracking.ts` module while this composer path builds the payload inline (hardcoding `aiOriginated: false`, `hasAssignee: false`, `attachmentsCount: 0`). Two construction sites for the same event will drift. Suggested fix: expose a `trackTicketCreated(...)` helper in `services/Analytics/ticketTracking.ts` and call it from both paths.

XyneSpaces (commented):
## Xyne Spaces Reviewer — PR #1865

**Reviewed:** 30 page-1 files hunk-by-hunk (all `components/` changes, 342 added lines) + cross-file contract checks on the page-2 core (`openSource` machine wiring, `useXyneAIStream` otel-counter move, `standaloneNavigate` state param, AppRoot's dual sidebar instances) · **Findings:** 0 blocking, 1 high, 3 medium, 1 low, 1 nit

**Coverage note:** the GitHub files endpoint paginated at 30 files, so the remaining page-2 files (9 hooks, 13+ routes dirs, 3 new `services/Analytics` modules, `xyneAIMachine`, `AIOnboardingContext`, `electronApp`, page-2 components) were verified at file level via git tree-diff of `cdff545` vs base `958d717` — they follow the same mechanical `trackSource` / `data-track-metadata` wiring reviewed on page 1, but their hunks were not individually re-walked this run. All changes are under `apps/dashboard/src`; no prisma/zero/migration paths in the diff.

### Findings
| 🟠 high | R7 / #204 | services/Analytics/trackSource.ts:1 | 3 new analytics modules + ~60 instrumented files, zero tests in the diff |
| 🟡 medium | R2 | components/AIScreen/AIComposer.tsx:980 | `SEND_MESSAGE` → `SEND_BUTTON_CLICK` rename breaks existing analytics consumers |
| 🟡 medium | R2 | components/Chat/XyneAISidebar/components/XyneAIInputBox.tsx:1978 | `SUBMIT_MESSAGE` → `SEND_BUTTON_CLICK` — second historical series severed |
| 🟡 medium | #242 | components/Chat/XyneAISidebar/XyneAISidebar.tsx:309 | mount-only open/close effect + `eslint-disable` hides the toggle-`visible` regression class |
| 🟢 low | #263 | components/Chat/ChatInput/ChatInput.tsx:1228 | `CREATE_TICKET_SUCCEEDED` emitted from a second hand-built site |
| ⚪ nit | R6 | components/Chat/ChatInput/ChatInput.tsx:1262 | `errorKind` hardcoded `'server'` for all failure kinds |

**Checked, not flagged:** the otel counters removed from `XyneAISidebar.handleSubmit` are present in `useXyneAIStream` (move verified); `xyneAIMachine`'s OPEN action reads `event.trackSource` → `context.openSource`; `createTicket` returns the created ticket so `created?.id` is safe.
**Process nit:** the PR template's self-check checkboxes are all unchecked.

### Recommended reviewer
**@Radheyshree** — this PR builds directly on their XYNE-55815 thread-panel work (the `xyneAIActor.send({ type: 'OPEN' ... })` calls in `ThreadPannel.tsx` this PR instruments with `trackSource`), and their current review load is light.

| Candidate | Overlap with this PR | CODEOWNERS? | Open PRs (authored) | Tie-breaker |
|-----------|----------------------|------------|---------------------|-------------|
| @Radheyshree | ThreadPannel.tsx (XYNE-55815, direct ancestor commit) | no — CODEOWNERS covers ACL paths only | 1 (#1866) | most recent non-author committer in the PR's lineage |
| @harshpreet931 | 4 recordings route dirs (XYNE-63111) | no | 1 | — |
| @monika-annamaneni | dashboard components (XYNE-63479) | no | 0 | — |
| @shubhamdhamija-juspay | (author — excluded) | — | — | — |

Per-candidate *requested-reviewer* load isn't computable via the available API surface; open-PR-authored counts shown instead. The formal review-request API isn't available in this toolset, so this @mention is the request — @Radheyshree please pick this up.

**Top issue before merge:** add tests for the new analytics modules (🟠), and confirm/migrate the two renamed send-button events (🟡).

apps/dashboard/src/services/Analytics/trackSource.ts
shubhamdhamija-juspay: not needed

apps/dashboard/src/components/Chat/XyneAISidebar/XyneAISidebar.tsx
```diff
@@ -592,6 +594,77 @@ const XyneAISidebar = ({
...
+  // hidden div so an in-flight stream survives closing the drawer: nobody is
+  // looking at it, so it must not count as an open (or a close).
+  // `source` is whichever surface dispatched OPEN (machine `openSource`), or
+  // the prop for embedded instances. The otel ask_ai_opened counter rides the
+  // same effect so every entry point counts, not just the two that used to
+  // call it directly. No label: the channel name for a DM is a person's name.
+  // The open is a single moment, so the effect keys on `visible` alone and
+  // reads everything else off a per-render snapshot ref: the props are
+  // attribution, not triggers, and a prop change must not re-open the panel.
+  const openedAtRef = useRef<number>(Date.now());
```
XyneSpaces: **[⚪ 1 Nitpick] R2 — dead ref.** `openedAtRef` is initialized here and assigned again inside the open effect (`openedAtRef.current = openedAt`), but never read anywhere: the `XYNE_AI_CLOSED` payload computes `msOpen` from the closure `openedAt` const captured at open time.
↳ FIX: delete the ref (the closure const is doing the work), or read `openedAtRef.current` in the cleanup if the intent was to survive effect re-runs — as written it's dead code.

XyneSpaces (commented):
## Xyne Spaces Reviewer — PR #1865 (re-review @ `ae169097`)

**Verdict:** Changes requested — one high finding remains (no tests); everything else from the first pass was addressed.
**Reviewed:** new commits `1d13111e` + `ae169097` vs previously reviewed `cdff545` — 10 changed files: AIChatThread, AIComposer, ChatInput, XyneAISidebar, XyneAIInputBox, useXyneAIStream, xyneAiTracking, ticketTracking (page-1/2 modified) + new ActivityItemCard, ActivitySupportTicket.

### Prior findings — resolved ✅
| Prior finding | Resolution in these commits |
|---|---|
| 🟡 `SEND_MESSAGE` → `SEND_BUTTON_CLICK` rename broke the historical series (AIComposer.tsx) | Reverted — button keeps `data-track-name='SEND_MESSAGE'`, now with `aiSendButtonTrackingMetadata` (includes `trigger: 'button'`) |
| 🟡 `SUBMIT_MESSAGE` → `SEND_BUTTON_CLICK` rename (XyneAIInputBox.tsx) | Reverted — panel button keeps `SUBMIT_MESSAGE`; `onSubmit` now reports `'button' \| 'enter'` up to `handleSubmit` |
| 🟡 mount-only OPENED/CLOSED effect + `eslint-disable` (XyneAISidebar.tsx) | Effect re-keyed on `[visible]` with an `if (!visible) return` mount guard — AppRoot's hidden `visible={false}` keep-alive copy correctly emits nothing — plus a per-render snapshot ref; the eslint-disable is gone. Handles both mount/unmount and toggle-visible patterns |
| ⚪ hardcoded `errorKind: 'server'` (ChatInput.tsx) | `trackTicketCreateFailed` now derives `network` / `validation` / `server` from HTTP status |
| 🟢 duplicate `CREATE_TICKET_SUCCEEDED` emit (ChatInput.tsx) | Single emit via shared `trackTicketCreateSucceeded` helper |

Also verified in the rework: `submitQuery`'s new 11th `streamOverrides` param aligns at every call site (sidebar passes 4×`undefined` then options; AIChatThread's options stay in the same trailing position); submission-time otel counters now fire inside `useXyneAIStream.submitQuery` so the /ai page and the panel count identically; the `trigger !== 'button'` guard in the hook prevents double-counting button clicks (button's own click row + manual `SEND_MESSAGE`); the desk-draft fork failure now `return`s instead of proceeding with a stale draft (good drive-by fix).

### Findings (new hunks only)
|---|---|---|---|
| 🟠 4 High | 204 | `services/Analytics/**` | **Standing** (not re-commented inline): ~1,000 lines of new tracking surface — trigger union, snapshot-ref open/close effect, send/button metadata builders, ticket helpers — and still zero test files anywhere in the PR |
| ⚪ 1 Nitpick | R2 | `XyneAISidebar.tsx:610` | dead `openedAtRef` (see inline comment) |

### Recommended reviewer
**@Radheyshree** — unchanged from the first pass: owns the XyneAI panel / ThreadPannel lineage this PR builds on (XYNE-55815 ancestor of `ThreadPannel.tsx`), only 1 other open PR in flight.

| Candidate | Owns (files in this PR) | Pending reviews | Tie-breaker |
|---|---|---|---|
| @Radheyshree | ThreadPannel.tsx lineage, XyneAISidebar area | 1 (#1866) | most recent non-author committer in this PR's lineage |
| @harshpreet931 | recordings routes | 1 | — |
| @monika-annamaneni | Chat area | 0 | — |

*(This bot has no reviewer-request API — please add **@Radheyshree** from the PR sidebar.)*

### Coverage method (re-review)
Page-1 (30 files) compared by blob SHA against `cdff545`: only the 4 modified + 2 new files above differ; the other 24 are byte-identical. Page-2: `AppRoot.tsx` and `xyneAIMachine.ts` verified byte-identical by blob SHA; `useXyneAIStream`, `xyneAiTracking`, `ticketTracking`, `XyneAIInputBox` verified by full-content fetch. Remaining page-2 files (routes/hooks/contexts) share no contract with the reworked code and are assumed unchanged — say the word if you want an explicit sweep of those too.

apps/dashboard/src/components/xyne-desk/EmailComposer/EmailComposer.tsx
```diff
@@ -445,6 +480,113 @@
...
+  trackCtxRef.current = trackCtx;
+  const trackTicketRef = useRef(trackTicket);
+  trackTicketRef.current = trackTicket;
+
+  // COMPOSER_OPENED / AUTO_DRAFT_SHOWN on mount, COMPOSER_ABANDONED on unmount.
+  // The composer is mounted only while it is open (the host toggles it), so
+  // mount and unmount are the open and the close. The send path sets sentRef
+  // so a successful send never also counts as an abandon.
+  useEffect(() => {
+    const composerSessionId = composerSessionIdRef.current;
```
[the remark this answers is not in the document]
shubhamdhamija-juspay: fixed

apps/dashboard/src/services/Analytics/trackSource.ts
```diff
@@ -0,0 +1,26 @@
...
+ * The navigating surface sets `state.trackSource` on `navigate()` / `<Link>`
+ * (useWorkspaceNavigate forwards NavigateOptions untouched). Two rules keep the
+ * attribution honest:
+ *   - POP means back/forward. React Router replays the ORIGINAL state on history
+ *     navigation, so without this check a back button would re-report whichever
+ *     surface the user first arrived from and inflate it.
+ *   - No state at all means a deep link or an unattributed caller, not an error.
+ */
+export function readTrackSource(locationState: unknown, navigationType: NavigationType): string {
+  if (navigationType === NavigationType.Pop) return 'history_pop';
```
sithaarth: **Hickey · high** — `NavigationType.Pop` means two different things here.

`useNavigationType()` returns `Pop` on the **initial render of a fresh page load**, not only on back/forward. So every deep link, pasted URL, new tab and hard refresh reports `source: 'history_pop'` instead of `direct`.

Knock-on inside this same PR: `SupportScreen.tsx:5169` only falls back to `?src=` when `stateSource === 'direct'`, so the `?src=new_tab` (cmd-click path) and `?src=citation` (`clawCitationUrl.ts:27`) branches are unreachable — dead code as shipped.

Fix: split the two meanings apart. Gate on POP **and** "past the first render", e.g. a `useRef(false)` mount latch set in an effect, and only treat POP as `history_pop` once the latch is true.

apps/dashboard/src/hooks/useXyneAIStream.ts
```diff
@@ -451,6 +472,58 @@ export const useXyneAIStream = ({
...
+      // ask; the run's outcome is RESPONSE_* from the stream manager.
+      // A button send is already a click row under the button's own name
+      // (SEND_MESSAGE on the page, SUBMIT_MESSAGE in the panel) with the run
+      // dims baked into its metadata, so it is not repeated here.
+      const trigger: XyneAiSendTrigger = isRegenerate
+        ? 'regenerate'
+        : isEditUserMessage
+          ? 'edit'
+          : (ov?.trigger ?? 'submit');
+      if (trigger !== 'button') {
```
sithaarth, line 488: **Hickey · high** — the `trigger !== 'button'` carve-out is defeated by the AI landing page.

`AIScreen.tsx:273` passes the landing composer's `onSubmit` through without the new 4th `trigger` argument. A button send on `/ai` therefore becomes `initialQuery` → `AIChatThread`'s auto-send, which forces `trigger: 'auto_send'` (`AIChatThread.tsx:1918`). This suppression only skips `trigger === 'button'`, so the send emits **two** `XyneAI/SEND_MESSAGE` rows: the button's own `data-track-metadata` row (`xyneAiTracking.ts:38`, trigger `button`) and the manual one (trigger `auto_send`).

Every first message from the AI landing page double-counts — exactly the case this carve-out exists to prevent. Fix: thread `trigger` through `AIScreen`'s `onSubmit` so the button path stays `button` end-to-end.

apps/dashboard/src/components/Tickets/TicketTable/useBulkTicketActions.ts
```diff
@@ -264,6 +304,16 @@ export const useBulkTicketActions = (): BulkTicketActions => {
         }
       }
       void reportBulkErrors(pending);
+      void Promise.all(pending).then(results => {
+        if (results.some(r => r === null)) {
+          trackTicketOutcome('TICKET_FIELD_UPDATED', null, {
+            surface: 'bulk',
+            field: 'tags',
+            bulkCount: capSelection(tickets).length,
```
sithaarth: **Hickey · medium** — `capSelection` complects "compute the capped selection" with "tell the user about the cap".

`bulkCount: capSelection(tickets).length` calls it a **second** time, and `capSelection` fires `toast.info("Bulk actions apply to N tickets at a time…")` as a side effect. Select more than `MAX_BULK_TICKETS` rows and bulk-apply tags: the cap notice appears twice, the second time seconds later once the mutations settle.

`applyUpdates` was refactored correctly (`const selection = capSelection(tickets)` hoisted once); `applyTags` was missed. Either hoist it the same way, or split the toast out of `capSelection` so calling it is pure.

apps/dashboard/src/components/RecapPanel/ProjectRecapPanel.tsx
```diff
@@ -34,6 +35,22 @@ const ProjectRecapPanel = (): ReactElement => {
...
+  const projectRecapViewedRef = useRef(false);
+  useEffect(() => {
+    if (projectRecapViewedRef.current || isLoading || recaps.length === 0) return;
+    projectRecapViewedRef.current = true;
+    globalClickTracker.trackManualEvent('RECAP_PANEL', 'RECAP_VIEWED', undefined, {
+      recapType: 'project',
+      projectCount: recaps.length,
+      channelCount: recaps.reduce((sum, r) => sum + (r.channelCount ?? 0), 0),
+      totalMessages: recaps.reduce((sum, r) => sum + (r.messageCount ?? 0), 0),
+      isToday: true,
```
sithaarth: **Hickey · low** — the `isToday` dimension means two different things depending on which panel emitted it.

`isToday: true` is hardcoded here while the comment two lines above says project recaps are *always yesterday's*. `RecapPanel.tsx` emits the **same** `RECAP_VIEWED` event with `isToday: !isHistoricalView`.

Any query grouping on `isToday` silently mixes "is this recap about today" with "always true for this surface". Fix: emit the honest value here, or rename the dimension in one of the two producers.

apps/dashboard/src/components/Tickets/TicketDetails/TicketDetails.tsx
```diff
@@ -1858,14 +1993,15 @@ export const TicketDetails: React.FC<TicketDetailsProps> = ({
           toast.error(result.error.message || errorFallback);
           return false;
         }
+        trackDetailsOutcome(update);
```
sithaarth: **Löwy · high** — `trackDetailsOutcome(update)` runs on the wrong clock: after `await zero.mutate(...).server`.

Zero applies the mutation **optimistically**, so by the time the await resolves `ticketRef.current` has already re-rendered to the new row. Every "before" dimension is therefore wrong:
- `previous: before.stageName` equals `to`
- `previous: before.priority` equals `to`
- `hadAssignee` reflects the post-change value
- `isBackward` compares `fromSeq`/`toSeq` derived from the same stage, so it is always `false`

The comment above this call ("read at call time so `previous` is the value the user saw") states the opposite of what the code does.

Fix: snapshot `before` **before** the await and pass the snapshot in.

Also flagged by the Hickey pass: the dispatcher at `1925-1994` braids *which field changed* with *how to describe it* — a chain of `if (typeof u['x'] === 'string')` with early returns, which silently drops multi-field updates (a `stageName` + `priority` update reports only the stage). A field→event descriptor table would fix both.

apps/dashboard/src/hooks/useDeskAIDraft.ts
```diff
@@ -12,6 +12,93 @@ import {
...
+
+export function getDeskDraftTrackingSnapshot(): Readonly<DeskDraftTracking> {
+  return deskDraftTracking;
+}
+
+/**
+ * Called by the composer on mount so every AIDraft event for this session
+ * joins back to its COMPOSER_OPENED / SEND_EMAIL_* rows.
+ */
+export function setDeskDraftTrackingSession(
```
sithaarth: **Löwy · medium** — this state's clock is per-composer-mount, but it lives in a module singleton.

`deskDraftTracking` is one module-level record and `setDeskDraftTrackingSession` resets it on **every** composer mount. But `SupportScreen` keeps `composeInstances` as an *array* (multiple compose windows), and a reply composer can be open at the same time. The last-mounted composer overwrites `composerSessionId` / `acceptedLength` / `refineCount` for all of them, so `DraftCard`'s Insert/Discard clicks and `SEND_EMAIL_SUCCEEDED`'s `aiDraftState` are attributed to the wrong composer.

`deskTracking.ts:newComposerSessionId` documents the opposite guarantee — "even when a compose modal and a reply composer are open on the same ticket at once".

Fix: move this into the composer instance (context or ref owned by the composer), not a module-level singleton.

sithaarth (commented):
## Spacetime review — PR #1865

Reviewed with the two-lens method from [The spacetime of code](https://kolu.dev/blog/hickey-lowy/): a **Hickey** pass (spatial — is this complected?) and a **Löwy** pass (temporal — what changes on a different clock?), run separately over the same diff. 114 files, ~5.4k added lines, almost entirely analytics instrumentation. Findings are inline; grouped summary below.

### Hickey (spatial / complecting)

**high**
- `services/Analytics/trackSource.ts:15` — `NavigationType.Pop` encodes both "back/forward" and "first render of a fresh load"; every deep link / new tab / refresh reports `history_pop`. Makes the `?src=new_tab` and `?src=citation` branches shipped in this PR unreachable.
- `hooks/useXyneAIStream.ts:488` — the `trigger !== 'button'` carve-out is bypassed because `AIScreen.tsx:273` drops the new `trigger` argument; every first message from the AI landing page emits two `SEND_MESSAGE` rows.

**medium**
- `Tickets/TicketTable/useBulkTicketActions.ts:312` — `capSelection` fuses "compute capped selection" with "toast the user"; `applyTags` calls it twice, so the cap notice shows twice.

**low**
- `RecapPanel/ProjectRecapPanel.tsx:50` — `isToday` means two different things across the two panels emitting `RECAP_VIEWED`.

**structural (no bug attached)**
- Four parallel outcome helpers (`trackTicketOutcome`, `trackDeskOutcome`, `trackDeskDraftOutcome`, plus ~40 ad-hoc `globalClickTracker.trackManualEvent` sites) each re-derive surface/source/ticket dimensions at the call site. `TicketDetails.trackDetailsOutcome` (`1925-1994`) then reimplements a mini dispatcher that braids *which field changed* with *how to describe it*, and silently drops multi-field updates. Worth a field→event descriptor table before this grows further.

### Löwy (temporal / volatility)

**high**
- `Tickets/TicketDetails/TicketDetails.tsx:1996` — outcome read after `await …server`, but Zero already applied optimistically. Every `previous:` dimension equals the new value and `isBackward` is always `false`. The comment above it claims the opposite.

**medium**
- `hooks/useDeskAIDraft.ts:58` — per-composer-mount state kept in a module singleton; multiple compose windows clobber each other's session.
- `xyne-desk/EmailComposer/EmailComposer.tsx:532` — `[]`-dep effect reads props from the first-render closure; the Zero-backed `ticketDraft` hasn't synced yet, so `AUTO_DRAFT_SHOWN` never fires.
- `Tickets/TicketListView/StagePicker.tsx:353` — outcome emitted on the optimistic clock; a rolled-back change records as success. Same pattern in `TicketDetails.tsx` (~2965, ~2978, ~3110) and `useTicketReferences.ts`.
- `hooks/useMarkEmailRead.ts:36`, `TicketDetails.tsx:2863` and `:5496`, `hooks/useDragAndDrop.ts:618` and `:694` — `.then(...)` added with no `.catch`; previously-`void`ed rejections now surface unhandled.

**low**
- `Chat/Summary.tsx:351` — `fetchStartedAtRef` never cleared, so a cache hit followed by a failure reports a latency from an unrelated earlier request.
- `xyne-desk/EmailComposer/EmailComposer.tsx:285` — `useRef(newComposerSessionId())` runs `crypto.randomUUID()` on every render and discards it.

apps/dashboard/src/components/xyne-desk/EmailComposer/EmailComposer.tsx
```diff
@@ -445,6 +480,113 @@ export const EmailComposer = ({
...
+      draft: draft ? { autoDraftStatus: draft.autoDraftStatus ?? null } : null,
+    }),
+    [channelPreference?.deskType, draft],
+  );
+  const trackCtxRef = useRef(trackCtx);
+  trackCtxRef.current = trackCtx;
+  const trackTicketRef = useRef(trackTicket);
+  trackTicketRef.current = trackTicket;
+
+  // COMPOSER_OPENED / AUTO_DRAFT_SHOWN on mount, COMPOSER_ABANDONED on unmount.
```
sithaarth: **Löwy · medium** — mount clock vs. Zero sync clock.

This `COMPOSER_OPENED` / `AUTO_DRAFT_SHOWN` effect has `[]` deps (eslint-disabled) but reads `hasAutoDraft`, `draft`, `conversationId` and `ticketId` from the **first-render** closure. `ticketDraft` comes from a Zero query (`SupportScreen.tsx:4748`) that starts empty; pressing `r`/`a` on arrival mounts the composer before that row syncs.

Result: `hadAutoDraft: false`, `AUTO_DRAFT_SHOWN` never fires, and both the unmount `COMPOSER_ABANDONED` and the send-time `aiDraftState` treat an auto-draft reply as `none`.

Fix: mirror these props into refs and read the refs — this file already does exactly that for `trackTicket` / `trackCtx`.

apps/dashboard/src/components/Tickets/TicketListView/StagePicker.tsx
```diff
@@ -351,6 +352,15 @@ export function StagePicker({
         void zero.mutate(
           mutators.ticket.update({ id: ticketId, stageName: next, updatedAt: Date.now() }),
```
sithaarth: **Löwy · medium** — the outcome event is emitted on the optimistic clock, not the confirmed one.

`trackTicketOutcome('TICKET_STAGE_CHANGED', …)` fires immediately after a bare `void zero.mutate(...)`, so a server-rejected and rolled-back stage change is still recorded as a successful outcome. The other two branches in this same function correctly gate on `serverResult?.type !== 'error'` / `surfaceMutationError(...).then(ok => …)`.

Same unconditional pattern elsewhere in the PR: `TicketDetails.tsx` (tags ~2965 and ~2978, dynamic field ~3110) and `useTicketReferences.ts` (`handleRemoveReference`, `handleReferenceRelationChange`).

apps/dashboard/src/hooks/useMarkEmailRead.ts
```diff
@@ -23,13 +24,28 @@ export function useMarkEmailRead(
...
+    void zero
+      .mutate(
+        mutators.emailRead.markAsRead({
+          id: uuidv4(),
+          ticketId,
+          lastReadEmailId: latestEmailId,
+          updatedAt: Date.now(),
+        }),
+      )
+      .client.then(() => {
```
sithaarth: **Löwy · medium** — `.client.then(...)` with no `.catch` turns a later-arriving client-mutator failure into an unhandled promise rejection.

Before this PR the result was `void`ed and never touched, so the rejection had no handler to miss. The sibling hook `useMarkTicketsAsRead.ts` deliberately uses `.client.catch(...)` for the same mutator family — match it.

apps/dashboard/src/components/Tickets/TicketDetails/TicketDetails.tsx
```diff
@@ -2702,15 +2850,24 @@ export const TicketDetails: React.FC<TicketDetailsProps> = ({
...
+      .mutate(
+        mutators.ticketStageEta.update({
+          id: entryId,
+          stageEta: newStageEtaDate.getTime(),
+          updatedAt: Date.now(),
+          ticketId: ticket.id,
+          stageId: currentStage.id,
+        }),
+      )
+      .server.then(result => {
```
sithaarth, line 2872: **Löwy · medium** — same class as `useMarkEmailRead.ts:36`: `void zero.mutate(...).server.then(...)` with no `.catch`.

`zeroMutationToast.ts` documents that `.server` resolves for application errors, but its own implementation still `try/catch`es it — i.e. transport-level rejection is expected. This site (and `:5496`, plus `useDragAndDrop.ts:618` and `:694`) previously `void`ed the mutation and now surfaces that rejection as unhandled.

apps/dashboard/src/components/Tickets/TicketDetails/TicketDetails.tsx
```diff
@@ -5298,16 +5485,17 @@ export const TicketDetails: React.FC<TicketDetailsProps> = ({
...
+                    const backwardUpdate = {
+                      id: ticket.id,
+                      stageName: backwardStageChange.stageName,
+                      ...(backwardStageChange.newStatus && {
+                        statusV2: backwardStageChange.newStatus,
                       }),
-                    );
+                      updatedAt: Date.now(),
+                    };
+                    void zero.mutate(mutators.ticket.update(backwardUpdate)).server.then(result => {
```
sithaarth: **Löwy · medium** — same missing `.catch` on `.server.then(...)` as `:2863`. See that comment.

apps/dashboard/src/hooks/useDragAndDrop.ts
```diff
@@ -593,15 +605,29 @@ export const useDragAndDrop = ({
...
+              .mutate(
+                mutators.ticket.update({
+                  id: activeTicket.id,
+                  stageName: newStageName,
+                  ...(newStatus && { statusV2: newStatus }),
+                  ...(kanbanPosition !== undefined && { kanbanPosition }),
+                  updatedAt: Date.now(),
+                }),
+              )
+              .server.then(result => {
```
sithaarth, line 619: **Löwy · medium** — same missing `.catch` on `.server.then(...)` as `TicketDetails.tsx:2863`. A transport rejection here becomes an unhandled promise rejection where the pre-PR code `void`ed it.

apps/dashboard/src/hooks/useDragAndDrop.ts
```diff
@@ -657,13 +683,22 @@ export const useDragAndDrop = ({
...
-          );
+          void zero
+            .mutate(
+              mutators.ticket.update({
+                id: activeTicket.id,
+                statusV2: newStatus,
+                updatedAt: Date.now(),
+              }),
+            )
+            .server.then(result => {
```
sithaarth, line 704: **Löwy · medium** — same missing `.catch` on `.server.then(...)` as `:618`.

apps/dashboard/src/components/Chat/Summary.tsx
```diff
@@ -335,6 +348,7 @@ export const Summary = (props: SummaryProps): ReactElement => {
       if (abortControllerRef.current) abortControllerRef.current.abort();
       const abortController = new AbortController();
       abortControllerRef.current = abortController;
+      fetchStartedAtRef.current = Date.now();
```
sithaarth, line 359: **Löwy · low** — `fetchStartedAtRef` outlives the request it measures.

It is set on fetch but never cleared, and cache hits deliberately don't set it. So a summary served from cache followed by a later failure reports `latencyMs` measured from an unrelated earlier request — potentially minutes.

Fix: reset it to `null` on the cache-hit path (and after each terminal outcome).

apps/dashboard/src/components/xyne-desk/EmailComposer/EmailComposer.tsx
```diff
@@ -242,14 +270,21 @@ export const EmailComposer = ({
...
   const isComposeMode = mode === 'compose';
   const features = resolveFeatures(mode, featureOverrides);
   // Current user — used to scope localStorage compose drafts so two users
   // sharing the same browser don't see each other's in-flight drafts.
   const { userID } = useAuthContextValues();
+  // One id per composer mount. Every row of the reply funnel (COMPOSER_OPENED,
+  // AUTO_DRAFT_SHOWN, the AIDraft run events, SEND_EMAIL_*, COMPOSER_ABANDONED)
+  // carries it, so a ticket replied to three times, or a compose modal open
+  // beside a reply, still joins exactly.
+  const composerSessionIdRef = useRef<string>(newComposerSessionId());
```
sithaarth: **Löwy · low** — `useRef<string>(newComposerSessionId())` evaluates on the render clock, not the mount clock.

The argument to `useRef` is evaluated on **every render**, so this calls `newComposerSessionId()` (a `crypto.randomUUID()`) each time and throws the result away.

Fix: lazy-init — `useRef<string | null>(null)` assigned on first read, or `useState(() => newComposerSessionId())[0]`.

sithaarth: **Löwy · low** — `useRef<string>(newComposerSessionId())` evaluates on the render clock, not the mount clock.

The argument to `useRef` is evaluated on **every render**, so this calls `newComposerSessionId()` (a `crypto.randomUUID()`) each time and throws the result away.

Fix: lazy-init — `useRef<string | null>(null)` assigned on first read, or `useState(() => newComposerSessionId())[0]`.

sithaarth (dismissed):

Mayankbansal-juspay (approved):

samitbarai (approved):

PR #665, opened 2026-08-17
Feature/show kb file status

OmThakur2511:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

Added visibility for kb like how much file is ingested processing and failed,
Sending notifications once ingestion is completed for that collection
Added support for to upload files, folder from the google drive link
Set this env -> ENABLE_DRIVE_IMPORT_WORKER=true in generic worker
Add redirect url - https://<your BACKEND_URL>/api/drive/oauth/google/callback
 read google drive scope on to the google oauth consent - https://www.googleapis.com/auth/drive.readonly
 Need to deploy vespa worker and docling service
 
## POT

https://github.com/user-attachments/assets/37136ba4-5503-4300-9b7f-23bbaed60bce
image

Screenshot 2026-08-18 at 5 47 45 PM

Screenshot 2026-08-18 at 5 48 17 PM

## Areas Touched

- [ ] `apps/backend` (API / worker)
- [ ] `apps/dashboard`
- [ ] `apps/electron`
- [ ] `apps/xyne-claw` / `apps/xyne-claw-auth`
- [ ] `packages/shared` or another shared package
- [ ] Infra (docker, scripts, nix, CI)

---

## Zero (Sync) Changes

- [ ] No Zero changes — **skip this section**
- [ ] Adds/modifies a **mutator**
- [ ] Adds/modifies the **schema** (tables, columns, relationships)
- [ ] Adds/modifies a **query or ACL**

If any of the above:

- [ ] Server (`apps/backend/src/zero/mutators.ts`) and client (`apps/dashboard/src/zero/mutators.ts`) mutators mirror each other — same namespace, name, and args schema
- [ ] No `Date.now()` / `uuid()` generated inside a mutator
- [ ] Optimistic client result matches the server result (no state flash on rebase)
- [ ] New tables exported in `createSchema()` and given a Query ACL registered in `query-acl-factory.ts`
- [ ] Matching Prisma model + migration, client regenerated

## Schema & Data Changes

- [ ] No schema changes — **skip this section**
- [ ] Prisma schema modified (`apps/backend/prisma` and/or `prisma-common`)
- [ ] Migration added
- [ ] Backfill / data migration included

If any of the above:

- [ ] Clients regenerated (`db:generate`, `db:common:generate`)
[...]

PR #1870, opened 2026-09-15
fix: XYNE-63610 move docling OCR staging from pod-local disk to GCS

XyneSpaces:
## Problem

The async OCR scheduler (`apps/backend/src/services/ingestion/docling/`) staged split parts, per-part result JSON and the manifest on **pod-local disk** (`/tmp/docling-async/<fileId>/…`). The scheduler roles (splitter / submitter / result / writer) run on **different pods** in production (`vespa-ingestion-k8s`, `vespa-ingestion-k8s-02`, …), so whenever the writer pod differed from the pod that stored a part's OCR result, aggregation failed terminally:

```
Vespa write failed for 8974e1d0-892e-4ac9-b92f-cd8840285f59 after 1 attempts:
ENOENT: no such file or directory, open '/tmp/docling-async/8974e1d0-892e-4ac9-b92f-cd8840285f59/results/00000.json'
```

ENOENT is not a retryable Vespa error (`isRetryableVespaError` matches only 429/timeout/5xx/econnreset), so the file failed permanently after one attempt and stayed unindexed (repro: "ECICI - CPU2CPU - Client Facing Specification V4.1 _Consumer CIR.pdf" in the SME Agent KB collection on 2026-09-09).

## Fix

Staging moves to the **default storage bucket (GCS/S3 via the existing storage service)** under a new `DOCLING_SCHEDULER_STAGING_PREFIX` (default `docling-staging`), so any scheduler pod can read what any other pod staged.

| File | Change |
|---|---|
| `scheduler/storage.ts` | `partKey` / `resultKey` / manifest become relative object keys; `writePartBuffer` / `writeJson` upload via the storage service; `readBuffer` / `readJson` route absolute paths (legacy rows) to local fs and everything else to object storage; `cleanupStage` deletes staged objects **and** best-effort removes the legacy local stage dir |
| `workers/scheduler.ts` | submitter's ENOENT self-heal re-splits from the GCS source and reads the **fresh staging key** after re-staging (stored `partPath` may be a legacy local path) |
| `config/env.ts` + `.env.example` | new `DOCLING_SCHEDULER_STAGING_PREFIX`; `DOCLING_ASYNC_STORAGE_ROOT` documented as legacy-read/cleanup only |
| `scheduler/__tests__/storage.test.ts` | new — 8 unit tests: key layout, GCS roundtrips, legacy local-path reads, cleanup scoping |

## Compatibility / rollout

- **No DB migration.** Legacy rows carry absolute local paths in `docling_async_parts.part_path` / `result_path`; `readBuffer`/`readJson` still read those from local disk. New rows carry relative staging keys.
- **Rolling-deploy window:** mixed old/new pods can still fail cross-pod handoffs for the duration of the deploy (old writer can't read new GCS keys; new writer can only read legacy local paths if it happens to be on the same pod). Deploy while the docling queue is drained if possible.
- **Terminally failed files** (like the ECICI CIR PDF above) need a re-upload to re-ingest after this ships.
- Consider a GCS lifecycle rule on the `docling-staging/` prefix as a safety net for orphaned staging (cleanup deletes on success/`keepTempResults=false`).

## Testing

- `pnpm --filter backend test -- src/services/ingestion/docling/scheduler/__tests__/storage.test.ts` → 8/8 pass.
- `tsc --noEmit` clean for all changed files (remaining repo errors are pre-existing `src/zero/*` codegen staleness).

> Note: no XYNE ticket exists for this fix yet (searched Spaces) — happy to retitle with the ticket key once one is created. Related: the same logical fix was made in the `xyne-search` repo (Bitbucket PR #139).


apps/backend/src/services/ingestion/docling/scheduler/storage.ts
```diff
@@ -20,52 +28,81 @@ import { config } from '@/config/env'
...
 }
 
 export const cleanupStage = async (fileId: string): Promise<void> => {
+  // Object-storage staging: delete every staged object under the file's prefix.
   try {
-    await fs.rm(root(fileId), { recursive: true, force: true })
+    const service = getStorageService()
+    const staged = await service.listFiles(`${rootKey(fileId)}/`)
+    for (const file of staged) {
+      await service.deleteFile(file.name)
```
XyneSpaces: **[🟡 should-fix] R2/R6 — cleanupStage: first failed delete aborts the rest, and the deletes are sequential.** The whole loop sits in one try/catch, so a single `deleteFile` failure (transient GCS error, permission blip) skips every remaining object and only warns — the stranded objects hold extracted document content and there is no retry path. Each `deleteFile` also costs two sequential round-trips in the GCS service (`file.exists()` then `file.delete()` — `packages/storage/src/gcsService.ts`), so large multi-part files clean up slowly. Suggested fix: `await Promise.allSettled(staged.map(f => service.deleteFile(f.name)))` — or a per-item try/catch that keeps going — logging individual failures. (The legacy-local `fs.rm` branch below is already independent, which is the right shape.)

apps/backend/src/config/env.ts
```diff
@@ -542,7 +542,14 @@ const envSchema = Joi.object({
   // Set to true to restore full-content feeds for migrated attachments. Only consulted on
   // the Slack migration paths — live uploads and KB/collections are unaffected.
   FILE_CONTENT_ENABLED: Joi.boolean().default(false),
-  // Staging on the LOCAL filesystem (a tmp folder in the container). Single-pod only.
+  // Docling scheduler staging: split parts + per-part OCR results are written to
+  // the DEFAULT storage bucket (GCS/S3) under this prefix, so any scheduler pod
+  // can read them back — the roles (splitter/submitter/result/writer) run on
+  // different pods in production.
+  DOCLING_SCHEDULER_STAGING_PREFIX: Joi.string().default('docling-staging'),
```
XyneSpaces: **[🟡 should-fix] R2 — validate the staging prefix at boot.** `isLegacyLocalPath()` routes any *absolute* ref to the local filesystem, and every new key is `path.posix.join(stagingPrefix, fileId)`. If someone sets `DOCLING_SCHEDULER_STAGING_PREFIX=/tmp/docling-async` (an easy copy-paste from the legacy var a few lines below), every generated key becomes absolute → every read silently routes back to pod-local disk → the exact ENOENT this PR fixes returns, but only in the environment with the typo. A `..` segment would likewise escape the intended prefix. Suggested fix: add a Joi check that rejects values that are absolute or contain a `..` segment, so a bad value fails fast at startup.

apps/backend/.env.example
```diff
@@ -489,6 +489,10 @@ DATABASE_READ_REPLICA_POOL_URL=postgresql://xyne:xyne123@localhost:5433/xyne_dev
 DEFAULT_WORKFLOW_EXECUTOR=xyne-code
 DEFAULT_WORKSPACE_ID=
 DOCLING_ASYNC_SCHEDULER_ENABLED=false
+# Staging prefix in the default storage bucket for split parts + OCR results
+# (shared across scheduler pods). Keep stable once set.
```
XyneSpaces: **[🟡 should-fix] R3/R6 — make the `docling-staging/` lifecycle rule part of this change, not a "consider".** Cleanup is best-effort and warn-only (and per the finding on `storage.ts` can abort midway); terminally-failed files never run cleanup at all; and staged objects contain document-derived content (OCR chunks of user files) in the shared **default** bucket. A bucket lifecycle rule on the `docling-staging/` prefix (e.g. delete objects older than 7 days) bounds retention for the orphan case. Suggested fix: add the rule to the deploy checklist / infra config in this PR, or file the infra ticket now and link it here so it doesn't get lost after merge.

apps/backend/src/services/ingestion/docling/scheduler/storage.ts
```diff
@@ -20,52 +28,81 @@ import { config } from '@/config/env'
 import { logger } from '@/utils/logger'
 
 const pad = (n: number) => String(n).padStart(5, '0')
-const root = (fileId: string) => path.join(config.doclingScheduler.storageRoot, fileId)
+
+/** Staging prefix for one file — a relative object key, never absolute. */
+const rootKey = (fileId: string) =>
+  path.posix.join(config.doclingScheduler.stagingPrefix, fileId)
```
XyneSpaces: **[⚪ nit / hardening] R1 — `fileId` flows unsanitized into the object-key joins.** *Assumption:* I have not verified the full provenance of `fileId` in this diff — if it is always the server-generated DB id (cuid) of the docling file row, this is theoretical. But `rootKey()`/`stagingKey()` do `path.posix.join(prefix, fileId, …)` with no guard, and `cleanupStage` lists + deletes everything under the resulting prefix. If any caller can ever pass an id containing `../`, the key escapes `docling-staging/` and cleanup would delete objects outside the staging prefix — in the shared default bucket that holds all workspace attachments (previously the blast radius was one pod's `/tmp`). Cheap guard: in `rootKey`, throw (or strip) when `fileId` contains `/`, `\`, or `..`.

apps/backend/src/services/ingestion/docling/scheduler/__tests__/storage.test.ts
```diff
@@ -0,0 +1,182 @@
...
+  stagingPaths,
+  writePartBuffer,
+  writeJson,
+  readBuffer,
+  readJson,
+  cleanupStage,
+} from '../storage'
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const storageFactory = jest.requireMock('@/services/storage/storageServiceFactory') as any
```
XyneSpaces: **[⚪ nit] R8 — `as any` + eslint-disable.** The mock's shape is fully declared ~30 lines up in this same file, so the `jest.requireMock` result can be typed directly (an explicit interface for `{ getStorageService; __store; __service }`) instead of `as any` + `// eslint-disable-next-line @typescript-eslint/no-explicit-any`. Purely style — the tests themselves are good coverage of the key-layout / legacy-path / cleanup-scoping contract.

XyneSpaces (commented):
Automated review by **Xyne Spaces Reviewer** — 5 files, 9 hunks checked against the review guidelines. Result: **3 🟡 should-fix, 2 ⚪ nits, 0 🔴 blocking**. Details inline below; verdict + recommended reviewer in the top-level summary comment.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1870

**Verdict:** Approve-worthy after fixes — no 🔴 blocking issues; 3 🟡 should-fix, 2 ⚪ nits
**Reviewed:** 5 files, 9 hunks (+277/−43) · **Findings:** 0 blocking, 3 should-fix (major), 2 nits

### Findings
| 🟡 should-fix | R2/R6 | `scheduler/storage.ts:91` | `cleanupStage`: first failed `deleteFile` aborts the remaining deletes (whole loop in one try/catch) and each delete costs 2 sequential round-trips in the GCS service — use `Promise.allSettled` or a per-item try/catch |
| 🟡 should-fix | R2 | `config/env.ts:549` | `DOCLING_SCHEDULER_STAGING_PREFIX` is unvalidated — an absolute or `..`-containing value makes every key absolute, silently routing reads back to pod-local disk (the exact ENOENT this PR fixes); add a Joi check |
| 🟡 should-fix | R3/R6 | `.env.example:493` | lifecycle rule on `docling-staging/` should be mandatory, not "consider" — cleanup is best-effort/warn-only and staged objects hold document-derived content in the shared default bucket |
| ⚪ nit | R1 *(assumption)* | `scheduler/storage.ts:34` | `fileId` flows unsanitized into the key joins + list/delete — pre-existing, but the blast radius is now the shared default bucket; cheap guard in `rootKey` |
| ⚪ nit | R8 | `scheduler/__tests__/storage.test.ts:64` | `as any` + eslint-disable; the mock shape is declared locally and can be typed |

Checked and cleared while reviewing: `listFiles` in the GCS adapter auto-paginates (so cleanup sees all objects, not just the first page), the fresh-key read after re-split (`workers/scheduler.ts:154`) is correct for legacy rows, and GCS per-object uploads are atomic so the lost tmp-then-rename is fine. Drive-by note (file not in this diff): the header comment in `workers/result.ts` still describes staging as living on the local filesystem — stale after this PR, worth a one-line follow-up.

### Recommended reviewer
**@PriyanshuC132** — owns the docling-scheduler runtime/config area (PR #1604 touched `submitPermits.ts`, the Bull/Redis queue prefixing and the same `.env.example`/env-var surface this PR changes), and currently has **0 pending review requests**.

| Candidate | Owns (area in this PR) | Default reviewer? | Pending reviews | Tie-breaker |
|-----------|------------------------|------------------|-----------------|-------------|
| @PriyanshuC132 | docling scheduler runtime + env config (PR #1604) | no CODEOWNER coverage | 0 | most recent author in the docling-scheduler area |
| @MukherjeeSourish | docling chunk-output contract, `chunks`/`chunks_pos` (PR #495) | no CODEOWNER coverage | unknown* | — |
| @OmThakur2511 | ingestion status/deploy surface (PR #665) | no CODEOWNER coverage | 0 | — |

\* the `review-requested:` search for this login returned all 360 open PRs (including this one, which has no requested reviewers) — the qualifier was not honored, so the count is not computable; cross-checked the 50 most recent open PRs via REST and none of the three candidates is a requested reviewer there.

**Signal caveats (nothing invented):** the repo's CODEOWNERS only covers ACL paths (`database/acl`, `tenant`, `zero/acl`, mutators) — none of this PR's files — so ranking used docling-area PR authorship + workload instead; per-file commit authorship couldn't be computed (the available commits API has no path filter), so docling-related PR authorship is the ownership proxy.

**@PriyanshuC132** — requesting your review. This run has no GitHub API path for requesting reviewers (no request-reviewers tool, no `gh` CLI), so this @mention is the request; please also see the inline comments on `storage.ts`, `env.ts`, `.env.example` and the new test file.

apps/backend/src/services/ingestion/docling/scheduler/storage.ts
PriyanshuC132: Fixed in `df0821dc` (before this round of review landed — the cleanup was rewritten wholesale in that commit).

`cleanupStage` now deletes in batches of 16 through `Promise.allSettled`, so one failed delete no longer skips the remaining objects, and the sequential round-trips are gone. Failures are counted and reported rather than swallowed by a single early `warn`:

```ts
for (let i = 0; i < staged.length; i += CLEANUP_CONCURRENCY) {
  const batch = staged.slice(i, i + CLEANUP_CONCURRENCY)
  const results = await Promise.allSettled(batch.map((file) => service.deleteFile(file.name)))
  failed += results.filter((result) => result.status === 'rejected').length
}
if (failed > 0) { logger.warn('[DOCLING_SCHEDULER] Stage cleanup left objects behind', { fileId, staged: staged.length, failed }) }
```

The legacy-local `fs.rm` branch you pointed at as the right shape is gone entirely: that `/tmp` dir dies with its pod, and cleanup runs on the writer, which is usually a different pod, so it was almost never deleting anything.

apps/backend/src/config/env.ts
PriyanshuC132: Fixed in `df0821dc`, by removing the failure mode rather than validating against it: `DOCLING_SCHEDULER_STAGING_PREFIX` is gone and the prefix is a module constant in `scheduler/storage.ts`.

```ts
const STAGING_PREFIX = 'docling-staging'
```

The typo you describe had a worse consequence than a bad startup value: staging keys are persisted on `docling_async_files.stage_dir` / `docling_async_parts.part_path`, so changing the prefix between deploys strands every in-flight file even when the value is perfectly valid. It was never a per-environment knob — environments are separated by bucket — so a constant removes both the typo surface and the strand-on-change surface. `isLegacyLocalPath` is also gone; the legacy check is now just `path.isAbsolute(ref)`, since the old `startsWith(storageRoot)` clause was unreachable (the root is itself absolute).

apps/backend/src/services/ingestion/docling/scheduler/storage.ts
PriyanshuC132: Fixed in `6749f3ca`. Agreed on the reasoning — the blast radius argument is the important part, since `cleanupStage` lists-and-deletes under whatever prefix it builds, and that is now the shared default bucket.

```ts
const assertSafeFileId = (fileId: string): string => {
  if (!fileId || /[/\\]/.test(fileId) || fileId.includes('..') || fileId === '.') {
    throw new Error(`Unsafe docling fileId for a staging key: ${JSON.stringify(fileId)}`)
  }
  return fileId
}
```

Called from `stagingKey`, so every key-building path (`stagingPaths`, `partKey`, `resultKey`, `cleanupStage`) goes through it. Throws rather than strips: a caller passing something unexpected is a bug worth surfacing, and the scheduler already treats a throw as a retry/terminal failure.

On your assumption — you were right that it is theoretical today. `fileId` is the `collection_items.fileId` / `message_attachments.id` DB id on both intake paths, never user input. The guard is there to keep that invariant from quietly changing.

apps/backend/src/services/ingestion/docling/scheduler/__tests__/storage.test.ts
PriyanshuC132: Moot now — the test file has been deleted (`cb69ff7f`).

For the record the nit was addressed first, in `10d914b1`, with explicit `MockStorageService` / `MockStorageFactory` interfaces replacing the `as any` + eslint-disable. It was then removed along with the whole file: nothing in CI runs jest (the backend job is `tsc` only, and `tsconfig.json` excludes `**/*.test.ts`), so we opted not to carry a suite the pipeline never executes.

apps/backend/.env.example
PriyanshuC132: Partly fixed in code (`6749f3ca`); the bucket rule itself needs the infra repo.

The biggest orphan source was the one you named — terminally-failed files never ran cleanup at all, because the writer is the only caller of `cleanupStage` and a failed file never reaches it. Both terminal transitions now drop the stage:

```ts
const cleanupStageAfterTerminalFailure = (fileId: string): void => {
  if (config.doclingScheduler.keepTempResults) return;
  void cleanupStage(fileId);
};
```

Wired into `failDoclingFile` and `failDoclingFileIfOwned`, detached and best-effort so a failed cleanup cannot fail the transition, and honouring `DOCLING_KEEP_TEMP_RESULTS` so the debugging escape hatch still works. Combined with the batched `allSettled` cleanup, the remaining orphan cases are narrow: a cleanup that fails every retry-less attempt, and a file recovered by the sync fallback (which returns before the terminal path runs).

The lifecycle rule is not something this repo can apply — there is no bucket/terraform config here, same as the k8s manifests the previous isolation commit had to defer. So: **could you file the infra ticket and link it on this thread?** The ask is a rule on the `docling-staging/` prefix of the default bucket, delete after 7 days. I have documented the requirement in `scheduler/storage.ts` so it is discoverable from the code, but I would rather it be tracked by someone with infra access than left as a comment.

PriyanshuC132:
All five review findings are addressed; each inline thread has a reply with the commit and the reasoning. Summary:

| Finding | Status | Commit |
| --- | --- | --- |
| 🟡 R2/R6 `cleanupStage` aborts on first failed delete, sequential deletes | Fixed — batched `Promise.allSettled` (16 at a time), failures counted and reported | `df0821dc` |
| 🟡 R2 `DOCLING_SCHEDULER_STAGING_PREFIX` unvalidated | Fixed differently — the env var is **removed**, the prefix is a module constant | `df0821dc` |
| 🟡 R3/R6 `docling-staging/` lifecycle rule should be mandatory | Partly fixed in code; bucket rule needs the infra repo — **thread left open** for the ticket link | `6749f3ca`, `3b01cba9` |
| ⚪ R1 `fileId` unsanitized into key joins | Fixed — `assertSafeFileId` guards every key-building path | `6749f3ca` |
| ⚪ R8 `as any` + eslint-disable in the test | Typed explicitly, then moot — the test file was removed | `10d914b1`, `cb69ff7f` |

Two of these went further than suggested, for reasons worth flagging:

**The staging prefix was deleted rather than validated.** A bad value was the smaller problem: staging keys are persisted on `docling_async_files.stage_dir` / `docling_async_parts.part_path`, so changing the prefix between deploys strands every in-flight file even when the value is perfectly valid. Environments are separated by bucket, so it was never a per-environment knob.

**The orphan-retention finding got a code fix too.** The review correctly noted that terminally-failed files never run cleanup — the writer is `cleanupStage`'s only caller and a failed file never reaches it. Both terminal transitions now drop the stage (`failDoclingFile`, `failDoclingFileIfOwned`), detached and best-effort, honouring `DOCLING_KEEP_TEMP_RESULTS`. That narrows the orphan cases to a cleanup that fails outright and a file recovered by the sync fallback — which is what the lifecycle rule still needs to bound.

Drive-by note also fixed (`64440eea`): the `workers/result.ts` header no longer claims results are stored on the local filesystem.

One thing the review missed while checking the cross-provider behaviour: the self-heal and the splitter's terminal-failure check matched only GCS/`ENOENT` wording, so an S3 `NoSuchKey` or Azure `BlobNotFound` miss propagated instead of re-staging. That is fixed in `df0821dc` as `isMissingObjectError`, matched narrowly so a Postgres `relation ... does not exist` cannot fail a file terminally.

PR #1895, opened 2026-09-16
feat: XYNE-63458 heic attachment support for spaces

amitavac-dot:


apps/backend/src/services/heicRenditionService.ts
```diff
@@ -0,0 +1,222 @@
...
+// Decode + encode are CPU- and memory-bound; an uncapped flood of first-time
+// requests would run one full-image decode per request and risk OOM-ing the
+// pod. Excess requests queue behind this limiter.
+const HEIC_MAX_CONCURRENCY = Number(process.env.HEIC_MAX_CONCURRENCY) || 2
+const conversionLimit = pLimit(HEIC_MAX_CONCURRENCY)
+
+// Cached in GCS, keyed on the sha256 of the *original* bytes — attachments
+// are immutable once uploaded, so entries are content-addressed and never
+// need to expire. Renditions land in the same bucket as the original.
+const GCS_CACHE_PREFIX = "heic-rendition-cache"
```
sithaarth: **Renditions are never deleted with their attachment.** The cache lives at a flat `heic-rendition-cache/<sha256>/{kind}.webp`, outside the workspace-scoped `attachments/` prefix. Every delete path — `zero/mutators.ts:4473-4477` and `zero/utils/attachmentEntityCleanup.ts:25-27` — removes only `attachment.url` and `attachment.thumbnailUrl`.

So: user sends an iPhone photo, someone opens it (rendition generated), the message is deleted. The original is gone, but a full-resolution WebP copy of that photo stays in the bucket forever — and the comment above says entries "never need to expire", so nothing else will reap it. That's a retention/GDPR problem, not just wasted storage.

Either store renditions under the attachment's own prefix so the existing delete logic sweeps them, or add explicit rendition cleanup to both delete paths.

sithaarth (commented):
Review of the HEIC attachment support. Inline comments cover the issues I think are worth acting on before merge — four correctness/cost problems in the rendition service and controller, three client-side bugs, one duplication issue. Two more findings are in files this PR doesn't touch, plus an answer on the Electron question:

**`packages/shared/src/utils/fileTypes.ts:41`** — `UNRENDERABLE_MIMETYPES` still contains `image/heic`/`image/heif`, so `classifyMediaKind` returns `'file'` and `CanvasFilePanel.tsx:117` still demotes a dropped HEIC to a non-renderable file chip while chat now renders it. The PR changed the fact but left this owner of the fact stale.

**`apps/electron/src/services/request-interceptor.ts:264`** — Electron strips the backend CSP and forces `script-src 'self' …` with no `'unsafe-eval'`. `heic2any@0.0.4` is asm.js + Emscripten embind, which calls `new Function` during init inside a blob Worker (inherits the document CSP), so the client-side draft preview should throw `EvalError` on desktop only. Server renditions (`?format=webp`, `/thumbnail`) are plain fetches and unaffected, so the photo appears correctly once sent — only the sender's pre-upload chip degrades. Either accept that (and confirm `setPreviewError(true)` renders something sensible) or add `'unsafe-eval'` there — note it has to be full `unsafe-eval`, not the narrower `wasm-unsafe-eval`, since the requirement is embind's `new Function`, not WASM. This one is static analysis, not runtime-verified.

**On whether Electron needs changes: no, dashboard-only is the right scope.** `apps/electron/src/app/config.ts` sets `useBundledUI: false` for every flavor, so `window/manager.ts:314` always loads the deployed dashboard — this PR ships to desktop with zero Electron changes. There is also no Electron-side attachment path at all (no drag/drop, paste, clipboard-image or file-dialog handlers in `apps/electron/src`); everything goes through the dashboard DOM and produces real `File` objects, which is what `heic2any` needs. Native APIs would be worse: no image capability exists in the main process today (no sharp/canvas/ffmpeg, and `nativeImage` can't read HEIC), `build.prod.json` ships Windows and Linux targets so macOS `sips`/`qlmanage` covers maybe half the install base, and the `onlyLoadAppFromAsar` + ASAR-integrity fuses with no `nodeGypRebuild` make a native `.node` decoder a real packaging project — for a decoder sharp's prebuilts don't ship anyway, for the same licensing reason this PR needed `heic-decode`.

apps/backend/src/services/heicRenditionService.ts
```diff
@@ -0,0 +1,222 @@
...
+ * `storage` is the bucket the original lives in (see getAttachmentStorage) —
+ * renditions are cached alongside it, content-addressed by the original's
+ * sha256.
+ */
+export async function getHeicRendition(
+    storage: typeof storageService,
+    storagePath: string,
+    kind: HeicRenditionKind,
+): Promise<Buffer> {
+    const buffer = await storage.getFileBuffer(storagePath)
```
sithaarth: **The cache key is derived from the content it's caching, so every hit still downloads the whole original.** `getFileBuffer(storagePath)` → sha256 → build `cachePath` → check cache. On a cache hit this has already pulled a multi-MB HEIC out of GCS and hashed it, purely to learn the name of a file it was about to read anyway.

This module's own docstring says attachments are immutable once uploaded — which means `storagePath` is *already* a stable key. Key on `(storagePath, kind)` (hash the path, not the bytes) and the hit path collapses to one `fileExists` + one read of a small WebP. The full download then only happens on a genuine miss, where it's needed regardless.

At current scale this is the difference between a thumbnail grid costing one small read per photo and costing a full-resolution download per photo, per request.

apps/backend/src/services/heicRenditionService.ts
```diff
@@ -0,0 +1,222 @@
...
+            pipeline = pipeline.resize({
+                width: THUMB_LONG_EDGE,
+                height: THUMB_LONG_EDGE,
+                fit: "inside",
+                withoutEnlargement: true,
+            })
+        }
+
+        const webp = await pipeline
+            .webp({ lossless: true, quality: 100, effort: 5 })
```
sithaarth: **`lossless: true` for the `full` rendition inflates payloads ~10×.** A 12MP iPhone HEIC is ~2 MB; lossless WebP of the same decoded pixels lands around 15-30 MB. Opening one photo full-screen ships that over the wire and stores it in the cache bucket.

The header comment ("no fidelity is lost beyond the HEIC's own compression") is accurate but argues the wrong way: the source is already lossy, so bit-exactness relative to the decoded pixels buys nothing visually for a display rendition. Lossless at `effort: 5` on 12MP is also seconds of CPU per image, under a global `pLimit(2)`.

Suggest `{ quality: 85 }` lossy for `full` — the original HEIC is still downloadable byte-exact for anyone who needs it.

apps/backend/src/services/heicRenditionService.ts
```diff
@@ -0,0 +1,222 @@
...
+const HEIC_EXTENSIONS = [".heic", ".heif", ".hif"]
+
+/**
+ * HEIC detection for an attachment. The client-supplied MIME type is checked
+ * first, but not every browser reports one for HEIC (Chrome on Linux reports
+ * application/octet-stream), so the filename extension is accepted as a
+ * fallback — the same dual check the client applies when deciding to request
+ * a rendition, so the two stay in agreement.
+ */
+export function isHeicAttachment(mimetype: string, originalFilename: string): boolean {
```
sithaarth: **`isHeicAttachment` and `toWebpFilename` exist twice and have already diverged.** The client copies are `apps/dashboard/src/services/heicAttachmentService.ts:32` and `:53`. The docstring here says the dual check is "the same dual check the client applies … so the two stay in agreement" — but this version strips MIME parameters (`.split(';')[0].trim()`) and the client's does a bare `toLowerCase()`, and the two filename derivations use different stem rules. Nothing enforces the agreement the comment asserts, and it isn't actually holding.

Same shape one level up: the rendition *spec* is written twice too — server is 1024px/lossless, client is 2048px/q0.9 — so the draft chip and the sent chip show materially different images of the same photo.

Moving the predicate, the filename rule and the spec constants into `@xyne/shared` and keeping two executors (sharp here, heic2any there) consuming one spec would make this a single point of change.

apps/backend/src/controllers/attachmentController.ts
```diff
@@ -395,7 +421,36 @@ export class AttachmentController {
...
+      // HEIC attachments have no upstream thumbnail (the image pipeline
+      // can't decode HEVC), so generate one lazily from the original on
+      // first request and serve the cached rendition afterwards.
       if (!attachment.thumbnailUrl) {
+        const filePath = normalizeStoragePath(attachment.url);
+
+        if (filePath && isHeicAttachment(attachment.mimetype, attachment.originalFilename)) {
+          try {
+            const service = getAttachmentStorage(attachment);
+            const thumbBuffer = await getHeicRendition(service, filePath, 'thumb');
```
sithaarth, line 488: **No negative caching — an unconvertible HEIC re-decodes forever.** Some HEICs can never convert: over `MAX_PIXELS` (50MP), animated `image/heic-sequence`, corrupt, or over 16383px on a side (WebP's hard dimension limit, which `MAX_PIXELS` doesn't catch for a panorama). Each of those runs a full libheif decode and fails, and nothing records the failure — so the next request runs it again.

`AttachmentPreview.tsx:82` then retries 5× with `forceRefresh`, so one bad photo in a channel is 5 full-resolution decodes queued against a 2-wide global limiter, starving thumbnail generation for everyone else on the pod.

Worth writing a small failure marker into the cache prefix (or caching the 500 for a while) so a permanent failure is paid for once.

apps/backend/src/controllers/attachmentController.ts
```diff
@@ -343,6 +344,31 @@ export class AttachmentController {
...
+          res.setHeader('Content-Length', webpBuffer.length);
+          setSafeDownloadHeaders(res, {
+            mimetype: 'image/webp',
+            filename: toWebpFilename(attachment.originalFilename),
+          });
+          setAttachmentCacheHeaders(res, attachment);
+
+          res.send(webpBuffer);
+          return;
+        } catch (error) {
```
sithaarth, line 402: **Silent fallback to the original HEIC produces a broken image with no error state.** On conversion failure this logs and falls through to the normal path, which serves the original HEIC bytes — but the client has already committed to the response being WebP: `FileViewerModal.tsx:482` and `attachmentViewerMachine.ts:314` both build `new File([blob], toWebpFilename(...), { type: 'image/webp' })` from whatever comes back.

So the browser gets HEIC bytes labelled `image/webp` and renders a broken ``, while the client's `.catch` never fires and no error UI is shown. The fallback is only useful for a caller that can handle HEIC, and on `?format=webp` there is none.

Returning 415 or 502 on this path would let the client's existing error handling take over.

apps/dashboard/src/components/ui/files/AttachmentPreview.tsx
```diff
@@ -68,19 +72,83 @@ export const AttachmentPreview: React.FC<AttachmentPreviewProps> = ({
   });
   const isTextFile = getMimeType(file) === 'text/plain' || getFileName(file).endsWith('.txt');
   const fileId = getFileId(file);
+  // HEIC can't render from local bytes in most browsers — drafts convert it
+  // client-side, uploaded HEICs use the server-generated WebP thumbnail.
+  const isHeic = isHeicAttachment(getMimeType(file), getFileName(file));
```
sithaarth: **The extension fallback is unreachable at the call sites that need it most.** `isHeicAttachment` accepts a `.heic` filename when the MIME type is missing — which the comments correctly identify as the Chrome-on-Linux `application/octet-stream` case. But reaching this line requires `category === 'image'`, and `getFileCategory` (`ui/utils/files.ts:71`) is MIME-prefix-only. Same for `isImageFile` (`Chat/MessageAttachment/utils.tsx:235`), which gates `MessageAttachment.tsx:178` and `FileViewerModal.tsx:84`.

So exactly the file the fallback exists for — a `.heic` uploaded as `application/octet-stream` — is classified `'file'` and never reaches any `isHeic` branch, in the chip, the message, or the viewer.

The extension check needs to move up into `getFileCategory`/`isImageFile` for the fallback to do anything.

apps/dashboard/src/services/heicAttachmentService.ts
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * Download URL for the WebP rendition of an HEIC attachment. Accepts either
+ * an attachment id or an already-resolved `/attachments/:id/download` URL.
+ */
+export function heicWebpDownloadUrl(source: string): string {
+  const base =
+    source.startsWith('/') || source.startsWith('http')
+      ? source
+      : `/attachments/${source}/download`;
+  return `${base}?format=webp`;
```
sithaarth: **Unconditional `?format=webp` breaks for callers that don't pass a bare id or a clean path.** Callers pass `file.fileUrl`, which is not always `/attachments/:id/download`: `CreateTicketModal.tsx:2757` passes `attachment.url`, `ChannelCommandMenu.tsx:2239` passes `internalUrl`, and `MessageAttachment.tsx:959` passes `''`.

A URL that already has a query string becomes `…?a=b?format=webp`, and the empty-string case hits the `else` branch and produces `/attachments/?format=webp/download`. Both silently fetch the wrong thing.

`const sep = base.includes('?') ? '&' : '?'` fixes the first; the empty/relative case needs a guard. The same concatenation is hand-rolled again at `FileViewerModal.tsx:166` and has the same issue.

sithaarth (dismissed):

Mohan0544 (dismissed):

apps/backend/src/services/heicRenditionService.ts
```diff
@@ -0,0 +1,321 @@
...
+// default limitInputPixels and comfortably above any real phone camera output
+// while keeping a maliciously large ispe header from OOM-ing the pod.
+const MAX_PIXELS = 50_000_000
+const WEBP_MAX_SIDE = 16383
+
+// Decode + encode are CPU- and memory-bound; an uncapped flood of first-time
+// requests would run one full-image decode per request and risk OOM-ing the
+// pod. Excess requests queue behind this limiter.
+const HEIC_MAX_CONCURRENCY = Number(process.env.HEIC_MAX_CONCURRENCY) || 2
+const conversionLimit = pLimit(HEIC_MAX_CONCURRENCY)
```
santhoshJus: **Blocking: the HEIC decode blocks the Node event loop, and `pLimit(2)` does not help.**

I pulled `heic-decode@2.1.0` and read `lib.js`:

```js
await libheif.ready;                    // the only real await
const decoder = new libheif.HeifDecoder();
const data = decoder.decode(buffer);    // synchronous WASM
// decodeImage(): image.display(...) -- synchronous WASM + callback
```

The `async` wrapper awaits module init and nothing else. Container parse, HEVC decode and RGBA conversion are all synchronous WASM calls on the main thread, so this limiter gives queueing, not parallelism — two "concurrent" slots can still only run one decode at a time, and the event loop is stalled for the full duration of each.

Failure mode: the first view of a channel with N iPhone photos fires N thumbnail requests, each a full-resolution decode. The pod stops serving everything else — Zero mutations, API calls, health probes — for the combined decode time. `MAX_PIXELS = 50M` bounds a single decode but not the aggregate stall.

`sharp`'s resize/encode is fine, it runs on the libuv threadpool. It is specifically `decode.all(...)` / `image.decode()` that needs to move off the main thread (`worker_threads` / piscina), or the rendition should be generated at upload time through the existing job queue (`vespaQueue` is already in this controller's neighbourhood) instead of on the request path.

apps/backend/src/services/heicRenditionService.ts
```diff
@@ -0,0 +1,321 @@
...
+}
+
+function cacheBasePath(storagePath: string): string {
+    const normalized = normalizeStoragePath(storagePath)
+    const pathHash = createHash("sha256").update(normalized).digest("hex")
+    return `${GCS_CACHE_PREFIX}/${pathHash}`
+}
+
+function isPermanentFailure(code: HeicRenditionError["code"]): boolean {
+    return code === "NOT_HEIC" || code === "TOO_LARGE"
```
santhoshJus, line 74: **Blocking: a misdetected file becomes a permanent hard error where it used to work.**

`isHeicAttachment` falls back to the filename extension, so anything named `.heic`/`.heif`/`.hif` routes into the rendition path regardless of its bytes — a renamed JPEG, or an AVIF whose *major* brand is `mif1` (it's in `HEIC_FTYP_BRANDS`, and heic-decode's own brand table has the same entry, so it is accepted and then fails to decode without AV1).

`decode.all` throws -> `NOT_HEIC` -> `isPermanentFailure` -> a failure marker written to GCS **with no expiry**. From then on `/thumbnail` returns 500 and `?format=webp` returns 502 forever, for a file that rendered fine before this PR. `AttachmentPreview.fetchHeicThumbnail` burns its 5 retries against the marker and lands on an error chip.

Two small fixes:

1. On `NOT_HEIC`, fall through to the existing non-rendition response (serve the original / return the current 404) instead of erroring. The predicate is a routing hint, not a guarantee.
2. Now that `isHeicBuffer` runs at ingestion and `uploadFiles` returns the corrected `mimeType`, the server can trust the stored MIME type and treat the extension fallback as advisory for rows written after this lands.

`TOO_LARGE` has the same shape — a permanently-known condition answered with a 500 the client retries 5 times. A fall-through (or 404/415) lets the UI show the normal file chip instead of an error state.

santhoshJus (commented):
Two blocking items on the rendition service; the rest of the review is non-blocking and I can add it if useful.

apps/backend/src/services/heicRenditionService.ts
amitavac-dot: Addressed. No longer blocking the main thread, separate queue and worker for heic attachment decoding.

apps/backend/src/config/env.ts
```diff
@@ -145,6 +145,7 @@ const envSchema = Joi.object({
   TAG_GENERATION_CONCURRENCY: Joi.number().integer().min(1).max(20).default(1),
   TAG_GENERATION_LLM_TIMEOUT_MS: Joi.number().integer().min(1000).default(120000),
   ENABLE_STITCH_WORKER: Joi.boolean().default(false),
+  ENABLE_HEIC_RENDITION_WORKER: Joi.boolean().default(true),
```
santhoshJus: **The blocking decode moved to the shared worker process, and this defaults it on.**

`heicRenditionQueue.ts`'s own docstring has it right: *"the decode blocks THIS process's event loop while it runs — a busy deployment should isolate it on a dedicated node that enables only ENABLE_HEIC_RENDITION_WORKER."* But with `default(true)`, out of the box it lands on `worker.ts`, which is not a single-purpose process — it hosts the stitch worker, scheduled messages, conversation ingestion, AI provisioning, social sync, workflows, and the workflow-step GCS sync queue.

So a burst of iPhone uploads now stalls all of those for seconds at a time instead of the API. `WORKER_CONCURRENCY = 2` doesn't help: the libheif decode is synchronous, so two "concurrent" jobs still decode one at a time on that process's single thread (the comment in the queue file says as much).

Every sibling flag defaults to false — `ENABLE_STITCH_WORKER`, `ENABLE_AI_PROVISIONING_WORKER`. Either match that and enable this only on a dedicated node, or run the decode in a `worker_threads`/piscina pool inside the processor so whichever process hosts it stays responsive. As written the stall is relocated, not removed.

apps/dashboard/src/machines/attachmentViewerMachine.ts
```diff
@@ -303,6 +308,17 @@ export const attachmentViewerMachine = createMachine(
...
           }
 
+          // HEIC cannot render from its original bytes in most browsers; fetch
+          // the backend's lossy (q85) WebP rendition instead (generated + cached
+          // server-side on first request).
+          if (isHeicAttachment(mimeType, fileName)) {
+            return fetchFile(
+              heicWebpDownloadUrl(attachmentId),
+              toWebpFilename(fileName),
+              'image/webp',
```
santhoshJus, line 318: **The NOT_HEIC/TOO_LARGE fall-through renders a broken image with no error state.**

The server side of the fall-through is right — `attachmentController.ts:429-436` serves the original with `attachment.mimetype`. The client throws that away: this passes a hardcoded `'image/webp'` into `fetchFile`, which does `new File([blob], fileName, { type: mimeType })`. Same at `FileViewerModal.tsx:482-484` and `:523-525`.

So when the server falls through, the original HEIC bytes arrive labelled `image/webp` and `` silently renders nothing — no error, no fallback chip. sithaarth flagged this exact shape earlier in the thread against the previous silent-fallback version; the server-side fix re-exposes it.

Fix: build the `File` from `blob.type` when the response isn't WebP, and drop to the non-image fallback.

The thumbnail path has the same problem in different clothing: the fall-through lands on the existing 404, and `AttachmentPreview.fetchHeicThumbnail` burns its 5 retries over ~8s before showing an error chip — for a renamed JPEG the normal image path would render fine. Falling back to `createPreviewUrl(attachmentId)` on a 404 closes that one.

santhoshJus (commented):
Re-reviewed at 4696316b. Both blocking items are addressed server-side — decode is off the request path and NOT_HEIC/TOO_LARGE now fall through instead of erroring. Two follow-ups from the fix itself; I also have four smaller ones (client 503/PENDING handling, the worker/reader bucket mismatch, stale comments) if you want them.

apps/backend/src/config/env.ts
amitavac-dot: Using 2 worker threads now which will spawn only upon first call. No env gate as this should be on every node. If impact on perf is evident we can use one thread. Both decoding heic->rgba and encoding rgba->webp are now handled by the worker and off the main event loop.

apps/dashboard/src/machines/attachmentViewerMachine.ts
amitavac-dot: Addressed.

santhoshJus (approved):

sithaarth (approved):

knaman2609 (approved):

PR #1914, opened 2026-09-16
feat: XYNE-63631 roll up Recap by unread window with thread-clustered…

MukherjeeSourish:
https://github.com/user-attachments/assets/b780825f-3cac-4e40-aa2c-e89eba916be7

## Type of Change
- [x] New feature

## Description
Recap rolls a channel's daily rows into one card, grouped by thread.

- **Thread grouping** — points citing the same conversation merge into a topic with
  a bold title and nested bullets. A topic appears only when its newest point falls
  inside the selected window (the anchor rule); older points ride along as dimmed
  prior context, clipped at the channel's last-seen day so read days never resurface.
- **Time ranges** — Yesterday / Last 7 days / Last 14 days / Custom. One fixed 30-day
  fetch backs every pill, so switching never refetches.
- **Yesterday means the newest recap that exists**, not the calendar date. Generation
  runs early each morning and skips quiet days; pinning to yesterday left the pane
  empty whenever it hadn't run. Every mode anchors there, keeping ranges nested.
- Read controls stay in the Yesterday pane — a backdated range would consume older
  unread days through the single lastSeenRecapDate watermark.

XYNE-63631

## Areas Touched
- [x] `apps/dashboard`

## Zero (Sync) Changes

Two read-only queries, mirrored in `apps/backend` and `packages/shared`:
`channelRecapsRange` (recapDate BETWEEN) and `conversationsByIds` (batch title
lookup; `initialMessage` gated on `visibleTo IS NULL OR visibleTo = ctx.userID`).
Both go through `defineQuery`, so `applyQueryACL` wraps them via the existing
`RecapsACL` / `ConversationsACL`. **No mutator changes** — mark-as-read calls the
same `markChannelRecapAsRead` + `markChannelAsViewed` pair as today.

- [x] Optimistic client result matches the server result
- [ ] New tables exported in `createSchema()` — n/a, no new tables

## Schema & Data Changes

## Config, Environment & URLs
- [x] No config changes — skip this section

## API Contract
- [x] No API changes

## How did you test it?
Manually against local Postgres + Zero, asserting on DB rows, not just the UI.

- Ranges: Yesterday "September, 16th" · Last 7 "Sep 10 – Sep 16" · Last 14
  "Sep 3 – Sep 16" · custom Sep 10→16.
- Latest-recap fallback: moved the five Sep 16 rows out of range — panes became
  Sep 15 / Sep 9–15 / Sep 2–15 and the subtitle followed. Before the change the
  pane read "Nothing important happened yesterday." Restored, all back to Sep 16.
- Custom clamps its end to the newest recap, so the chip never promises an empty day.
- Title fallback: stripped `topicTitle` from a row; titles resolved from the thread's
  initial message, truncated at 80 chars.
- Read state: mark-all at 08:40:09 set `lastViewedAt` = 09-17 08:40:09 on all five
  (production's full clear); mark read/unread round-trips; **clicking a citation opens
  the thread without marking anything read**; mark-all is absent in the range panes.
- Console clean apart from Google Fonts/APIs, unreachable in the test browser.

### Automated
- [x] Not covered by tests — the dashboard has no test runner wired up. The clustering
      logic is isolated in `RecapPanel.merge.ts` as a pure function so it can be
      unit-tested once one exists.

### Manual
**Users**
- [x] Single user

**Login**

**Run mode**

**Sync & resilience**
- [x] Reload / hard refresh — no duplicate or lost rows

**Surface & data**
- [x] Web browser
- [x] Empty state
- [x] Large / realistic data volume — 15 channels × 30 days (450 rows / 2,250 points /
      1,200 threads, no `topicTitle`): Yesterday ~300ms, Last 14 ~890ms, widest range
      ~4.5s for ~1,275 topics / 8,480 DOM nodes. No long tasks; the cost is unvirtualized
      render volume, not the queries.

Not covered: two users at once, multiple tabs, offline replay, Electron, narrow
viewport, dark theme.

## Checklist
      (`vite build` needs `NODE_OPTIONS=--max-old-space-size=8192`; pre-existing)
- [x] Commits follow `<type>: <TICKET-ID> <subject>`; branch is `feature/*`
- [ ] Backend `typecheck` + `build` pass — `tsc` reports pre-existing
      `users.activityStatus` errors from a stale Prisma client after the rebase.
      None are in files this PR touches.


Pradeesh333 (dismissed):

mamtha-v (dismissed):

apps/backend/src/zero/queries.ts
```diff
@@ -751,6 +751,28 @@ export const queries: AnyQueryRegistry = defineQueries({
...
+  // Batch-fetch conversations by ID — resolves recap topic titles in one query
+  conversationsByIds: defineQuery(
+    z.object({ conversationIds: z.array(z.string()) }),
+    ({ ctx, args: { conversationIds } }) => {
+      if (conversationIds.length === 0) {
+        return zql.conversations.limit(0);
+      }
+      return zql.conversations
+        .where(helpers =>
+          helpers.or(...conversationIds.map(id => helpers.cmp('conversationId', id))),
```
santhoshJus: why are we using map of 'ORs' instead of 'IN'.

apps/backend/src/zero/queries.ts
```diff
@@ -751,6 +751,28 @@ export const queries: AnyQueryRegistry = defineQueries({
...
+        )
+        .related('initialMessage', im =>
```
santhoshJus: this is denormalised - initial_message_md column in conversations table

apps/backend/src/zero/queries.ts
MukherjeeSourish: moving to IN

apps/backend/src/zero/queries.ts
MukherjeeSourish: using the denormalised method now

apps/backend/src/zero/queries.ts
```diff
@@ -5628,6 +5641,31 @@ dmChannelsLatestMessagesPaginated: defineQuery(
     },
   ),
 
+  channelRecapsRange: defineQuery(
```
santhoshJus, line 5632: This is unbounded. There should be some limit on the days or the number of channels? And recaps do not change at all, so why need a query for it?

PR #1921, opened 2026-09-16
feat: XYNE-63645 Recording Access Levels

abhinavjuspay:
Recording Access Levels - https://drive.google.com/file/d/1z9QIztLdM46zXRxMfIrZDrt2AY4tFAEF/view?usp=sharing


apps/backend/src/services/recordingSharingService.ts
```diff
@@ -177,12 +192,6 @@ export class RecordingSharingService {
     await this.runTransaction(async tx => {
       const recording = await this.loadManageableRecording(tx, callId, actor);
       this.assertRecordingOnly(recording, 'Link access');
-      if (recording.createdByUserId !== actor.userId) {
-        throw new RecordingSharingError(
-          'Only the recording creator can change link access',
-          403,
-        );
-      }
       await tx.call.update({
```
XyneSpaces, line 195: **Blocker — link visibility is now editor-changeable, with no audit trail.** The creator-only guard that used to sit right here was removed, so `loadManageableRecording`'s default `'edit'` now covers `setLinkVisibility`. An editor can flip a private recording to `PUBLIC`, which is workspace-wide view. Unlike grant/revoke, this path emits no activity and no notification, so the owner never finds out. Either keep link access owner-only, or emit a `recording_access_changed`-style activity here and notify the owner.

apps/backend/src/services/recordingSharingService.ts
```diff
@@ -878,18 +912,19 @@ export class RecordingSharingService {
...
-      throw new RecordingSharingError(
-        isRecording(call)
-          ? 'Only the recording creator or people it is shared with can manage sharing'
-          : 'Only people in this call, or people it is shared with, can share it',
-        403,
-      );
+
+    const recording = isRecording(call);
+    const needed: CallAccessLevel = recording ? required : 'view';
+    if (await callShareService.hasAtLeast(call, actor.userId, actor.workspaceId, needed, tx)) {
```
XyneSpaces, line 918: **Blocker — editors can mint other editors and revoke each other.** `grantAccess` and `revokeAccess` both come through here needing only `'edit'`, and the controller schema still accepts `EntityUserAccess.EDIT`, so an editor can promote anyone to editor and can revoke the grant of the editor who invited them. The owner has no way to restrict it. If that is the intended Docs-like model, please state it in the PR description; otherwise keep promotion-to-editor and revoking *someone else's* grant owner-only, and let editors grant `VIEW` only.

apps/backend/src/zero/queries.ts
```diff
@@ -3174,7 +3198,11 @@ export const queries: AnyQueryRegistry = defineQueries({
             .where('entityUserAccess', '!=', EntityUserAccess.REVOKED)
             .related('user')
             .related('userGroup')
-            .related('channel'),
+            .related('channel')
```
XyneSpaces, line 3201: **Blocker — private channel names still sync to the client; the mask is UI-only.** `RecordingShareModal` renders `'Private channel'` for channels the viewer cannot reach, with the comment "Zero does not ACL-filter related() subqueries" — but this `.related('channel')` is unconditional, so the private channel row (name included) is still shipped into every viewer's Zero client and is readable in devtools. Fix it in the query rather than the render: gate the related channel on `visibility = PUBLIC` OR `exists('channelMembers', m => m.where('userId', ctx.userID))`, mirroring the pattern already used for `shares` just above. Same change is needed in the `packages/shared/src/zero/queries.ts` mirror.

apps/backend/src/database/acl/tables/calls-acl.ts
```diff
@@ -94,12 +98,36 @@ export class CallsACL extends BaseQueryACL<
...
   async getMutateWhere(): Promise<Prisma.CallWhereInput> {
+    const editableRecordingIds = await this.getEditableRecordingIds()
+
     return {
       workspaceId: this.ctx.workspaceId,
       OR: [
         { createdByUserId: this.ctx.userId },
         { participants: { some: { userId: this.ctx.userId } } },
+        ...(editableRecordingIds.length
+          ? [{ callType: CallType.HEADLESS, id: { in: editableRecordingIds } }]
```
XyneSpaces, line 129: **Latent grant with no consumer.** Every call mutator that matters (`markMoment`, the notes-canvas one) does its own `call.createdByUserId !== authData.sub` check, so this branch changes nothing today. What it does do is widen every ACL-scoped Prisma mutate on `Call` — deletes included — for any future code that trusts the ACL alone, while the actual policy (delete = creator-only) lives in `callController.deleteRecording`. Either drop the widening, or add the editor case where the enforcement really lives so the two cannot drift.

packages/shared/src/utils/recordingAccess.ts
```diff
@@ -0,0 +1,87 @@
...
+    null,
+  );
+
+/**
+ * Stored grant level -> effective level. EDIT is the only elevation a share
+ * carries; every other live value reads as view, so neither a legacy ADMIN row
+ * nor an unrecognised one can quietly elevate someone. REVOKED never reaches
+ * here — callers filter it out before ranking.
+ */
+export const recordingAccessFromGrant = (entityUserAccess: string): RecordingAccessLevel =>
```
XyneSpaces, line 37: **Fails open on unknown values.** `entityUserAccess === EDIT ? 'edit' : 'view'` means `REVOKED` — and anything unrecognised — resolves to **view**. That is only correct because every caller happens to pre-filter revoked rows today. Map `REVOKED` and unknown values to `null` and keep the caller-side filtering as defence in depth. Related: existing `ADMIN` rows now resolve to `view`, so anyone who previously held ADMIN silently loses sharing management — this needs a backfill (`ADMIN -> EDIT`) or an explicit "we accept the downgrade" note in the description.

apps/backend/src/controllers/recordingEmailController.ts
```diff
@@ -176,8 +176,8 @@ export class RecordingEmailController {
       throw new RecordingEmailError('Recording not found', 404);
     }
 
-    const canView = await callShareService.canView(call, userId, workspaceId);
-    if (!canView) {
+    const required: CallAccessLevel = isRecording(call) ? 'edit' : 'view';
```
XyneSpaces, line 179: **Unannounced behaviour tightening.** This was `canView`; recordings now require `'edit'`, so viewers lose "email this recording". Looks deliberate, but it is a user-visible permission removal and should be called out in the PR description (same for the Google Docs list, which flipped the other way from owner-only to visible-to-all viewers).

apps/backend/src/services/recordingSharingService.ts
```diff
@@ -1069,17 +1078,21 @@ export class RecordingSharingService {
...
           create: {
             id: randomUUID(),
             canvasId,
             workspaceId,
-            role: CanvasRole.VIEWER,
+            role,
             ...targetFields,
           },
-          update: {},
+          update: { role },
```
XyneSpaces, line 1095: **This clobbers canvas roles granted outside the recording.** `update: {}` -> `update: { role }` means a person given EDITOR directly on a recording's summary canvas is demoted to VIEWER the next time anyone re-grants them recording VIEW. The comment above is honest about it and PRD §11.3 owns the reverse gap — please raise a follow-up ticket so it does not get lost.

apps/backend/src/zero/queries.ts
```diff
@@ -3079,6 +3079,18 @@ export const queries: AnyQueryRegistry = defineQueries({
...
+            .where('shareableEntityType', ShareableEntityType.NOTE_TAKER)
+            .where('entityUserAccess', '!=', EntityUserAccess.REVOKED)
+            .where(({ or, cmp, exists }) =>
+              or(
+                cmp('userId', ctx.userID),
+                exists('userGroupMemberships', m => m.where('userId', ctx.userID)),
+                exists('channelMembers', m => m.where('userId', ctx.userID)),
+              ),
+            ),
+        )
```
XyneSpaces, line 3093: **Not needed on this query.** `createdOatsRecordings` is the caller's own recordings, and access resolution short-circuits to `owner` before grants are read — so this `related('shares')`, with its nested `exists` on `userGroupMemberships`/`channelMembers`, is pure IVM fan-out per row on a paginated list. Safe to drop here; keep it on the shared-with-me query.

XyneSpaces (commented):
Reviewed the full diff. The direction is right: collapsing `canView` and the duplicated `hasActiveShare` implementations into a single `CallShareService.getEffectiveAccess` (with `tx` support, so the permission check and the write it guards see the same snapshot) removes a real drift risk, `shareEntityTypeFor` correctly keeps the `CALL` and `NOTE_TAKER` share namespaces apart, dropping `ADMIN` from the grant enum shrinks the surface, and sharing the resolver via `@xyne/shared` is what keeps the UI's `canEdit` from disagreeing with what the server enforces.

Three things I would not merge as-is, flagged inline: the dropped owner-only guard on link visibility (with no audit trail), editors being able to promote other editors and revoke each other, and the private-channel name leak that is masked in the modal but not in the Zero query.

Also worth addressing before merge:
- `recordingAccessFromGrant` fails open on unknown/`REVOKED` values, and legacy `ADMIN` grants have no migration path.
- "Leave" is a silent 200 no-op when access arrived via a group or channel — `revokeAccess` finds no `DIRECT_SHARE` row, `continue`s, and returns `{ shares: [] }` as success. The `canLeave` guard keeps the UI honest, but the API should signal the no-op.
- `listActiveGrants` fires two unbounded per-user reads (`userGroupMapping`, `channelParticipant`, neither filtered by workspace) on every `hasAtLeast`, and `getEditableRecordingIds` adds a third on every mutate. Pre-existing shape, but this PR multiplies it — worth memoising per request context.
- No tests. `packages/shared/src/utils/recordingAccess.ts` is a pure function and the ideal unit-test target: highest-wins ranking, PUBLIC -> view-never-more, unknown grant handling.
- Stale comment: `appResourceAccessService.ts:16` still references `entityAccessService.hasActiveShare`, which this PR deletes.
- PR hygiene: the description is a single Drive link with no written summary of the permission changes or migration, and no reviewers are requested.

apps/backend/src/services/recordingSharingService.ts
abhinavjuspay: intended behaviour

apps/backend/src/services/recordingSharingService.ts
abhinavjuspay: Intended

apps/backend/src/zero/queries.ts
abhinavjuspay: just a comma addition, not the .channel

apps/backend/src/database/acl/tables/calls-acl.ts
abhinavjuspay: not needed.

apps/backend/src/services/recordingSharingService.ts
abhinavjuspay: intended.

packages/shared/src/utils/recordingAccess.ts
abhinavjuspay: All callers strip revoked rows first

apps/backend/src/controllers/recordingEmailController.ts
abhinavjuspay: intended

apps/backend/src/zero/queries.ts
abhinavjuspay: done

abhinavjuspay: This query defines the row type for both lists, so shares can't leave one without leaving the other — recovering it takes an as unknown as cast plus a type that claims a field created rows won't have.

apps/backend/src/services/recordingSharingService.ts
Mayankbansal-juspay: **Ticket-link sync silently demotes canvas editors.**

Changing `update: {}` to `update: { role }` makes every grant sync authoritative over the canvas role — but `linkTicket` always calls `setAccess(..., EntityUserAccess.VIEW, RECORDING_SHARE_INTENT.TICKET_LINK)` (line ~600), which now reaches here with `access: VIEW`.

Scenario: a recording is shared with channel `#eng` as **Editor** → `canvas_participants(canvasId, channelId=#eng).role = EDITOR`. The owner then links the recording to a ticket that lives in `#eng`. `linkTicket` upserts the *same* `canvasId_channelId` row with `role: VIEWER`, so everyone in `#eng` loses edit on the summary/notes canvas even though their recording `EDIT` grant is untouched. The old `update: {}` preserved the role.

Suggest skipping the role write when `intent === TICKET_LINK` (or taking `max(existing, new)` rank) so a view-only ticket link can't clobber a real EDIT grant.

apps/backend/src/database/acl/tables/calls-acl.ts
```diff
@@ -94,12 +98,36 @@ export class CallsACL extends BaseQueryACL<
...
+        entityUserAccess: EntityUserAccess.EDIT,
+        OR: await this.getShareTargetOr(),
+      },
+      select: { entityId: true },
+    })
+    return shares.map((s) => s.entityId)
+  }
+
   async getMutateWhere(): Promise<Prisma.CallWhereInput> {
+    const editableRecordingIds = await this.getEditableRecordingIds()
```
Mayankbansal-juspay: **Per-mutation cost added on the calls hot path, with no mutator that can use it.**

`getMutateWhere()` is evaluated by `acl-extension` for every `calls` update/delete that goes through Zero. It previously did zero I/O; it now awaits `getEditableRecordingIds()`, which is 3 queries (`userGroupMapping.findMany`, `channelParticipant.findMany`, `entityAccess.findMany`). Those run on the high-frequency call-flow mutators too — `calls.join`, `calls.leave`, `updateCallParticipantPreview`, `markMoment`, lobby approve/reject — none of which touch a shared recording.

And nothing currently benefits from the widening: every mutator that writes a HEADLESS recording still hard-rejects non-creators (`linkNotesCanvas` and `markMoment` both do `if (call.createdByUserId !== authData.sub) throw new Error('Access denied')`), so an editor is refused one layer up regardless of what this filter permits.

Either wire those mutators to the new level check, or drop the widening until something needs it.

apps/backend/src/services/callShareService.ts
```diff
@@ -24,30 +55,94 @@ export class CallShareService {
...
+      // A public link is view-only by design: it never confers edit rights and
+      // never share management, however the viewer arrived at the recording.
+      if (call.visibility === CallVisibility.PUBLIC) levels.push('view');
+    } else if (await this.isCallAudience(call, userId)) {
+      // Regular calls are readable by the people who were in them; recordings
+      // are not, so this route is deliberately not offered to a recording.
+      levels.push('view');
+    }
+
+    const grants = await this.listActiveGrants(call, userId, workspaceId, tx);
```
Mayankbansal-juspay: **Public-link reads lost their zero-query short circuit.**

The old `canView` returned `true` immediately on `call.visibility === PUBLIC` without touching the DB. `getEffectiveAccess` now always falls through to `listActiveGrants`, which is 3 queries (`userGroupMapping`, `channelParticipant`, `entityAccess`).

That path is hot: `canViewRecordings` is called from `attachmentAccessService` (line ~156) and `callController` (line ~2746) on every media/transcript fetch, and from the new `hasAtLeast(..., 'view')` gates at lines 1379/2068/2172. Streaming a public recording's audio now costs 3 extra round-trips per request where it previously cost none.

Cheapest fix: short-circuit in `hasAtLeast` — if the levels collected so far already satisfy `required`, skip `listActiveGrants` (only `getEffectiveAccess`'s exact answer needs the full scan).

apps/dashboard/src/routes/RecordingDetailV2Screen/RecordingDetailV2Screen.tsx
```diff
@@ -479,6 +500,10 @@ export default function RecordingDetailV2Screen({
     queries.oatsRecordingByExternalId({ callId: recordingId ?? '' }),
     { enabled: !!recordingId },
   );
+  // Editors get every editing surface the owner has. Delete stays owner-only
+  // (PRD §5.2), and so does the Google Docs list, whose links live in the
+  // owner's Drive and are dead ends for anyone else (§11.4).
+  const { canEdit } = useRecordingAccessLevel(recordingRow);
```
Mayankbansal-juspay: **The owner can render read-only when the Zero row hasn't arrived.**

`canEdit` is now derived exclusively from `recordingRow` (the `oatsRecordingByExternalId` Zero result), replacing `const isOwner = recording.createdByUserId === currentUser?.id`, which came from the REST-loaded detail.

But the screen renders as soon as the REST `recording` state is set — `recordingService.getRecordingDetail` sets it independently at ~L671 and ~L695. Whenever REST resolves first (cold Zero cache after a reload, a recording that just replicated, a Zero query that is slow or errors), `recordingRow` is `undefined`, `useRecordingAccessLevel` returns `level: null`, and the recording's **own creator** sees it read-only: title not editable (L1266/`RecordingDetailV2Header`), no regenerate / template controls (L1306), no post-to-channel, post-to-email or Google Docs export (L1645/1660/1677), and no summary-model footer (L1445). The controls then pop in when Zero catches up.

Suggest keeping the REST fallback as a floor, e.g. `const canEdit = accessCanEdit || recording?.createdByUserId === currentUser?.id;` (the server re-checks every one of these actions anyway).

apps/backend/src/services/recordingSharingService.ts
```diff
@@ -242,8 +251,12 @@ export class RecordingSharingService {
           RECORDING_SHARE_INTENT.DIRECT_SHARE,
         );
         shares.push({ id: change.share.id, target, access: change.share.entityUserAccess });
-        if (change.activated && target.type !== 'channel') {
-          activities.push({ shareId: change.share.id, action: 'recording_shared' });
+        if (target.type !== 'channel') {
+          if (change.activated) {
+            activities.push({ shareId: change.share.id, action: 'recording_shared' });
+          } else if (change.levelChanged) {
+            activities.push({ shareId: change.share.id, action: 'recording_access_changed' });
```
Mayankbansal-juspay: **A level-only change creates an empty DM channel.**

The UI's "change access" action (`RecordingShareModal.handleChangeAccess`) re-issues `action: 'grant'`, so it lands in this same `grant()` flow. Before the transaction, `grant()` unconditionally resolves a DM for every user target:

```ts
for (const userId of userIds) {
  const channelId = await repositories.channels.findOrCreateDMChannel(actor.userId, [userId], ...);
}
```

For a pure promotion/demotion the share already carries a post, so `!asSharePost(change.share.metadata)` is false and nothing is ever posted into that DM. Net effect: promoting Bob from Viewer to Editor silently creates an empty 1:1 DM channel between the acting editor and Bob (if they had none), which then sits in both sidebars with no messages.

Worth resolving the DM lazily — only for targets that will actually get a post.

apps/backend/src/database/acl/tables/calls-acl.ts
abhinavjuspay: Those mutators don't reach this code — getMutateWhere is the Prisma-side gate, while Zero mutators write through tx.mutate and are gated by zero/acl/wrappers/mutator-wrapper.ts

Mayankbansal-juspay (dismissed):

Mayankbansal-juspay (dismissed):

samitbarai (dismissed):

PR #1966, opened 2026-09-17
fix: XYNE-63755 ticket api fix

HarimohanSharma11:
  - backend
  - dashboard

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


apps/backend/src/zero/queries.ts
```diff
@@ -5104,7 +5109,7 @@ dmChannelsLatestMessagesPaginated: defineQuery(
   // Reverse of applicationReleaseTicketsByReleaseId: the release(s) a dev ticket
   // belongs to. Keep in sync with the shared copy.
   applicationReleaseTicketsByDevTicketId: defineQuery(
-    z.object({ ticketId: z.string().min(1) }),
+    z.object({ ticketId: z.string() }),
```
PiyushJuspay, line 5212: why this change ?

apps/backend/src/apps/controllers/ticketController.ts
```diff
@@ -208,6 +208,8 @@ const TicketFiltersSchema = z
     createdBy: toArrayFilter(z.string().trim().min(1)).optional(),
     userGroupId: toArrayFilter(z.string().trim().min(1)).optional(),
     tags: toArrayFilter(z.string().trim().min(1)).optional(),
+    merchantId: toArrayFilter(z.string().trim().min(1)).optional(),
```
PiyushJuspay, line 211: why we need both merchantId and hasMerchantId ?

HarimohanSharma11: They filter different things:
 - merchantId: tickets for specific merchants, e.g. "merchantId": ["m_1", "m_2"] becomes merchantId IN ('m_1', 'm_2').
 - hasMerchantId: tickets that have any merchant (true → merchantId IS NOT NULL) or none (false → merchantId IS NULL).
 -  A list of IDs can't express "any merchant" or "no merchant". The main use case here is fetching all tickets linked to a merchant without knowing the IDs up front, so merchantId alone doesn't cover it. hasMerchantId: true also lets the search run workspace-wide without boardIds/projectId, since it's an indexed column.

apps/backend/src/zero/queries.ts
HarimohanSharma11: change removed fixed after rebase

apps/backend/src/zero/queries.ts
```diff
@@ -418,6 +419,10 @@ const applyKanbanTicketPageConditions = (
     query = query.where('ticketType', 'IN', filters.ticketTypes);
   }
 
+  if (filters?.merchantIds?.length) {
```
gyash1512: should we add index also for this @santhoshJus

santhoshJus (dismissed):

apps/dashboard/src/components/Tickets/TicketFilters/Submenus/MerchantIdSubmenu/MerchantIdSubmenu.tsx
```diff
@@ -0,0 +1,175 @@
...
+  // Refetched every time the submenu opens: merchants appear as tickets are created,
+  // so a cached list goes stale quickly. The list is shown only once the response is in
+  // (isFetching, not isLoading) — a cached page of a few thousand rows would otherwise
+  // read as current while the real one is still loading.
+  const { data: merchants, isFetching } = useQuery({
+    queryKey: ['merchants'],
+    queryFn: getMerchants,
+    staleTime: 0,
+    refetchOnMount: 'always',
+  });
```
santhoshJus: **Perf: every open re-downloads the whole merchants table, and the cache never gets used.**

`FilterChip` mounts the picker through a Radix popover with no `forceMount`, so this component mounts on each open. With `staleTime: 0` + `refetchOnMount: 'always'`, that's one full-table `findMany` per open, per user — `routes/merchants.ts` has no `where` and no `take`. And because the list is gated on `isFetching` rather than `isLoading`, the cached result is never shown: every open is a cold spinner, so React Query is being paid for and used for nothing.

The comment says the table is "small", but nothing here establishes that. `merchants` gets a row per distinct `merchantId` ever seen on a ticket (`ticketRepository.ts:174` upsert), is global (no tenant column), and is never pruned. Worth running `select count(*) from merchants` before committing to fetch-all — the same missing tenant scoping also means any authenticated user in any workspace receives every merchant id in the deployment.

The codebase already has the pattern for large lists: `TagsSubmenu` takes `onSearch` + `onLoadMore` and searches server-side. I'd add `/merchants?q=&limit=50` and put `q` in the query key (debounced). If it has to stay client-side, at minimum `staleTime: 5 * 60_000`, drop `refetchOnMount: 'always'`, and gate on `isLoading` so a reopen paints instantly.

apps/dashboard/src/components/Tickets/TicketFilters/Submenus/MerchantIdSubmenu/MerchantIdSubmenu.tsx
```diff
@@ -0,0 +1,175 @@
...
+    const matches = [...all].filter(mid => !query || mid.toLowerCase().includes(query));
+
+    const selected = new Set(selectedMerchantIds);
+    return matches
+      .sort((a, b) => {
+        const bySelected = Number(selected.has(b)) - Number(selected.has(a));
+        return bySelected !== 0 ? bySelected : a.localeCompare(b);
+      })
+      .slice(0, 100);
+  }, [merchants, searchQuery, selectedMerchantIds]);
```
santhoshJus: **Perf: this sorts all N ids before slicing to 100, on every keystroke, with no debounce.**

Each keystroke allocates a lowercase string for all N ids, then `.sort()`s the **full** match set with `localeCompare` (Intl collation), then discards everything past 100. A short query doesn't help — `"m"` matches nearly everything, so the first keystrokes are the worst case. There's no debounce here, unlike the sibling `TicketTypeSubmenu`, which debounces 300ms on a list orders of magnitude smaller.

Measured on Node 20 (a browser will be slower), cost of one call:

| N | current | selected-first + early-break |
|---|---|---|
| 2,000 | 2.1 ms | 0.01 ms |
| 10,000 | 10.3 ms | <0.01 ms |
| 50,000 | 67.6 ms | <0.01 ms |

The sort is entirely avoidable: `routes/merchants.ts` already returns `orderBy: { mid: 'asc' }`, so emit the selected ids first, then scan the server order and break at 100.

```ts
const lowered = useMemo(
  () => (merchants ?? []).map(m => ({ mid: m.mid, low: m.mid.toLowerCase() })),
  [merchants],
);

const { visibleIds, selectedSet } = useMemo(() => {
  const selected = new Set(selectedMerchantIds);
  const query = searchQuery.trim().toLowerCase();
  const out = selectedMerchantIds.filter(mid => !query || mid.toLowerCase().includes(query));
  for (const { mid, low } of lowered) {
    if (out.length >= 100) break;
    if (selected.has(mid)) continue;
    if (!query || low.includes(query)) out.push(mid);
  }
  return { visibleIds: out, selectedSet: selected };
}, [lowered, searchQuery, selectedMerchantIds]);
```

`selectedSet` also kills the per-row `selectedMerchantIds.includes()` further down.

santhoshJus (commented):
Three perf findings on the new Merchant ID filter — the first two I'd treat as blocking, the third is a one-liner that makes the second much worse than it looks in isolation.

apps/dashboard/src/components/Tickets/TicketsHeader/FilterValuePicker.tsx
```diff
@@ -198,6 +199,13 @@ export const FilterValuePicker = ({
           onChange={types => setKey('ticketTypes', types)}
         />
       );
+    case 'merchantIds':
+      return (
+        <MerchantIdSubmenu
+          selectedMerchantIds={filters.merchantIds || []}
```
santhoshJus: `filters.merchantIds || []` allocates a fresh array on every render whenever the filter is unset, so `selectedMerchantIds` changes identity every render — which busts the `visibleIds` `useMemo` inside `MerchantIdSubmenu`. That makes its O(N) set-build + sort run on **every** `TicketsHeader` re-render, not just on typing.

One-line fix: hoist a module-level `const NO_MERCHANTS: string[] = []` and pass that instead of the literal.

santhoshJus (approved):

samitbarai (approved):

PR #1978, opened 2026-09-17
feat:  XYNE-63913 transfer agent ownership + give editors access to memory and schedules

XyneSpaces:
1. Problem Description:
Two gaps in the agent sharing model in `xyne-claw-auth`.

(a) Agent ownership cannot be transferred. `Agent.ownerUserId` is set at creation and nothing in the codebase can move it, while the relation is `onDelete: Cascade` — deleting an owner's user row takes the agent, its tools, shares, prompt versions, provider credentials and MCP connections with it. The only workaround today is clone-and-rebuild, and `cloneAgentForUser` deliberately does not copy the Spaces app identity (`spacesAppId` / `spacesAppToken` / `signingSecret`), so every channel install has to be redone.

(b) Users added as editors are shown tabs they cannot use. The dashboard renders the Memory and Schedules tabs for anyone with `canEdit` (owner, EDITOR or CONTRIBUTOR — `frontend/src/v3/lib/agentPermissions.ts`), but the backends behind them checked `ownerUserId` alone: memory routes returned 403, and scheduled jobs are keyed by their creating user with every read clamped back to the requester, so the Schedules tab rendered an empty list with no error.

Reported in a Spaces thread — editors could not upload memory and could not see configured schedules. Traced from the agent detail page through `routes/memory.ts` and `routes/scheduled-jobs.ts`, which is where the UI permission rule and the backend permission rule diverge.

Ownership transfer — new `POST /api/v1/agents/:slug/transfer-ownership`. Immediate, with no acceptance step, owner/admin-only, same-org only (mirrors the cross-org guard on `POST /:slug/shares`). The agent row is kept, so app identity, webhooks, installs and existing schedules are untouched, and nothing on the dispatch path reads the owner (providers resolve from the invoking user plus agent-scoped rows in `lib/provider-resolution.ts`). The owner flip, removal of any share row held by the new owner, and demotion of the outgoing owner to EDITOR all run in one transaction — the demotion is required, not courtesy, because visibility derives from ownership OR a share row, so a personal-scope agent would otherwise vanish from the previous owner's list.

Editor access — memory maintenance routes (upload-md, upload-session, import, consolidate, enable/disable, backfill) now use the canonical `getAgentEditAccess` helper instead of comparing `ownerUserId`. `clear-all` and subsystem-delete deliberately stay owner/admin-only: both are irreversible and destroy work belonging to every user of the agent. Scheduled jobs gain an agent-scoped READ path — when `?agentSlug=` is supplied and the caller maintains that agent, the list is no longer clamped to self. Mutations (pause / resume / patch / delete) are unchanged and still go through `assertCanControlScheduledJob`, because a scheduled run executes under its creator's identity and credentials.

Audit — new `AGENT_OWNERSHIP_TRANSFERRED` event records actor, previous owner and new owner. With no acceptance gate this is the only durable trace of an agent (and the provider credentials and MCP connections that follow it, both keyed by `agentId`) changing hands.

UI — "Make owner" action on each contributor row in the Contributors tab, owner-only, behind a confirm dialog that states the transfer is immediate and that credentials move with the agent. To hand the agent to someone new, add them as a contributor first, then make them owner.

4. Documentation (link to docs created/updated for this change): N/A — route-level doc comments added inline.

5. DB Alter Details (if any):
`prisma/migrations/20260917120000_agent_ownership_transferred_audit_event/migration.sql` — `ALTER TYPE "AgentAuditEvent" ADD VALUE IF NOT EXISTS 'AGENT_OWNERSHIP_TRANSFERRED'`. Additive enum value only: no table rewrite, no row changes, safe to apply before the code that emits it, harmless if rolled back.

6. Data fix Details (if any): N/A

7. Environment variable Changes (if any): N/A

Frontend typecheck (`pnpm typecheck`, `tsc -b`) passes clean. The claw-auth backend typecheck could not be run in the build sandbox because `prisma generate` cannot reach `binaries.prisma.sh` there, so the generated client is absent and `tsc` reports the whole app as untyped — the same failure on an unmodified tree. Backend typecheck needs to be confirmed in CI.

Manual checks to run against a deployed build:
- An editor can upload a memory document and see the agent's schedules.
- An editor still cannot clear the memory bank or delete another user's scheduled job.
- A viewer still receives 403 on memory writes.
- After a transfer: the new owner is shown as owner, the previous owner remains an editor, the agent's Spaces app install still responds, and existing schedules continue to fire.
- The audit log records the transfer with both user ids.

9. Services to which this change is to be deployed: xyne-claw-auth (backend + dashboard).

10. Main PR link (if this PR is raised against a release branch): N/A — raised against the Claw integration line `feature/deploy-xyneclaw`, not a release branch.


Pradeesh333 (approved):

apps/xyne-claw-auth/backend/src/routes/memory.ts
```diff
@@ -38,6 +38,24 @@ import {
...
+ * editor OR contributor (frontend/src/v3/lib/agentPermissions.ts) — while these
+ * routes used to compare `ownerUserId` alone. The result was a tab that 403'd
+ * for exactly the people it was rendered for. This helper is the same rule the
+ * UI uses, via the canonical ACL helper.
+ *
+ * DESTRUCTIVE memory routes (clear-all, subsystem delete) deliberately do NOT
+ * use this: wiping a shared bank stays owner/admin-only.
+ */
+async function canMaintainAgentMemory(req: Request, agentSlug: string, userId: string): Promise<boolean> {
+  const access = await getAgentEditAccess(userId, agentSlug, getOrgId(req));
```
XyneSpaces, line 55: **[🔴 BLOCKING · R2] CLAW_ADMINs lose memory maintenance on 3 of the 7 re-gated routes.** `canMaintainAgentMemory` = `getAgentEditAccess(...)?.canEdit`, and `getAgentEditAccess` (middleware/agent-acl.ts) has **no admin path** — it checks only `ownerUserId` and EDITOR/CONTRIBUTOR share rows. The backfill (`memory.ts:2260`), enable (`:2437`) and disable (`:2487`) call sites dropped the previous `isClawAdmin(requesterId)` bypass, so a CLAW_ADMIN who is not the owner (and holds no share) now gets **403** on exactly the routes they could use before this PR — while the new error copy still promises "or an admin can …". The other four sites (consolidate `:1525`, import `:1634`, upload-md `:1787`, upload-session `:2660`) kept `!admin &&`, so the file is also internally inconsistent.

↳ FIX: mirror `canViewAgentSchedules` (scheduled-jobs.ts:136–139) — add `if (await isClawAdmin(userId)) return true;` as the first line of this helper (and adopt its `.catch(() => null)` guard so an ACL-lookup error degrades to 403 instead of a 500). PR is already merged — please land this as a follow-up on `feature/deploy-xyneclaw`.

apps/xyne-claw-auth/backend/src/routes/agents.ts
```diff
@@ -2204,6 +2204,125 @@ router.get("/:slug/shares", requireAgentOwnerContributorOrAdmin, async (req: Req
...
+// "agents I can see" filters (agentRepository.listVisible, agentCatalogService,
+// callable-agent-resolver). On a personal-scope agent the outgoing owner would
+// lose the agent entirely, so we demote them to an EDITOR share in the SAME
+// transaction. That is correctness, not courtesy.
+//
+// SECURITY: agent-scoped provider credentials and MCP connections are keyed by
+// agentId and therefore follow the agent. The new owner inherits key material
+// the previous owner pasted. With no acceptance gate, the audit record is the
+// only durable trace — hence AGENT_OWNERSHIP_TRANSFERRED with both user ids.
+router.post(
```
XyneSpaces, line 2234: **[🟠 MAJOR · R7] No tests for any of this.** The PR adds a transactional ownership endpoint (owner flip + share demotion + new-owner share cleanup + audit event + same-org guard) and rewrites the ACL on 7 memory routes and 3 scheduled-jobs routes — with zero test changes. These are exactly the paths where regressions are silent: this PR itself fixes a "tab rendered for people the route 403'd for" bug of that shape, and the admin regression at memory.ts:55 is another one that a test matrix would have caught before merge.

↳ FIX: at minimum, ACL tests for `canMaintainAgentMemory` / `canViewAgentSchedules` over the owner / EDITOR / CONTRIBUTOR / CLAW_ADMIN / outsider matrix, plus one integration test for the transfer endpoint (editor demotion of the outgoing owner, share-row cleanup for the new owner, cross-org rejection, self-transfer rejection).

apps/xyne-claw-auth/backend/src/routes/agents.ts
```diff
@@ -2204,6 +2204,125 @@ router.get("/:slug/shares", requireAgentOwnerContributorOrAdmin, async (req: Req
...
+            userId: previousOwnerId,
+            role: "EDITOR",
+            sharedBy: actorId,
+          },
+          update: { role: "EDITOR" },
+        });
+      }
+    });
+
+    await writeAuditLog({
```
XyneSpaces, line 2295: **[🟡 MEDIUM · R2] The audit write sits outside the transaction.** If `writeAuditLog` throws here (DB hiccup, constraint), the endpoint 500s **after** ownership has already been committed — the client sees failure and may retry for a transfer that actually happened. The PR's own SECURITY comment calls this record "the only durable trace" (there is no acceptance gate), so its durability shouldn't be weaker than the transfer itself.

↳ FIX: move the write inside the `$transaction` callback if `writeAuditLog` can accept the `tx` client; otherwise wrap it in try/catch that logs at CRITICAL and never fails the request after a committed transfer.

apps/xyne-claw-auth/backend/src/routes/agents.ts
```diff
@@ -2204,6 +2204,125 @@ router.get("/:slug/shares", requireAgentOwnerContributorOrAdmin, async (req: Req
...
+        previousOwnerUserId: previousOwnerId,
+        newOwnerUserId,
+        keepPreviousOwnerAsEditor,
+        byAdmin: ctx.isAdmin && !ctx.isOwner,
+      },
+    });
+
+    // Ownership is part of the indexed agent document's visibility, so refresh
+    // it best-effort. A stale index entry must never fail the transfer itself.
+    syncAgentToIndexBestEffort(agent.id, agent.orgId, agent.slug);
```
XyneSpaces, line 2311: **[🟡 MEDIUM · R6] The new owner is never told.** The transfer is immediate and unilateral, and the new owner silently inherits agent-scoped provider credentials and MCP connections (this PR's own security note) plus every install, webhook and schedule tied to the agent — no notification, no email, no Spaces ping. Notifying the new owner was proposed in the linked discussion but didn't land with the endpoint.

↳ FIX: enqueue a best-effort notification to `newOwnerUserId` on success — same pattern as `syncAgentToIndexBestEffort` right below (never fail the transfer on it).

apps/xyne-claw-auth/frontend/src/v3/components/agent-detail/tabs/ContributorsTab.tsx
```diff
@@ -151,6 +154,43 @@ export function ContributorsTab({ agent, userId, permissions }: Props) {
...
+        showSnackbar({
+          variant: "success",
+          title: `${target.user.name} is now the owner`,
+          description: "You keep editor access to this agent.",
+        });
+        const updated = await listAgentShares(agent.slug, userId);
+        setShares(updated);
+        // Re-read the agent so the owner row (and the page's permissions)
+        // reflect the new owner without a reload.
+        const refreshed = await getAgentDetail(agent.slug).catch(() => null);
```
XyneSpaces, line 179: **[🟢 LOW · R6] Silent `.catch(() => null)` on the agent re-read.** If `getAgentDetail` fails post-transfer, the tab still updates its share list but the page-level owner/permissions state silently goes stale (e.g. the "Make owner" button may keep rendering for the outgoing owner until a reload), and nothing is logged or surfaced. (The one-arg call itself is correct — `getAgentDetail(slug: string)`.)

↳ FIX: log the failure and/or surface a non-blocking "ownership moved — reload to refresh" hint instead of swallowing.

apps/xyne-claw-auth/frontend/src/v3/components/agent-detail/tabs/ContributorsTab.tsx
```diff
@@ -358,6 +408,21 @@ export function ContributorsTab({ agent, userId, permissions }: Props) {
...
         </div>
       )}
-    </div>
+          <ConfirmDialog
+        open={transferTarget !== null}
+        onOpenChange={(open) => { if (!open) setTransferTarget(null); }}
+        title={`Make ${transferTarget?.user.name ?? ""} the owner?`}
+        description={
+          "This takes effect immediately — they do not have to accept. They inherit the agent's stored provider credentials and MCP connections, " +
+          "and can manage who it is shared with. You stay on as an editor. The agent keeps its installs, webhooks and schedules."
```
XyneSpaces, line 417: **[⚪ NIT · R8] Dialog copy is wrong for admin-initiated transfers.** "You stay on as an editor" (and the success snackbar's "You keep editor access…" at `:173`) is only true when the **owner** clicks. A CLAW_ADMIN also reaches this dialog (`canShare` treats admin as owner), and then it's the *previous owner* who becomes an editor — the admin was never the owner.

↳ FIX: branch the copy on whether the actor is the owner, e.g. "The previous owner stays on as an editor."

apps/xyne-claw-auth/frontend/src/v3/components/agent-detail/tabs/ContributorsTab.tsx
```diff
@@ -358,6 +408,21 @@ export function ContributorsTab({ agent, userId, permissions }: Props) {
...
+        }
+        confirmLabel="Transfer ownership"
+        onConfirm={() => {
+          const target = transferTarget;
+          setTransferTarget(null);
+          if (target) void handleTransfer(target);
+        }}
+      />
+</div>
```
XyneSpaces, line 426: **[⚪ NIT · R8] Formatting drifted.** The `<ConfirmDialog>` block is indented 10 spaces deep while the closing `</div>` sits at column 0 — prettier would rewrite this whole region. Run the formatter on this file (and check why it slipped past the format check).

XyneSpaces (commented):
**Xyne Spaces Reviewer** — post-merge review of `819075c` (PR merged as `897c79f` while this review was in flight). 7 inline findings below (1 blocking — needs a follow-up fix on `feature/deploy-xyneclaw`); full summary + reviewer recommendation in the follow-up comment.

Migration check: single `ALTER TYPE "AgentAuditEvent" ADD VALUE IF NOT EXISTS …` — nothing unsafe to surface.

XyneSpaces:
## Xyne Spaces Reviewer — PR #1978

**Verdict:** Blocking issues found — the PR merged while this review was in flight (merge `897c79f`, 08:50 UTC); the 🔴 below needs a follow-up fix on `feature/deploy-xyneclaw`.
**Reviewed:** 8 files, 27 hunks · **Findings:** 1 blocking, 1 major, 2 medium, 1 low, 2 nits · **Migration:** `ALTER TYPE … ADD VALUE IF NOT EXISTS` only — nothing unsafe to surface.

### Findings
| 🔴 blocking | R2 | `backend/src/routes/memory.ts:55` | CLAW_ADMIN regression — backfill/enable/disable (`:2260`/`:2437`/`:2487`) dropped the admin bypass; `getAgentEditAccess` has no admin path → admins 403 while the error copy still says "or an admin can…" |
| 🟠 major | R7 | `backend/src/routes/agents.ts:2234` | New transactional transfer endpoint + 10 re-gated routes, zero tests |
| 🟡 medium | R2 | `backend/src/routes/agents.ts:2295` | Audit write sits outside the transaction — can 500 after ownership already moved |
| 🟡 medium | R6 | `backend/src/routes/agents.ts:2311` | New owner is never notified of the transfer (proposed in the linked discussion, dropped) |
| 🟢 low | R6 | `tabs/ContributorsTab.tsx:179` | Silent `.catch(() => null)` on the agent re-read → stale owner UI |
| ⚪ nit | R8 | `tabs/ContributorsTab.tsx:417` | "You stay on as an editor" copy is wrong for admin-initiated transfers |
| ⚪ nit | R8 | `tabs/ContributorsTab.tsx:426` | Misindented ConfirmDialog block (prettier) |

*(all paths under `apps/xyne-claw-auth/`; frontend ones under `frontend/src/v3/components/agent-detail/`)*

**Verified non-issues** (checked, so nobody re-litigates them): the frontend `x-user-id` header on `transferAgentOwnership` is inert for browser callers — `mountRequestContext` strips inbound `x-user-id` without a valid S2S key and `requireAuth` re-stamps it from the verified session (`http/routes.ts`); the `getAgentDetail(agent.slug)` call at ContributorsTab.tsx:179 matches the one-arg signature; the post-transfer `listAgentShares` refresh does **not** 403 for the demoted owner (`requireAgentOwnerContributorOrAdmin` accepts EDITOR shares).

### Recommended reviewer
**@anuragdvd** — strongest claw-auth familiarity in the pool: authored #1824/#1825 (CodeQL hardening of claw-auth + claw-shared), #1823 (claw folder sync), and is the top committer on this PR's base branch. Carries the heaviest review queue (392 open review requests — likely inflated by the repo's auto-request pool), so **@Revanthvenkatpasupuleti** (0 pending, merged agent PR #1901) is the capacity alternate if that queue is real.

| Candidate | Familiarity evidence (this PR's area) | CODEOWNER? | Pending reviews | Tie-breaker |
|-----------|----------------------------------------|-----------|-----------------|-------------|
| @anuragdvd | #1824, #1825 (claw-auth CodeQL), #1823 (claw sync), top base-branch committer | no entry for `apps/xyne-claw-auth/**` | 392 | strongest file familiarity |
| @Revanthvenkatpasupuleti | merged agent PR #1901 | no | 0 | lowest load |
| @Shailendra-03 | 5 claw-auth-adjacent PRs (XYNE-59714 workflow/connectors) | no | 336 | — |
| @gyash1512 | unrelated memory PR | no | 397 | — |

Signals that couldn't be computed (disclosed, not invented): commits-by-path authorship (no path-filtered commit listing available to this reviewer), CODEOWNERS (no entry matches the changed paths), and this PR carried no requested reviewers — the ranking above is merged claw-auth PR authorship + workload only. Since the PR is already merged, a formal review request is moot; **@anuragdvd** is tagged here as the suggested follow-up-fix owner. (This reviewer's toolset also cannot issue GitHub review requests, and no `gh` CLI is available — the @mention here plus the Spaces-side tag is the full extent of assignment.)

PR #1984, opened 2026-09-17
XYNE-63805 allow editing project code with confirmation warning

sirajshaik-code:
## What

Makes a project's **code** (the ticket-ID prefix, e.g. \`EUL\`) editable after creation, instead of only at creation time.

## Why

Teams sometimes need to rename a project's code, but changing it is high-impact and easy to misunderstand — so the change is gated behind an explicit confirmation dialog.

## Changes

- `project.update` mutators (shared + backend) now accept an optional `code`, sanitized via `sanitizeProjectCode` and validated for format + workspace-uniqueness (mirroring the existing name check).
- `ProjectForm` shows the code field in edit mode (pre-filled), sending `code` only when changed.
- `ProjectsListView` shows a destructive confirmation dialog before applying a code change.

## Impact the warning communicates

- Only **new** tickets adopt the new prefix; existing tickets keep their IDs (project ends up with mixed codes).
- Numbering **continues** — it does not reset to `-0001`.
- Existing ticket links and search by the old code keep working.
- Commit/release analysis links PRs by the project's **current** code, so PRs referencing the old code may stop matching automatically.

## Notes

Not build/typechecked locally — this checkout has no \`node_modules\`. Recommend running \`pnpm install && pnpm --filter dashboard typecheck\` in CI.


sirajshaik-code:
moving the Project code now at the board level . 
will discuss and come up with a new pr

PR #1972, opened 2026-09-17
feat: XYNE-64084 Scheduled Call Channel Pill

ishaanrawat-netizen:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/backend` (API / worker)
- [ ] `apps/dashboard`
- [ ] `apps/electron`
- [ ] `apps/xyne-claw` / `apps/xyne-claw-auth`
- [ ] `packages/shared` or another shared package
- [ ] Infra (docker, scripts, nix, CI)

---

## Zero (Sync) Changes

- [ ] No Zero changes — **skip this section**
- [ ] Adds/modifies a **mutator**
- [ ] Adds/modifies the **schema** (tables, columns, relationships)
- [ ] Adds/modifies a **query or ACL**

If any of the above:

- [ ] Server (`apps/backend/src/zero/mutators.ts`) and client (`apps/dashboard/src/zero/mutators.ts`) mutators mirror each other — same namespace, name, and args schema
- [ ] No `Date.now()` / `uuid()` generated inside a mutator
- [ ] Optimistic client result matches the server result (no state flash on rebase)
- [ ] New tables exported in `createSchema()` and given a Query ACL registered in `query-acl-factory.ts`
- [ ] Matching Prisma model + migration, client regenerated

## Schema & Data Changes

- [ ] No schema changes — **skip this section**
- [ ] Prisma schema modified (`apps/backend/prisma` and/or `prisma-common`)
- [ ] Migration added
- [ ] Backfill / data migration included

If any of the above:

- [ ] Clients regenerated (`db:generate`, `db:common:generate`)
- [ ] No new Postgres enum or enum value — enums are frozen (`pnpm run test:enum`)
- [ ] Backward compatible with deployed code, or rollout order noted below
- [ ] Backfill is idempotent

## Config, Environment & URLs

- [ ] No config changes — **skip this section**
- [ ] Env variable added / renamed / removed
- [ ] **Base URL, host, port, or origin changed** (API, Zero, Vespa, Claw, LiveKit, LiteLLM, …)
- [ ] CORS / allowed origins / OAuth redirect URIs changed
- [ ] Feature flag or runtime config changed

If any of the above:

[...]

PR #1854, opened 2026-09-15
feat: XYNE-63458 Server-side lossless HEIC to WebP conversion for attachment previews

XyneSpaces:
## Problem
iPhone photos attached in chat upload fine as `image/heic` (backend allowlist already permits it), but previews are broken for receivers: browsers cannot rasterize HEIC, and the backend serves it as an opaque `application/octet-stream` download (`SAFE_INLINE_MIME_TYPES` in `safeAttachmentDownload.ts`). Image tiles and the attachment viewer fail to render, showing broken/placeholder previews. Ticket: XYNE-63458.

## Approach: server-side, download-time conversion (originals preserved)
Manual downloads and previews are now deliberately different:

- **Manual download button** (no `format` param): serves the **original HEIC bytes, byte-identical** — unchanged behavior.
- **Preview/open attachment**: dashboard fetches `GET /api/attachments/:id/download?format=webp`. The backend detects HEIC/HEIF mimetype, transcodes **losslessly** (heic-decode → raw RGBA → sharp `.webp({ lossless: true })`), and serves it **inline** as `image/webp` with a `.webp` filename (image/webp is in the safe-inline list, so it renders in tiles/lightbox).
- **Fallback**: an undecodable/corrupt HEIC falls back to serving the original bytes with the standard safe headers — the preview degrades to the download tile instead of erroring.

## Changes (12 files, backend + desktop dashboard only)
Backend:
- `apps/backend/src/utils/heicToWebp.ts` (new): `isHeicMimetype`, `toWebpFilename`, `convertHeicToWebp` (lossless)
- `apps/backend/src/controllers/attachmentController.ts`: `?format=webp` branch in `downloadAttachment`
- `apps/backend/src/types/heic-decode.d.ts` (new): module type shim
- `apps/backend/src/utils/heicToWebp.test.ts` (new) + fixture: 5 unit tests
- `apps/backend/package.json`: adds `sharp` + `heic-decode` (backend only, no frontend deps)

Dashboard (desktop surfaces; mobile/canvas out of scope):
- `apps/dashboard/src/services/clients/fileFetchService.ts`: `isHeicMimeType` + `withPreviewFormat` — appends `format=webp` to `/download` URLs only for HEIC mimes; applied in `fetchFile` and `createPreviewUrl` (new optional `mimeType`)
[...]

PR #2016, opened 2026-09-18
fix: XYNE-62937 image copy fails with "Failed to copy image" across all message views

XyneSpaces:
## Problem
Clicking **Copy** on an image attachment shows `Failed to copy image` — in the channel message view, in the thread view, and via the message action menu.

## Root cause
`navigator.clipboard.write()` is only permitted while the **transient user activation** from the originating click is still valid. Every image-copy path awaited something before reaching the write:

- `MessageAttachment.tsx` `ActionTray.handleCopyImage` awaited an `axios.get(...)` for the blob before calling `copyImage`.
- `ChatBubble.tsx` `handleCopyImage` awaited `fetchFile(...)`.
- `clipboardUtils.copyImageToClipboard` additionally awaited `convertToPngBlob()` (a canvas decode + `toBlob` re-encode) for any non-PNG image.

By the time `navigator.clipboard.write()` ran, the activation had expired and the browser rejected it. The rejection was caught and surfaced as the generic toast. This is why it reproduced everywhere the bytes had to be fetched first, and why non-PNG images failed even where the blob was already in memory.

## Fix
`copyImageToClipboard` now accepts `Blob | Promise<Blob>` and passes the promise directly to `ClipboardItem`, so `navigator.clipboard.write()` is invoked synchronously inside the click handler while the browser resolves the bytes afterwards. A fallback retries with the resolved blob for engines that reject a promise value in `ClipboardItem`.

Call sites updated so they no longer await before calling `copyImage`:
- `apps/dashboard/src/utils/clipboardUtils.ts`
- `apps/dashboard/src/hooks/useClipboard.ts`
- `apps/dashboard/src/components/Chat/MessageAttachment/MessageAttachment.tsx`
- `apps/dashboard/src/components/Chat/ChatBubble/ChatBubble.tsx`

`FileViewerModal` and `flowUI/nodes/ImageNode` needed no change — they call `copyImage` synchronously and now also benefit from the lazy PNG conversion. Duplicate error toasts at the call sites were removed; `useClipboard` already reports failure with the underlying error message.

## Validation
- `tsc --noEmit -p tsconfig.app.json`: no new errors in the touched files (pre-existing unrelated errors in SDLC/Support/Workflow screens remain).
- `eslint` on the four files: 0 errors.
- `prettier --check` on the four files: clean.
- Manual: copy a PNG and a JPEG attachment from the channel view, the open-thread view, the message action menu, and the file viewer.


apps/dashboard/src/utils/clipboardUtils.ts
```diff
@@ -345,14 +345,38 @@ const convertToPngBlob = (blob: Blob): Promise<Blob> => {
...
+ * the image, or the canvas PNG re-encode below). So the blob is never awaited here:
+ * the promise is handed straight to ClipboardItem, which lets
+ * navigator.clipboard.write() run synchronously inside the click handler while the
+ * browser resolves the bytes afterwards.
+ *
+ * Converts non-PNG images to PNG first, as browsers only support image/png for
+ * clipboard writes.
  */
-export const copyImageToClipboard = async (blob: Blob): Promise<void> => {
+export const copyImageToClipboard = async (source: Blob | Promise<Blob>): Promise<void> => {
```
XyneSpaces, line 361: R204 (🟠 High — Add unit tests for new functionality and edge cases) — the new `Blob | Promise<Blob>` lazy-`ClipboardItem` + fallback-retry behavior in `copyImageToClipboard` ships with no test coverage, yet it is now the single clipboard path used by every message view — exactly the shared-util surface where this class of regression resurfaces.

Suggested fix: add a unit test alongside the existing util-test precedent (`messageLink.test.ts`, #1393; `heicToWebp.test.ts`, #1854), stubbing `navigator.clipboard.write` / `ClipboardItem`, covering: (1) write succeeds with the lazy promise; (2) first write rejects → fallback retry runs; (3) blob promise rejects → the surfaced error carries the real cause (see the line-377 finding). Per R211/R219, assert a concrete outcome (copy result / thrown error), not just that a toast fired.

apps/dashboard/src/utils/clipboardUtils.ts
```diff
@@ -345,14 +345,38 @@ const convertToPngBlob = (blob: Blob): Promise<Blob> => {
...
+  );
+  // Prevent an unhandled rejection if the write path below bails out before awaiting it.
+  pngBlobPromise.catch(() => undefined);
+
+  try {
+    await navigator.clipboard.write([new ClipboardItem({ 'image/png': pngBlobPromise })]);
+  } catch (writeError) {
+    // Some engines reject a Promise value in ClipboardItem. Retry with the resolved
+    // blob; if resolving is what failed, surface the original write error instead.
+    const pngBlob = await pngBlobPromise.catch(() => {
```
XyneSpaces, line 377: R79 (🟡 Medium — Always log errors in catch blocks instead of swallowing them silently) — `pngBlobPromise.catch(() => { throw writeError; })` discards the real rejection reason (blob fetch / canvas PNG-conversion failure) and substitutes `writeError`. When the actual problem is fetching or converting the image, the user's toast blames the clipboard write, and nothing anywhere logs the true cause — the failure becomes undebuggable from production data.

Suggested fix: capture the rejection reason and log it with this file's existing structured logger (same pattern as `markdownToHtml`), e.g. `logger.error(LogEvent.FRONTEND_ERROR, 'copyImageToClipboard: image blob conversion failed', { error })`, and/or rethrow an error that chains the real cause so the toast description reflects it instead of masking it behind `writeError`.

apps/dashboard/src/utils/clipboardUtils.ts
```diff
@@ -345,14 +345,38 @@ const convertToPngBlob = (blob: Blob): Promise<Blob> => {
...
+      throw writeError;
+    });
+    await navigator.clipboard.write([new ClipboardItem({ 'image/png': pngBlob })]);
```
XyneSpaces, line 380: R76 (🟠 High — Do not retry non-transient failures; filter them out of retry logic explicitly) — applied to this clipboard retry: the `catch` retries `navigator.clipboard.write` unconditionally, so non-transient failures (permission denied / `NotAllowedError`, unsupported feature) are retried exactly like the one case the fallback exists for (an engine rejecting a promise value in `ClipboardItem`). A doomed retry also compounds the error-masking at line 377.

Two asks:
- Gate the retry on the failure signature it exists for (construction/type rejection of the promise value) and rethrow everything else directly.
- Assumption: the PR says the fallback targets "engines that reject a promise value in ClipboardItem" but doesn't name one, and I could not verify any such engine in this run (web search unavailable); Safari — the motivating engine per the PR — is the one that *requires* the promise pattern. Verify: name the engine(s) where the resolved-blob retry actually helps, and note it in a comment (also R132 — add an inline comment for the custom-retry deviation). Note that the retry runs after `await pngBlobPromise`, and by this PR's own root-cause analysis the await consumes transient activation — so on any engine that both rejects promise values and enforces activation, the fallback fails identically.

XyneSpaces (commented):
Automated review by **Xyne Spaces Reviewer** against the team review guidelines — 3 inline findings below (all in `apps/dashboard/src/utils/clipboardUtils.ts`); verdict, findings table, and reviewer recommendation are posted as a PR comment.

XyneSpaces:
## Xyne Spaces Reviewer — PR #2016

**Verdict:** Changes requested — no blocking/security issues, but address the findings below before merge (all three inline findings are in `clipboardUtils.ts`).
**Reviewed:** 4 files, 4 hunks (+47/−22) · **Findings:** 0 critical/blocking · 2 high · 1 medium · 1 low

### Findings
| 🟠 High | R204 | apps/dashboard/src/utils/clipboardUtils.ts:361 | New lazy-ClipboardItem + fallback-retry behavior ships with no tests |
| 🟠 High | R76 | apps/dashboard/src/utils/clipboardUtils.ts:380 | Ungated `write()` retry — non-transient failures (permission etc.) get retried; fallback target engine unverified |
| 🟡 Medium | R79 | apps/dashboard/src/utils/clipboardUtils.ts:377 | Real blob error is swallowed and replaced with `writeError` — nothing logs the true root cause |
| 🟢 Low | R294 | branch / PR title | No ticket ID in the branch name (`fix/image-copy-clipboard-user-activation`) or the PR title |

What checks out: the call-site hunks are sound — `copyImage` in `useClipboard` never rejects (it catches internally and returns a boolean), so the fire-and-forget `void copyImage(...)` calls in `ChatBubble.tsx` / `MessageAttachment.tsx` and the removal of the duplicated call-site error toasts are correct; the `useClipboard.ts` change is a pure type widening. The root-cause analysis (transient user activation consumed by pre-write awaits) matches the diff.

### Recommended reviewer
**@amitavac-dot** — most recent human contributor to `MessageAttachment.tsx` (this PR's main call site — #1895, HEIC/image handling), 0 pending review requests in this repo, active in Spaces on 17/9.

|-----------|-------------------------|------------|-----------------|-------------|
| @amitavac-dot | MessageAttachment.tsx (#1895, open) | no | 0 | active in Spaces 17/9 |
| @harshpreet931 | MessageAttachment.tsx — built the original image-copy buttons (#527, #414, #415) | no | 322 | strongest file-owner, but review queue saturated |
| @HarimohanSharma11 | ChatBubble.tsx (#1856, merged) | no | 0 | most recently active in Spaces (18/9) |
| @ishaanrawat-netizen | ChatBubble.tsx (#1972, open) | no | 0 | — |
| @KushagarGoel1 | clipboardUtils.ts (#2, stale signal) | no | 328 | stale ownership + saturated |

No CODEOWNERS entry covers any changed file (`.github/CODEOWNERS` maps only ACL paths), so ranking = file ownership (GitHub PR history per changed path) → current workload → Spaces activity. @harshpreet931 knows this feature best (he built the image-copy buttons this PR fixes) but is a requested reviewer on 322 open PRs — no capacity — so **@amitavac-dot** gets the pick on direct file familiarity + actual capacity. (PR author `XyneSpaces` is a bot account — excluded; @Dnyaneshwari-04 touched MessageAttachment.tsx via #914 but is the requester of this review — excluded.)

⚠️ This automation has no reviewer-request API available, so @amitavac-dot is tagged here but not formally requested — please add him via the UI or `gh pr edit 2016 --add-reviewer amitavac-dot`.

PR #2021, opened 2026-09-18
fix:  XYNE-63877 stop mid-playback stalls on high-bitrate attachment streaming

sithaarth:
## Problem

Playing a high-bitrate attachment (4K and up) stalled repeatedly mid-playback.

The cause was per-chunk overhead on `GET /api/attachments/:id/stream`, not bandwidth. Every range request re-ran the attachment lookup, the workspace/participant authorization queries and two storage HEAD calls before a single byte moved — and the endpoint capped every open-ended range response at 1MB, so a 4K stream at 25-45 Mbps needed several of those round trips per second of video. The player's buffer could not keep up, `<video>` raised an error, and the client's retry loop restarted playback.

## Changes

All in `apps/backend/src/controllers/attachmentController.ts`.

- **Open-ended range cap 1MB → 16MB** (`RANGE_CHUNK_SIZE`). The player aborts the response once it has buffered enough, so this bounds per-connection work rather than read-ahead. Roughly one request per 3-5s of 4K video instead of several per second.
- **Resolve once per playback.** New `resolveStreamTarget()` memoises the attachment row, the authorization verdict and the file size per `(userId, workspaceId, attachmentId)` for 30s using the existing `cacheManager`. Transcripts are excluded, since they are rewritten in place and a remembered size would go stale.
- **Dropped a redundant `fileExists()` HEAD** — `getFileMetadata()` already fails on a missing object, so the two calls asked storage the same question.
- **Storage streams are destroyed on client disconnect** (`res.on('close')`). Previously each seek abandoned an open GCS read stream that kept pulling bytes nobody would read; scrubbing a long video leaked one per seek.
- **Range edge cases.** `start >= fileSize` or an empty file now returns `416` with `Content-Range: bytes */N` instead of a `206` carrying a negative `Content-Length` that left the player waiting forever. `end < start` returns `400`.
- Per-chunk `logger.info` replaced with one line per resolve.

## Reviewer note: authorization is now cached for 30s

`assertAttachmentAccess` no longer runs on every chunk. If a user loses access to a channel mid-playback, they can continue streaming that one attachment for up to 30 more seconds. That is the explicit trade for removing database latency from in front of each chunk. Only successful resolutions are cached, so denials are always recomputed. Tune `STREAM_TARGET_TTL_SECONDS` down if that window is too wide — even 5s removes most of the per-chunk cost.

## Not included

- Signed-URL redirect + CDN offload for the stream path. Larger win, but it puts a bearer-capable URL in the client and bypasses the participant check for the URL's lifetime — worth deciding separately.
- Codec/container support (MKV, AVI, ProRes) and adaptive streaming. Those need a transcode step and are a separate change; no frontend file is touched here.

## Testing

- `eslint` clean on the changed file.
- `tsc --noEmit`: 0 errors in this file. The 17 repo-wide errors are pre-existing Prisma client/schema drift in `userActivityStatusService`, `workflowsV2/adapters/persistence` and `deskMetricsController`, all untouched here.
- Manual check to run: play a 4K attachment and watch DevTools Network — expect a handful of `206`s rather than one per second, and one `Resolved attachment stream` log line per playback instead of one per chunk.


apps/backend/src/controllers/attachmentController.ts
```diff
@@ -488,46 +575,43 @@ export class AttachmentController {
...
-      }
+      const { attachment, filePath, fileSize } = resolved.target;
+      const service = getAttachmentStorage(attachment);
 
-      const filePath = normalizeStoragePath(attachment.url);
-      if (!filePath) {
-        res.status(404).json({ error: 'Attachment not yet uploaded' });
-        return;
-      }
+      const pipeToResponse = async (options?: { start: number; end: number }): Promise<void> => {
```
amitavac-dot, line 587: (race) Abort during createReadStream setup still leaks the storage stream
attachmentController.ts:587-596:
const stream = await service.createReadStream(filePath, options);   // ← await; client can abort here
res.on('close', () => { stream.destroy?.(); });                      // ← listener attached too late
If the client disconnects while the await is in flight (on GCS, createReadStream does a file.exists() HEAD first — tens of ms; exactly the seek-storm window this PR targets), res has already emitted 'close' before the listener is attached. The listener then never fires, stream.pipe(res) writes into a destroyed response, Node unpipes on the destination error, and the GCS stream keeps draining to nowhere — the exact leak the PR intends to fix, in the very scenario (rapid scrubbing) cited in the description.
One-line fix after the await:
const stream = await service.createReadStream(filePath, options);
if (res.destroyed) { (stream as { destroy?: () => void }).destroy?.(); return; }
res.on('close', ...);
Better yet, replace the hand-rolled {close, error} wiring with pipeline(stream, res) from node:stream/promises — it checks destination-destroyed state, destroys the source on destination close, and unifies the error path. That's precisely this pattern.

apps/backend/src/controllers/attachmentController.ts
```diff
@@ -472,6 +499,66 @@ export class AttachmentController {
...
+    attachmentId: string,
+    userId: string,
+    workspaceId: string | undefined,
+  ): Promise<StreamTargetResult> {
+    const cacheKey = `attachment-stream:${userId}:${workspaceId ?? 'none'}:${attachmentId}`;
+    const cached = cacheManager.get<ResolvedStreamTarget>(cacheKey);
+    if (cached) {
+      return { ok: true, target: cached };
+    }
+
```
amitavac-dot, line 524: Cache is per-process — depends on pod routing and stampedes on the first burst
- Per-pod memoisation: cacheManager is an in-memory Map (utils/cacheManager.ts:10). With multiple backend replicas, a playback's chunk requests can land on different pods and each pays a full resolve per 30 s. Functionally fine (the documented security trade holds per pod), but the manual test's success signal — "one Resolved attachment stream line per playback" — only holds for a single pod or sticky sessions. Worth a sentence in the PR so QA doesn't chase ghosts.
- No promise coalescing: on a cold cache, the player's initial parallel range requests all miss simultaneously and each run the full resolve (attachment lookup + participant query + HEAD). Storing the in-flight promise in the cache (or a Map<key, Promise>) collapses this. The first byte of playback is exactly where the stall manifested, so this is the highest-leverage cheap addition.

devesh-juspay (dismissed):

devesh-juspay (approved):

Navviiii (approved):

PR #2043, opened 2026-09-18
feat: XYNE-62786 ui for channel board mappings

sakthin-juspay:
  - backend
  - dashboard

## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [x] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [x] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

Screenshot 2026-09-18 at 16 12 57
Screenshot 2026-09-18 at 16 13 20
Screenshot 2026-09-18 at 16 13 46

---

## Checklist


apps/backend/src/zero/mutators.ts
```diff
@@ -2081,6 +2109,84 @@ export function createMutators(
...
+          }
+          if (channel.scopeType !== ChannelScopeType.DEFAULT) {
+            throw new Error('Boards can only be linked to channels');
+          }
+
+          // Workspace/org admins and owners, the channel's creator, or a channel
+          // admin. Enforced here rather than only in the UI — the button is hidden
+          // for everyone else, but hiding a button is not a permission check.
+          const isPrivileged =
+            authData.role === WorkspaceRole.ADMIN ||
```
PiyushJuspay: move these checks to ACLs. Also keep it to channel Admin and LISTPROJECTS resource admin

packages/shared/src/zero/mutators.ts
```diff
@@ -1398,6 +1398,83 @@ export const mutators = defineMutators({
...
+        }
+        if (channel.scopeType !== ChannelScopeType.DEFAULT) {
+          throw new Error('Boards can only be linked to channels');
+        }
+
+        // Workspace/org admins and owners, the channel's creator, or a channel
+        // admin. Enforced here rather than only in the UI — the button is hidden
+        // for everyone else, but hiding a button is not a permission check.
+        const isPrivileged =
+          ctx.role === WorkspaceRole.ADMIN ||
```
PiyushJuspay: no need to check in client.

apps/dashboard/src/components/Chat/ChannelInformation/LinkBoardsDialog.tsx
```diff
@@ -0,0 +1,308 @@
...
+      if (next.has(projectId)) next.delete(projectId);
+      else next.add(projectId);
+      return next;
+    });
+  };
+
+  const handleSave = async (): Promise<void> => {
+    if (newBoardIds.length === 0) return;
+    setIsSaving(true);
+    try {
```
PiyushJuspay, line 196: no need for try/catch on mutators

apps/backend/src/zero/acl/tables/channel-board-mappings-acl.ts
```diff
@@ -1,20 +1,76 @@
...
+    if (!board || board.workspaceId !== this.ctx.workspaceId) {
+      throw new MutationACLError(
+        'Channel-board mapping insert failed: board not found in this workspace',
+        'channel_board_mappings',
+      );
+    }
+
+    // Linking is open to a channel's own admins, and to LISTPROJECTS resource
+    // admins, who administer boards across the workspace.
+    if (await this.isChannelAdmin(args.channelId, tx)) {
```
santhoshJus, line 54: **Security — this gate is self-grantable, so it isn't an authorization boundary.**

The rule here is "channel admin **or** LISTPROJECTS admin", but channel admin can be granted to oneself. `ChannelParticipantsACL.canUpdate` returns early when `userParticipationData?.userId === this.ctx.userID`, so any channel member can run:

```ts
channel_participants.update({ id: <their own row>, role: ChannelRole.ADMIN })
```

and then link **any board in the workspace** to that channel — which is exactly what this ACL is meant to prevent.

Either drop `isChannelAdmin` from the insert/delete path and gate on project admin only, or fix `ChannelParticipantsACL.canUpdate` so a participant cannot escalate their own `role`.

apps/backend/src/zero/mutators.ts
```diff
@@ -1285,6 +1285,46 @@ export function createMutators(
             updatedAt: timestamp,
           });
 
+          // The promoter becomes the new channel's admin, mirroring channel creation
+          // where the creator is ADMIN. Group DM participants are all MEMBER, and any
+          // participant may promote — without this the board-mapping inserts below
+          // would be rejected by ChannelBoardMappingsACL (channel admin or projects
+          // admin), rolling back the whole promotion for ordinary members.
+          if (participant.role !== ChannelRole.ADMIN) {
```
santhoshJus, line 1293: **Security/behaviour — `promoteToChannel` silently self-elevates the promoter.**

Setting the promoter to `ChannelRole.ADMIN` means any GROUP_DM participant who promotes a group DM permanently gains full channel-admin rights — rename, archive, remove participants, link boards — as an undisclosed side effect of an unrelated action.

This appears to exist only to satisfy `ChannelBoardMappingsACL.canInsert` for the mapping rows written later in this same transaction. A narrow, explicit ACL bypass scoped to this mutator (or writing the mappings through a system context) is much tighter than granting a permanent role the user didn't ask for and can't see.

apps/backend/src/zero/mutators.ts
```diff
@@ -2083,6 +2123,49 @@ export function createMutators(
           });
         },
       ),
+      // Link boards to a channel (channel_board_mappings). Additive only: this never
+      // removes a mapping, so a board that is already linked is silently skipped
+      // rather than erroring — the unique (channelId, boardId) index would reject it,
+      // and a partially-applied batch is worse than an idempotent one.
+      linkBoards: defineMutator(
```
santhoshJus, line 2130: **Perf — full ACL re-evaluation per inserted row, up to 100× inside one write transaction.**

`args` allows up to 100 boards, and `transaction-wrapper.ts:102` runs `ChannelBoardMappingsACL.canInsert` on **every** `insert`. Each pass redoes a channel lookup, a board lookup, a participant lookup, plus `hasProjectAdminAccess`'s `resources` / `resource_access` / `user_group_mappings` / group `resource_access` queries — roughly 6–8 reads × 100 inserts, all while holding the write transaction.

The channel-scoped and user-scoped parts are loop-invariant. Hoisting them (resolve channel + participant + project-admin set once, then only the per-board check inside the loop), or memoizing per-mutation, turns this from ~700 reads into ~10.

apps/dashboard/src/routes/KanbanBoardScreen/KanbanBoardScreen.tsx
```diff
@@ -526,6 +529,20 @@ const KanbanBoardScreen: React.FC<BoardKanbanScreenProps> = ({
...
+  // Boards linked to this channel. The only source of channel→board truth; there
+  // is no fallback to the channel's project, so an empty set genuinely means
+  // "no boards configured" (ChannelTicketsTab renders that empty state).
+  // Declared here rather than beside the other board queries because the filters
+  // machine's INIT effect below reads channelDefaultBoardId in its dep array.
+  const channelBoards = useChannelBoards(channelId);
+  // Channels have no "All Boards" state: exactly one board is always selected,
+  // seeded through the filters machine. Everything downstream (paged query,
+  // counts, Vespa pushdown) then scopes off that single board as it always has.
+  const channelDefaultBoardId = channelBoards.boards[0]?.id ?? null;
```
santhoshJus, line 541: **Correctness — this ignores `isDefault` and picks the *oldest* mapping instead.**

`channelBoards.boards[0]` is ordered by link creation (`boardsByChannel`), so this disagrees with the canonical server-side resolver `resolveChannelDefaultBoard` (`apps/backend/src/utils/channelDefaultBoard.ts:36`), which orders `[{ isDefault: 'desc' }, { createdAt: 'asc' }]`. `channelRepository.create` (`channelRepository.ts:92-98`) deliberately flags the caller-supplied `defaultBoardId`, so the two genuinely diverge whenever the default isn't the oldest board — the Tickets tab then opens on the wrong board while every server consumer uses the other one.

`boardsByChannel` doesn't order by `isDefault` today, so the hook can't express this yet; the query needs `orderBy(isDefault desc, createdAt asc)` (or the hook needs to expose the raw mapping rows) before this line can be fixed.

santhoshJus (commented):
Automated review of the channel→board mappings change (findings 1–10 as inline comments).

The change makes `channel_board_mappings` the sole channel→board source and stops reading `channel.projectId` on the client; most of the risk clusters there.

**Blocking, in my view:**
- The `channel_board_mappings` insert ACL is gated on a self-grantable role, so any channel member can link any workspace board (comment on `channel-board-mappings-acl.ts`).
- A channel with zero linked boards can never satisfy `channelScopeReady`, leaving the Streams board surface stuck in a permanent syncing state with no way out.
- The channel default board ignores `isDefault`, silently disagreeing with the server's canonical resolver.

The rest are correctness/perf issues worth fixing before this becomes the only path. Three lower-severity items (multi-board ticket-report navigation, `locked` derivation in `LinkBoardsDialog`, optimistic `claimDefault` flash) are not filed as inline comments.

apps/dashboard/src/routes/KanbanBoardScreen/KanbanBoardScreen.tsx
```diff
@@ -912,8 +942,13 @@ const KanbanBoardScreen: React.FC<BoardKanbanScreenProps> = ({
   // Don't query until:
   // 1. Machine is initialized (filters loaded from URL/storage)
   // 2. For workspace views: seeding is complete AND a board is picked
+  // A channel view carries no projectId, so its only scope is the selected board.
+  // Querying before one is seeded would send an unscoped request that returns
+  // every ticket in the workspace, so hold until a board is in the filters.
+  const channelScopeReady = !channelId || (filters.boards?.length ?? 0) > 0;
```
santhoshJus, line 948: **Correctness — a channel with zero linked boards can never become ready.**

If `channelId` is set and no boards are linked, `filters.boards` stays empty, so `channelScopeReady` is permanently `false`. That latches `workspaceViewReady`, the legacy `ticketsQueryV2` (line ~1875) and both pagination paths (~3779/3780) off forever.

`ChannelTicketsTab` handles this case, but `Surfaces.tsx:180` mounts `KanbanBoardScreen` with `channelId` directly — so a boardless channel in a Streams board surface renders a permanent syncing state with no empty state and no "Link Boards" affordance, i.e. no way for the user to get out of it.

Gate on the synced-and-empty case explicitly, e.g. `!channelId || channelBoards.hasBoards === false || (filters.boards?.length ?? 0) > 0`, and render the empty state when synced with no boards.

apps/dashboard/src/routes/KanbanBoardScreen/KanbanBoardScreen.tsx
```diff
@@ -1297,17 +1335,34 @@ const KanbanBoardScreen: React.FC<BoardKanbanScreenProps> = ({
...
+  // from the board in view — a linked board can belong to any project.
+  const scopedProjectId =
+    effectiveProjectId ??
+    channelBoards.boards.find(board => board.id === filteredSingleBoardId)?.projectId ??
+    null;
+
+  // A create-ticket modal can be seeded from this view only when there is a live
+  // channel AND something to source boards from: the route's project, or the
+  // channel's own linked boards.
+  const hasSeedableChannelContext =
```
santhoshJus, line 1363: **Reads `hasBoards` without `isSynced` — the modal remounts under the user and loses typed state.**

On a cold load `channelBoards.hasBoards` is `false` (mappings not yet synced), so the **channel-picker** `CreateTicketModal` instance (~line 5492) mounts. When the mappings arrive this flips to `true` and the **seeded** instance (~line 5468) mounts instead. Those are two different component instances, so React unmounts one and mounts the other, discarding everything the user has typed.

Reachable via the `enableUrlSync` deep link (`?createTicket=…`) and by clicking Create on a cold page. The hook exposes `isSynced` for exactly this — `channelBoards.isSynced && channelBoards.hasBoards`, and hold the modal closed until synced.

apps/dashboard/src/routes/KanbanBoardScreen/KanbanBoardScreen.tsx
```diff
@@ -1991,6 +2078,11 @@ const KanbanBoardScreen: React.FC<BoardKanbanScreenProps> = ({
       // Use workspaceSelectedBoards which is always enabled for workspace views
       return uniqueProjectIds(workspaceSelectedBoards ?? []);
     }
+    // A channel's linked boards can span projects, so take every project they
```
santhoshJus, line 2081: **Correctness — a multi-project channel now gets *no* tags query at all.**

Returning every project the linked boards span breaks both branches: `shouldUseSingleProjectQuery` requires exactly one project, and `shouldUseMultiProjectQuery` is `isMyTicketsView || isWorkspaceView`, both false for a channel. So a channel whose linked boards cross two or more projects enables neither query and the Tags filter is permanently empty. Previously a channel had exactly one `projectId` and always took the single-project path.

`useVespaTagSearch` has the same issue — it only passes `tagsProjectIds[0]`, so tag *search* silently covers one project out of N.

Either include the channel case in `shouldUseMultiProjectQuery` or let the multi-project tags query accept the channel's project set.

apps/dashboard/src/components/Tickets/CreateTicketModal/CreateTicketModal.tsx
```diff
@@ -2474,6 +2487,18 @@ export const CreateTicketModal: React.FC<CreateTicketModalProps> = ({
               >
                 {field => {
                   if (ticketKind === 'release') return null;
+                  // A channel with no linked boards would otherwise render an empty
+                  // picker and only explain itself via "Board is required" on submit.
+                  // Guarded on a channel actually being chosen — in the pick-a-channel
+                  // variant there is none yet, and the boards are empty for that reason.
+                  if (effectiveChannelId && boardOptions.length === 0) {
```
santhoshJus, line 2494: **Two false positives on this empty state.**

1. **While syncing.** `useChannelBoards` exposes `isSynced` precisely so an unsynced empty array isn't read as "no boards" (see the doc comment on the hook), and this caller ignores it — so "No boards are configured for this channel" flashes on every modal open for channels that do have boards.
2. **Release-only channels.** `boardOptions` filters out `isReleaseBoard` (line ~1851), so a channel whose linked boards are all release boards hits this branch too.

Both tell the user to link a board that is already linked. Condition on `channelBoards.isSynced && !channelBoards.hasBoards` for the genuine empty case, and give the release-only case its own message.

apps/dashboard/src/components/Chat/ChatBubble/ChatBubble.tsx
```diff
@@ -189,6 +190,9 @@ export const ChatBubble: React.FC<ChatBubbleProps> = ({
   const { setSkipMarkAsRead } = React.useContext(ConversationTabContext);
   const { isMobile } = usePlatform();
   const channel = useChannel(channelId);
+  // Tickets need a board to land on, and a channel's boards come from
+  // channel_board_mappings — a channel with none can't create one.
+  const { hasBoards: channelHasBoards } = useChannelBoards(channelId);
```
santhoshJus, line 195: **Perf — this puts a Zero subscription on every message bubble.**

`ChatBubble` is the hottest component in the app and renders once per message. Each instance adds a `boardsByChannel` subscription plus a `queryCacheActor` selector and persist effect, for a value that is **constant per channel** — and `ConversationPanelV2` already computes it. A long conversation means hundreds of duplicate hook instances all resolving the same answer.

It also runs for DM channels, where `scopeType !== DEFAULT` already suppresses ticket creation, so the work is pure waste there.

Pass `channelHasBoards` down via props/context from `ConversationPanelV2` instead of subscribing per bubble.

apps/dashboard/src/hooks/useChannelBoards.ts
```diff
@@ -0,0 +1,48 @@
...
+  isSynced: boolean;
+  hasBoards: boolean;
+}
+
+/**
+ * Channel → boards, read exclusively from `channel_board_mappings`.
+ *
+ * This is the single source of truth. `channel.projectId` is deprecated and is
+ * deliberately NOT consulted here, and there is no fallback to
+ * `boardsListByProject`: a channel with no mapping rows genuinely has no boards
```
santhoshJus, line 25: **Drift risk — mappings are now authoritative, but nothing back-fills a mapping for a board created in an existing project.**

Removing the `boardsListByProject` fallback (here, plus `Info.tsx:149` and `CreateTicketModal.tsx:515`) makes `channel_board_mappings` the only source. But every `channelBoardMapping.createMany` site is on **channel or integration creation**:

- `channelRepository.ts:98`
- `google-auth.ts:1313`
- `app-store.ts:294`
- `google-play.ts:622`
- `SdlcHubService.ts:361`
- `microsoftDeskService.ts:376`
- `workspaceGeneralChannel.ts:86`

So a board added to a project *after* its channels already exist gets no mapping and silently never appears in those channels' Info panel, board picker, or ticket creation — where it used to show up automatically via the fallback. There's no UI signal; it just looks like the board doesn't exist.

Needs either a dual-write on board creation (map the new board into the project's existing channels) or a documented migration/back-fill, plus a note that linking is now explicit.

sakthin-juspay: already backfilled in channel_boards_mappings bro,  currently used in production

apps/backend/src/zero/mutators.ts
sakthin-juspay: expected

apps/backend/prisma/generated/zero/schema.ts
santhoshJus: Remove this

packages/shared/src/zero/queries.ts
```diff
@@ -3169,9 +3169,14 @@ export const queries = defineQueries({
   boardsByChannel: defineQuery(
     z.object({ channelId: z.string() }),
     ({ args: { channelId } }) => {
+      // isDefault first, then oldest link — identical to the server-side resolver
+      // resolveChannelDefaultBoard (channelDefaultBoard.ts). Consumers treat the
+      // first row as the channel's default board, so the two must agree or the UI
+      // opens on a different board than inbound email/Slack flows write to.
       return zql.channel_board_mappings
```
santhoshJus: don't call this query on channel level. fetch channel_board_mappings with limit 1

apps/backend/src/zero/acl/tables/channel-board-mappings-acl.ts
```diff
@@ -1,20 +1,117 @@
...
+ * runs canInsert on EVERY insert with a fresh ACL instance — so without this the
+ * channel, participant and projects-admin lookups repeat once per board, holding
+ * the write transaction open for hundreds of reads.
+ *
+ * Keyed on the Transaction object, so entries die with the transaction and an
+ * authorization decision can never be reused across requests.
+ */
+const txMemo = new WeakMap<object, Map<string, Promise<unknown>>>();
+
+function memoizeForTx<T>(
```
santhoshJus: remove this

apps/dashboard/src/components/Chat/ChannelInformation/ChannelNoBoardsEmptyState.tsx
```diff
@@ -0,0 +1,67 @@
...
+ * the board view, which such a channel never reaches, so without this a boardless
+ * channel could never be given a board. Rendered both by the Tickets tab and by
+ * KanbanBoardScreen itself, since other hosts (Streams surfaces, SDLC) mount the
+ * screen with a channelId directly and bypass the tab.
+ */
+export const ChannelNoBoardsEmptyState = ({
+  channelId,
+}: ChannelNoBoardsEmptyStateProps): ReactElement => {
+  const channel = useChannel(channelId);
+  const canLinkBoards = useCanLinkChannelBoards(channelId);
```
santhoshJus: make an API call for this

santhoshJus (changes requested):

apps/dashboard/src/components/Chat/ChannelInformation/ChannelNoBoardsEmptyState.tsx
sakthin-juspay: changed into api-call

apps/dashboard/src/components/Chat/ChatInput/ChatInput.tsx
```diff
@@ -340,6 +341,9 @@ const ChatInputInner = forwardRef<InputBoxHandle, ChatInputProps>(
       excludeSelf: false,
     });
     const channel = useChannel(channelId);
+    // Tickets need a board to land on, and a channel's boards come from
+    // channel_board_mappings — a channel with none can't create one.
+    const { hasBoards: channelHasBoards } = useChannelHasBoards(channelId);
```
santhoshJus: why is it again called here? Can't you call it on a channel level and pass it to these components?

apps/dashboard/src/routes/KanbanBoardScreen/KanbanBoardScreen.tsx
```diff
@@ -479,7 +502,7 @@ const KanbanBoardScreen: React.FC<BoardKanbanScreenProps> = ({
 
   const logEntityTiming = (entityName: string): void => {
     // Lazy reset: when context changes, reset timings without a useEffect
-    const sessionKey = `${viewMode}-${channelId}-${effectiveProjectId}`;
```
santhoshJus: why is this changed? projectIdParam does not exist for channel kanban view right?

AnkurAbraham (approved):

PiyushJuspay (approved):

santhoshJus (approved):

PR #72, opened 2026-08-03
Fix/desk fixes

rahul1841:
1. fix for status filter
2. added createdat column in all 3 views
3. showing ticket sub rather than ticket id for smaller scree

image
image
image

PR #2089, opened 2026-09-20
feat: XYNE-64135 add Frequently Used emoji section to the reaction picker

XyneSpaces:
1. Problem Description:
Xyne Spaces has no "Frequently Used" section in the emoji picker. Every reaction starts from the full grid or a search, so the emojis a person actually uses are never one click away (the Slack behaviour asked for in #general).

Enhancement, raised in the product thread on frequently used emojis/stickers. `emoji-picker-react` does ship a suggested category, but it only counts clicks made inside the picker grid and resolves entries through the unicode dataset, so it never reflects reactions applied from the hover toolbar/drawers and it drops Spaces custom emojis entirely.

- New `apps/dashboard/src/utils/frequentEmojis.ts`: a localStorage-backed usage counter keyed by the exact reaction token the app already stores (a unicode char, or `custom:<emojiId>:<name>`), ranked by count with most-recent-use as tie-break, capped at 50 entries, cross-tab aware via the `storage` event.
- New `apps/dashboard/src/hooks/useFrequentEmojis.ts`: `useSyncExternalStore` wrapper returning the top N tokens.
- New `apps/dashboard/src/components/Chat/FrequentEmojiRow/FrequentEmojiRow.tsx`: the "Frequently Used" strip rendered above the picker. It renders nothing until the user has reacted at least once, so the picker is unchanged for new users. Custom emojis render through the existing `renderEmoji` helper.
- Usage is recorded in `useReactions.toggleReaction` (`apps/dashboard/src/hooks/useReaction.ts`) — the single choke point every picker and drawer already funnels through — and only on add, not on un-react.
- Wired into the three reaction surfaces: `HoverActionsToolbar` (desktop), `AddReactionDrawerMobile` and `AddReactionActionView` (mobile). Each of those now applies a reaction through one shared `applyReaction` helper, so the picker grid and the new row cannot drift on toggle semantics or close behaviour.

4. Documentation (link to docs created/updated for this change): N/A

5. DB Alter Details (if any): N/A — ranking is per device in localStorage, no schema change.

6. Data fix Details (if any): N/A

7. Environment variable Changes (if any): N/A

Verified manually in a sandbox dev environment (dashboard on :5173, backend in test mode) with a Playwright-driven walkthrough: the picker opens with no Frequently Used section for a new user, reacting from the grid makes the section appear, a single click in that row applies the reaction, and the ranking survives a page reload. `pnpm typecheck` passes and `eslint` reports 0 errors on the touched files. A recorded walkthrough is attached to the originating Spaces thread.

9. Services to which this change is to be deployed: dashboard (frontend only).

10. Main PR link (if this PR is raised against a release branch): N/A

## Update — inline quick reactions

In addition to the Frequently Used section inside the picker, the message hover toolbar now carries the user's top 3 emojis as one-click buttons (padded with a default set of 👍 ✅ 👀 so the strip is always 3 wide, including for users who have never reacted). This was added at the request of the originating Spaces thread. Reviewer feedback from the first pass is addressed in commit 4f886c7f.


XyneSpaces (commented):
Requesting review from @samit403, @anuragdvd, @harshpreet931.

XyneSpaces (commented):
Xyne Spaces Reviewer — inline findings. Full verdict, findings table, and reviewer recommendation in the summary comment.

apps/dashboard/src/components/Chat/HoverActionsToolbar/HoverActionsToolbar.tsx
```diff
@@ -231,32 +241,28 @@ export const HoverActionsToolbar: React.FC<HoverActionsToolbarProps> = ({
...
-                hasReacted,
-              });
-              handleEmojiOpenChange(false);
-            }}
-            customEmojis={customEmojis || []}
-            previewConfig={{ showPreview: true }}
-          />
+          <div className='w-[320px]'>
+            <FrequentEmojiRow onSelect={applyReaction} />
+            <EmojiPicker
```
XyneSpaces, line 255: **[major] R5 (duplicate/drift)** — The new app strip *and* emoji-picker-react's built-in "Frequently Used" category now render together: the POT screenshots show the header twice with near-identical content. Worse, they count differently — the library only counts in-grid clicks while our counter counts every surface — so once they diverge the two rows will visibly disagree. Fix in this PR (one line per surface): pass a `categories` prop on the `EmojiPicker` instances in all three surfaces that excludes the library's `suggested` category. The author already flagged this as a known follow-up; it should ship fixed, not duplicated.

apps/dashboard/src/utils/frequentEmojis.ts
```diff
@@ -0,0 +1,116 @@
+/**
```
XyneSpaces, line 1: **[major] R7 (guidelines #204/#211)** — This module carries the feature's core logic — count + last-used tie-break, the 50-entry cap with tail eviction, corrupted-JSON recovery, `custom:<id>:<name>` token keys — with no unit tests. Please add a vitest spec covering: tie-break ordering (equal counts → most recent first), cap eviction, `isEntry` filtering of malformed entries, and a `recordEmojiUse` → `getFrequentEmojis` round-trip for both unicode and custom tokens. The attached POT covers the happy path; the ranking/eviction edges are what the next PR will silently break.

apps/dashboard/src/utils/frequentEmojis.ts
```diff
@@ -0,0 +1,116 @@
...
+
+const EMPTY: FrequentEmojiEntry[] = [];
+
+let cache: FrequentEmojiEntry[] | null = null;
+const listeners = new Set<() => void>();
+
+const isEntry = (value: unknown): value is FrequentEmojiEntry => {
+  if (!value || typeof value !== 'object') return false;
+  const entry = value as Partial<FrequentEmojiEntry>;
+  return typeof entry.emoji === 'string' && !!entry.emoji && typeof entry.count === 'number';
```
XyneSpaces: **[nit] R2 (guideline #72)** — `isEntry` validates `emoji` and `count` but not `lastUsedAt`. A corrupted entry with a non-numeric `lastUsedAt` produces `NaN` in the `byUsage` tie-break and unstable ordering. One-liner: also require `typeof entry.lastUsedAt === 'number'`.

apps/dashboard/src/components/Chat/HoverActionsToolbar/HoverActionsToolbar.tsx
```diff
@@ -231,32 +241,28 @@ export const HoverActionsToolbar: React.FC<HoverActionsToolbarProps> = ({
...
+              style={{
+                width: '320px',
```
XyneSpaces: **[nit] R8 (guidelines #176/#195)** — `width: '320px'` is now redundant: the wrapping `<div className='w-[320px]'>` already fixes the width. Drop it from the style object (keeping the `--epr-*` custom properties, which legitimately need inline style) so the width has a single source of truth.

apps/dashboard/src/components/Chat/FrequentEmojis/FrequentEmojiRow.tsx
```diff
@@ -0,0 +1,55 @@
...
+    >
+      <span className='text-[11px] font-medium uppercase tracking-wide text-muted-foreground'>
+        Frequently Used
+      </span>
+      <div className='flex flex-wrap items-center gap-1'>
+        {frequentEmojis.map(emoji => (
+          <button
+            key={emoji}
+            type='button'
+            onClick={() => onSelect(emoji)}
```
XyneSpaces, line 41: **[nit] R8 (guideline #163)** — Inline arrow handler inside `.map()`. Acceptable at this size, but if the row grows, extract a small `EmojiButton` child (or a single `data-emoji` handler) so a re-render doesn't rebuild handlers for every button.

apps/dashboard/src/utils/frequentEmojis.ts
```diff
@@ -0,0 +1,122 @@
+/**
```
XyneSpaces, line 1: **[major] R7 / Guidelines-204 (🟠 High)** — This PR adds a new module with real logic (ranking by count + recency, 50-entry cap, corrupt-JSON filtering, storage-event resync) plus a hook and two components, but ships no tests. The added data-testids (`frequent-emoji-row`, `inline-quick-reaction-*`) are also unused without them. Please add unit tests for `frequentEmojis.ts` at minimum: sort order with count ties broken by lastUsedAt, MAX_TRACKED truncation, `isEntry` filtering of corrupt localStorage payloads, and the default→learned strip transition.

apps/dashboard/src/utils/frequentEmojis.ts
```diff
@@ -0,0 +1,122 @@
...
+ * never appear. Reactions in Spaces are applied from several surfaces (hover toolbar, mobile
+ * drawer, message actions drawer) and may be custom, so usage is tracked here instead, keyed
+ * by the exact token the reaction layer stores: a unicode char, or `custom:<emojiId>:<name>`.
+ *
+ * Storage is localStorage — per device, not synced across devices. A server-side counter would
+ * need a new table and a mutator; this keeps the first cut cheap and reversible.
+ */
+
+const STORAGE_KEY = 'xyne_frequent_emojis';
+
```
XyneSpaces: **[minor] R8 / Guidelines-249+261 (🟡 Medium)** — localStorage keys should be defined in (or at least re-exported from) the centralised settings/constants module so all consumers share one source of truth, rather than inline at the usage site. Move `xyne_frequent_emojis` (and ideally the display-limit/default-set constants) to the shared constants file.

apps/dashboard/src/components/Chat/FrequentEmojiRow/FrequentEmojiRow.tsx
```diff
@@ -0,0 +1,55 @@
...
+  limit,
+  className = '',
+}: FrequentEmojiRowProps): ReactElement | null => {
+  const frequentEmojis = useFrequentEmojis(limit);
+
+  if (frequentEmojis.length === 0) return null;
+
+  return (
+    <div
+      className={`flex flex-col gap-1 border-b border-border px-3 py-2 ${className}`}
```
XyneSpaces: **[nit] R8 / Guidelines-176 (🟡)** — Team convention is `cn()` for composed Tailwind classes instead of template-string interpolation. Suggest `className={cn('flex flex-col gap-1 border-b border-border px-3 py-2', className)}`.

apps/dashboard/src/components/Chat/FrequentEmojiRow/FrequentEmojiRow.tsx
```diff
@@ -0,0 +1,55 @@
...
+        {frequentEmojis.map(emoji => (
+          <button
+            key={emoji}
+            type='button'
+            onClick={() => onSelect(emoji)}
+            title={getEmojiDisplayName(emoji)}
+            aria-label={getEmojiDisplayName(emoji)}
+            data-track-category='MESSAGE'
+            data-track-name='FREQUENT_EMOJI_SELECTED'
+            className='flex size-7 items-center justify-center rounded hover:bg-accent'
```
XyneSpaces: **[nit] R8 / Guidelines-155 (🟡)** — The new emoji buttons (here and in `InlineQuickReactions.tsx:41`) have `hover:bg-accent` but no `:focus-visible` equivalent, so keyboard users get no visual feedback when tabbing through the strip. Add e.g. `focus-visible:bg-accent focus-visible:outline-none` (or the standard focus ring).

XyneSpaces (commented):
Automated review pass (Xyne Spaces Reviewer) — findings inline; summary follows as a top-level comment.

XyneSpaces (commented):
## Xyne Spaces Reviewer — PR #2089

**Reviewed:** 7 files, 8 hunks (all) · **Findings:** 0 blocking, 3 major, 3 nits · **Migration gate:** skipped — no `prisma/**` paths touched.

### Findings
| major | R5 | HoverActionsToolbar.tsx:246 | The app's new **Frequently Used** strip and emoji-picker-react's built-in suggested category render together — the header is duplicated in the POT screenshots. They also count differently (library = in-grid clicks only; ours = every reaction surface), so the two rows will diverge. Fix in this PR: exclude the `suggested` category via the picker's `categories` prop on all three surfaces. |
| major | R7 | utils/frequentEmojis.ts:1 | Core ranking logic (count + recency tie-break, 50-entry cap, corrupted-JSON recovery) ships with no unit tests — guidelines #204/#211. POT covers the happy path only. |
| major | R5 | PR-wide — *assumption* | XYNE-54237 / PR #9423 is landing a separate emoji frequency ranking (for `:`-autocomplete, incl. custom-emoji tracking). Two independent "frequently used" counters can disagree (storage + semantics). Not verified against #9423's code — reviewers, please confirm whether one usage store should back both. |
| nit | R2 | utils/frequentEmojis.ts:37 | `isEntry` doesn't validate `lastUsedAt` → `NaN` in the tie-break on corrupted storage. |
| nit | R8 | HoverActionsToolbar.tsx:248 | `width: '320px'` duplicates the new `w-[320px]` wrapper — keep one source of truth. |
| nit | R8 | FrequentEmojiRow.tsx:40 | Inline arrow handler inside `.map()` — extract a child component if the row grows. |

5 inline comments posted on `e92b31cc`.

### Recommended reviewer
**Samit Barai** (samit.barai@juspay.in) — recommended primary. ⚠️ Reviewers were NOT attached programmatically: this reviewer has no request-review API in its toolset, and GitHub user-search was unhealthy so name→login mapping couldn't be verified. Please add the three reviewers manually from the Reviewers panel.

|-----------|--------------------------|------------|-----------------|-------------|
| **Samit Barai** | n/a — per-path author signal unavailable with current tools | no (see note) | n/a — PR-list API doesn't expose extractable per-candidate counts | Active most recently (last seen 19/9) and engaged on the originating feedback thread ("how do I check this experience?") |
| Anurag Dwivedi | n/a | no | n/a | CC'd on the same thread; last seen 12/9 |
| Harshpreet Singh | n/a | no | n/a | CC'd on the same thread; last seen 14/9 |

**CODEOWNERS note:** `.github/CODEOWNERS` covers only ACL/mutator paths — none of this PR's files match, so no code-owner gate applies; the pick is driven by product context + recency.

### Process notes
- **Ticket key:** XYNE-54237 already exists and was referenced by Xyne Doctor in the originating `#xyne-spaces-feedback` thread as the tracking ticket for this gap — recommend retitling branch/PR as `feat: XYNE-54237 ...`. Its stated scope is autocomplete ranking, so either widen it or create a small child ticket for the picker section (neither this reviewer nor the authoring agent has a ticket-creation tool).
- PR description, typecheck, and POT are solid; the toggle-semantics consolidation into one `applyReaction` per surface is the right call.

XyneSpaces:
## Xyne Spaces Reviewer — PR #2089

**Verdict:** Changes requested (light) — one major: no unit tests for the new tracking module (`frequentEmojis.ts`); everything else is nits. The refactor is clean: one shared `applyReaction` per surface so grid and strip can't drift, usage recorded once at the `useReactions` choke point on add-only, localStorage reads are shape-validated, storage is capped, cross-tab sync and `useSyncExternalStore` cleanup are correct.
**Reviewed:** 8 files, all hunks · **Findings:** 0 blocking · 1 major · 1 minor · 3 nits
**Migration safety:** N/A — no Prisma schema/migration files touched.

### Findings
| major | R7 / Guidelines-204 🟠 | `apps/dashboard/src/utils/frequentEmojis.ts:1` | New ranking/persistence module + hook + 2 components ship with zero unit tests; added data-testids are unused without tests |
| minor | R8 / Guidelines-249 🟡 | `apps/dashboard/src/utils/frequentEmojis.ts:15` | `xyne_frequent_emojis` localStorage key defined inline, not in the centralised settings/constants module |
| nit | R8 / Guidelines-176 🟡 | `FrequentEmojiRow.tsx:29` | className composed via template string instead of `cn()` |
| nit | R8 / Guidelines-155 🟡 | `FrequentEmojiRow.tsx:45` + `InlineQuickReactions.tsx:41` | Buttons have `hover:` styling but no `:focus-visible` state for keyboard users |
| nit | R8 / Guidelines-294 🟢 | PR-wide | Branch/PR title carry no XYNE ticket key (author's note says one is pending) |

### Recommended reviewer
**@samitbarai (Samit Barai)** — named primary reviewer for this exact PR in the `#xyne-spaces-feedback` feature thread; please join **@anuragdvd (Anurag Dwivedi)** and **@harshpreet931 (Harshpreet Singh)** as co-reviewers. Heads-up: the earlier review comment pinged `@samit403` — that handle is a personal Gmail, not Samit's GitHub; the correct login is `@samitbarai`.

> Note: my toolset has no request-reviewers API, so I'm notifying via this mention: @samitbarai @anuragdvd @harshpreet931

| Candidate | Owns (areas in this PR) | Default reviewer? | Pending review requests | Tie-breaker |
|-----------|--------------------------|-------------------|-------------------------|-------------|
| Samit Barai · @samitbarai | Thread-assigned reviewer (emoji/reactions) | No — CODEOWNERS only covers ACL/mutator paths, none touched here | 0 | Named primary reviewer in the feature thread |
| Harshpreet Singh · @harshpreet931 | Composer trigger boundaries (adjacent), thread co-reviewer | No | 0 | Most recently active on `main` (4 of last 100 commits) |
| Anurag Dwivedi · @anuragdvd | Thread co-reviewer | No | 0 | Recent co-author activity on `main` |
| Vishal M / Milind Pithadia / Himansh Varma | Emoji-selector, picker UI, custom-emoji rendering (per Xyne Doctor report) | No | — | No confirmed GitHub login → couldn't request |

Signals unavailable: per-path commit history (no path filter in my commits tool) and Spaces presence. Pending-review counts came from scanning every open PR's `requested_reviewers` (all empty).

apps/dashboard/src/components/Chat/HoverActionsToolbar/HoverActionsToolbar.tsx
```diff
@@ -231,32 +245,28 @@ export const HoverActionsToolbar: React.FC<HoverActionsToolbarProps> = ({
...
-                emoji: emojiName,
-                hasReacted,
-              });
-              handleEmojiOpenChange(false);
-            }}
-            customEmojis={customEmojis || []}
-            previewConfig={{ showPreview: true }}
-          />
+          <div className='w-[320px]'>
+            <FrequentEmojiRow onSelect={applyReaction} />
```
harshpreet931: the picker already has a "Frequently Used" section. emoji-picker-react shows one by default and we never turn it off. so two rows, same title, different emojis. can we pass `categories` without SUGGESTED here, AddReactionDrawerMobile.tsx:78 and AddReactionActionView.tsx:50

apps/dashboard/src/components/Chat/HoverActionsToolbar/HoverActionsToolbar.tsx
```diff
@@ -207,6 +218,9 @@ export const HoverActionsToolbar: React.FC<HoverActionsToolbarProps> = ({
       key={`hover-actions-toolbar-${messageId}`}
       className={`absolute ${placement === 'below' ? 'top-1' : '-top-7'} right-4 z-50 p-1 flex items-center gap-1 rounded-lg border border-border bg-popover shadow-md`}
     >
+      {/* Frequently used emojis, one click each — then the full picker */}
+      {onEmojiPickerOpenChange && <InlineQuickReactions onSelect={applyReaction} />}
```
harshpreet931: new always-on 3 button strip for everyone, not mentioned in the PR description. intended?

apps/dashboard/src/components/Chat/HoverActionsToolbar/HoverActionsToolbar.tsx
```diff
@@ -179,6 +181,15 @@ export const HoverActionsToolbar: React.FC<HoverActionsToolbarProps> = ({
     onEmojiPickerOpenChange?.(open);
   };
 
+  // One path for every emoji this toolbar can apply — the picker grid and the
+  // Frequently Used row — so the toggle semantics and close behaviour cannot drift.
+  const applyReaction = (emojiName: string): void => {
+    const hasReacted = !!user && (reactionsData[emojiName] || []).includes(user.id);
+
+    toggleReaction({ messageId, emoji: emojiName, hasReacted });
+    handleEmojiOpenChange(false);
```
harshpreet931, line 189: fires even when the popover was never open, so a quick reaction click sends a fake "picker closed" to MessageHoverToolbar.tsx:190. harmless today. can we guard with `if (emojiOpen)`

apps/dashboard/src/components/Chat/HoverActionsToolbar/HoverActionsToolbar.tsx
```diff
@@ -207,6 +218,9 @@ export const HoverActionsToolbar: React.FC<HoverActionsToolbarProps> = ({
       key={`hover-actions-toolbar-${messageId}`}
       className={`absolute ${placement === 'below' ? 'top-1' : '-top-7'} right-4 z-50 p-1 flex items-center gap-1 rounded-lg border border-border bg-popover shadow-md`}
```
harshpreet931: 8 buttons becomes 11, no wrap and no max width. worth a look in a narrow thread panel

apps/dashboard/src/components/Chat/HoverActionsToolbar/HoverActionsToolbar.tsx
```diff
@@ -231,32 +245,28 @@ export const HoverActionsToolbar: React.FC<HoverActionsToolbarProps> = ({
...
+                ['--epr-emoji-size' as string]: '22px',
+                ['--epr-emoji-gap' as string]: '4px',
+              }}
+              theme={emojiPickerTheme}
+              emojiStyle={EmojiStyle.NATIVE}
+              onEmojiClick={emoji => {
+                // For custom emojis, store the emojiId with a prefix
+                applyReaction(
+                  emoji.isCustom
+                    ? `custom:${emoji.emoji}:${emoji.names[0] || 'custom'}`
```
harshpreet931: still inline, 8 copies of this string now and customEmojiUtils.tsx has no serialiser. can we add one

apps/dashboard/src/utils/frequentEmojis.ts
```diff
@@ -0,0 +1,122 @@
...
+ * *inside* the picker and it resolves entries through the unicode dataset, so custom emojis
+ * never appear. Reactions in Spaces are applied from several surfaces (hover toolbar, mobile
+ * drawer, message actions drawer) and may be custom, so usage is tracked here instead, keyed
+ * by the exact token the reaction layer stores: a unicode char, or `custom:<emojiId>:<name>`.
+ *
+ * Storage is localStorage — per device, not synced across devices. A server-side counter would
+ * need a new table and a mutator; this keeps the first cut cheap and reversible.
+ */
+
+const STORAGE_KEY = 'xyne_frequent_emojis';
```
harshpreet931: not scoped to user or workspace, and neither logout nor workspace switch clears it. so org A's custom emojis show up in org B, and clicking one saves a reaction that renders as a broken `:name:`. can we key it by workspaceId + userId

apps/dashboard/src/utils/frequentEmojis.ts
```diff
@@ -0,0 +1,122 @@
...
+  const current = existing.find(entry => entry.emoji === emoji);
+  const next = (
+    current
+      ? existing.map(entry =>
+          entry.emoji === emoji ? { ...entry, count: entry.count + 1, lastUsedAt: now } : entry,
+        )
+      : [...existing, { emoji, count: 1, lastUsedAt: now }]
+  )
+    .sort(byUsage)
+    .slice(0, MAX_TRACKED);
```
harshpreet931: the cap runs after the append and sorting is by count, so once 50 are tracked a new emoji enters at count 1, sorts last and gets dropped. every time, so it can never climb. can we slice to MAX_TRACKED - 1 and keep the one we just used

apps/dashboard/src/utils/frequentEmojis.ts
```diff
@@ -0,0 +1,122 @@
...
+
+const EMPTY: FrequentEmojiEntry[] = [];
+
+let cache: FrequentEmojiEntry[] | null = null;
+const listeners = new Set<() => void>();
+
+const isEntry = (value: unknown): value is FrequentEmojiEntry => {
+  if (!value || typeof value !== 'object') return false;
+  const entry = value as Partial<FrequentEmojiEntry>;
+  return typeof entry.emoji === 'string' && !!entry.emoji && typeof entry.count === 'number';
```
harshpreet931: doesnt check `lastUsedAt` (non-optional at L32) or NaN counts. nothing crashes, you just lose the tie-break. `Number.isFinite` on both

apps/dashboard/src/utils/frequentEmojis.ts
```diff
@@ -0,0 +1,122 @@
...
+  cache = next;
+  try {
+    window.localStorage.setItem(STORAGE_KEY, JSON.stringify(next));
+  } catch {
+    // Keep the in-memory ranking even when persistence fails.
+  }
+  emit();
+};
+
+/** Top `limit` emojis, most used first. Returns a stable reference between changes. */
```
harshpreet931: says "top `limit`" but theres no limit param, it returns all 50. thats describing useFrequentEmojis

apps/dashboard/src/utils/frequentEmojis.ts
```diff
@@ -0,0 +1,122 @@
...
+export const getFrequentEmojis = (): FrequentEmojiEntry[] => read();
+
+export const clearFrequentEmojis = (): void => {
```
harshpreet931: no callers, so users cant reset their own history. wire it up or drop it

apps/dashboard/src/utils/frequentEmojis.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+  emit();
+};
+
+/** Subscribe for `useSyncExternalStore`. Also follows writes made in other tabs. */
+export const subscribeToFrequentEmojis = (listener: () => void): (() => void) => {
+  listeners.add(listener);
+
+  const onStorage = (event: StorageEvent): void => {
+    if (event.key !== null && event.key !== STORAGE_KEY) return;
```
harshpreet931: one listener per subscriber instead of `emit()` at L63, so a cross tab write reparses twice. just inconsistent

apps/dashboard/src/components/Chat/FrequentEmojiRow/InlineQuickReactions.tsx
```diff
@@ -0,0 +1,50 @@
...
+ * set until the user has reacted, so the toolbar looks the same on day one and the
+ * strip never collapses to nothing.
+ */
+export const InlineQuickReactions = ({
+  onSelect,
+  limit = 3,
+}: InlineQuickReactionsProps): ReactElement | null => {
+  const frequentEmojis = useFrequentEmojis(limit);
+  const emojis =
+    frequentEmojis.length > 0 ? frequentEmojis : DEFAULT_QUICK_REACTIONS.slice(0, limit);
```
harshpreet931: all or nothing, not top up. after your first ever reaction the strip drops from 3 buttons to 1. and it resorts on every click so the button under your cursor changes. can we merge with the defaults and freeze the order

apps/dashboard/src/components/Chat/FrequentEmojiRow/InlineQuickReactions.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <button
+          key={emoji}
+          type='button'
+          onClick={() => onSelect(emoji)}
+          title={`React with ${getEmojiDisplayName(emoji)}`}
+          aria-label={`React with ${getEmojiDisplayName(emoji)}`}
+          data-testid={`inline-quick-reaction-${emoji}`}
+          data-track-category='HOVER_ACTIONS_TOOLBAR'
+          data-track-name='INLINE_QUICK_REACTION'
+          className='flex size-7 items-center justify-center rounded hover:bg-accent'
```
harshpreet931: toggle with no active styling, so clicking an emoji you already reacted with silently removes it. the chip gets this right at MessageBubble.tsx:1850

apps/dashboard/src/components/Chat/FrequentEmojiRow/InlineQuickReactions.tsx
```diff
@@ -0,0 +1,50 @@
...
+    <div className='flex items-center gap-0.5' data-testid='inline-quick-reactions'>
+      {emojis.map(emoji => (
+        <button
+          key={emoji}
+          type='button'
+          onClick={() => onSelect(emoji)}
+          title={`React with ${getEmojiDisplayName(emoji)}`}
+          aria-label={`React with ${getEmojiDisplayName(emoji)}`}
+          data-testid={`inline-quick-reaction-${emoji}`}
+          data-track-category='HOVER_ACTIONS_TOOLBAR'
```
harshpreet931: no data-track-metadata. every other toolbar action passes it (HoverActionsToolbar.tsx:284), so these events have no messageId

apps/dashboard/src/components/Chat/FrequentEmojiRow/InlineQuickReactions.tsx
```diff
@@ -0,0 +1,50 @@
+import type { ReactElement } from 'react';
```
harshpreet931: separate component living inside `FrequentEmojiRow/`. own folder, or rename the shared one

apps/dashboard/src/hooks/useReaction.ts
```diff
@@ -41,8 +42,13 @@ export const useReactions = (): UseReactionsReturn => {
           }),
         );
 
-        // Every picker and drawer funnels through here, so this is the one
-        // place a reaction is counted.
+        // Every picker and drawer funnels through here, so this is the one place a
+        // reaction is counted — including for the Frequently Used row. Removals are not
+        // counted: un-reacting is a correction, not a preference.
+        if (!hasReacted) {
```
harshpreet931, line 54: `zero.mutate` is bare, so a server rejection rolls back silently (zeroMutationToast.ts:15) but the count is already written. minor

apps/dashboard/src/hooks/useReaction.ts
```diff
@@ -41,8 +42,13 @@ export const useReactions = (): UseReactionsReturn => {
           }),
         );
 
-        // Every picker and drawer funnels through here, so this is the one
-        // place a reaction is counted.
+        // Every picker and drawer funnels through here, so this is the one place a
```
harshpreet931: true for message reactions, i checked every caller. but Call/hooks/useReactions.ts is a different one that doesnt. can we say "every message reaction picker"

apps/dashboard/src/components/Chat/FrequentEmojiRow/FrequentEmojiRow.tsx
```diff
@@ -0,0 +1,55 @@
...
+      </span>
+      <div className='flex flex-wrap items-center gap-1'>
+        {frequentEmojis.map(emoji => (
+          <button
+            key={emoji}
+            type='button'
+            onClick={() => onSelect(emoji)}
+            title={getEmojiDisplayName(emoji)}
+            aria-label={getEmojiDisplayName(emoji)}
+            data-track-category='MESSAGE'
```
harshpreet931: `MESSAGE` here, `HOVER_ACTIONS_TOOLBAR` on the sibling in the same toolbar. pick one

apps/dashboard/src/components/Chat/FrequentEmojiRow/FrequentEmojiRow.tsx
```diff
@@ -0,0 +1,55 @@
+import type { ReactElement } from 'react';
+import { useFrequentEmojis } from '../../../hooks/useFrequentEmojis';
+import { getEmojiDisplayName, renderEmoji } from '../../../utils/customEmojiUtils';
+
+interface FrequentEmojiRowProps {
+  /** Receives the stored reaction token — unicode char, or `custom:<emojiId>:<name>`. */
+  onSelect: (emoji: string) => void;
+  /** How many to show. Defaults to the shared display limit. */
+  limit?: number;
+  className?: string;
```
harshpreet931: className and limit (L20) are never passed by any of the 3 call sites. dead on arrival

apps/dashboard/src/components/Chat/AddReactionDrawerMobile/AddReactionDrawerMobile.tsx
```diff
@@ -28,6 +29,14 @@ const AddReactionDrawerMobile = ({
     }
   };
 
+  // Shared by the picker grid and the Frequently Used row.
+  const applyReaction = (emojiName: string): void => {
```
harshpreet931, line 37: same code as HoverActionsToolbar.tsx:186. the comment says it stops the paths drifting, but two copies is how they drift. can we lift it

apps/dashboard/src/components/Chat/AddReactionActionView/AddReactionActionView.tsx
```diff
@@ -31,8 +33,21 @@ const AddReactionActionView = ({ handleEmojiSelect, customEmojis }: AddReactionA
     }
   };
 
+  // The row hands back a stored reaction token; split it back into the shape the
+  // parent already serialises, so both paths produce the identical `custom:` string.
+  const handleFrequentSelect = (emoji: string): void => {
```
harshpreet931: splits the token apart just so MessageActionsDrawer.tsx:135 can rebuild the same string. can we pass it straight through like the other two

XyneSpaces:
## Review feedback addressed — pushed as `4f886c7f`

Thanks @harshpreet931 — every inline thread is answered below. `pnpm typecheck`, `eslint` (0 errors) and a new `pnpm test` (16 tests) are green.

### Correctness
| Comment | Change |
|---|---|
| ranking not scoped to user/workspace — org A's custom emojis leak into org B | Storage key is now `xyne_frequent_emojis:<workspaceId>:<userId>`, built by `useFrequentEmojiScope()` from the route workspace + authed user. The store keeps one parsed list per scope, so a workspace switch reads a different ranking instead of clearing one. |
| cap eviction drops every newcomer so it can never climb | Eviction now applies to the *other* entries only (`slice(0, MAX - 1)`) and the just-used emoji is always kept. Covered by three tests, including "repeated use moves it up the ranking". |
| `isEntry` doesn't check `lastUsedAt` / NaN counts | Both fields now go through `Number.isFinite`. Test seeds a corrupt payload and asserts only the valid entry survives. |
| picker already has its own "Frequently Used" | All three surfaces now pass `categories={EMOJI_PICKER_CATEGORIES}` (new `utils/emojiPickerCategories.ts`) which omits the library's `suggested` category, so the heading appears once and comes from one counter. |
| `onEmojiPickerOpenChange` fires when the popover was never open | `handleEmojiOpenChange` returns early when the state is unchanged, so an inline quick reaction no longer emits a fake "picker closed". |
| inline strip is all-or-nothing and re-sorts under the cursor | `useInlineQuickReactions` tops the learned emojis up with the defaults so the strip is always 3 buttons, and the component freezes the list for the lifetime of the toolbar (it unmounts on pointer-out, so the next hover shows the new ranking). |
| toggle with no active styling | Buttons take `hasReacted` and render the chip's `border-action-primary bg-accent` treatment plus `aria-pressed`, and the label switches to "Remove … reaction". |

### Structure
| Comment | Change |
|---|---|
| `applyReaction` duplicated across surfaces | Lifted to `hooks/useApplyReaction.ts` (`applyReaction` + `hasReacted`); the hover toolbar uses it. `AddReactionDrawerMobile` still closes its own drawer around the parent-supplied `toggleReaction`, so it keeps that one wrapper by design. |
| `custom:${id}:${name}` written inline in 8 places | Added `toEmojiToken()` to `customEmojiUtils` (inverse of `parseCustomEmoji`) and used it on the three reaction surfaces. |
| `AddReactionActionView` splits the token just so the drawer can rebuild it | It now passes the token straight through; `MessageActionsDrawer.handleEmojiSelect` takes a `string`. |
| separate component inside `FrequentEmojiRow/` | Both components moved to `components/Chat/FrequentEmojis/`. |
| `className` / `limit` props never passed | Removed. |
| track category `MESSAGE` vs `HOVER_ACTIONS_TOOLBAR` | Both are `HOVER_ACTIONS_TOOLBAR` now. |
| no `data-track-metadata` | Both components take `messageId` and emit it. |
| localStorage key should live in the shared constants module | Key prefix, cap, display limit and the default set moved to `constants/settings.ts`. |
| `cn()` instead of template strings; `focus-visible` states | Both applied to the new buttons. |
| `clearFrequentEmojis` has no callers | Removed — per-scope keys make the reset case moot. |
| `getFrequentEmojis` comment says "top limit" | Comment corrected; slicing belongs to the hook. |
| one storage listener per subscriber | Single module-level listener that clears the caches and calls `emit()`. |
| "every picker and drawer" is only true for message reactions | Comment now says message reaction pickers and notes `Call/hooks/useReactions.ts` is separate. |

### Tests
`apps/dashboard` had no test runner, so this adds `vitest` + `vitest.config.ts` + `tsconfig.test.json` + an eslint override for specs, and `src/utils/__tests__/frequentEmojis.test.ts` with 16 tests: scope partitioning, unicode + custom round-trip, count ranking, recency tie-break, cap eviction (3 cases), corrupt storage (4 cases), subscriber notify/unsubscribe, and snapshot stability for `useSyncExternalStore`. Run with `pnpm --filter xyne-spaces-dashboard test`.

### Not changed, deliberately
- **`zero.mutate` is bare, so a rejected mutation leaves the count written.** Agreed and now documented at the call site. The count is the user's own ranking hint, not shared state, and wiring a rollback would mean threading mutation results through the one place all surfaces share. Happy to do it if you'd rather.
- **XYNE-54237 / PR #9423 landing a second frequency tracker.** Still open and worth a decision before both merge: the picker ranking is keyed on reaction tokens (`custom:<id>:<name>`), autocomplete ranks shortcodes, so unifying means one store keyed on emoji identity. I'd rather resolve that as a follow-up than block this PR.
- **Toolbar width with 11 buttons.** The toolbar now wraps (`flex-wrap`, `max-w-[calc(100%-2rem)]`, right-aligned) instead of overflowing a narrow thread panel.

Note on evidence: the sandbox dev shell wouldn't render the dashboard for a fresh screen recording this run — it reproduces on the pre-change commit `33863ef8` as well, so it isn't a regression from these fixes. Verification here is typecheck + eslint + the 16 unit tests; happy to attach a fresh walkthrough once the dev shell is usable.

Mohan0544 (approved):

Mohan0544 (dismissed):

harshpreet931 (dismissed):

Mohan0544 (dismissed):

harshpreet931 (dismissed):

Mohan0544 (approved):

satvikbatra (approved):

apps/dashboard/package.json
```diff
@@ -20,6 +20,8 @@
     "lint": "eslint .",
     "lint:errors-only": "eslint . --quiet",
     "lint:fix": "eslint . --fix",
+    "test": "vitest run",
```
XyneSpaces, line 23: **[nit] R7** — Nothing type-checks the new test files: they are excluded from `tsconfig.app.json` (per the eslint comment added in this PR), the `typecheck` script only runs `tsconfig.app.json`, and vitest transpiles without type-checking — so a type error in the suite would surface only through eslint's typed rules. Suggested fix: add `"typecheck:test": "tsc --noEmit --project tsconfig.test.json"` next to `typecheck` (and wire it into CI if the suite runs there). *(Post-merge follow-up — PR already merged.)*

apps/dashboard/src/components/Chat/FrequentEmojis/FrequentEmojiRow.tsx
```diff
@@ -0,0 +1,60 @@
...
+            title={getEmojiDisplayName(emoji)}
+            aria-label={getEmojiDisplayName(emoji)}
+            data-track-category='HOVER_ACTIONS_TOOLBAR'
```
XyneSpaces, line 44: **[nit] R6** — This row also renders inside the mobile add-reaction drawer (`AddReactionDrawerMobile`) and the message-actions drawer (`AddReactionActionView`), so picks from those surfaces are now categorised `HOVER_ACTIONS_TOOLBAR` — the category no longer identifies the surface in analytics. Consider taking the category as a prop (defaulting to `HOVER_ACTIONS_TOOLBAR` for the toolbar popover) or renaming to an action-scoped category like `FREQUENT_EMOJI` with the surface in the name. *(Post-merge follow-up — PR already merged.)*

XyneSpaces (commented):
## Xyne Spaces Reviewer — PR #2089 (re-review)

**Verdict:** Approve-worthy after nits — already merged today 15:02Z (merge commit `3cc2ee29`), so the two nits below are post-merge follow-ups, not merge blockers.
**Reviewed:** 19 files, every new/changed hunk since my last review at `33863ef8` (commits `4f886c7f`, `3bf9d05e` → head `5b2ba6b5`) · **Findings:** 0 blocking, 0 major, 2 nits · **Migration gate:** skipped — no `prisma/**` paths touched.

### Prior findings — all resolved at `5b2ba6b5`
- ✅ **R5 duplicate "Frequently Used" category** — library's SUGGESTED row switched off on all three pickers via shared `EMOJI_PICKER_CATEGORIES`.
- ✅ **R7 no tests** — 207-line suite covering ranking, count tie-break, per-workspace+user scope partitioning, cap eviction (incl. newcomer-climbs), corrupted storage, and quota failures.
- ✅ **R5 counter drift** — single funnel (`useApplyReaction` + `useReactions.toggleReaction`); comment now scoped to message reactions, call reactions documented as a separate hook.
- ✅ **my nits** — `isEntry` validates count+`lastUsedAt` with `Number.isFinite`; constants centralised in `constants/settings.ts`; `cn()` + focus-visible ring; unified 320px wrapper.
- ✅ **from @harshpreet931's review** — workspace+user-scoped storage keys; `toEmojiToken` serialiser; toolbar `flex-wrap` + `max-w`; `open === emojiOpen` popover guard; inline strip top-up (never collapses 3→1) with ranking frozen for the toolbar's lifetime; `aria-pressed` + active styling; `data-track-metadata` with messageId; folder renamed `FrequentEmojis/`; dead `clearFrequentEmojis` dropped; one shared storage listener; cap evicts from others so newcomers can climb.
- ℹ️ `zero.mutate` rollback (his "minor") is now a documented trade-off in `useReaction.ts` — count written optimistically, a server rejection leaves the count. Reasonable for a per-device convenience ranking.

### Findings (new hunks only)
| nit | R7 | apps/dashboard/package.json:23 | New test files are excluded from `tsconfig.app.json` and nothing runs `tsc` against `tsconfig.test.json` — see inline comment. |
| nit | R6 | FrequentEmojiRow.tsx:44 | Drawer-surface picks are categorised `HOVER_ACTIONS_TOOLBAR` — see inline comment. |

### Recommended reviewer
**@harshpreet931** — deepest reviewer of exactly these files (~20 inline comments across all 19), his review was dismissed by the rebase so confirming the fixes is naturally his, and he is the most recently active candidate in Spaces (today). ⚠️ No review request was sent: this bot has no request-review API in its toolset, and the PR is already merged — @mentioned here instead; please confirm your comments are addressed.

| @harshpreet931 | reviewed all 19 files (~20 inline comments; review dismissed by rebase) | no — CODEOWNERS covers only ACL/mutator paths | 0 in fetched open-PR list | most recently active in Spaces (today) |
| @samitbarai | named primary reviewer on the originating feature thread | no | 0 in fetched open-PR list | prior-round recommendation |
| @anuragdvd | co-reviewer on feature thread | no | 0 in fetched open-PR list | — |
| @satvikbatra | approved at head + authored fix commit | no | 0 in fetched open-PR list | already approved — no action needed |

*Workload caveat:* pending-review counts are a lower bound from one page (25 most recent open PRs) of the PR list; `review-requested:` search was unreliable this session (matched ~all open PRs).

PR #2102, opened 2026-09-21
feat: XYNE-63011 Calls tab new version re

sayantanp-eng:
  - dashboard

## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Description

XYNE-63011: new version of the Calls list, behind a per-device toggle.

- Added `CallHistoryV2Screen` with new `CallCard`, `UpcomingCallRowV2` / `UpcomingCallsListV2`, `CallsEmptyStateIllustration`, and `CallAskAIModal` components.
- Added `useCallsVersion` hook and `CallsRoute` to switch between the old (`CallHistoryScreen`) and new (`CallHistoryV2Screen`) list at `/calls`; wired into `AppRoot` routing.
- Added a "Use new calls experience" toggle in Preferences (`usePreferencesState`) so users can opt in per device.
- Extracted a shared `dateGroupedList` util (Today / Yesterday / This week / Last week / This month / by-month grouping) and switched `CallHistoryV2Screen`, `RecordingsV2Screen`, `CallAskAIModal`, and `RecordingAskAIModal` to use it, replacing the bespoke grouping logic in `RecordingsV2.utils.ts`.
- Fixed the shared grouping util so a future-dated item (e.g. an upcoming scheduled call) no longer gets swept into "This week" — it now falls through to its month group.
- Extracted `callHistoryItem.utils.ts` out of `CallHistoryScreen` and simplified `CallHistorySearchPanel`.
- Removed a stray `openAI()` auto-open call on `CallDetailScreen` mount.
- Passed `calls` initial context selections through to the Ask AI sidebar's transcript selection.

## POT:

https://github.com/user-attachments/assets/fe12f69f-24f4-4c6e-8aeb-3948457dabe0

https://github.com/user-attachments/assets/1bc626b4-bc74-486e-b81f-a1d2cb23e47b

## Areas Touched

- [x] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [x] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


apps/dashboard/src/routes/CallHistoryScreen/CallHistoryScreen.tsx
```diff
@@ -551,20 +413,7 @@ const CallHistoryScreen = (): ReactElement => {
...
-      setShowMinLoader(false);
-      return;
-    }
-    setShowMinLoader(true);
-    const timer = setTimeout(() => setShowMinLoader(false), 10000);
-    return () => clearTimeout(timer);
-  }, [isLoading]);
-
-  const showRecentCallsLoader = isLoading || (showMinLoader && (calls?.length ?? 0) === 0);
+  const showRecentCallsLoader = isLoading;
```
XyneSpaces, line 416: **High — this changes the DEFAULT screen, which the toggle does not protect.**

The old code kept a minimum loader for up to 10s specifically so the screen would not flash the empty state while the Zero query warms up:

```ts
const showRecentCallsLoader = isLoading || (showMinLoader && (calls?.length ?? 0) === 0);
```

Now `showRecentCallsLoader = isLoading`, so as soon as Zero reports "not loading" with an empty result the v1 user sees **"No Calls Yet"** and then watches their call list pop in. Because `useCallsVersion` defaults to `'v1'`, that is every user until they opt in — the largest blast radius in this PR, and it is not mentioned in the description.

If the intent is that V2 does not want the min-loader, please leave `CallHistoryScreen` untouched and keep the simplification local to `CallHistoryV2Screen`. If the removal is deliberate for both, say so in the PR description and call out the empty-state flash as an accepted trade-off.

XyneSpaces (commented):
## Architecture review — Calls V2 behind a per-device toggle

Reviewed as: *a workspace member opens `/calls`, sees Upcoming + Recents, filters/searches, and sends calls to Ask AI* — with the new screen gated by a localStorage toggle that defaults to `v1`.

Overall the shape is right: `CallsRoute` mirrors the existing `RecordingsRoute` pattern, the V2 screen reuses `useCallHistory` rather than forking data access, and pulling the date-grouping logic into a shared `dateGroupedList` util removes a duplicate implementation. `tsc --noEmit` is clean on `apps/dashboard` and `eslint` produces only pre-existing-style warnings on the touched files.

The material concerns are (1) a behaviour change to the **default (v1)** screen that is not covered by the toggle, (2) V2 dropping the calendar sync / re-auth affordance entirely, (3) time-derived UI that is frozen at render, and (4) telemetry that cannot tell v1 from v2 — which is exactly what you need to decide whether to flip the default.

### Findings by severity

**High**
1. `CallHistoryScreen.tsx` — the 10s minimum loader was deleted. This ships to every user on the default path and can flash "No Calls Yet" while the Zero query warms up.

**Medium**
2. V2 never mounts `useCalendarSync`, so the sync button, the re-auth countdown and the OAuth-return handling on `/calls` disappear for anyone on v2.
3. Upcoming "today"/"now" is computed once per data change, not on a clock — a long-lived Calls tab goes stale across midnight and at a call's start minute.
4. V2 logs/metrics carry `source: 'CallHistoryScreen'`, so adoption and load-latency of v2 cannot be separated from v1.
5. Ask AI lets a user select calls with no transcript; the footer promises "Answers will cite only these calls".
6. `filteredRecentCallsNoGcal` / `filteredMissedCalls` are rebuilt every render, which invalidates `callRows` and defeats `Virtuoso`'s memoization on a potentially long list.

**Low**
7. Label selection is silently ignored while a search is active and silently re-applies when the search clears.
8. `useCallsVersion` is a literal copy of `useRecordingVersion`; neither syncs across tabs, and neither is observable server-side.
9. No tests were added — `groupItemsByDate` is now a pure util with four call sites and clear boundary cases; it is the cheapest thing to lock down.
10. An unrelated padding change in `XyneCalendarSidebar` ships to v1 users too.

Details inline.

apps/dashboard/src/routes/CallHistoryV2Screen/CallHistoryV2Screen.tsx
```diff
@@ -0,0 +1,980 @@
...
+  }, [endedCallsCount]);
+
+  useEffect(() => {
+    if (queryDetails.type === 'unknown') {
+      callHistoryLoadStartTimeRef.current = Date.now();
+    } else if (queryDetails.type === 'complete') {
+      if (callHistoryLoadStartTimeRef.current !== null) {
+        const duration = Date.now() - callHistoryLoadStartTimeRef.current;
+        logger.info(Event.CALL_HISTORY_LOADED, {
+          source: 'CallHistoryScreen',
```
XyneSpaces, line 338: **Medium — v1 and v2 are indistinguishable in telemetry.**

This block is copied verbatim from the v1 screen, including `source: 'CallHistoryScreen'` on both the log line and the `dataLoadDuration` metric attributes (also at lines 346–348 and again on the error path at 357–368).

With a per-device localStorage toggle there is no server-side record of who is on which version, so these events are the *only* signal you have. As written you cannot answer "how many people turned v2 on", "is v2 slower to first paint than v1", or "which screen produced this error" — which are precisely the questions that gate flipping the default.

Suggest `source: 'CallHistoryV2Screen'`, or better, keep `source` and add a `callsVersion: 'v2'` attribute so the two series are directly comparable.

apps/dashboard/src/routes/CallHistoryV2Screen/CallHistoryV2Screen.tsx
```diff
@@ -0,0 +1,980 @@
...
+  }, [visibleScheduledCalls]);
+
+  const filteredRecentCalls = useMemo(() => {
+    if (!hasCallSearch) return calls;
+    return vespaRecentCallRows.filter(call =>
+      isVisibleInCallList(call, user?.id, showChannelCalls),
+    );
+  }, [calls, hasCallSearch, showChannelCalls, user?.id, vespaRecentCallRows]);
+
+  const filteredRecentCallsNoGcal = filteredRecentCalls?.filter(
```
XyneSpaces, line 426: **Medium — unstable identities defeat the virtualized list.**

`filteredRecentCallsNoGcal` (here) and `filteredMissedCalls` (line 457) are plain expressions, so they get a new array identity on **every** render. They are both dependencies of the `displayRecentCalls` memo, which feeds `callRows` (line 515), which is the `data` prop of `Virtuoso`.

Net effect: every parent render — a search keystroke, a Zero poke, a modal open — rebuilds the grouped row array and hands `Virtuoso` a brand-new `data` reference, so the `useMemo`s on lines 489 and 515 buy nothing. On an account with a long call history that is avoidable work on the interaction path.

Wrap both in `useMemo` with the same dependency sets the downstream memos already declare.

apps/dashboard/src/routes/CallHistoryV2Screen/CallHistoryV2Screen.tsx
```diff
@@ -0,0 +1,980 @@
...
+        <div className='sticky left-0 top-0 z-[60] hidden h-0 w-fit md:block'>
+          <div className='h-[52px] w-fit'>
+            <AppNavigator />
+          </div>
+        </div>
+      )}
+      <div className='w-full flex flex-col items-center'>
+        {/* Sticky header */}
+        <div className='max-w-[820px] w-full sticky top-0 bg-background z-50 flex flex-col px-6 pt-8 sm:px-8'>
+          <CallHistorySearchPanel
```
XyneSpaces, line 584: **Medium — v2 silently drops calendar sync and, more importantly, the re-auth prompt.**

V1 mounts `useCalendarSync(user?.id)` and passes `calendarProvider` / `isSyncing` / `syncMessage` / `reauthCountdown` / `onCalendarSync` into this panel. V2 passes `variant='v2'` and replaces that whole affordance with a sidebar-toggle button, and never calls `useCalendarSync` anywhere.

Three user-visible consequences for someone who opts into v2:
- No manual "Sync Google/Microsoft Calendar" action.
- **No re-authorization prompt.** `useCalendarSync` is what surfaces `syncMessage.reauth` and drives the `reauthCountdown` redirect to `initCalendarOAuth`. On v2 a user whose calendar token has expired gets no signal at all — their external calendar events just quietly stop appearing.
- The OAuth return handler (`parseCalendarOAuthReturn` → strips the query params and triggers `shouldAutoSync`) also lives in that hook. If the OAuth flow redirects back to `/calls`, v2 leaves the params in the URL and skips the auto-sync.

`XyneCalendarSidebar` mounts the same hook, but only while the sidebar is open, so it does not cover this.

Please either mount `useCalendarSync` in `CallHistoryV2Screen` (even without the button, for the re-auth + OAuth-return effects) or state explicitly where that responsibility now lives.

apps/dashboard/src/routes/CallHistoryV2Screen/CallHistoryV2Screen.tsx
```diff
@@ -0,0 +1,980 @@
...
+      case 'missed':
+        filtered = filteredMissedCalls || [];
+        break;
+      case 'recurring':
+        filtered = base.filter(c => c.isRecurring);
+        break;
+      default:
+        filtered = base;
+    }
+    if (selectedLabels.length > 0 && !hasCallSearch) {
```
XyneSpaces, line 502: **Low — hidden filter state.**

`isLabelFilterDisabled = hasCallSearch` (line 455) disables the control, and this guard makes the selection a no-op while searching — both reasonable given Vespa rows carry `labels: []`. But `selectedLabels` is never cleared.

From the user's side: pick a label, start typing a search, the label chip greys out and its effect disappears; clear the search and the list silently re-narrows to a filter they set a while ago. `hasCallFilters` also stays true, so the empty state reads "0 results / no calls match" rather than "no calls with this label".

Cheapest fix: clear `selectedLabels` when a search becomes active (and say so on the disabled control's tooltip).

apps/dashboard/src/routes/CallHistoryV2Screen/CallHistoryV2Screen.tsx
```diff
@@ -0,0 +1,980 @@
...
+  const handleConfirmAskAIContext = useCallback((selected: Call[]): void => {
+    xyneAIActor.send({
+      type: 'OPEN',
+      contextType: 'general',
+      threadInfo: null,
+      startFreshChat: true,
+      initialContextSelections: {
+        canvases: [],
+        recordings: [],
+        calls: selected.map(call => ({
```
XyneSpaces, line 549: **Low — the Ask AI context chip will not be able to navigate back to the transcript.**

`AskAIInitialContextSelections['calls']` declares `conversationId`, and `XyneAISidebar` maps these straight onto `SelectedTranscript`. Downstream, `ContextPickerPanel.toAttachedContext` only attaches `threadId` when `transcript.conversationId` is present, and `navigateToTranscript` is what uses it.

Here only `channelId` is forwarded, so every call sent from this modal produces a context item with no `threadId`. Selection and `callIds` still work, but the chip loses its "go to the transcript" behaviour compared with the recordings path. If `useCallHistory` rows carry the transcript conversation id, forward it; if they genuinely do not, drop `conversationId` from the machine type so the contract does not imply something the only producer never sets.

apps/dashboard/src/routes/CallHistoryV2Screen/components/UpcomingCallsListV2.tsx
```diff
@@ -0,0 +1,165 @@
...
+  calls,
+  onJoinCall,
+  onEditCall,
+  onCancelCall,
+  currentUserId,
+  day,
+}: UpcomingCallsListV2Props): React.JSX.Element {
+  const allUsers = useUsers();
+
+  const targetDay = useMemo(() => day ?? new Date(), [day]);
```
XyneSpaces, line 55: **Medium — "today" and "now" are frozen; the Calls tab is a long-lived route.**

Two time values are captured and then never refreshed:

- `targetDay = useMemo(() => day ?? new Date(), [day])` — with `day` undefined (how the screen uses it) this is pinned to mount time. A tab left open across midnight keeps filtering `isSameDay(..., targetDay)` against *yesterday*, so today's calls never appear and yesterday's never leave.
- `const now = Date.now()` inside the `todaysCalls` memo only re-reads when `calls` or `targetDay` change, and the same is true of the active-first sort on line 63.

The screen header has the same issue — `format(new Date(), 'EEE, MMMM d')` on `CallHistoryV2Screen.tsx:622` is evaluated at render.

In practice Zero pokes hide some of this, but the correctness of the section should not depend on unrelated data churn. A shared minute-tick (`useNow(60_000)`) fed into `targetDay`, the memo and `isActiveCall` makes the behaviour deterministic.

apps/dashboard/src/routes/CallHistoryV2Screen/components/UpcomingCallRowV2.tsx
```diff
@@ -0,0 +1,257 @@
...
+  const visibleChannels = useAllVisibleChannels();
+  const canEdit = isOwner || canEditScheduledCallParticipants(call, currentUserId, visibleChannels);
+
+  const currentUserParticipant = call.participants?.find(p => p.userId === currentUserId);
+  const isUserInvited = !!currentUserParticipant;
+  const hasCurrentUserJoined = currentUserParticipant?.response === InvitationResponse.ACCEPTED;
+  const currentCallId = useSelector(roomActor, state => state.context.externalId);
+  const isUserInThisDevice = currentCallId === call.externalId;
+
+  const isActive =
```
XyneSpaces, line 68: **Medium (same root cause as the list) — the row never promotes itself to "live".**

`isActive` calls `Date.now()` during render with nothing scheduling a re-render at `startsAt`. For a call scheduled at 3:00pm, the "now" pulse badge (line 156), the green Join styling (line 199) and the hiding of the Edit/Delete menu (line 207) all flip only when something *else* re-renders this component.

So a user sitting on the Calls tab at 2:59 watches nothing happen at 3:00 — and the Edit/Delete menu stays available on a call that has effectively started. Drive `isActive` off the same shared ticker as the parent list rather than off render timing.

apps/dashboard/src/routes/CallHistoryV2Screen/components/CallAskAIModal.tsx
```diff
@@ -0,0 +1,415 @@
...
+    return searchable.filter(entry => entry.haystack.includes(needle));
+  }, [searchable, query]);
+
+  const participantsByCallId = useMemo(
+    () => new Map(searchable.map(entry => [entry.call.id, entry.participants])),
+    [searchable],
+  );
+
+  const rows = useMemo(
+    () => buildDateGroupedRowsFromItems(filtered.map(entry => entry.call)),
```
XyneSpaces, line 101: **Medium — a user can build a context set with nothing in it.**

Every non-calendar call is selectable here; nothing checks `call.transcript`. The screen passes `calls` unfiltered (`CallHistoryV2Screen.tsx:816`), which includes missed calls and 0-second calls that have no transcript to ground an answer on. The footer meanwhile promises *"Answers will cite only these calls"* and the CTA reads *"Ask across N calls"*.

Failure path from the user's side: select 25 calls, hit send, get an answer with no grounding and no explanation of why. The signal is already available and already used elsewhere — `CallCard` computes `hasTranscript = Boolean(call.transcript)` to disable "Download Transcript".

Suggest either disabling non-transcript rows with a "No transcript" hint, or keeping them selectable but showing a count ("18 of 25 have transcripts") before confirm. Related: the counter on line 241 reads `of {calls.length}` (the unfiltered total) while the select-all it sits next to operates on the *filtered* set.

apps/dashboard/src/utils/dateGroupedList.ts
```diff
@@ -0,0 +1,169 @@
...
+      monthGroups.set(monthKey, [item]);
+    }
+  }
+
+  const groups: DateGroup<T>[] = FIXED_GROUP_DEFINITIONS.flatMap(({ id, label }) => {
+    const groupedItems = fixedGroups.get(id) ?? [];
+    return groupedItems.length > 0 ? [{ id, label, items: groupedItems }] : [];
+  });
+
+  for (const [monthKey, monthItems] of monthGroups) {
```
XyneSpaces, line 104: **Low — good extraction; please lock the boundaries down with tests.**

The consolidation is the right call and the future-item fix (line 71) is a real bug fix. Two things worth pinning:

- Month groups are emitted in `Map` insertion order, so correct output depends on callers passing a newest-first list. `RecordingsV2Screen` and `CallHistoryV2Screen` do (both come off sorted sources), but the doc comment on line 43 explicitly says the function does not sort — that invariant is now load-bearing across four call sites with nothing enforcing it. Either sort the month groups by key descending here, or state the precondition in the JSDoc.
- Sites now depending on this: `CallHistoryV2Screen`, `RecordingsV2Screen`, `CallAskAIModal`, `RecordingAskAIModal`. This PR adds ~2,970 lines and no tests. `groupItemsByDate` is pure and takes an injectable `now` — a small table test over today / yesterday / this-week / last-week boundaries, the month fallthrough and a future-dated item would cover the highest-risk logic in the change for very little effort.

Also worth noting for the Ask AI modal specifically: it is fed scheduled (future) calls, which by design now fall through to a month heading — so an upcoming call can appear under "September 2026" alongside old calls. Confirm that reads the way you want.

apps/dashboard/src/hooks/useCallsVersion.ts
```diff
@@ -0,0 +1,38 @@
...
+const DEFAULT_CALLS_VERSION: CallsVersion = 'v1';
+
+const listeners = new Set<() => void>();
+
+const subscribe = (listener: () => void): (() => void) => {
+  listeners.add(listener);
+  return () => listeners.delete(listener);
+};
+
+const getSnapshot = (): string => localStorage.getItem(CALLS_VERSION_STORAGE_KEY) ?? '';
```
XyneSpaces, line 16: **Low — the third copy of this pattern, and the rollout lever it gives you is weak.**

This file is a character-for-character copy of `useRecordingVersion.ts` apart from the storage key and the default. Now that there are two, a small factory (`createDeviceVersionToggle(storageKey, defaultVersion)`) removes the copy and gives both the two things they are missing:

- **No cross-tab sync.** `saveVersion` notifies in-process listeners only; there is no `window.addEventListener('storage', ...)`. Flip the toggle in Settings in one tab and the other tab stays on the old screen until reload — confusing when a user is comparing them side by side.
- `subscribe`'s cleanup returns `listeners.delete(listener)` (a boolean) from a `() => void`. Harmless, but explicit `{ listeners.delete(listener); }` is clearer.

Architecturally: a localStorage-only flag means there is no server-side kill switch and no adoption metric — if v2 turns out to be broken for a subset of users, the only remedy is a redeploy, and you cannot tell how many people are affected. That is acceptable for an opt-in preview (and matches the recordings precedent), but it should be a conscious decision recorded in the PR, and it makes finding 2 (telemetry that distinguishes the versions) more important, not less.

apps/dashboard/src/components/Chat/XyneCalendarSidebar/XyneCalendarSidebar.tsx
```diff
@@ -873,7 +873,7 @@ const XyneCalendarDayView = memo(
         onDragEnd={onDragEnd}
         onDragCancel={onDragCancel}
       >
-        <div ref={scrollContainerRef} className='min-h-0 flex-1 overflow-y-auto px-3 pb-7'>
+        <div ref={scrollContainerRef} className='min-h-0 flex-1 overflow-y-auto px-3 py-4'>
```
XyneSpaces, line 876: **Low — unrelated change, ships to v1 users.**

`pb-7` → `py-4` in the calendar day view is not gated by the calls toggle and is not covered by the PR description. It also changes both the top and bottom padding (adding `pt-4`, reducing the bottom from `1.75rem` to `1rem`) inside a scroll container sized to `TIMELINE_HOUR_HEIGHT * 24`, which affects where the 11pm–midnight block can scroll to.

If it is needed for the v2 sidebar toggle, mention it in the description; otherwise it belongs in its own PR so a layout regression here is not bisected to "Calls V2".

apps/dashboard/src/routes/CallHistoryScreen/callHistoryItem.utils.ts
```diff
@@ -167,6 +306,30 @@ export function getParticipantDisplayData(
...
+  otherParticipantCount: number,
+): string {
+  const [firstParticipantName, secondParticipantName] = displayNames;
+  if (!firstParticipantName || otherParticipantCount <= 1) {
+    return firstParticipantName || 'Unknown';
+  }
+  if (otherParticipantCount === 2) {
+    return `${firstParticipantName}, ${secondParticipantName}`;
+  }
+  return `${firstParticipantName}, ${secondParticipantName} & ${otherParticipantCount - 2} other${
```
Mayankbansal-juspay: **Renders the literal string `undefined` when fewer names resolve than the participant count.**

`displayNames` and `otherParticipantCount` come from two different sources and can disagree. In `CallHistoryV2Screen/components/CallCard.tsx`:

```ts
const otherParticipantCount = Math.max(participantCount - (currentUserParticipant ? 1 : 0), otherParticipants.length);
const participantUsers = getPreviewParticipantUsers(call.participantPreviewUserIds, allUsersData, currentUserId);
const participantDisplayNames = participantUsers.length > 0 ? participantUsers.map(...) : fallback;
```

`getPreviewParticipantUsers` drops any preview id that is not in `allUsers` (external guests carry no user row, deactivated/unsynced users resolve to nothing), while `otherParticipantCount` is derived from the server-side `participantCount`.

Concrete case: an ended 3-person call (me + Alice + an external guest). `otherParticipantCount === 2`, but `displayNames === ['Alice']`. The `otherParticipantCount === 2` branch then returns `` `${'Alice'}, ${undefined}` `` → the card title and meta line read **"Alice, undefined"**. With `otherParticipantCount === 3` it reads "Alice, undefined & 1 other".

The V1 `CallCard` this replaced avoided it by using `participantDisplayNames.slice(1, 2).join(', ')`, which yields an empty string rather than `undefined`. Worth falling back on `displayNames.length` when it is smaller than `otherParticipantCount`, or bailing out to the single-name branch when `secondParticipantName` is missing.

apps/dashboard/src/routes/CallDetailScreen/CallDetailScreen.tsx
```diff
@@ -330,7 +330,6 @@ export default function CallDetailScreen(): ReactElement {
         aiOpenedRef.current = false;
       }
     });
-    openAI();
     return (): void => {
       subscription.unsubscribe();
       xyneAIActor.send({ type: 'CLOSE' });
```
Mayankbansal-juspay: **Removing the auto-open leaves an asymmetric teardown: this effect still closes the AI sidebar unconditionally, but no longer reopens it.**

Before, the effect body called `openAI()` and the cleanup called `CLOSE`, so the pair was symmetric — any re-run or unmount closed and immediately reopened. With `openAI()` gone, only the `CLOSE` half remains, so this effect now closes a sidebar it never opened.

Two concrete cases:

1. On the new V2 calls list the user clicks **Ask AI**, picks call context, and the sidebar opens. They then click a call row to open its detail, and navigate back. Unmounting `CallDetailScreen` fires the cleanup → `xyneAIActor.send({ type: 'CLOSE' })` → the sidebar and the context the user just assembled are silently dropped.
2. On a deep-linked call detail (no navigation state, so `call` comes from `useCachedQuery`), `call` is a fresh object on every row materialisation. Deps are `[call, isMobile, openAI]`, so any update to the call row (transcript arrives, participant leaves, summary lands) re-runs the effect, and the cleanup closes the sidebar the user opened with the page's own **Ask AI** button mid-conversation.

Also note `openAI` is now dead weight in the dependency array — the effect body no longer references it, so it only serves to re-trigger the teardown.

apps/dashboard/src/routes/CallHistoryScreen/CallHistoryScreen.tsx
```diff
@@ -552,20 +413,7 @@ const CallHistoryScreen = (): ReactElement => {
...
```
Mayankbansal-juspay: **Dropping the min-loader reintroduces the empty-state flash it was added to prevent** — and this is the V1 screen, which the PR description says is otherwise untouched.

`usePaginatedCalls` computes `isLoading = queryDetails.type !== 'complete' && accumulatedCalls.length === 0`, and the accumulator is only filled by an effect that runs *after* the render in which `queryDetails.type` flips to `'complete'`:

```ts
useEffect(() => {
  if (!firstPage || queryDetails.type !== 'complete') return;
  queryCacheActor.send({ type: 'MERGE_CALL_HISTORY_PAGE', ... });
}, [firstPage, queryDetails.type, participantOnly]);
```

So on the render where the query resolves, `isLoading` is already `false` while `calls` is still empty → `displayRecentCalls.length === 0` → the `No Calls Yet` empty state renders before the merged page arrives. That is exactly the window the deleted comment described ("so the screen doesn't flash the empty state while the Zero query is still warming up").

`CallHistoryV2Screen.tsx:321` copies the same `showRecentCallsLoader = isLoading` line, so V2 inherits it too. If the 10s floor was too blunt, a `queryDetails.type !== 'complete' || !calls` guard would still cover the gap.

apps/dashboard/src/routes/CallHistoryV2Screen/components/UpcomingCallsListV2.tsx
Mayankbansal-juspay: **`targetDay` is frozen at mount, so "Upcoming" is wrong for any tab left open across midnight.**

`day` is never passed by `CallHistoryV2Screen`, so `useMemo(() => day ?? new Date(), [day])` evaluates `new Date()` once and never again — the deps array is `[undefined]` forever. `todaysCalls` then keeps filtering with `isSameDay(call.startsAt, <the day the tab was opened>)`, and `getEmptyStateTitle(targetDay)` keeps using that day's hour.

Concrete: a dashboard tab opened Monday evening still shows Monday's calls under "Upcoming" on Tuesday morning and hides Tuesday's, while the section header right above it (`format(new Date(), 'EEE, MMMM d')` in `CallHistoryV2Screen.tsx`) re-renders and says Tuesday — so the header and the list disagree. The greeting also stays stuck on "Winding down for the day".

The list already re-renders whenever `calls` changes, so recomputing the day per render (or ticking it on an interval) would be enough.

apps/dashboard/src/routes/RecordingsV2Screen/components/RecordingAskAIModal.tsx
```diff
@@ -422,7 +413,8 @@ const RecordingOption = ({
           </span>
         )}
         <span className='mt-0.5 block truncate text-xs text-muted-foreground'>
-          {formatParticipants(participants)} · {formatRecordingTimestamp(recording.startedAt)}
+          {formatParticipantNames(participants.map(participant => getUserDisplayName(participant)))}{' '}
```
Mayankbansal-juspay: Small unintended copy regression from swapping the local `formatParticipants` for the shared `formatParticipantNames`.

The deleted helper joined the two-name case with `" & "`:

```ts
if (names.length <= 2) return names.join(' & ');   // "Asha & Ben"
```

`formatParticipantNames` delegates to `buildParticipantSummary(names, names.length)`, whose two-name branch returns `` `${first}, ${second}` `` → **"Asha, Ben"**. Zero and 3+ participants are unchanged, so every two-participant recording row in this modal quietly changes separator. Harmless if intentional, but it isn't mentioned in the PR description and the modal's own list rows elsewhere still use the `&` form.

apps/dashboard/src/routes/CallHistoryScreen/CallHistoryScreen.tsx
sayantanp-eng: Reverted back the manual loader code.

apps/dashboard/src/routes/CallHistoryV2Screen/CallHistoryV2Screen.tsx
sayantanp-eng: Done, source changed to 'CallHistoryV2Screen'.

apps/dashboard/src/routes/CallHistoryV2Screen/CallHistoryV2Screen.tsx
sayantanp-eng: useCalendarSync mounted

apps/dashboard/src/routes/CallHistoryV2Screen/CallHistoryV2Screen.tsx
sayantanp-eng: Done.

apps/dashboard/src/routes/CallHistoryV2Screen/CallHistoryV2Screen.tsx
sayantanp-eng: Added
```
useEffect(() => {
    if (hasCallSearch) setSelectedLabels([]);
  }, [hasCallSearch]);
```

apps/dashboard/src/routes/CallHistoryV2Screen/CallHistoryV2Screen.tsx
sayantanp-eng: the call's chat thread lives in callUpdatesChannel when set (falling back to its own channel), and its conversationId is stashed in metadata rather than a real column.

apps/dashboard/src/routes/CallHistoryScreen/callHistoryItem.utils.ts
sayantanp-eng: Added secondParticipantName back, although in the design reference only one user + n being used.

apps/dashboard/src/routes/CallDetailScreen/CallDetailScreen.tsx
sayantanp-eng: Cleanup now guarded by aiOpenedRef.current — only closes the sidebar if this screen's own "Ask AI" button opened it. A sidebar opened from the v2 calls list (or anywhere else) survives navigating into/out of a call detail

apps/dashboard/src/routes/CallHistoryScreen/CallHistoryScreen.tsx
sayantanp-eng: Added back already.

apps/dashboard/src/routes/CallHistoryV2Screen/components/UpcomingCallsListV2.tsx
sayantanp-eng: Added.
```
useEffect(() => {
    const timer = setInterval(() => setNow(Date.now()), 60_000);
    return () => clearInterval(timer);
  }, []);
```

apps/dashboard/src/routes/RecordingsV2Screen/components/RecordingAskAIModal.tsx
sayantanp-eng: No fix, stale.

apps/dashboard/src/routes/CallHistoryV2Screen/components/UpcomingCallRowV2.tsx
sayantanp-eng: Not needed.

apps/dashboard/src/hooks/useCallsVersion.ts
sayantanp-eng: Intended.

Mayankbansal-juspay (approved):

shubhamdhamija-juspay (approved):

apps/dashboard/src/routes/CallHistoryV2Screen/components/CallAskAIModal.tsx
```diff
@@ -0,0 +1,415 @@
...
+    [filteredIds, selectedIds],
+  );
+
+  const handleClose = useCallback((): void => {
+    onOpenChange(false);
+  }, [onOpenChange]);
+
+  const handleConfirm = useCallback((): void => {
+    if (selectedIds.size === 0) return;
+    onConfirm(calls.filter(call => selectedIds.has(call.id)));
```
XyneSpaces: **[🟡 Medium] Correctness / UX consistency** (flagged by the earlier architecture review as Medium; still unaddressed at this head) — `handleConfirm` forwards every selected call, including calls without a transcript, while the modal footer (line 313) promises “Answers will cite only these calls” — a selected call with no transcript gives the assistant nothing to cite. `CallCard` already gates its “Download Transcript” action on transcript availability, so the modal is inconsistent with the rest of the screen. Suggested fix: exclude no-transcript calls from the selectable set (or render them disabled with a hint) — or, at minimum, soften the footer copy so it doesn't promise citations for calls that cannot be cited.

apps/dashboard/src/routes/CallHistoryV2Screen/components/UpcomingCallsListV2.tsx
```diff
@@ -0,0 +1,173 @@
...
+const COLLAPSED_COUNT = 3;
+const COLLAPSE_TRANSITION = { duration: 0.26, ease: [0.4, 0, 0.2, 1] as const };
+
+function isActiveCall(call: Call, now = Date.now()): boolean {
+  return (
+    call.status === CallStatus.ACTIVE ||
+    call.status === CallStatus.IN_PROGRESS ||
+    (call.status === CallStatus.SCHEDULED &&
+      Boolean(call.startsAt) &&
+      new Date(call.startsAt!).getTime() <= now &&
```
XyneSpaces: **[🟡 Medium] Rule #63 — avoid `!` non-null assertions** — `new Date(call.startsAt!)` here and `call.endsAt!` at line 22 assert non-null right after Boolean guards, and the same pattern repeats in `UpcomingCallRowV2.tsx`. If the value can be absent at runtime, the assertion hides that from the type system; if the guard already proves presence, the `!` is redundant. Suggested fix: narrow explicitly (e.g. `call.startsAt !== null && new Date(call.startsAt)`) so the compiler enforces the guard instead of trusting it.

apps/dashboard/src/routes/CallHistoryV2Screen/CallHistoryV2Screen.tsx
```diff
@@ -0,0 +1,1016 @@
...
+  const filteredMissedCalls = useMemo(
+    () =>
+      (hasCallSearch
+        ? filteredRecentCalls?.filter(call => isMissedCallForUser(call, user?.id))
+        : missedCalls
+      )?.filter(call => !isExternalCalendarEvent(call)),
+    [hasCallSearch, filteredRecentCalls, missedCalls, user?.id],
+  );
+
+  const activeCallsForUpcoming = useMemo(() => {
```
XyneSpaces: **[🟡 Medium] Rule #143 — include all values read inside useMemo in its dependency array** — `activeCallsForUpcoming` (here) and `hasUpcomingCallsToday` (line 513) read the clock (`Date.now()` / `new Date()`) inside the memo but depend only on the call lists, so both freeze at whatever time the lists last changed: a SCHEDULED call whose `startsAt` has already passed won't move into the active set, and the “today” gate can go stale across midnight — even though the screen keeps a 60s `now` ticker (line 91) for exactly this staleness problem. `UpcomingCallsListV2` re-derives active state with its own ticker, which limits the visible impact — hence medium, not high. Suggested fix: read the component's `now` state inside both memos and add it to their dependency arrays.

apps/dashboard/src/routes/CallHistoryScreen/CallHistorySearchPanel.tsx
```diff
@@ -262,33 +337,35 @@ export function CallHistorySearchPanel({
...
+                htmlFor='channel-calls-toggle'
+                className='hidden md:flex items-center gap-1 text-sm text-muted-foreground whitespace-nowrap cursor-pointer select-none'
+              >
+                See thread calls
+                <Info className='size-3.5' />
+              </label>
+            </Tooltip>
+            <Switch
+              id='channel-calls-toggle'
+              checked={showChannelCalls ?? false}
```
XyneSpaces: **[⚪ Nit] Rule #270 — remove dead code** — the `?? false` fallback here and the `?? (() => {})` at line 357 are dead: both call sites (v1 `CallHistoryScreen` and v2 `CallHistoryV2Screen`) pass these props, and the `variant` union types them as required — the fallbacks only exist to satisfy the union type. Suggested fix: drop the fallbacks and type the props as required on both variants (or make them explicitly optional if a future caller is genuinely expected to omit them).

apps/dashboard/src/hooks/useCallsVersion.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const parseVersion = (raw: string): CallsVersion =>
+  raw === 'v2' || raw === 'v1' ? raw : DEFAULT_CALLS_VERSION;
+
+const saveVersion = (version: CallsVersion): void => {
+  localStorage.setItem(CALLS_VERSION_STORAGE_KEY, version);
+  listeners.forEach(listener => listener());
+};
+
+export const useCallsVersion = (): {
```
XyneSpaces: **[⚪ Nit] Rule #263 — extract duplicated logic** — `useCallsVersion` is a line-for-line re-shape of the existing `useRecordingVersion` (useSyncExternalStore over a module-level listener set plus localStorage read/subscribe); only the storage key and default value differ. Suggested fix: extract a shared `useStoredVersionToggle(key, fallback)` hook and have both call it, so the next version toggle doesn't grow a third copy.

apps/dashboard/src/utils/dateGroupedList.ts
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * Groups a newest-first list without sorting or discarding entries.
+ * More specific calendar periods take precedence over broader ones.
+ *
+ * @example
+ * groupItemsByDate(items, now).map(group => group.label);
+ * // ['Today', 'This week', 'June 2026']
+ */
+export function groupItemsByDate<T extends DateGroupable>(
```
XyneSpaces: **[🟡 Medium] Rule #211 — add unit tests for new utility functions** — `dateGroupedList.ts` is a new ~169-line pure util with real date-boundary logic (Today/Yesterday/This week/Last week/This month grouping, future-item fallthrough, `findNearestVisibleItem`) and is now consumed by four screens (Calls v2, Recordings v2, both Ask-AI modals) — but the PR adds no tests for it. Suggested fix: a small unit suite over the grouping boundaries (today/yesterday/week edges, a future-dated item landing in its month group, empty input, nearest-visible lookup) — it's pure, so it's cheap to cover.

XyneSpaces (commented):
## Xyne Spaces Reviewer — PR #2102

**Verdict:** Changes requested — no blocking issues; 4 medium findings worth fixing before merge, 2 nits.
**Reviewed:** 26 files, all hunks (~3,109 additions / 436 deletions) at head `b8e6efdf` · **Findings:** 0 blocking, 4 medium, 2 nits

All findings are inline below. Everything else in the diff checked out clean — in particular, the items the earlier architecture review asked for are addressed on this head: the 10s minimum loader while the Zero query warms up, `useCalendarSync` actually mounted, telemetry `source` wired, memoized search-result filtering, the label filter disabled during search, 60s `now` tickers on the screen and in `UpcomingCallsListV2`, and the shared `dateGroupedList` util replacing the RecordingsV2 copy.

### Findings
| 🟡 medium | #211 | apps/dashboard/src/utils/dateGroupedList.ts:50 | new 169-line pure date-grouping util, no unit tests |
| 🟡 medium | #63 | apps/dashboard/src/routes/CallHistoryV2Screen/components/UpcomingCallsListV2.tsx:20 | `startsAt!` / `endsAt!` non-null assertions after Boolean guards |
| 🟡 medium | #143 | apps/dashboard/src/routes/CallHistoryV2Screen/CallHistoryV2Screen.tsx:497 | clock read inside useMemo without `now` in deps (also :513) |
| 🟡 medium | — | apps/dashboard/src/routes/CallHistoryV2Screen/components/CallAskAIModal.tsx:161 | no-transcript calls selectable although the footer promises citations |
| ⚪ nit | #270 | apps/dashboard/src/routes/CallHistoryScreen/CallHistorySearchPanel.tsx:356 | dead `?? false` / `?? (() => {})` fallbacks |
| ⚪ nit | #263 | apps/dashboard/src/hooks/useCallsVersion.ts:27 | useCallsVersion duplicates useRecordingVersion |

### Recommended reviewer
**@harshpreet931** — reviewed both sibling PRs for this surface (#2038 calls-v2 design, #1788 recordings-v2) and has the lowest pending review load of the candidates.

| Candidate | Knows this area via | CODEOWNER match? | Pending review requests | Tie-breaker |
|-----------|---------------------|------------------|-------------------------|-------------|
| @harshpreet931 | reviewed #2038 + #1788 (iterated with the author, re-approved after fixes) | no | 343 | covers both the calls and recordings surfaces this PR touches |
| @knaman2609 | reviewed #2038 (calls design) | no | 367 | — |
| @samitbarai | reviewed #1788 (recordings) | no | 367 | — |
| @anuragdvd | reviewed #1788 (recordings) | no | 404 | — |
| @Mayankbansal-juspay | authored #2038 + #1788 | no | — | already approved this PR — not re-requested |

Notes: `.github/CODEOWNERS` in this repo covers only ACL paths and call-infrastructure (`services/Call/`, call join hooks) — none of this PR's 26 dashboard files match, so no CODEOWNER is eligible and ranking is by sibling-PR review history plus current review load. Pending-review counts come from GitHub search (`is:pr is:open review-requested:<login>`; the repo currently has 519 open PRs) and include team auto-requests, so treat them as relative rather than absolute. The PR author (@sayantanp-eng) and requester (@shubhamdhamija-juspay, already approved) are excluded by rule.

Note on the reviewer request itself: this bot does not have API access to formally add a reviewer on the PR (`requested_reviewers` is currently empty), so please consider this an explicit review request — **@harshpreet931** — and the author can also add you via the Reviewers panel in the GitHub UI.

PR #2106, opened 2026-09-21
fix: unread badge unification

arpitmourya-blr:
Summary 
Unread badges disagreed across surfaces: the dock badge and workspace switcher showed numbers that didn't match the sidebar rails, DM unreads were invisible outside the DM shelf, and DM channels the user had already viewed stayed at 0 unread forever. This PR fixes the frozen DM unread counter and the client/server count-definition drift behind it, unifying every unread surface on one shared rule set: workspace count = dm + bell + call, with the dock rendering the sum over all workspaces.
After the fix: dock = Σ workspace counts; the switcher trigger shows the active workspace's count (a dot when only other workspaces have unread); each rail shows its own shelf count — all derived from the same shared rules.

RCA
Root Cause

Two independent defects, both required fixing:
1. Frozen DM counter. handleUnreadCount skips recompute when channel_stats.lastActivityAt <= lastViewedAt. Nothing on the Zero message-insert path bumped lastActivityAt — only channel creation, membership changes, calls, and the REST conversation path did. For any DM channel the recipient had ever viewed, the guard was never satisfied and channel_user_status.unreadCount froze at 0 permanently.
2. Count-definition drift. GET /activity/workspace-counts counted raw unread activity rows per user — reactions (added_v2), removed, missed calls, closed-channel rows, legacy direct_message, and SKIP classifications were all included — while the client rails applied their own hand-maintained exclusion lists. Two independent definitions of "what counts as unread" drifted apart. Additionally, the response was per-user rows that the client folded into a Map keyed by workspaceId (last-write-wins), so members holding two identities in one workspace got the wrong count.

Missed Scenario

* Frozen counter: the guard in handleUnreadCount and every existing lastActivityAt writer were individually correct; the bug was the interaction — the Zero conversations side-effect never wrote the column. No test asserted the invariant "a message insert must make lastActivityAt > lastViewedAt reachable", and reviewing either file in isolation could not reveal it.
* Drift: backend and frontend counts were never compared against the same dataset. Each side's filters evolved independently with no parity check, so the divergence only became visible when users compared the dock against the rails.
Impact
* Dock badge inflated by reactions/missed-call/closed-channel/legacy rows — never matching the rails.
* DM unreads absent from the workspace switcher and dock (the endpoint had no dm component).
* Viewed DM channels permanently showed 0 unread.
* Multi-identity members: workspace counts clobbered (last-write-wins).
* Cross-workspace unreads missing from the dock total (active workspace only).

Fix

* Frozen counter (Phase 0): the Zero conversations side-effect bumps channel_stats.lastActivityAt to the conversation's createdAt before recomputing, as a guarded updateMany (lastActivityAt < createdAt) — never throws on a missing stats row and never moves the timestamp (or DM ordering) backward. DM/GROUP_DM channels only.
* Server counts (Phase 1a): getWorkspaceActivityCounts rewritten to dmCount + bellCount + callCount:
    * dm = open channel_user_status.unreadCount minus unread top-level GROUP_DM mention rows (the bell counts mentions — no double counting), floored at 0 per channel;
    * bell per shared BELL_COUNT_RULES, with a per-user (userId, channelId) closed-pair post-filter — closed state is per-user on channel_user_status, so a plain channelId notIn would over-exclude other identities' activities in the same channel;
    * call = unread missed_call rows (identical semantics to the Calls rail query).
    * Identities merged server-side per workspace; userId dropped from the response.
* Shared rules: bellCountRules.ts + countPredicates.ts in @xyne/shared — one definition of "what counts", imported by backend where-clauses and dashboard hooks alike.
* Single poller (Phase 1b): WorkspaceUnreadCountsProvider (30s poll, visibilitychange, unread:refetch event); the switcher migrated off its private fetch.
* Client rails (Phase 1c): rails derive from the shared predicates (closed-channel parity via the visible-channels set, which the server already filters per-user); DM rail gains a numeric badge; reactions render a dot only when the numeric badge is 0; useZero's mutate chokepoint emits unread:refetch after read mutators so poll-fed badges refresh immediately.
* Dock (Phase 1d): dock badge source switchable via VITE_DOCK_BADGE_POLL_SOURCE — default off, so dock behavior is unchanged by default; flip after comparing legacy vs polled totals in dev.
* Switcher UI: trigger shows the active workspace's count; a red dot appears when the active count is 0 but another workspace has unread (makes the dock total explainable); badge/dot hidden while the popover is open.

API note: /activity/workspace-counts response changed from [{workspaceId, userId, count}] to [{workspaceId, count}], and count semantics changed from raw unread activities to dm+bell+call. No consumers exist outside the dashboard (verified across the monorepo).

Prevention

* Single source of truth: BELL_COUNT_RULES / countPredicates in @xyne/shared — backend and frontend can no longer maintain independent definitions.
* Parity tests: packages/shared/test/unreadCountParity.test.mjs (12 tests) pins the client predicate and the rules shape on golden fixtures; activityService.test.ts (8 tests) pins the Prisma where-clause translation of the same constants plus sum/closed/subtraction/merge invariants.
* Regression tests: conversations-handler.test.ts (6 tests) pins the lastActivityAt bump, its position before the recompute, and the DM-only scope.
* Accepted limits (documented): the parity pin is translation-level (mocked Prisma), not full-semantics against a live DB; the server's queries are not a cross-query snapshot — a mark-read landing mid-poll self-corrects on the next tick/refetch.

Testing

* Backend jest (conversations-handler.test.ts, activityService.test.ts): 14/14 passing.
* Shared: tsc build + parity suite: 12/12 passing.
* Typecheck: dashboard clean; backend at the pre-existing 62-error baseline with zero errors in touched files. Lint: clean on all touched files.
* Edge cases covered: empty identities; per-user closed state (another identity's rows kept when only one closed the channel); null-channelId ticket rows always count; GROUP_DM subtraction floored at 0; legacy null-isThreadActivity mention rows (server isThreadActivity: { not: true } matches client !== true); multi-identity merge; cross-workspace split; bump-before-recompute ordering; non-DM and missing-conversation no-ops.
* Not yet run (needs a live environment, tracked as follow-up): cross-tab read on the same channel, mark-read offline + reconnect, fresh load after disconnect; dev comparison of legacy vs polled dock totals before flipping VITE_DOCK_BADGE_POLL_SOURCE.
* Confirmed-intended side effect: the DM shelf now sorts by actual last message (most-recent-on-top), since lastActivityAt is bumped on message inserts.


arpitmourya-blr:
will raise again

PR #2015, opened 2026-09-18
feat: move encrypted-fields config to enc-owned DB table

sithaarth:
## Type of Change

- [x] Refactoring

## Description

Removes the hardcoded `encryptedFieldsConfig` from `@xyne/shared`. The encryption service now owns that config in its own DB table (`encryption.encrypted_field_configs`) and serves it over s2s; the backend and browser consume it at runtime. Companion private PR: juspay/xyne-spaces-private#337 (enc service + backend overlay).

Why: the same decision ("which columns are encrypted") lived in three places — the shared constant bundled into browser + backend, a hardcoded mirror inside the encryption service, and the s2s endpoint `GET /api/internal/encryption/fields-config` the enc service polled at boot. Now one owner, one shape (`Record<table, { fields: string[]; enforceClientEncryption }>`), readers cache a 15-min snapshot.

- `packages/shared`: `zero/encrypted-fields.ts` deleted; `zero/client-transaction-wrapper.ts` (dead `wrapClientTransaction` + friends) replaced by `zero/query-validation.ts` exporting `validateQueryWhereClause(query, encryptedFields)`, `EncryptedFieldQueryError`, `EncryptedTableConfig`. `useZero` passes the config it already holds from `useEncryptionConfig()`; `useEncryptionConfig` / `key-exchange` reuse the exported type.
- `apps/backend`: s2s `fields-config` endpoint deleted; `EncryptionProvider.getEncryptedFieldsConfig()` added to the contract (public no-op returns `{}` when encryption is off); `DatabaseClient.connect()` warms it so a misconfigured deploy fails at boot, not on the first query.

Net −166 lines.

## Areas Touched

- [x] `apps/backend` (API / worker)
- [x] `packages/shared` or another shared package

## Zero (Sync) Changes

- [x] No Zero changes — **skip this section**

## Schema & Data Changes

- [x] No schema changes — **skip this section** (the new table lives in the encryption service's Prisma project, private repo)

## Config, Environment & URLs

- [x] No config changes — **skip this section**

## API Contract

- [x] Existing contract modified (shape, status codes, auth) — `GET /api/internal/encryption/fields-config` (s2s-only) removed; its only caller was the encryption service, which no longer calls it after juspay/xyne-spaces-private#337.

[...]

PR #2117, opened 2026-09-21
feat: add @xyne/cache snapshot cache package

sithaarth:
## What

New zero-dependency workspace package `packages/cache` (`@xyne/cache`). No consumer is wired in this PR; it only creates the package and registers it in the workspace.

v1 ships one primitive, `createSnapshotCache`: a keyed map that always holds the whole set from its source.

- `init()` loads once and throws on failure, so boot fails loudly.
- A background timer reloads every `refreshEveryMs` and replaces the snapshot atomically (never merges), so deleted source rows disappear.
- `get` / `snapshot` are synchronous, safe on hot paths.
- A failed refresh keeps the last-good snapshot and emits `refresh-failed`.
- `refreshNow()` joins any in-flight load and never throws. `stop()` clears the timer.
- No eviction by design: for config-shaped data a miss must mean "not in the source".
- Timers are `unref`'d. Clock is injectable, so tests need no fake-timer framework.
- Observability is an `onEvent` callback with a typed union; the package imports no logger or metrics library.

## Why

The encrypted-fields config cache (#2015 / private #337) was hand-rolled twice with identical policy and different loaders. A survey of the repo found about 30 more hand-rolled caches repeating the same decisions ad hoc: timers never unref'd, no in-flight dedupe, empty results cached forever after an error, no events. The duplicated part is the freshness policy, not storage, so that is what this package owns. Services own the loader, the interval, and event wiring.

## Structure

Each concern is its own small module and none of them know each other: `clock`, `single-flight`, `slot`, `scheduler`, `attempt`, `events`. `snapshot-cache.ts` only composes them. Roughly 140 lines of source.

## Also changed

- `pnpm-workspace.yaml`: adds `packages/cache` (explicit member list).
- root `build:shared`: adds `--filter @xyne/cache`.
- `pnpm-lock.yaml`: only the new importer block (9 lines). A full `pnpm install` on `main` currently rewrites unrelated peer suffixes, so that churn was deliberately left out.

## Verification

```
pnpm --filter @xyne/cache typecheck
pnpm --filter @xyne/cache test        # 16 tests, node --test against dist
pnpm install --frozen-lockfile --filter @xyne/cache   # "Lockfile is up to date"
```

## Follow-up (not in this PR)

Migrate the two encrypted-fields caches onto `createSnapshotCache`. Wiring notes for the backend (tsconfig paths, jest moduleNameMapper) and the encryption service Dockerfile (COPY lines plus an explicit package build because the install uses `--ignore-scripts`) are in the plan.


sithaarth:
Superseded: replaced by a thin lru-cache wrapper landing directly in #2015.

PR #2119, opened 2026-09-21
feat: XYNE-64120 Log logout reasons and timestamps

shubhamdhamija-juspay:
  - backend

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates
- [ ] Documentation
- [ ] CI/CD

## Areas Touched

- [ ] `apps/dashboard`
- [ ] `apps/electron`

---

## Zero (Sync) Changes

If any of the above:

## Schema & Data Changes

- [ ] Migration added

If any of the above:

- [ ] Backfill is idempotent

## Config, Environment & URLs

If any of the above:

## API Contract

- [ ] No API changes
- [ ] New endpoint

---

### Automated

- [ ] Not covered by tests 

### Manual

**Users**
- [ ] Single user

**Login**
- [ ] Google
- [ ] Microsoft

**Run mode**
- [ ] Test mode (`dev:test`)
- [ ] Docker compose
- [ ] Electron desktop

**Surface & data**
- [ ] Web browser
- [ ] Electron desktop
- [ ] Empty state

---

## Checklist


apps/backend/src/services/sessionRefreshValidator.ts
```diff
@@ -223,6 +223,25 @@ export async function isRefreshAllowed(session: LoadedSession | null): Promise<b
...
+  // 401. Ending the session records the LOGOUT once; later requests with the
+  // same cookie hit status_EXPIRED / status_REVOKED above and record nothing.
+  // provider_revoked is recorded by the deactivation cleanup below.
+  if (session) {
+    const sessionId = session.id;
+    const endSession =
+      verdict.reason === 'refresh_token_expired'
+        ? userSessionService.expireSession(sessionId)
+        : verdict.reason === 'org_member_left' || verdict.reason === 'workspace_user_left'
+          ? userSessionService.revokeSession(sessionId, 'MEMBER_LEFT')
```
sithaarth: **Regression: refresh denial on member-left now permanently revokes the session.** `validateSessionForRefresh` checks status before the `leftAt` checks, and `revokeSession` hard-sets REVOKED (terminal — nothing resets it on re-add). On base the row was left untouched on denial, so when `invitationService.ts:733-757` or `zero/mutators.ts:16022` clears `leftAt` on re-invite, the user's open tabs/devices resumed refreshing transparently. With this change, any background auto-refresh landing during the removal window kills the session for good — after reinstatement every device needs a full SSO re-login. Nothing else in the codebase revokes sessions on member removal, and this write is fire-and-forget from a function documented as "Pure decision — no cookie/JWT side effects". Suggest: record the LOGOUT event without flipping the session row, or make re-invite restore eligibility.

apps/backend/src/services/userSessionService.ts
```diff
@@ -212,16 +268,30 @@ export class UserSessionService {
...
+      // guarded update makes that atomic, so concurrent logouts emit one row.
+      const { count: wasLive } = await this.prisma.userSession.updateMany({
+        where: { id: sessionId, status: SessionStatus.ACTIVE },
+        data: {
+          status: SessionStatus.REVOKED,
+          updatedAt: new Date(),
+        },
+      });
+
+      const session = await this.prisma.userSession.update({
```
sithaarth: **Unguarded second `update` after the guarded `updateMany` — throws P2025 and 500s logout.** With a stale `user_session_id` cookie for a cascade-deleted row: `updateMany` returns count 0 (fine), then `update({where:{id}})` throws P2025, `revokeSession` rethrows, and `authV2Controller.ts:1486` awaits it unguarded — handler jumps to its catch, client gets a 500, and `clearCookie('user_session_id')`, the `xyne_ws_*_token` clears, and `revokeSessionKey` are all skipped, leaving an un-clearable cookie. Also, when the row exists but was already EXPIRED, this line silently stamps it REVOKED, so `user_sessions.status` disagrees with the only LOGOUT event emitted (TOKEN_EXPIRED). Use `findUnique` + early return when `wasLive === 0`, exactly as `expireSession` (369-382) already does.

apps/backend/src/controllers/authV2Controller.ts
```diff
@@ -2406,6 +2409,9 @@ export class AuthV2Controller {
             accessToken: currentSession.accessToken ?? undefined,
             deviceInfo: JSON.stringify({ userAgent: req.headers['user-agent'], timestamp: new Date().toISOString(), appVersion: req.headers['x-app-version'] }),
             ipAddress: req.ip || req.connection.remoteAddress || undefined,
+            // Already signed in — a session for the workspace they just created.
```
sithaarth: **The new `WORKSPACE_CREATED`/platform telemetry is unreachable — this `createSession` always fails with P2002.** `schema.prisma:849` declares `refreshToken String @unique` (global). This block loads `currentSession` via a pure `findUnique` and passes `refreshToken: currentSession.refreshToken` straight into a plain `prisma.userSession.create` — the old row is never revoked/deleted and no new token is minted, so the insert is a guaranteed unique-constraint violation. The catch below swallows it, `newSessionId` stays null, and the cookie re-set at 2436 never fires: workspace created, no session issued, no LOGIN event. The P2002 itself is pre-existing on base (byte-identical reuse and swallow) — the PR adds telemetry to a block that never completes. Fix by minting a fresh refresh token or revoking the old row first.

apps/backend/src/services/userSessionService.ts
```diff
@@ -252,8 +329,19 @@ export class UserSessionService {
   /**
    * Revoke all sessions for a user
    */
-  async revokeAllUserSessions(userId: string): Promise<void> {
+  async revokeAllUserSessions(userId: string, reason: LogoutReason): Promise<void> {
     try {
+      // Captured before the update so each session that was actually live gets
+      // its own LOGOUT row (already expired/revoked ones are not logouts).
+      const liveSessions = await this.prisma.userSession.findMany({
```
sithaarth: **Read-then-write race: concurrent callers emit duplicate LOGOUT rows.** `revokeSession` got an atomic guarded `updateMany` ("concurrent logouts emit one row"), but the actually-concurrent path didn't. When a user's Google account is revoked with ~8 requests in flight, each enters `isRefreshAllowed`, gets `provider_revoked`, and `sessionRefreshValidator.ts:247` fires `handleDeactivatedUser` fire-and-forget with no in-flight guard. All 8 run this `findMany` before any `updateMany` commits, all observe ACTIVE, all call `trackLogout` — 8 identical PROVIDER_REVOKED LOGOUT rows for one session. Same shape when `changePassword` races the client's own /logout (one PASSWORD_CHANGED + one USER_LOGOUT for a single session). Do the guarded `updateMany` first and track only the rows it actually flipped.

apps/backend/src/services/userSessionService.ts
```diff
@@ -252,8 +329,19 @@ export class UserSessionService {
...
+        where: {
+          userId,
+          status: SessionStatus.ACTIVE,
+          refreshTokenExpiry: { gt: new Date() },
```
sithaarth: **ACTIVE-but-token-lapsed sessions are revoked with no LOGOUT row, ever.** This capture filters on `refreshTokenExpiry > now`, but the `updateMany` at 345 filters on `userId` alone. An abandoned device's session (ACTIVE, expiry lapsed, `expireSession` never fired because no request came) gets flipped to REVOKED on password change without a LOGOUT being written — and none can ever be written afterwards since the row is no longer ACTIVE. That LOGIN permanently has no matching LOGOUT, breaking the 1:1 pairing the sessionId change exists to guarantee. Note `revokeSession` (ACTIVE-only predicate) *would* have emitted one for the same session — the three methods disagree on what "live" means.

apps/backend/src/services/userSessionService.ts
```diff
@@ -90,6 +137,15 @@ export class UserSessionService {
       logger.info(`✅ [${sessionCreateId}] === Session Creation Complete ===`);
       
       logger.info(`Created new session for user: ${session.userId}`);
+
+      const loginMethod = sessionData.loginMethod ?? session.user.authProvider;
+      void userActivityTrackingService.trackLogin(session.userId, {
```
sithaarth: **`trackLogin` fires on every session insert, so auto-login paths count as fresh sign-ins.** Only `createWorkspaceAuth` got a distinguishing `loginMethod`. The auto-login branch of `loginWorkspace` (authV2Controller.ts:1590-1616, commented "AUTO-LOGIN FLOW: Use existing session", reached from WorkspaceSelectionScreen and AcceptInvitation) hits `createSession` at 1679 with no `loginMethod`, falling back to `user.authProvider` → `eventLabel='GOOGLE'`. `communityWorkspaceController.ts:371` does the same and wasn't touched by this PR at all. Every workspace pick / invite accept by an already-signed-in user counts as a fresh GOOGLE sign-in, inflating logins-by-method for multi-workspace users, and the extra LOGIN gets no matching LOGOUT unless that second session is independently revoked. Consider a distinct method (e.g. `AUTO_LOGIN`) on those paths.

apps/backend/src/services/userSessionService.ts
```diff
@@ -24,13 +39,45 @@ export interface UpdateSessionData {
...
 
   constructor() {
     this.prisma = DatabaseClient.getInstance();
   }
 
+  private trackLogout(session: EndedSession, reason: LogoutReason): void {
+    void userActivityTrackingService.trackLogout(session.userId, {
+      sessionId: session.id,
+      reason,
+      platform: resolveSessionPlatform(session.deviceInfo),
```
sithaarth: **`CreateSessionData.platform` is never persisted, so LOGIN and LOGOUT report different platforms for the same session.** `platform` is read only at line 145 — never written to a column (schema has none) nor merged into `deviceInfo`. Concrete case: a native mobile client sends `x-platform: mobile` with an okhttp/Dart/CFNetwork UA (no 'mobile' substring). With >1 workspace it isn't auto-logged-in (gated on `workspaces.length===1`) and reaches `createSession` at 1679: LOGIN records MOBILE, but the `deviceInfo` built at 1672-1677 holds no platform hint, so at logout `resolveSessionPlatform` falls back to WEB. Logouts-by-platform is systematically wrong for mobile. `microsoftAuthController.ts:1089/1434` and `authV2Controller.ts:165` already do it right by embedding platform inside `deviceInfo` — do the same here (or persist the hint).

apps/backend/src/services/userSessionService.ts
```diff
@@ -262,13 +350,42 @@ export class UserSessionService {
         },
       });
 
+      liveSessions.forEach((session) => this.trackLogout(session, reason));
```
sithaarth: **Rewrites every historical row, then unbounded un-awaited INSERT fan-out.** The `updateMany` at 345-351 matches `where:{userId}` with no status guard — a user with 500 accumulated rows rewrites all 500 (dead tuples for autovacuum) to revoke the 3 that were live, on every password change/reset/deactivation; adding `status: ACTIVE` to the predicate is cheap given the existing indexes. Then this line starts N detached promises with no await and no concurrency bound, and `trackLogout → saveActivityEvent` is a single-row INSERT each (no `createMany`). 40 live sessions → 40 concurrent INSERTs on the shared Prisma pool from a path nobody awaits — pool pressure under load, events lost at shutdown. Compounding: `emailAuthController.ts:760-762` loops users strictly sequentially and this PR doubled the round trips per iteration inside the HTTP request.

apps/backend/src/services/userActivityTrackingService.ts
```diff
@@ -13,12 +13,14 @@ class UserActivityTrackingService {
...
     platform?: Platform;
+    // Real UserSession.id when the event belongs to one; a random id otherwise.
+    sessionId?: string;
     contextMetadata?: Record<string, unknown>;
   }): Promise<void> {
     try {
       const event = {
         user_id: params.userId,
-        session_id: uuidv4(),
+        session_id: params.sessionId || uuidv4(),
```
sithaarth: **`session_id` now mixes two id spaces.** AUTH rows carry a `UserSession.id` while every other event carries a per-tab uuidv4 (`useActivityTracking.ts:42`, `globalClickTracker.ts:24`, plus the `params.sessionId || uuidv4()` fallback here). One 30-day UserSession spanning 40 tabs yields 2 AUTH rows under one id and thousands of click rows under 40 unrelated ids — `COUNT(DISTINCT session_id)`, events-per-session, and any AUTH↔behavioral join silently return plausible-looking wrong numbers, with nothing but `event_category` to discriminate. Consider a separate column (e.g. `auth_session_id`) or a prefix. Related: `loginMethod` is an open `string` (accepts 'EMAIL', 'TEST', 'WORKSPACE_CREATED', any authProvider value) while `LogoutReason` beside it is a closed union — a typo'd method compiles and fragments the breakdown the union was designed to prevent.

sithaarth (commented):
Automated review (Claude Code). 9 inline comments below; one more finding is on a line outside the diff:

**`userSessionService.ts:417` — tracking attached to four hand-picked methods, not the status transition itself.** `cleanupExpiredSessions` performs exactly the transition `expireSession` now instruments (`updateMany({where:{status:ACTIVE, refreshTokenExpiry:{lt:now}}, data:{status:EXPIRED}})`) and emits nothing, and `updateSession` still accepts `status?: SessionStatus`, so any caller can reach REVOKED/EXPIRED untracked. `cleanupExpiredSessions` has no caller today, which makes it the obvious home for a future nightly cron — the day someone wires it up, most session endings become EXPIRED rows with no LOGOUT event. A single private `endSessions(where, status, reason)` helper (guarded updateMany + fetch-affected + trackLogout) would collapse the four copies and close the back doors. Related: `revokeSessionByRefreshToken` has zero callers repo-wide yet gained a required `reason` param, a `findMany` over a `@unique` column, and a third liveness predicate — delete it or route it through the shared helper.

devesh-juspay (approved):

sithaarth (approved):

PR #2125, opened 2026-09-21
fix: XYNE-64141 keep the SDLC hub and page across reloads and frame resets

ameer030:
## Problem

Reloading while inside SDLC lost the selected hub: a browser reload at bare `/:workspaceId/sdlc` and the frame's own **Reload SDLC Hub** button both booted the lane with no hub id, so `SdlcScreen` fell back to the alphabetically first hub (`channels[0]`).

## Root Cause

Hub identity lives only in the URL segment `/:workspaceId/sdlc/:channelId/...`. The frame's last location was kept solely in `frameLocationRef` (SdlcFrameHost) — a memory-only `useRef`. Both reload paths destroyed it before the fallback could use it:

1. **Browser reload**: the ref dies with the page, so the bare-`/sdlc` resume logic (PR #1727) had nothing to resume, and the iframe booted at `/sdlc-app/ws/sdlc` with no hub id.
2. **Reload SDLC Hub (frame reset)**: the reset handler intentionally nulled the ref and rebuilt the iframe at the bare app root (`/ws/sdlc`), discarding the hub by design.

Either way the lane's no-`channelId` redirect (`SdlcScreen`) resolved the hub from `channels[0]` — the alphabetically first hub (`orderBy name asc`), not the last-visited one. No durable storage of the last hub/page existed anywhere.

## Fix

- **`lastSdlcLocation.ts`** (new): durable per-workspace store of the frame's last-visited SDLC path (localStorage; only paths naming a hub are stored — bare roots and Workflows paths never overwrite a good resume target)
- **`SdlcFrameHost.tsx`:
  - every frame route report persists the path
  - a bare `/sdlc` first-visit boot resumes the stored page instead of the hub-less root (the durable version of the in-session `frameLocationRef` resume)
  - **Reload SDLC Hub** rebuilds the frame at the page it was showing (+ `?_reset=` cache-bust) instead of the app root; tooltip updated to match
- **`SdlcScreen.tsx`**: the no-`channelId` redirect prefers the stored hub (when still a member of `channels`) over `channels[0]` — covers bare frame boots the host never sees (chained resets, NotFound → `/sdlc`)

## Test plan

- [ ] Hub B open → browser reload at bare `/ws/sdlc` → resumes hub B's last page (not first hub)
- [ ] Hub B wiki page open → click Reload SDLC Hub → same page, fresh frame session
- [ ] First ever visit (no storage) → still lands on `channels[0]` overview
- [ ] Stored hub deleted/unjoined → falls back to `channels[0]`


ameer030 (commented):
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 0 |
| 🟡 Better to fix | 1 |
| 🔵 Nit | 1 |

### 🔁 Re-check
- ✅ **Verified OK:** resume-on-reload and reset-on-current-page flows traced end to end (`SdlcFrameHost` mount/reset handlers ↔ `useSdlcFrameBridge` route reports); dead-hub/unjoined stored path validated against `channels` membership in `SdlcScreen` with working `channels[0]` fallback; workflows paths correctly never persisted; `workspaceId` keyed storage; reset remount (`key={resetCount}`) and `isReady`/`frameLocationRef` lifecycle intact.

### 📌 Verdict
The feature works, but `sdlcHubIdOf`'s regex mis-extracts the hub id for sectionless URLs with a query/hash and silently reintroduces the wrong-hub fallback this PR fixes — fix the regex before merging.

apps/dashboard/src/routes/SdlcScreen/lastSdlcLocation.ts
```diff
@@ -0,0 +1,31 @@
+/** Durable per-workspace last-visited SDLC path; frameLocationRef is memory-only. */
+
+const KEY_PREFIX = 'sdlc:last-location:';
+
+/** Extracts the hub (channel) id from an SDLC path; null for bare roots and non-SDLC paths. */
+export function sdlcHubIdOf(path: string): string | null {
+  return path.match(/^\/[^/]+\/sdlc\/([^/]+)/)?.[1] ?? null;
+}
```
ameer030, line 15: 🟡 **W01 — sdlcHubIdOf swallows query/hash into the hub id, so the stored-hub redirect never matches channels**

`sdlcHubIdOf` captures `([^/]+)` after `/sdlc/`, which greedily includes a trailing `?query` or `#hash` when the path has no further segment (e.g. `/ws/sdlc/abc?canvas=x` → hub id `abc?canvas=x`). `SdlcScreen`'s no-`channelId` redirect then compares that against `item.id` and silently falls back to `channels[0]` — the wrong-hub behavior this PR is fixing, for exactly the sectionless hub URLs the app otherwise renders with a default section.

- `lastSdlcLocation.ts:8` — `path.match(/^\/[^/]+\/sdlc\/([^/]+)/)`; with nothing between the hub id and `?`/`#`, the capture runs to end of string.
- `SdlcScreen.tsx:540` — `channels.find(item => item.id === sdlcHubIdOf(storedPath))`; any captured query/hash guarantees a mismatch → `fallbackId`.
- Reachability: `AppRoot.tsx:1645` routes `sdlc/:channelId` (no section) to SdlcScreen, and `SdlcScreen.tsx:274-280` defaults `section` to `'overview'` without rewriting the URL — so `/ws/sdlc/<hubId>` is a stable, reportable location. Canvas selection lives in `?canvas=` (`SdlcScreen.tsx:559`), making `/ws/sdlc/<hubId>?canvas=<id>` a realistic stored value.
- Impact path: the PR body states this redirect "covers bare frame boots the host never sees (chained resets, NotFound → /sdlc)"; in those boots the in-frame SdlcScreen lands the user on the first hub instead of the stored one.

</details>

Stop the capture at query/hash boundaries:

```ts
return path.match(/^\/[^/]+\/sdlc\/([^/?#]+)/)?.[1] ?? null;
```

</details>

rules: BUG-C30
agent: omp
run: pr-2125/20260921T134009Z
</details>

apps/dashboard/src/routes/SdlcScreen/SdlcFrameHost.tsx
```diff
@@ -111,6 +121,7 @@ const SdlcFrameHost = (): ReactElement | null => {
       if (message.type !== SDLC_FRAME_MESSAGE.route) return;
 
       frameLocationRef.current = message.path;
+      if (workspaceId) setLastSdlcLocation(workspaceId, message.path);
```
ameer030: 🔵 **N01 — Frame route reports persist the transient `_reset` cache-bust into localStorage**

After **Reload SDLC Hub**, the frame boots at `…&_reset=<ts>` and its first route report includes that query, so the persisted resume target (and the next reset's base URL) accumulate the transient `_reset` param. Search strings do carry real state (e.g. `folder=`), so strip only `_reset`.

- `useSdlcFrameBridge.ts:247` sends `path = location.pathname + location.search + location.hash` — search included — and the bridge's `parentLocationRef` starts null (`useSdlcFrameBridge.ts:218`), so the frame's very first report after a reset boot carries `?_reset=<ts>`.
- `SdlcFrameHost.tsx:113` builds the reset src with `_reset`; `SdlcFrameHost.tsx:124` persists the reported path verbatim via `setLastSdlcLocation`; the next reset then stacks another (`…?_reset=t1&_reset=t2`).
- Nothing reads `_reset` (only writers in `SdlcFrameHost.tsx`/`SdlcWindow.tsx`), so impact is durable-state hygiene plus address-bar junk via the route→navigate sync (`SdlcFrameHost.tsx:127-128`).

</details>

In `setLastSdlcLocation` (or before calling it), drop the param:

```ts
const url = new URL(path, window.location.origin);
url.searchParams.delete('_reset');
localStorage.setItem(keyFor(workspaceId), url.pathname + url.search + url.hash);
```

</details>

rules: BUG-C27
agent: omp
run: pr-2125/20260921T134009Z
</details>

ameer030:
🔵 **N01 — SdlcFrameResetMessage contract comment still says the frame remounts at the SDLC root**

The diff changed the reset handler to remount the frame at its current page, but the interface doc for the reset message still reads "destroy this frame and mount a fresh one at the SDLC root" — the opposite of the new behavior. This file is the shared contract between the two bundles, so the next person wiring a reset sender will read the stale version.

- `apps/dashboard/src/routes/SdlcScreen/sdlcFrameMessages.ts:38` — stale comment.
- New behavior: `apps/dashboard/src/routes/SdlcScreen/SdlcFrameHost.tsx:114-118` stays at `frameLocation` when it names a hub.
</details>

Reword, e.g. "destroy this frame and mount a fresh one at the frame's current page (or the SDLC root when none)".
</details>

rules: DOCS-C04
agent: omp
run: pr-2125/20260921T142229Z
</details>

ameer030 (commented):
## 📝 Xyne PR Review — Summary

### 📊 Results
| Category | Count |
|----------|-------|
| 🔴 Blocking | 0 |
| 🟡 Better to fix | 1 |
| 🔵 Nit | 1 |

### 🔁 Re-check
- 💬 **Existing threads:** 2 fixed ✅ — both prior findings (greedy `sdlcHubIdOf` capture, persisted `_reset`) verified fixed in this head
- ✅ **Verified OK:** postMessage origin/source + root-relative path shape still enforced; stored value shape-validated on read (BUG-C27); dead/deleted stored hub converges to `channels[0]` via the membership check and NotFound → `/sdlc`; the `!workspaceId` guard in the redirect is dead-code-safe since every SDLC route mounts under `:workspaceId`; ready→route message ordering resumes without clobbering the frame; old/new lane bundle skew degrades to prior behavior

### 📌 Verdict
The resume logic is sound, but the reset path appends the `_reset` cache-bust after any URL fragment — fix that (and the stale contract comment) before merge.

apps/dashboard/src/routes/SdlcScreen/SdlcFrameHost.tsx
```diff
@@ -97,10 +109,13 @@ const SdlcFrameHost = (): ReactElement | null => {
...
-        // request a true cache-bypassing reload.
-        const root = `/${workspaceId}/sdlc`;
-        initialSrcRef.current = `${SDLC_APP_BASE_PATH}${root}?_reset=${Date.now()}`;
+        // Timestamp forces a fresh document rather than a cached one. Stay on the
+        // frame's current page instead of dropping to the hub root.
+        const frameLocation = frameLocationRef.current;
+        const root =
+          frameLocation && sdlcHubIdOf(frameLocation) ? frameLocation : `/${workspaceId}/sdlc`;
+        const separator = root.includes('?') ? '&' : '?';
+        initialSrcRef.current = `${SDLC_APP_BASE_PATH}${root}${separator}_reset=${Date.now()}`;
```
ameer030, line 118: 🟡 **W01 — Reset cache-bust is appended after the URL fragment, so hash-bearing pages bypass it and pollute the stored path**

`initialSrcRef.current = ${SDLC_APP_BASE_PATH}${root}${separator}_reset=...` appends the cache-bust to the end of `root`, but `root` comes from `frameLocationRef.current`, which the lane reports with `pathname + search + hash`. On any page carrying a fragment — this lane uses `#origin`/`#messageId` scroll targets — the `_reset` parameter lands *inside* the fragment, not the query string.

Two consequences: (1) the `_reset` timestamp never reaches the document URL, so the fresh-document cache-bust the timestamp exists for is a no-op on exactly those pages; (2) `withoutResetParam` strips via `url.searchParams.delete('_reset')`, which cannot see a fragment — so when the remounted frame reports its route, the polluted hash (`#messageId&_reset=1737…`) survives, gets mirrored into the dashboard's address bar, and is persisted into `sdlc:last-location:*`. Every later resume then carries the stale garbage fragment.

- `apps/dashboard/src/routes/SdlcScreen/SdlcFrameHost.tsx:114-118` — `root` may contain `#…`; `_reset` is concatenated after it.
- `apps/dashboard/src/routes/SdlcScreen/useSdlcFrameBridge.ts:246` — the lane reports `pathname + search + hash`, so `frameLocationRef` regularly holds a hash (see also the comment at SdlcFrameHost.tsx:148: "The hash carries #origin/#messageId scroll targets").
- `apps/dashboard/src/routes/SdlcScreen/lastSdlcLocation.ts:15-17` — `withoutResetParam` only deletes from `searchParams`; a `_reset` in the fragment is kept.
- Callers checked: the route handler (SdlcFrameHost.tsx:127-133) passes the polluted path to both `setLastSdlcLocation(workspaceId, reported)` and `navigate(reported)`, so storage and the address bar both pick it up. A hash-only path (`/ws/sdlc/<hub>/page#msg`) is the worst case: the entire `?_reset=ts` becomes fragment text.
</details>

Insert the cache-bust before the fragment instead of after the whole path:

```ts
const hashIndex = root.indexOf('#');
const beforeHash = hashIndex === -1 ? root : root.slice(0, hashIndex);
const hash = hashIndex === -1 ? '' : root.slice(hashIndex);
const separator = beforeHash.includes('?') ? '&' : '?';
initialSrcRef.current = `${SDLC_APP_BASE_PATH}${beforeHash}${separator}_reset=${Date.now()}${hash}`;
```

A unit test for the three shapes (no query, with query, hash-only) in `lastSdlcLocation.ts`-adjacent code would lock this in.
</details>

rules: BUG-C30
agent: omp
run: pr-2125/20260921T142229Z
</details>

ameer030: Checked reachability before changing this, and I don't think the fragment case is reachable today.

Nothing in the app produces an SDLC path with a fragment:

- `BaseRoute` is only `/chat/dm | /chat/dir | /chat/bookmarks | /chat/drafts-sent | /chat/activity` (`useRouteContext.ts:3-8`). Every `#origin=` / `#messageId=` navigate in the repo resolves under `/chat/*` — none of them under `/:workspaceId/sdlc/...`.
- The SDLC routes are `sdlc/:channelId`, `sdlc/:channelId/:section`, `sdlc/:channelId/workflows/*` and `sdlc/:channelId/tickets/:ticketId` (`AppRoot.tsx:1644-1674`). All lane state is carried in query params — `?wiki=`, `?canvas=`, `?folder=` — so the lane never sets a hash on its own routes.
- The comment at `SdlcFrameHost.tsx:148` ("The hash carries #origin/#messageId scroll targets") came in with `a5a24f867`, before this lane existed. It documents generic path forwarding, not an SDLC hash producer, so it isn't evidence that one exists.

That leaves a hand-typed or hand-pasted URL as the only way in, which doesn't justify the extra branching here. Leaving it as is. If SDLC ever grows anchor deep-links (wiki headings being the obvious candidate), this becomes real and the suggested split-at-`#` fix is the right one.

ameer030:
> 🔵 **N01 — SdlcFrameResetMessage contract comment still says the frame remounts at the SDLC root**

Fixed in 7fdc65af3.

Reworded to drop the boot path entirely rather than describe the new one, because the two hosts now disagree: `SdlcFrameHost.tsx:114-118` remounts at the frame's current page, while `SdlcWindow.tsx:62` still boots at `/:workspaceId/sdlc`. A shared contract that names either target is wrong for the other host.

```diff
-/** Frame → parent: destroy this frame and mount a fresh one at the SDLC root. */
+/** Frame → parent: destroy this frame and mount a fresh one; the host picks the boot path. */
```

Mohan0544 (approved):

utkarshpandey6 (approved):

nikunjJuspay (approved):

PR #2134, opened 2026-09-21
fix: scope board custom fields to the board they were configured on

XyneSpaces:
## Problem

Every release board in a workspace is bound to the SAME seeded, workspace-wide form. `saveReleaseBoardConfig` (`apps/backend/src/zero/mutators.ts:7749-7802`) looks up `xyne_release_specs_form` / `xyne_release_version_specs_form` by name + workspaceId and points every main release board and every application board at that one `formId`.

Consequence: a custom field added from ONE release board's edit screen is written to that shared form, so it shows up on every other release board's Create Release modal and, when not optional, blocks creation there. Per-board `customFieldVisibility` did not help, because the mandatory gate in `CreateTicketModal` was fed the unfiltered field list — a field hidden on a board still blocked submission with a tooltip naming a field the user could not see.

## Fix

1. **Fork-on-write (`apps/backend/src/services/boardCustomFieldsService.ts`, new).** The first time a board edits its custom fields, if its form is shared with other contexts, the board gets its own copy of the same fields, its `forms_context_mapping` row is repointed at the copy, and that board's existing ticket `form_entity_values` are moved onto the copy's field ids (the seeded release fields are legacy rows, so their ids change on clone; already global-backed fields keep their ids and only the `formId` moves). Boards still bound to the shared form are untouched. Idempotent and a no-op for any board whose form is already private, i.e. every non-release board.
   New endpoint: `POST /api/boards/:boardId/custom-fields/detach-shared-form` (authenticated, board must be in the caller's workspace).
2. **Board edit screen** calls that endpoint before updating fields and re-keys the edited field list onto the forked form's ids.
3. **Create-ticket gate** now uses the board's visible field list for both the mandatory-field tooltip and submit validation, so a field hidden on a board can never block creation there.

## Validation

- `tsc --noEmit` clean for the dashboard; backend typecheck shows only pre-existing errors in `src/workflowsV2/agents/*` (untouched by this change).
- Needs manual verification on staging: add a custom field on one release board, confirm it does not appear on other release boards, and confirm existing release tickets on the edited board still show their previously captured `branch` / `deployedCommitId` / `newCommitId` values after the fork.


apps/backend/src/services/boardCustomFieldsService.ts
```diff
@@ -0,0 +1,202 @@
...
+ * This service forks that shared form the first time a board edits its custom fields:
+ * the board gets its own copy of the exact same fields, its mapping is repointed at the
+ * copy, and the board's existing ticket values are moved onto the copy's field ids so
+ * nothing already captured is lost. Boards still bound to the shared form are untouched.
+ *
+ * A board whose form is already private (only one context mapping) is left exactly as
+ * it is — this is a no-op for every non-release board.
+ */
+export class BoardCustomFieldsService {
+  async detachSharedBoardTicketForm(
```
XyneSpaces, line 51: **[blocking] R2 — Non-atomic fork/repoint flow.** `detachSharedBoardTicketForm` creates a form, clones fields, repoints `form_entity_values` in batches, and updates `formContextMapping` as separate DB calls. If anything fails after `createWithFields`, the board is left with a dangling private form while its mapping still points at the shared form, or values are partially moved. Wrap the entire flow in `db.$transaction` (interactive/serializable if concurrent forks are possible) and pass the transaction client into repository calls, or restructure so the mapping is updated atomically with the value repoint.

apps/backend/src/services/boardCustomFieldsService.ts
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    const sourceForm = await repositories.forms.findFormWithFields(mapping.formId);
+    if (!sourceForm) {
+      return { formId: mapping.formId, detached: false, repointedValues: 0, fields: [] };
+    }
+
+    const newForm = await repositories.forms.createWithFields({
+      formName: `${board.name} Custom Fields`,
+      formDescription: `Custom fields for ${board.name}`,
```
XyneSpaces, line 86: **[blocking] R2/R3 — Field cloning drops columns.** The spread in `createWithFields` only preserves `fieldName`, `fieldType`, `options`, `isOptional`, and `parentOptionId`. Other `form_fields`/`form_field_definitions` properties (`description`, `placeholder`, `sequenceNumber`, `defaultValue`, validation config, etc.) are lost during the fork, silently regressing the board’s custom fields. Clone all scalar properties from the source field row.

apps/backend/src/routes/boards.ts
```diff
@@ -7,4 +7,11 @@ const boardController = new BoardController();
 // Create a new board with stages
 router.post('/', boardController.createBoard);
 
+// Give this board its own copy of a custom-fields form it currently shares with other
+// boards, so editing its fields cannot change theirs.
```
XyneSpaces, line 11: **[major] R1/R8 — New mutating endpoint lacks validation and explicit auth/ACL middleware.** `POST /:boardId/custom-fields/detach-shared-form` registers the raw controller. Add a Zod schema for `:boardId` (UUID) and confirm this router is mounted behind authentication and board-level ACL middleware; otherwise unauthenticated or cross-workspace calls can reach the controller.

apps/backend/src/controllers/boardController.ts
```diff
@@ -152,4 +153,47 @@ export class BoardController {
       res.status(500).json({ error: 'Internal server error' });
     }
   };
+
+  /**
```
XyneSpaces, line 157: **[minor] R8 — Manual param validation instead of shared Zod middleware.** Consider moving the `boardId` presence/UUID check into the route-level Zod validator so controllers stay thin and validation is consistent across board routes.

apps/dashboard/src/components/Board/BoardEditScreen/BoardEditScreen.tsx
```diff
@@ -1292,13 +1292,59 @@ const BoardEditScreen = ({
...
+              if (!forked) return field;
+              const { membershipId: _previousMembershipId, ...rest } = field;
+              return {
+                ...rest,
+                id: forked.fieldId,
+                ...(forked.membershipId ? { membershipId: forked.membershipId } : {}),
+              };
+            });
+          }
+        } catch {
```
XyneSpaces, line 1339: **[minor] R6 — Error is surfaced to the user but not logged.** The bare `catch { toast.error(...); return; }` hides the underlying failure from Sentry/ops. Capture the error with `logger.error` or Sentry before returning so incidents can be investigated.

XyneSpaces (commented):
## Xyne Spaces Reviewer — PR #2134

**Verdict:** Changes requested — blocking correctness/data-loss issues need fixing before merge.
**Reviewed:** 5 files · ~320 hunks · **Findings:** 2 blocking, 3 major, 3 minor

### Findings
| blocking | R2 | `apps/backend/src/services/boardCustomFieldsService.ts:51` | Fork/create/repoint/mapping steps are not wrapped in a DB transaction; a mid-flight failure leaves a partially-forked form and inconsistent ticket values. |
| blocking | R2/R3 | `apps/backend/src/services/boardCustomFieldsService.ts:86` | Cloned fields copy only a subset of `form_fields` columns (`fieldName`, `fieldType`, `options`, `isOptional`, `parentOptionId`), dropping `description`, `placeholder`, `sequenceNumber`, defaults, validation rules, etc. |
| major | R1/R8 | `apps/backend/src/routes/boards.ts:11` | New mutating endpoint has no Zod path-param/body validation and relies on the caller for auth/ACL; verify it sits behind authentication and board ACL middleware. |
| major | R7 | `apps/backend/src/services/boardCustomFieldsService.ts` (new) | New service and endpoint have no automated tests covering the happy path, partial failures, or concurrent forks. |
| major | R4 | `apps/backend/src/services/boardCustomFieldsService.ts:132` | `formContextMapping.update` runs outside a transaction and after the value repoint; races can duplicate forms or repoint onto a stale fork. |
| minor | R6 | `apps/dashboard/src/components/Board/BoardEditScreen/BoardEditScreen.tsx:1339` | Catch block shows a toast but never logs or reports the underlying error, hindering incident investigation. |
| minor | R8 | `apps/backend/src/controllers/boardController.ts:157` | `:boardId` is validated manually in the controller; prefer a shared Zod validation middleware. |
| minor | R6 | `apps/backend/src/services/boardCustomFieldsService.ts:137` | Template-literal logging injects user-controlled values into logs (already flagged by CodeQL). Use structured `logger.info(message, { boardId, actorUserId, … })`.

### Recommended reviewer
**@devesh-juspay** — not the PR author, has recent backend security/auth context, and 4 open review requests in this repo. CODEOWNERS does not cover these paths, so I picked the strongest available backend reviewer. *(I don’t have a `request_reviewers` tool in this session, so please add @devesh-juspay as a reviewer manually.)*

| Candidate | Owns files in this PR | Default reviewer? | Pending reviews (open PRs) | Tie-breaker |
|-----------|-----------------------|-------------------|---------------------------|-------------|
| @devesh-juspay | recent backend security/auth work | no (CODEOWNERS N/A for these paths) | 4 | active Juspay GH account, not the author |
| @devesh-rastogi | — | no | 0 | could not confirm repo access / mapping to Spaces profile |

Top blocking issue: the fork-and-repoint flow must be atomic (`db.$transaction`) and clone every field column to avoid data loss.

